PR #1, opened 2016-02-24
swupd-client: Add Curl bottom layer error messages

bradTpeters:
In each of these Curl error cases, we have no choice but to terminate
operation. We cannot programmatically work around these issues,
therefore this patch adds error msg details so user can pursue a
solution


bradTpeters:
Done. Should have just left this patch as it was v1. Simple solution for a simple problem

phmccarty:
I did a squash rebase of this branch and merged to master. Thanks!

PR #2, opened 2016-02-25
Fix massive output spam

tmarcu:
During a swupd verify --install, files are expected to be missing and must
be "fixed," so do not print out that it is missing/fixed in this case.


bryteise:
+1

PR #3, opened 2016-02-25
Search feature

bradTpeters:
Patch v2, incorporates feedback by William Douglas

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -52,9 +52,9 @@ char *bin_paths[] = {
 
 static void print_help(const char *name) {
 	printf("Usage:\n");
-	printf("	swupd %s 'search_term'\n", basename((char*)name));
-	printf("		'search_term': A substring of a binary or library name\n");
-	printf("		Return: Name of bundle or bundles in which this binary is found\n\n");
+	printf("	swupd %s [Options] 'search_term'\n", basename((char*)name));
+	printf("		'search_term': A substring of a binary, library or filename (default)\n");
+	printf("		Return: Bundle name : filename matching search term\n\n");
 
 	printf("Help Options:\n");
 	printf("   -h, --help              Display this help\n");
--- a/src/search.c
+++ b/src/search.c
@@ -237,11 +237,10 @@ void do_search(struct manifest *MoM, char search_type, char *search_term)
 	struct file *file;
 	struct file *subfile;
 	struct manifest *subman = NULL;
-	int i, ret, current_version;
+	int i, ret;
 	bool done_with_bundle, done = false;
 	bool hit = false;
 
-	current_version = read_version_from_subvol_file(path_prefix);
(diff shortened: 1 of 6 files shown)


bradTpeters:
New v3 pushed, should incorporate all feedback. Thanks all!

src/search.c
```diff
@@ -0,0 +1,506 @@
...
+
+int search_main(int argc, char **argv)
+{
+	int ret = 0;
+	int lock_fd;
+	struct manifest *MoM = NULL;
+
+	if (!parse_options(argc, argv) ||
+		create_required_dirs()) {
+		free_globals();
```
phmccarty: init_globals() is called the first time a few lines down, so you'll want to avoid calling it here.

src/swupd.c
```diff
@@ -39,6 +39,7 @@ static struct subcmd commands[] = {
 	{ "update", "Update to latest OS version", update_main },
 	{ "verify", "Verify content for OS version", verify_main },
 	{ "check-update", "Checks if a new OS version is available", check_update_main},
+	{ "search", "Search Clear Linux for a binary or library", search_main },
```
phmccarty: This command does not appear to be specific to Clear Linux. Do we need to mention Clear Linux here?

src/search.c
```diff
@@ -0,0 +1,506 @@
...
+	char *tarfile, *untard_file, *url;
+
+	while (list) {
+		file = list->data;
+		list = list->next;
+
+		string_or_die(&untard_file, "%s/%i/Manifest.%s", STATE_DIR, file->last_change,
+						file->filename);
+
+		string_or_die(&tarfile, "%s/%i/Manifest.%s.tar", STATE_DIR, file->last_change,
```
phmccarty: The tarfile string allocated here is unused in this function, and the memory is not freed.

bradTpeters:
Archana, lmk when your back in. We should discuss the load_manifests() question, and whether a reversing of the arguments order makes sense

src/swupd.c
bradTpeters: The search is specifically over the manifests/bundles provided by Clear Linux, so I thought it made sense.

bradTpeters:
And one more

bradTpeters:
Cleanup from prev patch added.

src/swupd.c
phmccarty: On second thought, I'm okay with leaving this for now.

src/search.c
phmccarty: Comment addressed :-)

src/swupd.c
bradTpeters: Great

phmccarty:
Looks good to me. We can iterate from this point. I did a squash/merge to master.

PR #5, opened 2016-03-01
Fix installation of missing boot files with bundle-add

phmccarty:
A user was running into issues installing the 'kernel-container' bundle
on Clear Linux. Namely, all missing files marked as boot were not
installed.

The root issue is that the bundle-add action does not set the "fix"
global variable before calling ignore(), so ignore() mistakenly returns
true when considering boot files that are missing, but not marked
deleted.

Since there are only three callers of ignore(), I have removed the
global variable and instead pass in the required values of "fix" to the
new second argument to ignore, "fix_missing". Passing "true" for the
bundle-add action corrects the return value of ignore() in this case.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,67 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: true
+AlignConsecutiveAssignments: false
+AlignEscapedNewlinesLeft: false
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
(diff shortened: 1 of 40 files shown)


phmccarty:
Looking at this again today, I'm not seeing any reason to keep the special boot file handling in ignore(). Does anyone have any insight into why that handling is still needed?

amshinde:
I don't see any reason for the special boot handling in ignore(). Looks like the only place we need to check for boot files is in remove_orphaned_files() while doing verify --fix so as to not remove any boot files that have been marked as deleted.

phmccarty:
I'm running into another bug with running 'verify --fix' for boot files. I'll prepare a fix for that, and then open a new PR, also with functional tests.

phmccarty:
I cleaned up this patch and wrote some functional tests as well. Closing this one and opening a new PR.

PR #9, opened 2016-03-07
Fix build failure on Yocto

avalluri:
Building on Yocto install phase certificate files are being installed twice as included in
_DATA twice. We can use EXTRA_DIST than dist_.


phmccarty:
Looks good, thanks!

PR #11, opened 2016-03-08
Right usage of AC_ARG_ENABLE on bzip2

avalluri:


avalluri:
Updated the PR to make bzip2 default selection.

phmccarty:
Rebased and pushed to master.

PR #13, opened 2016-03-09
Add compatibility with libarchive's bsdtar command

rojkov:
Since GNU tar fails to extract files with xattrs preserved when
Integrity Measurement Architecture (IMA) is enabled some vendors
may choose to install libarchive-based tar (bsdtar) on their embedded
devices, so the swupd server needs to be able to create archives
in its format.

This patch adds one compile-time options --enable-bsdtar that is used
to enable/disable GNU tar specific options.

Signed-off-by: Dmitry Rozhkov <dmitry.rozhkov@linux.intel.com>

PS. I understand the team may not want to support two flavors of tar, but as users of the tool we have a use case for that and have to maintain our patches ourselves. And since our development policy requires us to submit PRs upstream for the patches we maintain I submit this PR.

### DIFF
--- a/src/staging.c
+++ b/src/staging.c
@@ -148,8 +148,8 @@ int do_staging(struct file *file, struct manifest *MoM)
 			ret = -errno;
 			goto out;
 		}
-		string_or_die(&tarcommand, "tar -C %s " TAR_PERM_ATTR_ARGS " -cf - '%s' 2> /dev/null | "
-					   "tar -C %s%s " TAR_PERM_ATTR_ARGS " -xf - 2> /dev/null",
+		string_or_die(&tarcommand, TAR_COMMAND " -C %s " TAR_PERM_ATTR_ARGS " -cf - %s 2> /dev/null | "
+			      TAR_COMMAND " -C %s%s " TAR_PERM_ATTR_ARGS " -xf - 2> /dev/null",
 			      rename_tmpdir, base, path_prefix, rel_dir);
 		ret = system(tarcommand);
 		if (WIFEXITED(ret)) {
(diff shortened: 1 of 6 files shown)


rojkov:
rebased to the latest master, resolved conflicts, force pushed.

barbieri:
ping?

src/staging.c
```diff
@@ -148,8 +148,8 @@ int do_staging(struct file *file, struct manifest *MoM)
 			ret = -errno;
 			goto out;
 		}
-		string_or_die(&tarcommand, "tar -C %s " TAR_PERM_ATTR_ARGS " -cf - '%s' 2> /dev/null | "
-					   "tar -C %s%s " TAR_PERM_ATTR_ARGS " -xf - 2> /dev/null",
+		string_or_die(&tarcommand, TAR_COMMAND " -C %s " TAR_PERM_ATTR_ARGS " -cf - %s 2> /dev/null | "
```
phmccarty: The single quotes were present to protect filenames with spaces or other special characters interpreted by the shell. However, because this prohibits filenames with single quotes, I want to make this more robust in the future.

I will restore the single quotes in a followup commit.

phmccarty:
I rebased and pushed this commit to master, with a followup commit to restore the single quotes.

src/staging.c
barbieri: ok, we're suffering from quotes and other kind of injections possible with `system()`. My recommendation is for you guys to do a review and replace these with non-`system()` variants where you would avoid calling `sh -c "string"` that is implied from `system()`, using a `fork()` + `execv()` + `waitpid()`

PR #14, opened 2016-03-10
Update source file formats with clang v3.8

bryteise:


bryteise:
Looks like some files were missed, redoing.

PR #15, opened 2016-03-10
Update source file formats with clang v3.8

bryteise:
Fixing missed files from https://github.com/clearlinux/swupd-client/pull/14


phmccarty:
Looks good. Pushed.

PR #16, opened 2016-03-11
Bundle add include

bryteise:
This patchset adds support for the client to install bundle includes once this feature is in the server.


phmccarty:
This looks like a good starting point. I will merge, and let's iterate from here.

PR #18, opened 2016-03-15
Search testing

bradTpeters:


ikeydoherty:
RFC 3986 already covers port numbers, should this not just be handled in the URI decode?

```
      authority   = [ userinfo "@" ] host [ ":" port ]
```

ikeydoherty:
..Feel somewhat guilty for mentioning it now I see the size of the diff.. :)

bryteise:
@ikeydoherty Oh so late on that one... I completely forgot about that since I have too many different things the port gets passed to but yea all the port additions for testing the client were not required =(. I'll likely make a patch of shame removing them at some point when I get over my depression.

ikeydoherty:
@bryteise yeah I feel bad for bringing it up now :S I notice Github now has comment reactions.. where's the "crying into a cup" one ..

test/functional/search/content-check-negfull-path/setup.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+rm -f web-dir/*/*.tar
+touch target-dir/test-file
```
bryteise: These don't seem to be used for anything and aren't cleaned up in a teardown script so they'll show up in git status as noise after running make check.

bryteise:
+1 outside of my comment about the setup script touching files.

bryteise:
@ikeydoherty I don't know but party in your comment now XP.

ikeydoherty:
lol

bradTpeters:
bryteise - Sure I can cleanup the touch code; is the rm needed? These were copied from one of the existing test cases and repurposed for search. I didn't do any real cleanup

bradTpeters:
Ikeydoherty - I'll leave the port cleanup to bryteise, since that same usage exists in every primary swupd command

bryteise:
@bradTpeters The rm isn't needed since you don't add any new files to the target-dir for the purpose of the test so all you need to do is remove the touch commands for all the tests.

test/functional/search/content-check-negfull-path/setup.sh
phmccarty: The guideline to follow here is: If files need to be added/removed at runtime for the test, the reverse operation should be applied in teardown.sh. Otherwise, the files should be committed to the tree.

In this case, it appears that "target-dir/test-file" is unused by the test, so this line can be removed entirely.

test/functional/search/content-check-negfull-path/setup.sh
```diff
@@ -0,0 +1,10 @@
...
+touch target-dir/usr/share/clear/bundles/test-bundle
```
phmccarty: Similarly for this line. Unused file?

test/functional/search/content-check/setup.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
phmccarty: This "content-check" test is not run. Should it be added to the functional-tests script?

test/functional/search/content-check-negwrongpath/setup.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
phmccarty: This test is also not being run. Should it be added to the test suite?

phmccarty:
Aside the few comments/questions I had, these changes look fine. +1

test/functional/search/content-check-negfull-path/setup.sh
bradTpeters: Yeah I'll clean these up. They were inherited from the test case I reused

test/functional/search/content-check-negwrongpath/setup.sh
bradTpeters: Good eye, I wrote the test case for this but pointed it at the wrong folder

bradTpeters:
Branch has been updated with feedback

bryteise:
Looks good, +1.

phmccarty:
Looks good to me.

bradTpeters:
v3 with that final cleanup

phmccarty:
Looks good, thanks. I squashed the v2 and v3 commits into the functional test commit, and pushed to master.

PR #19, opened 2016-03-15
Update include

bryteise:
This patch set adds support for swupd update to pull in included bundles. It also has a few fixups and cleanups found along the way.


phmccarty:
Looks good, pushed.

PR #22, opened 2016-03-25
Store default URLs and format in config files; cleanups for URL handling

phmccarty:
This patch series changes where default URL and format values are stored; instead of hardcoding the values in the swupd binary, they are now stored in config files. It also cleans up the handling of the URL global variables.


src/globals.c
```diff
@@ -60,20 +60,95 @@ bool download_only;
...
+	}
+
+	/* the file should contain exactly one line */
+	line[0] = 0;
+	if (fgets(line, LINE_MAX, file) == NULL) {
+		if (ferror(file)) {
+			printf("Error: Unable to read data from %s\n", path);
+			return -1;
+		}
+		if (feof(file)) {
```
bryteise: Since these both return do you want this to be an else?

src/globals.c
```diff
@@ -60,20 +60,95 @@ bool download_only;
...
+			printf("Error: Contents of %s are empty\n", path);
+			return -1;
+		}
```
bryteise: Need to close the file before exiting this function.

src/globals.c
```diff
@@ -152,34 +152,41 @@ void set_version_url(char *url) {
...
-		string_or_die(&format_string, "%s", SWUPD_DEFAULT_FORMAT);
-		return true;
-	}
 
-	// allow "staging" as a format string
-	if ((strcmp(userinput, "staging") == 0)) {
+		// otherwise, expect a positive integer
+		errno = 0;
+		version = strtoull(userinput, NULL, 10);
+		if ((errno < 0) || (version <= 0)) {
```
bryteise: Since you are calling strtoull version can't be negative.

bryteise: version check should be version == 0 here.

src/helpers.c
```diff
@@ -783,14 +783,34 @@ int verify_fix_path(char *targetpath, struct manifest *target_MoM)
 	return ret;
 }
 
-/* In case URL has the file:// prefix, use libcurl's file protocol to copy
- * files from the local filesystem instead of downloading from a web server.
- * The local_download global variable defaults to false and is used to
- * determine how to handle libcurl results after a *_perform.
+/* In case both version_url and content_url have the file:// prefix, use
+ * libcurl's file protocol to copy files from the local filesystem instead of
+ * downloading from a web server.  The local_download global variable defaults
```
bryteise: global variables*

src/helpers.c
```diff
@@ -783,14 +783,34 @@ int verify_fix_path(char *targetpath, struct manifest *target_MoM)
...
+
+	if (strncmp(version_url, "file://", 7) == 0) {
+		version_local = true;
+	}
+	if (strncmp(content_url, "file://", 7) == 0) {
+		content_local = true;
+	}
+
+	/* protocols must match for download logic */
+	if ((version_local && !content_local) || (!version_local && content_local)) {
```
bryteise: Wouldn't version_local == content_local be the simpler check?

src/search.c
```diff
@@ -106,6 +110,20 @@ static bool parse_options(int argc, char **argv)
 			set_version_url(optarg);
 			set_content_url(optarg);
 			break;
+		case 'c':
```
bryteise: Need to add c and v to getopt_long string.

src/globals.c
```diff
@@ -149,9 +149,21 @@ void set_version_url(char *url) {
...
 }
 
-bool set_format_string(char *userinput)
+static bool is_valid_integer_format(char *str)
 {
 	int version;
+	errno = 0;
+
+	version = strtoull(str, NULL, 10);
+	if ((errno < 0) || (version <= 0)) {
```
bryteise: version == 0

src/globals.c
bryteise: Ah sorry this was fixed in a later patch.

src/clr_bundle_add.c
```diff
@@ -45,6 +45,8 @@ static void print_help(const char *name)
 	printf("Help Options:\n");
 	printf("   -h, --help              Show help options\n");
 	printf("   -u, --url=[URL]         RFC-3986 encoded url for version string and content file downloads\n");
+	printf("   -c, --contenturl=[URL]  RFC-3986 encoded url for content file downloads\n");
+	printf("   -v, --versionurl=[URL]  RFC-3986 encoded url for version string download\n");
```
bradTpeters: A good argument could be made for adding these flags to the top-level swupd command, rather than each sub-command.  1) The end-result is to set global config variables, and 2) these config options apply to every sub-command

src/globals.c
```diff
@@ -60,51 +60,156 @@ bool download_only;
...
-	// otherwise, expect a positive integer
+void set_version_url(char *url) {
+	int ret;
+
+	ret = set_url(&version_url, url, default_version_url_path);
+	if (ret < 0) {
+		printf("\nDefault version URL cannot be read. Use the -v option instead.\n");
+		exit(EXIT_FAILURE);
+	}
+}
```
bradTpeters: One possibility here which I think makes sense:

When a user passes a cmdline config option for content or version url, write it out to the state/config file, overwriting the prior value. Could make this a user-specific file, so that a given user can set their default without altering default settings for other users. This implementation detail would allow a user to modify their default behavior 'on the fly', and retain that behavior between runs

src/clr_bundle_add.c
barbieri: This is a good approach

src/clr_bundle_rm.c
```diff
@@ -46,7 +46,10 @@ static void print_help(const char *name)
 	printf("   -h, --help              Show help options\n");
 	printf("   -p, --path=[PATH...]    Use [PATH...] as the path to verify (eg: a chroot or btrfs subvol\n");
 	printf("   -u, --url=[URL]         RFC-3986 encoded url for version string and content file downloads\n");
+	printf("   -c, --contenturl=[URL]  RFC-3986 encoded url for content file downloads\n");
```
barbieri: Maybe mark the line above with "(deprecated, use --contenturl instead?)"

src/check_update.c
```diff
@@ -37,6 +37,7 @@ static void print_help(const char *name)
 	printf("Help Options:\n");
 	printf("   -h, --help              Show help options\n");
 	printf("   -u, --url=[URL]         RFC-3986 encoded url for version string and content file downloads\n");
+	printf("   -v, --versionurl=[URL]  RFC-3986 encoded url for version string download\n");
```
barbieri: Also add --contenturl?

src/globals.c
barbieri: This would only make sense if --path is used, otherwise it would conflict to have different users using different values.

src/clr_bundle_add.c
tmarcu: The idea here was the configs that store the default URLs should never be modified by a client, they are created by the server as update content and remain ~stateless, so each subcommand requires no flags by default and just reads the files. The flags should be used in very specific cases to override the configs - generally for testing custom content.

src/check_update.c
bryteise: For check-update there isn't content it would ever use so the option doesn't make sense here I don't think.

barbieri: My bad, ok

src/clr_bundle_add.c
barbieri: My plan is to use "levels of quality" with different versionurl, like latest build, QA-approved (manually tested) and milestones. They would only provide different numbers of the same content-stream/url.

Then how to do that with your suggested approach?

I'd expect an overlay in /etc, where /usr would default to something (i.e.: latest build) and user can set /etc to another (i.e.: qa-approved). We'd cook images with one or another pre-set, that is, the image would come with a state in /etc. this is not optimal, so suggestions are welcome.

bryteise: The /etc overlay approach is fine though adding an /etc value to the image on creation is unfortunate. I am not remembering if we had decided to add the /etc detection and use to this later or not though. @phmccarty would know =).

src/globals.c
phmccarty: I'm okay either way. I'll add the else.

src/globals.c
phmccarty: The bigger problem is that "version" is an int, not an unsigned type, so I'll make that correction too.

src/helpers.c
phmccarty: Yes, I will change it.

src/search.c
phmccarty: Good catch.

src/globals.c
phmccarty: Will change this too, and fix the variable type.

src/clr_bundle_rm.c
phmccarty: bundle-remove needs both version and content URLs, so the -u option is available for convenience.

src/check_update.c
phmccarty: If check-update ever needs to download update content, then we can add a -c option at that time.

src/clr_bundle_add.c
phmccarty: @bradTpeters I agree that some commandline options should move to toplevel, but I would like to hold off on doing that until we rework the subcommand implementation so that all option parsing happens in swupd.c.

phmccarty: @barbieri I'm not sure having an /etc overlay would work well for swupd... If the default config in /usr/share/defaults/swupd changed (e.g. FORMAT changed), how would updates continue to work if the /etc config is overriding the /usr config? The /etc config would not inherent the format bump, and manual intervention would be needed to ensure updates could continue past the format change.

barbieri: I was not thinking about format, just content and version urls. But in that sense, the user configuration would have the same issue, or more since with /etc you have a single place to run a post-inst script that converts the format if needed.

Be it in /etc or not, I'd like a best practice to allow us to stick with different versions of the same stream (daily/nightly builds, manually qa-tested and major milestones).

barbieri: ideas? :-)

phmccarty: @barbieri To support that setup, I think it would be nice for swupd to understand a prioritized config scheme, /usr/share/defaults/swupd/??-versionurl, where swupd will choose the config the greatest priority. That way, you can have different bundles for every quality level, each of which installs a different ??-versionurl file. Would this scheme work for you?

src/globals.c
phmccarty: @bradTpeters For a better user experience, I like this idea... I think it would be okay to cache the URLs in the state dir (maybe /var/lib/swupd/{contenturl,versionurl}).

bryteise: I'd suggest adding a print out that says what url(s) and format is being used each time swupd runs.

phmccarty:
Added some more commits incorporating fixes from review comments.

src/globals.c
phmccarty: Print them out in general? I'm not certain the end user would be interested in this info, since it is more of an internal detail. Maybe we could add a --print-config option?

bryteise: My worry is admin 1 runs sudo, admin 2 wonders why their versions don't match the expected. I'd actually suggest adding a seperate swupd command to populate the state dir rather than piggypacking on the options used in any command that change default system behavior.

src/clr_bundle_add.c
barbieri: interesting! So I'd have 10-versionurl (milestone), 50-versionurl (qa-approved) and 90-versionurl (nightly)?

this looks good to me.

phmccarty: Yes, that's what I'm thinking :-)

src/globals.c
phmccarty: I think we should revisit this issue after swupd-client (soon) gains a configurable state directory.

phmccarty:
I pushed these changes to master after squashing a couple of the commits I added this morning.

src/clr_bundle_add.c
barbieri: Nice! @kad and @pohly can you track this for future inclusion in ostroproject?

@phmccarty please let us know when that gets implemented.

PR #24, opened 2016-03-30
Adding script to create autocompletion function for complete command.

jrguzman-intel:


bryteise:
I think the project is mostly aligned on 2 space indents for shell scripts.

scripts/swupd-completion.sh
```diff
@@ -0,0 +1,139 @@
...
+SCRIPTNAME="swupd.bash"
+COMPLETIONFUNCTION="_swupd"
+SWUPDCOMMAND="./swupd"
+
+optionsfromhelp=""
+subcommandsfromhelp=()
+optionsfromsubcommand=()
+
+#Checks if swupd commad exists in current shell
+if [ -a $SWUPDCOMMAND ] ; then
```
phmccarty: I wasn't aware of the -a test option :-)  Since -e is more commonly used, can you use it instead?

scripts/swupd-completion.sh
```diff
@@ -0,0 +1,139 @@
...
+	if [[ $line == "Help Options:" ]]; then
+	    nexttype="options"
+	fi
+	if [[ $line == "Subcommands:" ]]; then
+	    nexttype="subcommand"
+	fi
+	if [[ $line == "" ]]; then
+	    currenttype=""
+	fi
+	case $currenttype in
```
phmccarty: Since the value of $currenttype may be the empty string "", I would add another case for the empty string (even if you just run "break").

scripts/swupd-completion.sh
```diff
@@ -0,0 +1,139 @@
...
+    while read line
+    do
+	# Check when is inside of help options to start read them
+	if [[ $line == "Help Options:" || $line == "Application Options:" ]]; then
+	    nexttype="options"
+	fi
+	if [[ $line == "" ]]; then
+	    currenttype=""
+	fi
+	case $currenttype in
```
phmccarty: Similarly for this case block, I think there should be a case for the empty string.

phmccarty:
@bryteise I would like to standardize on 2-space indents for the shell scripts, yes.

scripts/swupd-completion.sh
bryteise: May want to have a not exist check and exit that way if somebody moves a test before --help it won't bomb out oddly without the command check.

phmccarty:
@jrguzman-intel Aside from the comments I added inline, and William's suggestion to change the indent level, I think these changes look good. I tested the autocompletion, and I'm not seeing any functional problems with it.

phmccarty:
Looks good, +1. I will merge shortly.

phmccarty:
I squashed the last three commits into the original and merged. Thanks!

PR #25, opened 2016-04-01
Fix update when include manifests are not new

bryteise:
Previously include manifests that were not newer than the current
version would be silently ignored even if they had not been previously
installed causing bundles to miss included dependencies.

Fix this by adding a is_tracked field to struct file that is nonzero
when the bundle the file is contained in is marked as tracked via having
a /usr/share/clear/bundles/$name file.

The is_tracked == 0 version of the file will be dropped in the
consolidate_manifest phase if that filename from another bundle has
is_tracked != 0 to avoid downloading and installing an already existing
file.

The is_tracked == 0 version of the file will then be added to the update
list from the consolidated manifest if that file was not previously
installed on the system (in addition to the other cases it would already
have been added to the list for).


tpepper:
Sure stat() is cheap, but you really don't want to call that hundreds of thousands of times down inside set_tracked_files().  Instead load_manifests() could check that the bundle is a tracked bundle or not?  Then with one stat per bundle, set_tracked_files() might have a signature, eg:

set_tracked_files(struct manifest *manifest, bool is_tracked);

Assuming a zero'd manifest, you could even drop the second argument and only call the function for bundles for which you will unconditionall set the per file flag == 1.

bryteise:
Ah I was doing something like that originally but got bit by manifest allocation and forgot about the stat when I changed it X(.

I think going back to change alloc_manfiest to default setting is_tracked to 1 and then set_tracked_files can just return if the manifest is tracked.

bryteise:
Fixed in #26

PR #27, opened 2016-04-04
Includes fixesv2

bryteise:
This change adds support for includes to verify as well as fixes update when a newly included manifest was not updated in the version being updated to.


src/manifest.c
```diff
@@ -556,6 +574,10 @@ int load_manifests(int current, int version, char *component, struct file *file,
 		}
 	}
 
+	if (component) {
+		set_tracked_files(*manifest);
```
phmccarty: If component == "MoM", this call is a bit weird, but because MoM manifests have an empty files list, should be okay.

phmccarty: and is_tracked_bundle() will fail, unless someone has bundle named "MoM" :-)

bryteise:
Updated with request #28

PR #28, opened 2016-04-04
Includes fixesv3

bryteise:
Updates from v2 (closes #27) :
 - Add comment for what happens when set_untracked_manifest_files is called with "MoM" component.
 - Rename set_tracked_files function -> set_untracked_manifest_files to correct functions purpose.
 - Remove incorrect check for is_tracked_bundle when downloading packs
 - Update format path of testing files to formatstaging

---

This change adds support for includes to verify as well as fixes update when a newly included manifest was not updated in the version being updated to.


phmccarty:
Looks good, will merge shortly.

PR #29, opened 2016-04-05
Always handle errors from recurse_manifest()

phmccarty:
Some call sites of recurse_manifest() did not handle return codes. Add handling for those, and also fix a functional test that was broken after adding the proper error handling.


bryteise:
+1

PR #30, opened 2016-04-05
Fix check for symlinks in a dirname path

phmccarty:
The intended behavior here is to check if the directory path for the
current file contains any symlinks, not the full path itself. By not
checking the directory path only, verify --fix would not delete any
symlinks that were marked deleted.

Also, because the directory path is being examined now, we can obtain
the directory file descriptor (dirfd) and use unlinkat(2) to delete
files and directories instead of unlink/rmdir; doing so avoids race
conditions in that paths may change between calling realpath() and
unlink().

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -652,7 +652,7 @@ int get_dirfd_path(const char *fullname)
 	char *dir;
 	char *real_path = NULL;
 
-	tmp = strdup(fullname);
+	string_or_die(&tmp, "%s", fullname);
 	dir = dirname(tmp);
 
 	fd = open(dir, O_RDONLY);
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -644,23 +644,46 @@ void delete_motd(void)
 	}
 }
 
-int is_dirname_link(const char *fullname)
+int get_dirfd_path(const char *fullname)
 {
 	int ret = -1;
+	int fd;
+	char *tmp = NULL;
+	char *dir;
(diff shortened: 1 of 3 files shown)


src/helpers.c
```diff
@@ -644,23 +644,46 @@ void delete_motd(void)
...
+int get_dirfd_path(const char *fullname)
 {
 	int ret = -1;
+	int fd;
+	char *tmp = NULL;
+	char *dir;
 	char *real_path = NULL;
-	real_path = realpath(fullname, NULL);
+
+	tmp = strdup(fullname);
```
bryteise: If this strdup fails...

src/helpers.c
```diff
@@ -644,23 +644,46 @@ void delete_motd(void)
...
+	dir = dirname(tmp);
```
bryteise: dir == "."

src/helpers.c
```diff
@@ -644,23 +644,46 @@ void delete_motd(void)
...
+
+	fd = open(dir, O_RDONLY);
```
bryteise: And this open will get the wrong value.

src/helpers.c
phmccarty: I will do string_or_die() here instead for now.

phmccarty:
Added a commit to address review comments.

bryteise:
+1

amshinde:
+1

phmccarty:
Merged

PR #31, opened 2016-04-06
fix segfault in bundle-remove on invalid bundle

fandeli:
When trying to remove a bundle that has not
valid name contained into /usr/share/clear/bundles
a segfault arise. This patch solves that by
properly validating the bundle name is valid
within MoM.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -272,6 +272,12 @@ int remove_bundle(const char *bundle_name)
 		goto out_free_curl;
 	}
 
+	if (!search_bundle_in_manifest(current_mom, bundle_name)) {
+		printf("Bundle name is invalid, aborting removal\n");
+		ret = EBUNDLE_REMOVE;
+		goto out_free_mom;
+	}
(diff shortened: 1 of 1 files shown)


src/bundle.c
```diff
@@ -272,6 +272,11 @@ int remove_bundle(const char *bundle_name)
 		goto out_free_curl;
 	}
 
+	if (!search_bundle_in_manifest(current_mom, bundle_name)) {
+		printf("Bundle name is invalid, aborting removal\n");
```
phmccarty: Since the "ret" variable is not set for this error condition, can we set ret to EBUNDLE_REMOVE in this case?

fandeli: of course!

fandeli: Done.

phmccarty:
Looks good! I'll push to master shortly.

phmccarty:
Merged. Thanks, Jaime.

PR #32, opened 2016-04-07
Fix a hashdump crash

phmccarty:
My recent change that removed the init_globals() call from the hashdump code led to a crash if --basepath was not passed on the command line.

To fix, refactor the path_prefix init code to a new function, and explicitly call this function for hashdump.

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -212,20 +212,28 @@ bool set_format_string(char *userinput)
 	return true;
 }
 
-bool init_globals(void)
+/* Passing NULL for PATH will either use the last --path argument given on the command
+ * line, or sets the default value ("/").
+ */
+bool set_path_prefix(char *path)
 {
 	struct stat statbuf;
 	int ret;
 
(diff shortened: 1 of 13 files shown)


src/globals.c
```diff
@@ -212,42 +212,58 @@ bool set_format_string(char *userinput)
...
+
+		/* ensure path_prefix is absolute, at least '/', ends in '/',
+		 * and is a valid dir */
+		if (tmp[0] != '/') {
 			char *cwd;
 
 			cwd = get_current_dir_name();
 			if (cwd == NULL) {
 				printf("Unable to getwd() (%s)\n", strerror(errno));
 				return false;
```
bryteise: Free tmp?

phmccarty: Yes, good catch

phmccarty:
Added the missing free() for the get_current_dir_name() error case

phmccarty:
Merged

PR #33, opened 2016-04-07
Reduce noise for updates and bundle-adds

bryteise:
Stop printing a message when bundles are included and going to be
installed or if bundles are already included and going to be skipped
when doing the add_subscriptions call.

Instead update the install_bundles call to notify users when an
installed bundle was attempted to be be installed again.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -348,7 +348,6 @@ int add_subscriptions(struct list *bundles, int current_version, struct manifest
 
 		file = search_bundle_in_manifest(mom, bundle);
 		if (!file) {
-			printf("%s bundle name is invalid, skipping it...\n", bundle);
 			continue;
 		}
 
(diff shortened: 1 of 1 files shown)


src/bundle.c
```diff
@@ -413,7 +410,7 @@ int install_bundles(struct list *bundles, int current_version, struct manifest *
 
 	if (ret) {
 		if (ret == 1) {
-			printf("There are no pending bundles to install, exiting now\n");
+			printf("%s bundle already installed, exiting now\n", bundle);
```
phmccarty: There is no bundle variable in this function.

bryteise: Want to know how I know I didn't run make check... I clearly was in the middle of something else with this change X(.

bryteise:
Updated in #34

PR #34, opened 2016-04-08
Reduce noise for updates and bundle-adds

bryteise:
Fixes #33 use of an undefined variable, also corrects a bug in detecting already installed updates and adds a functional test to ensure that detecting already installed bundles is done correctly.

---

Stop printing a message when bundles are included and going to be
installed or if bundles are already included and going to be skipped
when doing the add_subscriptions call.

Instead update the install_bundles call to notify users when an
installed bundle was attempted to be be installed again.


src/bundle.c
```diff
@@ -348,7 +348,6 @@ int add_subscriptions(struct list *bundles, int current_version, struct manifest
 
 		file = search_bundle_in_manifest(mom, bundle);
 		if (!file) {
-			printf("%s bundle name is invalid, skipping it...\n", bundle);
```
phmccarty: I would prefer to keep this printf, since it is useful for 'bundle-add', and shouldn't ever occur for 'update'.

phmccarty: I will add this line back after merging.

phmccarty:
Merged

PR #35, opened 2016-04-11
add --statedir option to alternate state directory

fandeli:
This option overrides the global setup for state_dir
variable that is now used to find out where swupd
is going to do work (state dir).

If not --statedir option is passed along, the global
STATE_DIR definition value is going to be used instead,
this last one can also be changed at compilation time.

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -50,6 +50,7 @@ char *path_prefix = NULL; /* must always end in '/' */
 char *mounted_dirs = NULL;
 char *bundle_to_add = NULL;
 struct timeval start_time;
+char *state_dir = NULL;
 
 /* NOTE: Today the content and version server urls are the same in
  * all cases.  It is highly likely these will eventually differ, eg:
@@ -169,6 +170,29 @@ static bool is_valid_integer_format(char *str)
 	return true;
 }
 
+bool set_state_dir(char *path)
+{
+	if (path) {
+		if (path[0] != '/') {
+			printf("statepath must be a full path starting with '/', not '%c'\n", path[0]);
+			return false;
+		}
+
+		if (state_dir) {
+			free(state_dir);
+		}
+
+		string_or_die(&state_dir, "%s", path);
+	} else {
+		if (state_dir) {
+			return true;
+		}
+		string_or_die(&state_dir, "%s", STATE_DIR);
+	}
+
+	return true;
+}
+
 bool set_format_string(char *userinput)
 {
 	int ret;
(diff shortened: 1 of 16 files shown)


src/helpers.c
```diff
@@ -144,18 +144,18 @@ int create_required_dirs(void)
 	int ret = 0;
 	int i;
 	char *dir;
-#define STATE_DIR_COUNT 3
+#define state_dir_COUNT 3
```
phmccarty: Since this one is a local macro, I would prefer that it remain named STATE_DIR_COUNT

src/globals.c
```diff
@@ -212,13 +213,37 @@ bool set_format_string(char *userinput)
...
 
 	gettimeofday(&start_time, NULL);
 
+	if (state_dir == NULL) {
+		string_or_die(&state_dir, "%s", STATE_DIR);
+	}
+
 	/* insure path_prefix is absolute, at least '/', ends in '/',
 	 * and is a valid dir */
 	if (path_prefix != NULL) {
```
phmccarty: Latest code in master sets path_prefix in the set_path_prefix() function, so there are likely merge conflicts here. Can you rebase this patch?

src/helpers.c
fandeli: hey, I sed'd that as well?... good catch, let me fix. That one, not
intended to be changed.

src/globals.c
```diff
@@ -212,13 +213,37 @@ bool set_format_string(char *userinput)
 	return true;
 }
 
+bool set_state_dir(char *dir)
```
phmccarty: My preference is for set_state_dir() to handle the state_dir global analogous to the way set_format_string() handles format_string. This way, the interface that we use for handling globals will be more consistent.

In this case, I'm thinking init_globals() would call set_state_dir(NULL), which means that either the value passed to --statedir is set, or the default value (STATE_DIR). If the argument to set_state_dir() is not NULL, then we can do validation and set the user-defined value.

src/globals.c
```diff
@@ -212,13 +213,37 @@ bool set_format_string(char *userinput)
...
+{
+	if (dir == NULL || dir[0] == '\0') {
+		return false;
+	}
+
+	/* TODO: more validation of input? */
+	if (dir[0] != '/') {
```
phmccarty: Not really sure what extra validation we need here...

Restricting the value to a full path is good, and because we create the directory in create_required_dirs(), we don't need an existence check.

src/globals.c
fandeli: The patch handles stuff in the way you described, or maybe I missed something else?

phmccarty: In init_globals(), I would like to call set_state_dir(NULL), but that's not possible in the current state, because dir==NULL will return false. I'm suggesting to make NULL a special value, just like it is for the other set_* functions.

fandeli: all right

fandeli: done.

phmccarty:
Looks good to me. I'll merge shortly.

phmccarty:
Merged, thanks!

PR #36, opened 2016-04-12
Fix for manifest consolidation

phmccarty:
In the case where the same file is deleted in all bundles, but it was deleted in various versions for different bundles, the manifest consolidation step should choose the newest version among these deleted entries. More details on the justification are in the commit description.

Also, I added a functional test that covers the scenario where I ran into this bug.


bryteise:
+1

PR #38, opened 2016-04-12
Add support for verify --install -m latest

bryteise:
Instead of specifying a version number argument to -m allow passing the
latest string which will query for the latest version and use that as
the install target version.


phmccarty:
+1

PR #39, opened 2016-04-12
Consolidate initialization actions; store lock file in state directory

phmccarty:
To ease maintenance of swupd initialization steps, call all necessary functions for common actions in swupd_init().

Also, store the lock file in the state directory, now that the state directory is guaranteed to exist before creating the file. This also prevents multiple readers/writers for a given state directory.


bryteise:
+1  the consolidation was a little rough to follow with the free_globals moved from main.c to update.c but it everything seems right.

phmccarty:
Yes, sorry about that. I will split the moving of free_globals into a separate commit.

phmccarty:
I split that change out into a separate commit, and merged.

Issue #41, opened 2016-04-13
swupd download slow when using IMA


pohly:
swupd client 3.3.0 constantly opens and closes the file it is writing to during downloads (pack-from-0 in my test). That behavior incurs a huge (*) performance penalty when running on a device where IMA is active with on-device hashing, because each close() implies hashing the file content to update the security.ima xattr.

(*) a few seconds when using just curl vs. half an hour and still not done with swupd client.

swupd should open each file once and keep it open until the download is complete.

pohly:
I'm working on a patch...

PR #42, opened 2016-04-14
downloads: minimize syscalls to improve performance

pohly:
This fixes https://github.com/clearlinux/swupd-client/issues/41

Beware that I have only tested a slightly different patch against
2.87. Porting it to the current master code required some manual
conflict resolution and I haven't had the time to try the result.

### DIFF
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -84,6 +84,7 @@ struct file {
 	int last_change;
 	struct update_stat stat;
 
+	unsigned int fd_valid : 1;
 	unsigned int is_dir : 1;
 	unsigned int is_file : 1;
 	unsigned int is_link : 1;
@@ -104,6 +105,7 @@ struct file {
 
 	char *staging; /* output name used during download & staging */
 	CURL *curl;    /* curl handle if downloading */
+	int fd;        /* file written into during downloading, unset when fd_valid is false */
 };
 
 extern bool download_only;
@@ -192,6 +194,7 @@ extern void swupd_curl_set_current_version(int v);
 extern void swupd_curl_set_requested_version(int v);
 extern double swupd_query_url_content_size(char *url);
 extern size_t swupd_download_file(void *ptr, size_t size, size_t nmemb, void *userdata);
+extern CURLcode swupd_download_file_complete(CURLcode curl_ret, struct file *file);
 extern int swupd_curl_get_file(const char *url, char *filename, struct file *file,
 			       char *tmp_version, bool pack);
 #define SWUPD_CURL_LOW_SPEED_LIMIT 1
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -84,7 +84,6 @@ struct file {
 	int last_change;
 	struct update_stat stat;
 
-	unsigned int fd_valid : 1;
 	unsigned int is_dir : 1;
 	unsigned int is_file : 1;
 	unsigned int is_link : 1;
@@ -105,7 +104,7 @@ struct file {
 
 	char *staging; /* output name used during download & staging */
 	CURL *curl;    /* curl handle if downloading */
-	int fd;        /* file written into during downloading, unset when fd_valid is false */
+	FILE *fh;      /* file written into during downloading */
 };
 
 extern bool download_only;
@@ -193,7 +192,7 @@ extern void swupd_curl_cleanup(void);
 extern void swupd_curl_set_current_version(int v);
 extern void swupd_curl_set_requested_version(int v);
 extern double swupd_query_url_content_size(char *url);
-extern size_t swupd_download_file(void *ptr, size_t size, size_t nmemb, void *userdata);
+extern CURLcode swupd_download_file_start(struct file *file);
 extern CURLcode swupd_download_file_complete(CURLcode curl_ret, struct file *file);
 extern int swupd_curl_get_file(const char *url, char *filename, struct file *file,
 			       char *tmp_version, bool pack);
--- a/src/curl.c
+++ b/src/curl.c
@@ -169,35 +169,57 @@ size_t swupd_download_file(void *ptr, size_t size, size_t nmemb, void *userdata)
 	const char *outfile;
 	int fd;
 	FILE *f;
-	size_t written;
+	size_t written, remaining;
 
 	outfile = file->staging;
+	if (file->fd_valid) {
+		fd = file->fd;
+	} else {
+		fd = open(outfile, O_CREAT | O_RDWR | O_CLOEXEC | O_APPEND, 00600);
+		if (fd < 0) {
+			fprintf(stderr, "Cannot open file for write \\*outfile=\"%s\",strerror=\"%s\"*\\\n",
+				outfile, strerror(errno));
+			return -1;
+		}
+		file->fd = fd;
+		file->fd_valid = 1;
+	}
 
-	fd = open(outfile, O_CREAT | O_RDWR, 00600);
-	if (fd < 0) {
-		printf("Error: Cannot open %s for write: %s\n",
-		       outfile, strerror(errno));
+	/* handle short writes with repeated write() calls */
+	for (remaining = size * nmemb; remaining; remaining -= written) {
+		written = write(fd, ptr, size*nmemb);
+		if (written < 0) {
+			if (errno == EINTR) {
+				written = 0;
+				continue;
+			}
+			fprintf(stderr, "write error \\*outfile=\"%s\",strerror=\"%s\"*\\\n",
+				outfile, strerror(errno));
+			return -1;
+		}
+	}
+
+	if (fdatasync(fd)) {
+		fprintf(stderr, "fdatasync \\*outfile=\"%s\",strerror=\"%s\"*\\\n", outfile, strerror(errno));
 		return -1;
 	}
 
-	f = fdopen(fd, "a");
-	if (!f) {
-		printf("Error: Cannot fdopen %s for write: %s\n",
-		       outfile, strerror(errno));
-		close(fd);
-		return -1;
+	return size*nmemb;
+}
+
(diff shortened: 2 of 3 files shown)


src/curl.c
```diff
@@ -157,39 +157,58 @@ static size_t swupd_download_version_to_memory(void *ptr, size_t size, size_t nm
...
+		}
+		file->fd = fd;
+		file->fd_valid = 1;
+        }
+
+        /* handle short writes with repeated write() calls */
+        for (remaining = size*nmemb; remaining; remaining -= written) {
+		written = write(fd, ptr, size*nmemb);
+		if (written < 0) {
+			LOG_ERROR(file, "write error", class_file_io,
```
barbieri: write() can fail on `EINTR`, which is not a fatal error and you could try again

include/swupd.h
```diff
@@ -84,6 +84,7 @@ struct file {
 	int last_change;
 	struct update_stat stat;
 
+	unsigned int fd_valid : 1;
```
barbieri: no need for this, once you `close(fd)`, make the `fd = -1` and use that as `fd_valid`

pohly: The problem is that there is no explicit initializer function for the struct where one could set -1 reliably in all places; one can only rely on members being cleared (calloc). That's why "fd_valid" != 0 works, but "fd != -1" does not.

src/curl.c
pohly: True, need to add that.

include/swupd.h
barbieri: well.. then use zero. While zero is a valid fd, in this scenario it will never happen for the file you're writing. At least you have the one that is providing you with the data (socket or local file), also the swupd-client is not closing stdout/stderr, which are used to log and communicate with the user.

phmccarty: @pohly I am open to adding file initializer helper functions, for both a struct file on the stack, and dynamically allocated.

src/curl.c
```diff
@@ -157,39 +157,58 @@ static size_t swupd_download_version_to_memory(void *ptr, size_t size, size_t nm
...
-	struct file *file = (struct file *)userdata;
+	struct file *file = (struct file*)userdata;
 	const char *outfile;
 	int fd;
 	FILE *f;
-	size_t written;
+	size_t written, remaining;
 
 	outfile = file->staging;
+        if (file->fd_valid) {
```
phmccarty: The swupd codebase uses tabs+spaces for indentation.

src/curl.c
```diff
@@ -157,39 +157,58 @@ static size_t swupd_download_version_to_memory(void *ptr, size_t size, size_t nm
...
+		fd = file->fd;
+        } else {
+		fd = open(outfile, O_CREAT | O_RDWR | O_CLOEXEC | O_APPEND, 00600);
+		if (fd < 0) {
+			LOG_ERROR(file, "Cannot open file for write", class_file_io,
+				  "\\*outfile=\"%s\",strerror=\"%s\"*\\", outfile, strerror(errno));
```
phmccarty: LOG_* macros are not longer used, so this can just be a printf().

src/curl.c
```diff
@@ -157,39 +157,58 @@ static size_t swupd_download_version_to_memory(void *ptr, size_t size, size_t nm
 /* curl easy CURLOPT_WRITEFUNCTION callback */
 size_t swupd_download_file(void *ptr, size_t size, size_t nmemb, void *userdata)
 {
-	struct file *file = (struct file *)userdata;
+	struct file *file = (struct file*)userdata;
```
phmccarty: The space before * is the style enforced.

src/download.c
```diff
@@ -164,6 +164,7 @@ static void free_curl_list_data(void *data)
 {
 	struct file *file = (struct file *)data;
 	CURL *curl = file->curl;
+	swupd_download_file_complete(CURLE_OK, file);
```
phmccarty: The codebase is not consistent here, but I would like to see an explicit (void) cast to indicate an unused return value.

src/curl.c
```diff
@@ -227,7 +246,6 @@ int swupd_curl_get_file(const char *url, char *filename, struct file *file,
 			}
 		}
 		local->staging = filename;
-
```
phmccarty: extra line removed?

src/curl.c
```diff
@@ -157,39 +157,58 @@ static size_t swupd_download_version_to_memory(void *ptr, size_t size, size_t nm
...
+			LOG_ERROR(file, "Cannot open file for write", class_file_io,
+				  "\\*outfile=\"%s\",strerror=\"%s\"*\\", outfile, strerror(errno));
+			return -1;
+		}
+		file->fd = fd;
+		file->fd_valid = 1;
+        }
+
+        /* handle short writes with repeated write() calls */
+        for (remaining = size*nmemb; remaining; remaining -= written) {
```
phmccarty: For style, we're adding spaces before and after all arithmetic operators.

include/swupd.h
```diff
@@ -104,6 +105,7 @@ struct file {
 
 	char *staging; /* output name used during download & staging */
 	CURL *curl;    /* curl handle if downloading */
+	int fd;        /* file written into during downloading, unset when fd_valid is false */
```
phmccarty: I would prefer to use fd == -1 to indicate an invalid file descriptor, with the requisite addition of struct file initializers.

src/curl.c
barbieri: In soletta we provide a check-style script that calls uncrustify and if diff is non-0 we complain. Recommend doing so and using a semaphore-ci to ensure that on PR.

barbieri: https://github.com/solettaproject/soletta/blob/master/tools/check-style.py

bryteise:
How about an alternative that uses the default curl handler for CURLOPT_WRITEFUNCTION and open the FILE * prior to initiating the download?

phmccarty:
@bryteise That sounds reasonable, I think. So we only need to set CURLOPT_WRITEDATA to the FILE * we open for file->staging, because the default callback is fwrite()...

pohly:
+1 for using the default curl write handler. I didn't do that because I wanted to keep the existing semantic (= file only opened when transfer at least has started), but if that's not a problem, then it's certainly easier to open in advance.

Can you guys fix it that way? I'm a bit swamped right now.

bryteise:
So @phmccarty attempted the default curl write handler and because of how we queue downloads we end up with too many open file descriptors so it seems like @pohly has the best approach.

phmccarty:
@bryteise Right. I forgot to post the latest status here. The current async download code flow does not work well with the default write handler, so I would need to refactor so make it work.

pohly:
@rojkov was so kind to address the review feedback (https://github.com/pohly/swupd-client/pull/1).

As this is now seen as the right approach, I merged those fixes, squashed, rebased and force-pushed them in this PR. Please consider merging. No manual testing by myself, but @rojkov ran tests.

include/swupd.h
phmccarty: @pohly @rojkov Any thoughts on this approach? I really think fd initialized to -1 would be better than using the separate fd_valid field.

include/swupd.h
pohly: On Thu, 2016-04-21 at 14:19 -0700, Patrick McCarty wrote:
> In include/swupd.h:
> 
> > @@ -104,6 +105,7 @@ struct file {
> >  
> >  	char *staging; /* output name used during download & staging */
> >  	CURL *curl;    /* curl handle if downloading */
> > +	int fd;        /* file written into during downloading, unset when fd_valid is false */
> 
> I would prefer to use fd == -1 to indicate an invalid file descriptor,
> with the requisite addition of struct file initializers.

That would indeed be the cleaner solution, but introducing such an
initializer in all the right places IMHO goes beyond the scope of this
patch and should better be done by someone who knows the code.

So in other words, it would be good if someone could do that, but please
don't expect me or Dmitry to do it ;-}

phmccarty: That's fair. I will add it to my TODO list then.

src/curl.c
```diff
@@ -169,35 +169,57 @@ size_t swupd_download_file(void *ptr, size_t size, size_t nmemb, void *userdata)
...
+	size_t written, remaining;
 
 	outfile = file->staging;
+	if (file->fd_valid) {
+		fd = file->fd;
+	} else {
+		fd = open(outfile, O_CREAT | O_RDWR | O_CLOEXEC | O_APPEND, 00600);
+		if (fd < 0) {
+			fprintf(stderr, "Cannot open file for write \\*outfile=\"%s\",strerror=\"%s\"*\\\n",
+				outfile, strerror(errno));
```
phmccarty: When testing this patch for a large update of Clear Linux, I am running into this error condition repeatedly. The specific open(2) error is EMFILE (Too many open files).

I think there is a bug in the hysteresis logic where too many files are being added to the download queue and not being processed fast enough, thus reaching the max open fd limit when the files are open for the entirety of the download. I will debug this issue first.

pohly:
@phmccarty when testing this change in May, you ran into a "bug in the hysteresis logic" which you wanted to debug. Has that been fixed? Can this PR be merged now?

phmccarty:
@pohly The hysteresis logic fix is merged (see #133).

I would rather use the default curl write handler (fwrite(3)) when downloading to files, as discussed in previous comments on this PR. But I haven't yet investigated the proper fix to use the default handler everywhere in the code. Some refactoring of the download code may be needed as well.

pohly:
@phmccarty @tmarcu @bryteise: I've changed the code such that it uses the default curl write handler. I've intentionally not squashed or rebased, because the old commit was working and preserving it IMHO is useful.

Please review and merge.

phmccarty:
@pohly I agree that keeping the original commit is helpful.

+1 from me. Thanks!

PR #48, opened 2016-04-20
Cleanups

ikeydoherty:
Couple of quick cleanups rebased against the current master

### DIFF
--- a/include/swupd-error.h
+++ b/include/swupd-error.h
@@ -19,7 +19,7 @@
 #define EFULLDOWNLOAD 17	/* full_download problem */
 #define ENET404 404		/* download 404'd */
 #define EBUNDLE_INSTALL 18      /* Cannot install bundles */
-#define EREQUIRED_DIRS 19	/* Cannot create required dirs */
-#define ECURRENT_VERSION 20	/* Cannot determine current OS version */
+#define EREQUIRED_DIRS 19       /* Cannot create required dirs */
+#define ECURRENT_VERSION 20     /* Cannot determine current OS version */
 
 #endif
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -320,8 +320,8 @@ int add_subscriptions(struct list *bundles, int current_version, struct manifest
 	bool new_bundles = false;
 	char *bundle;
 	int ret;
-        int retries = 0;
-        int timeout = 10;
+	int retries = 0;
+	int timeout = 10;
 	struct file *file;
 	struct list *iter;
 	struct manifest *manifest;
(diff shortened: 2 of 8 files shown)


include/swupd-error.h
```diff
@@ -19,7 +19,7 @@
 #define EFULLDOWNLOAD 17	/* full_download problem */
 #define ENET404 404		/* download 404'd */
 #define EBUNDLE_INSTALL 18      /* Cannot install bundles */
-#define EREQUIRED_DIRS 19	/* Cannot create required dirs */
-#define ECURRENT_VERSION 20	/* Cannot determine current OS version */
+#define EREQUIRED_DIRS 19       /* Cannot create required dirs */
+#define ECURRENT_VERSION 20     /* Cannot determine current OS version */
```
phmccarty: I am really confused by clang-format's decision here... Maybe it's only looking at EBUNDLE_INSTALL line? Regardless, I would prefer that these lines be tab-aligned.

src/bundle.c
```diff
@@ -339,23 +339,23 @@ int add_subscriptions(struct list *bundles, int current_version, struct manifest
 			continue;
 		}
 
-retry_manifest_download:
+	retry_manifest_download:
```
phmccarty: I would prefer labels always start on column 1. I hope that the rest of the code agrees. :)

src/verify.c
```diff
@@ -788,7 +788,7 @@ int verify_main(int argc, char **argv)
 		ret = EXIT_FAILURE;
 	}
 
-	/* this concludes the critical section, after this point it's clean up time, the disk content is finished and final */
+/* this concludes the critical section, after this point it's clean up time, the disk content is finished and final */
```
phmccarty: I remember seeing this clang-format behavior before, and I think I worked around it by placing the comment *after* the label. Then clang-format didn't adjust indentation.

src/staging.c
```diff
@@ -148,8 +148,7 @@ int do_staging(struct file *file, struct manifest *MoM)
 			ret = -errno;
 			goto out;
 		}
-		string_or_die(&tarcommand, TAR_COMMAND " -C %s " TAR_PERM_ATTR_ARGS " -cf - '%s' 2> /dev/null | "
-			      TAR_COMMAND " -C %s%s " TAR_PERM_ATTR_ARGS " -xf - 2> /dev/null",
+		string_or_die(&tarcommand, TAR_COMMAND " -C %s " TAR_PERM_ATTR_ARGS " -cf - '%s' 2> /dev/null | " TAR_COMMAND " -C %s%s " TAR_PERM_ATTR_ARGS " -xf - 2> /dev/null",
```
phmccarty: This change is made (I think) because clang-format doesn't understand that TAR_COMMAND is a string literal. If it did, it would reindent to align the two TAR_COMMANDs here.

phmccarty:
+1 on the second patch

phmccarty:
I'm not incredibly happy with some of the style changes from clang-format, but I'll tolerate it for now.

Merged.

PR #49, opened 2016-04-20
Add symlink match to Search

bradTpeters:
Match search term against symlinks as well as files. This is
necessary in order to find hits for perfectly valid searches


phmccarty:
Looks good, will merge shortly.

phmccarty:
Merged.

PR #51, opened 2016-04-25
Fix manifest iteration

bradTpeters:
search_file_in_manifest() is not starting its iteration from
list_head, leading to incomplete manifest search and failing
verify_fix_path()

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -1170,7 +1170,7 @@ struct file *search_bundle_in_manifest(struct manifest *manifest, const char *bu
 	struct list *iter = NULL;
 	struct file *file;
 
-	iter = manifest->manifests;
+	iter = list_head(manifest->manifests);
 	while (iter) {
 		file = iter->data;
 		iter = iter->next;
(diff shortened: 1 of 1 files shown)


bryteise:
+1 though having a test that verifies the newly correct behavior (because it failed before this patch) would be an improvement.

src/manifest.c
```diff
@@ -1170,7 +1170,7 @@ struct file *search_bundle_in_manifest(struct manifest *manifest, const char *bu
 	struct list *iter = NULL;
 	struct file *file;
 
-	iter = manifest->manifests;
+	iter = list_head(manifest->manifests);
```
phmccarty: I agree that the list_head() is needed here, but this function is not called by verify_fix_path()... Can you update the commit description to reflect that this is fixing a general issue, not just for verify_fix_path()?

bradTpeters: Revised commit msg, discussing search_bundle_in_manifest() separately

phmccarty:
Looks good. I will follow up with a functional test.

phmccarty:
Merged

PR #52, opened 2016-04-25
Fix deletion of directory trees with 'verify --fix'

phmccarty:
For safety, 'verify --fix' does not use swupd_rm() (like 'update') when
deleting files, etc. Instead, each deleted entry is considered
individually.

So for the deletion step to work correctly with 'verify --fix', files
contained within directories must be listed before the directories
themselves. A simple solution is to make the manifest reverse filename
sorted (opposite of the current filename sorted order).

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -76,6 +76,21 @@ int file_sort_filename(const void *a, const void *b)
 	return 0;
 }
 
+int file_sort_filename_reverse(const void *a, const void *b)
+{
+	struct file *A, *B;
+	int ret;
+	A = (struct file *)a;
+	B = (struct file *)b;
+
+	ret = strcmp(A->filename, B->filename);
+	if (ret) {
+		return -ret;
+	}
+
+	return 0;
+}
+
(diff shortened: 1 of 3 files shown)


phmccarty:
I will follow up later with a functional test.

src/manifest.c
```diff
@@ -76,6 +76,21 @@ int file_sort_filename(const void *a, const void *b)
...
+	int ret;
+	A = (struct file *)a;
+	B = (struct file *)b;
+
+	ret = strcmp(A->filename, B->filename);
+	if (ret) {
+		return -ret;
+	}
+
+	return 0;
```
bryteise: Probably can just return -ret always as -0 is fine.

bryteise:
Looks good.

src/manifest.c
phmccarty: I will make this change, thanks.

phmccarty:
Merged. I addressed William's comment, squashing that change into the original commit. And I added a functional test.

PR #53, opened 2016-04-26
Add DBus wrapper around swupd client

rojkov:
Simple daemon offering a system-bus D-Bus interface for
asynchronously triggering/executing Clear Linux swupd-client
utility from another process.

The D-Bus API reflects functionality offered by command line
interface of the Clear linux swupd-client utility (`swupd`).

Functionality is needed by Ostro OS architecture where it must
be possible to trigger system software update actions from a
containerized process. Thus, direct fork() & exec() doesn’t work.

Since D-Bus access is enabled by enabling access to a single
socket, this can be arranged for a containerized process.

This patch adds two binaries: swupdd and swupdctl. The former
is a D-Bus service wrapping `swupd`. It starts on demand
and exits after timeout (30 sec). The latter is a drop-in
replacement for `swupd` that communicates with the D-Bus
wrapper via D-Bus.


data/org.O1.swupdd.conf
```diff
@@ -0,0 +1,23 @@
...
+
+  <!-- This configuration file specifies the required security policies
+       for the swupd to work. -->
+
+  <!-- Only user root can own the fwupd service -->
+  <policy user="root">
+    <allow own="org.O1.swupdd.Client"/>
+  </policy>
+
+ <!-- Allow anyone to call into the service - we'll reject callers using PolicyKit -->
```
barbieri: is this the case? If not, remove the comment and I'd allow only user root and maybe wheel/admin group

dbus-wrapper/list.c
```diff
@@ -0,0 +1,287 @@
...
+
+#define _GNU_SOURCE
+
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "list.h"
+
+static struct list *list_append_item(struct list *list, struct list *item)
```
barbieri: can't we use the list that already exist in the repo?

barbieri:
looks good to me, can you also work on integrating this with their unit tests? The best case is to use their existing tests with out wrapper, once it passes we can propose:
 - hide existing `swupd` from `$PATH`;
 - rename/symlink `swupdctl` to `swupd` in `$PATH` (drop-in replacement!)

data/org.O1.swupdd.conf
rojkov: fixed. Only root user, wheel and adm groups are allowed to use the service now,

dbus-wrapper/list.c
rojkov: fixed. reworked to use the existing list code.

rojkov:
I've added a couple of squashable commits addressing the comments.

I haven't tried to run the functional tests in Ostro yet. Hopefully I'll have time for that on Friday or next week.

barbieri:
@rojkov now that ostro 1.0 is out, can you get to the tests? 

IMO we can also start to apply this PR to Ostro nightly builds if you're comfortable with that, make sure to hide the swupd and exposing the dbus client in its place so our QA tests that.

rojkov:
@barbieri I got reallocated to another project. @igor-stoppa is going to continue on this.

I think we need to upgrade swupd to the latest&greatest in Ostro first (it's v2.87 currently). Then this patch can be applied for the nightly builds. And before upgrading swupd in Ostro we'd better have https://github.com/clearlinux/swupd-client/pull/42 merged upstream.

Yesterday I tried to run the functional tests on our target device: the problem is that the tests are supposed to be run on a host system from a build directory which is not really suitable for embedded development. After tweaking `tests/functional/swupdlib.bash` I managed to run the tests against the D-Bus wrapper `swupdctl` and got 23/49 pass rate. If run against the actual executable `swupd` the pass rate is 24/49. Again such low pass rate is due to targeting a normal host system where e.g. bashisms and normal coreutils are allowed.

barbieri:
Do you execute the tests using a local DBus? You should play with dbus-daemon launch and export the variables so you don't reach the actual system bus, rather going with the local dbus-daemon for the test. It can use a specific dbus & policy configuration files to allow to run as non-root if that's the requirement.

Also, please run these tests in the expected environment, maybe get clearlinux and do it there. @phmccarty what do you recommend to get 100% pass rate?

bryteise:
@barbieri @igor-stoppa ClearLinux would be easiest though installing the dependencies wouldn't be too hard for any distro. Just our version of bsdiff (https://github.com/clearlinux/bsdiff) installed manually and everything else should be in any reasonable distro.

Unfortunately the tests require root at the moment due to tar xattr extraction (sudo is used in the tests that need them).

Makefile.am
```diff
@@ -46,6 +46,13 @@ SWUPD_COMMON_SOURCES = \
 	src/scripts.c \
 	src/bundle.c
 
+SUBDIRS = dbus-wrapper
```
bryteise: So I don't love subdirs and we haven't started to use them until this point but I can understand the reasoning for doing it this way. Overall I'd still prefer to avoid adding subdirs for one piece at this point but I could be convinced otherwise.

bryteise:
Just a comment before merging, that this needs to be given a run through clang-format.

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+		sd_bus_error_set_errnof(error, -r, "Can't read value of '%s'", optname);
+		return r;
+	}
+	r = sd_bus_message_exit_container(m);
+	if (r < 0) {
+		sd_bus_error_set_errnof(error, -r, "Can't exit variant container of '%s'", optname);
+		return r;
+	}
+
+	r = asprintf(&option, "--%s", optname);
```
bryteise: Do you really want to be handling the ENOMEM case? It isn't something we tend to do elsewhere (mostly just abort in string_or_die).

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+	if (r < 0) {
+		sd_bus_error_set_errnof(error, ENOMEM, "Can't allocate memory for '%s'", optname);
+		return -ENOMEM;
+	}
+
+	*strlist = list_append_data(*strlist, option);
+	*strlist = list_append_data(*strlist, strdup(value));
```
bryteise: In fact these could very well fail due to memory allocation so it isn't consistent within the code base. Prefer the use of string_or_die over asprintf.

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+	r = sd_bus_message_exit_container(m);
+	if (r < 0) {
+		sd_bus_error_set_errnof(error, -r, "Can't exit variant container of '%s'", optname);
+		return r;
+	}
+
+	if (value) {
+		char *option = NULL;
+
+		r = asprintf(&option, "--%s", optname);
```
bryteise: Similarly here, replace with string_or_die.

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+		sd_bus_error_set_errnof(error, -r, "Can't read value of '%s'", optname);
+		return r;
+	}
+	r = sd_bus_message_exit_container(m);
+	if (r < 0) {
+		sd_bus_error_set_errnof(error, -r, "Can't exit variant container of '%s'", optname);
+		return r;
+	}
+
+	if (asprintf(&option, "--%s", optname) < 0) {
```
bryteise: string_or_die

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+		sd_bus_error_set_errnof(error, ENOMEM, "Can't allocate memory for '%s'", optname);
+		return -ENOMEM;
+	}
+	*strlist = list_append_data(*strlist, option);
+
+	if (asprintf(&value_str, "%i", value) < 0) {
```
bryteise: string_or_die

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+				 * requests are still processed before we really exit. */
+
+				char *match = NULL;
+				const char *unique;
+				r = sd_bus_get_unique_name(bus, &unique);
+				if (r < 0) {
+					ERR("Failed to get unique D-Bus name: %s", strerror(-r));
+					return r;
+				}
+				r = asprintf(&match,
```
bryteise: string_or_die

dbus-wrapper/dbus_client.c
```diff
@@ -0,0 +1,328 @@
...
+} command_ctx_t;
+
+static int on_child_output_received(sd_bus_message *message, void *userdata, sd_bus_error *error)
+{
+	const char *output;
+	int r;
+
+	r  = sd_bus_message_read(message, "s", &output);
+	if (r < 0) {
+		ERR("Can't read client's output: %s", strerror(-r));
```
bryteise: ERR is not something used in swupd_client.

dbus-wrapper/dbus_client.c
```diff
@@ -0,0 +1,328 @@
...
+	}
+
+	/* TODO: do we really need return_value? */
+	r = sd_bus_message_read(reply, "b", &return_value);
+	if (r < 0) {
+		ERR("Failed to parse response message: %s", strerror(-r));
+		goto finish;
+	}
+
+	if (!return_value) {
```
bryteise: I'm a little confused here, the return value corresponds to the dbus return value? So the read was able to work fine but the dbus call had some sort of issue indicated by the return_value?

dbus-wrapper/dbus_client.c
```diff
@@ -0,0 +1,328 @@
...
+	}
+
+	r = sd_bus_message_read(reply, "b", &success);
+	if (r < 0) {
+		ERR("Failed to parse reply: %s", strerror(-r));
+		goto finish;
+	}
+finish:
+	if (r < 0 || !success) {
+		ERR("Failed to cancel. Not running?");
```
bryteise: So the only purpose of this block seems like to check and see if abort() is needed because we've lost the daemon service. Could it not exit normally in this case (I guess similar for the above) with just a print statement saying it couldn't talk with the daemon? (no need for all the "failed to get event loop" messages which aren't likely helpful for end users).

dbus-wrapper/dbus_client.c
```diff
@@ -0,0 +1,328 @@
...
+	r = sd_bus_add_match(bus,
+			     NULL, /* bus slot */
+			     "type='signal',"
+			     "interface='org.O1.swupdd.Client',"
+			     "member='RequestCompleted',"
+			     "path='/org/O1/swupdd/Client'",
+			     on_request_completed,
+			     event /* user data */);
+	if (r < 0) {
+		ERR("Failed to add handler for ChildOutputReceived signal: %s", strerror(-r));
```
bryteise: Should be RequestCompleted I think.

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+	sd_bus_slot *slot = NULL;
+	sd_event *event = NULL;
+	sigset_t ss;
+	int r;
+
+	memset(&context, 0x00, sizeof(daemon_state_t));
+
+        r = sd_event_default(&event);
+        if (r < 0) {
+                ERR("Failed to allocate event loop: %s", strerror(-r));
```
bryteise: Again ERR should probably go away.

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+		r = sd_event_get_state(event);
+		if (r < 0) {
+			ERR("Failed to get event loop's state: %s", strerror(-r));
+			return r;
+		}
+		if (r == SD_EVENT_FINISHED) {
+			break;
+		}
+
+		r = sd_event_run(event, exiting ? (uint64_t) -1 : (uint64_t) TIMEOUT_EXIT_SEC * 1000000);
```
bryteise: I'm not seeing it documented well but what does sd_event_run return when it hits the timeout?

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+	r = sd_bus_message_read(m, "b", &force);
+	if (r < 0) {
+		sd_bus_error_set_errnof(ret_error, -r, "Can't read 'force' option");
+		return r;
+	}
+
+	if (force) {
+		kill(child, SIGKILL);
+	} else {
+		kill(child, SIGTERM);
```
bryteise: This says swupd-client should probably start handling SIGTERM to be a little more graceful in shutdown.

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+			r = bus_message_read_option_bool(m, argname, strlist, error);
+			if (r < 0) {
+				return r;
+			}
+		} else if (is_in_array(argname, opts_int)) {
+			r = bus_message_read_option_int(m, argname, strlist, error);
+			if (r < 0) {
+				return r;
+			}
+		} else {
```
bryteise: What is this else case handling exactly? Should this be treated as an input error instead of skipped?

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+	int status = 0;
+	int r;
+
+	method_t child_method = context->method;
+	assert(child_method);
+
+	if (si->ssi_code == CLD_EXITED) {
+		status = si->ssi_status;
+	} else {
+		DEBUG("Child process was killed");
```
bryteise: Don't want to print in a signal handler.

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+
+	context->child = 0;
+	context->method = METHOD_NOTSET;
+
+	r = sd_bus_emit_signal(context->bus,
+			       "/org/O1/swupdd/Client",
+			       "org.O1.swupdd.Client",
+			       "RequestCompleted", "si", _method_str_map[child_method], status);
+	if (r < 0) {
+		ERR("Can't emit D-Bus signal: %s", strerror(-r));
```
bryteise: Same.

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+
+	/* Reap the zomby */
+	assert(si->ssi_pid == context->child);
+	pid_t ws = waitpid(si->ssi_pid, &child_exit_status, 0);
+	assert(ws != -1);
+
+	context->child = 0;
+	context->method = METHOD_NOTSET;
+
+	r = sd_bus_emit_signal(context->bus,
```
bryteise: I didn't verify this fully but from my brief investigation this looks to malloc and is very much not safe in a signal handler.

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+	}
+
+finish:
+	/* No more events for this handler are expected */
+	close(fd);
+	sd_event_source_unref(s);
+	return r;
+}
+
+static int on_child_exit(sd_event_source *s, const struct signalfd_siginfo *si, void *userdata)
```
bryteise: This might be something you could use the https://www.freedesktop.org/software/systemd/man/sd_event_add_child.html instead of using a signal handler to avoid having to otherwise handle this case.

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+	int r;
+
+	fdm = posix_openpt(O_RDONLY|O_NOCTTY);
+	if (fdm < 0) {
+		ERR("Can't open pseudo terminal: %s", strerror(errno));
+		return -errno;
+	}
+	r = grantpt(fdm);
+	if (r != 0) {
+		ERR("Can't grant pseudo terminal: %s", strerror(errno));
```
bryteise: Should close the open fd I think for these errors.

dbus-wrapper/swupdd-main.c
```diff
@@ -0,0 +1,985 @@
...
+		r = sd_bus_emit_signal(context->bus,
+				       "/org/O1/swupdd/Client",
+				       "org.O1.swupdd.Client",
+				       "ChildOutputReceived", "s", buffer);
+		if (r < 0) {
+			ERR("Failed to emit signal: %s", strerror(-r));
+		}
+
+		return 0;
+	} else if (count < 0) {
```
bryteise: Does count == 0 make sense for this and it should be handled as just closing the fd?

dbus-wrapper/swupdd-main.c
rojkov: True, I'll fix that.
Finally I've got some time to address the comments.

dbus-wrapper/dbus_client.c
rojkov: Yet I don't like using pure printf()s for error reporting, because in a bigger code base grepping for the error message may result in multiple findings. It's easier to modify the macro to include file name and line number then the failure point becomes obvious.

I'll remove file and line number from the macro.

dbus-wrapper/dbus_client.c
rojkov: Originally the boolean return values were in the D-BUS API spec which I used without modifications (see arguments `result` in https://github.com/ostroproject/swupd-daemon/blob/master/interfaces/org.O1.swupdd.Client.xml). This looks redundant indeed because errors can be communicated with D-Bus errors. That's why I've added the TODO comment.

dbus-wrapper/dbus_client.c
rojkov: Probably we could just abruptly abort the program in this case. But what if we add on-exit hooks at some point meaning to free allocated resources gracefully before exit. Also I think it's a good practice to exit the event loop with sd_event_exit() because it gives a chance to queued events to be handled before exit.
And if for some reason the event loop gets inaccessible don't we want to let the user know that's something is wrong and probably needs debugging?

dbus-wrapper/dbus_client.c
rojkov: Yep, will fix that.

dbus-wrapper/swupdd-main.c
rojkov: From https://www.freedesktop.org/software/systemd/man/sd_event_run.html#
"`sd_event_run()` returns a positive, non-zero integer if an event source was dispatched, and zero when the specified timeout hit before an event source has seen any event, and hence no event source was dispatched."

dbus-wrapper/swupdd-main.c
rojkov: Makes sense. Will report EINVAL instead of silent skipping.

dbus-wrapper/swupdd-main.c
rojkov: It's an event handler, not a signal handler.

dbus-wrapper/swupdd-main.c
rojkov: The initial implementation used `sd_event_add_child()`. IIRC this didn't generate events in case the children was killed and the event source for signals was more reliable.

dbus-wrapper/swupdd-main.c
rojkov: Good point. Will fix

dbus-wrapper/swupdd-main.c
rojkov: Yep, `count == 0` would be an error. Will add error message.

rojkov:
* fixed typos;
* used `string_or_die()` instead of `aprintf()`;
* stricter policy on unrecognized options in the daemon;
* added closing file descriptors upon errors;
* added error notification upon empty input in EPOLLIN;
* ran the code through `clang-format`;
* got rid of debugging info in output;
* rebased and force pushed.

bryteise:
So this looks pretty good to me.

@phmccarty @tmarcu @bradTpeters Could we get another look over, I'm interested on further opinions on dbus-wrapper/dbus_client.c:on_command_cancel function's error handling but otherwise I don't have any additional comments that should be addressed.

andred:
It's a pity that this makes the client dependent on systemd. It's not possible to use systemd in every environment, and so far swupd works fine without systemd (with one minor change to update_triggers()). I haven't looked at the patches, but is it really impossible to achieve what is being achieved here by using libdbus directly, or maybe glib?

barbieri:
this doesn't require systemd per se, just the library it provides -- which is more efficient and smaller than libdbus, that's why it's being used by other projects.

If a real situation where we  need the dbus wrapper and not installing systemd, we can create a static library to link against and live with that.

andred:
I wasn't clear enough. But the whole systemd project (including libsystemd.so) doesn't even compile against anything but glibc, in particular the musl c library, without numerous patches (none of which seem to be going to be accepted upstream by either party). Given embedded devices often use something else than glibc, it'd be a shame if swupd locked itself out by depending on it. I realise this place here is probably not the right to have an architectural discussion like this, and swupd is (mainly / only?) targeting Clearlinux with its own use-cases, and swupd shouldn't be limited because some other, unrelated use-case might break, but swupd seemed like a good choice for system updates of embedded devices so far. That is my main concern.

One last - even if libsystemd's dbus wrapper is more efficient, does it matter? Are there that many messages being sent that it actually matters?

Would a patch that allows selectively building against libsystemd or libdbus or libglib-dbus be acceptable?

barbieri:
indeed there are not many messages, my point is that due that reason libdbus is being deprecated in favor of libsystemd to reach DBus... at least for projects that rely on an external library. AFAIU Qt and Glib talk directly to the bus.

libglib-dbus is horrible to use, I've used it in few projects and it's the worst DBus bind I ever used. If we need to go small, use libell https://git.kernel.org/cgit/libs/ell/ell.git/ it was created by connman/ofono/... team after long time experience with glib-dbus, which lead them to create their own dbus wrapper with glib... but then they just used a subset and libell was born to be just that subset.

tmarcu:
Please open another issue if there is proposed work going forward with this feature.

PR #54, opened 2016-04-26
Signature verification

ocallesp:
need code review


src/signature.c
```diff
@@ -54,111 +54,289 @@
...
+		EVP_PKEY_free(pkey);
+		X509_free(cert);
 		initialized = false;
 	}
 }
 
+bool get_certificates(void)
+{
+	/* TODO - get paths and certificate names from Linux environment */
+
```
bradTpeters: Todo? What's the plan here?

ocallesp: are going to read a Linux environment variable to get the directory and certificate names o
set as a constant in source code?

src/clr_bundle_add.c
```diff
@@ -53,6 +53,7 @@ static void print_help(const char *name)
 	printf("   -l, --list              List all available bundles for the current version of Clear Linux\n");
 	printf("   -x, --force             Attempt to proceed even if non-critical errors found\n");
 	printf("   -S, --statedir          Specify alternate swupd state directory\n");
+    printf("   -V, --verify-sign       Enable signature verification, default value is disabled\n");
```
elmatterino: whitespace mismatch with rest of file

src/clr_bundle_add.c
```diff
@@ -67,14 +68,15 @@ static const struct option prog_opts[] = {
 	{ "format", required_argument, 0, 'F' },
 	{ "force", no_argument, 0, 'x' },
 	{ "statedir", required_argument, 0, 'S' },
+    { "verify-sign", no_argument, 0, 'V' },
```
elmatterino: whitespace mismatch with rest of file

src/verify.c
```diff
@@ -68,6 +68,7 @@ static const struct option prog_opts[] = {
 	{ "quick", no_argument, 0, 'q' },
 	{ "force", no_argument, 0, 'x' },
 	{ "statedir", required_argument, 0, 'S' },
+	{ "verify-sign", no_argument, 0, 'V' },
```
elmatterino: nitpick on grammar. proper term would be "verify-sig" or "verify-sigs" if multiple possible.

src/signature.c
```diff
@@ -46,115 +46,301 @@
...
+        fprintf(stderr, "Failed fopen %s\n", data_filename);
+        goto error;
+    }
+
+    result = validate_signature(fp_data, fp_sig);
+    fclose(fp_data);
+    fclose(fp_sig);
+
+    return result;
+
```
elmatterino: remove 142-154 and let flow fall through error cleanup. they are redundant with the error cleanup.

src/clr_bundle_add.c
elmatterino: see later grammar comment.

src/signature.c
```diff
@@ -46,115 +46,301 @@
 
 #warning "TODO pick signing scheme"
 #if defined(SWUPD_LINUX_ROOTFS)
-#define IMPL IMPL_FAKE
+#define IMPL IMPL_REAL
```
elmatterino: Why would this be anything other than IMPL_REAL (or an option at all?) Isn't checking the signatures controlled by the command line option? An incorrect build config could easily cause these checks to be bypassed, while telling the calling script that everything is good.

src/signature.c
```diff
@@ -46,115 +46,301 @@
...
 
-static X509_STORE *x509_store = NULL;
 
 bool signature_initialize(const char *ca_cert_filename)
 {
 	if (initialized) {
 		return true;
 	}
+
 	OpenSSL_add_all_algorithms();
```
elmatterino: Unless you want to accept signatures that use broken hash algorithms -- e.g. MD5 and SHA1 -- it is prudent to be more specific about what you will accept.

src/verify.c
ocallesp: signatures would be more appropriate since swupd-client will check all signatures for each file.

src/signature.c
ocallesp: this is for making coverity happy

src/signature.c
ocallesp: will remove the IMPL_FAKE ...it will be dead code

src/signature.c
ocallesp: we can use EVP_add_cipher to add specific ciphers, the algorithm used to encrypt the private key is need for functions like PEM_read_PrivateKey

src/signature.c
phmccarty: I would like the certs to live in a directory under /usr/share, perhaps in the same directory as the SSL certs. Though that directory name should not be Clear Linux specific, so we can rename it to /usr/share/swupd/certs, or similar.

src/verify.c
phmccarty: "verify-signatures" would also be okay with me. Can you add a configure option to enable the signature verification code at build time? So, if a distro wants to always verify signatures, they can update packaging files and not have to pass the -V option to swupd.

src/signature.c
```diff
@@ -54,111 +54,289 @@
...
+	return false;
+}
+
 bool signature_verify(const char *data_filename, const char *sig_filename)
 {
-	BIO *bio_data = NULL;
-	BIO *bio_sig = NULL;
-	PKCS7 *pkcs7 = NULL;
-	int ret;
+    FILE *fp_data = NULL;
```
phmccarty: The indentation is off in this function, and in the other new functions below. Can you adjust to uses tabs+spaces, which is the swupd style convention?

src/manifest.c
```diff
@@ -518,9 +518,11 @@ static int retrieve_manifests(int current, int version, char *component, struct
...
 	}
 
-	if (!signature_download_and_verify(url, filename)) {
-		unlink(filename);
-		goto out;
+	if (signing_enabled) {
+		if (!signature_download_and_verify(url, filename)) {
+			unlink(filename);
+			goto out;
+		}
```
phmccarty: Though the signature-related function calls sites have not been updated yet, the only Manifest.MoM.tar will be signed (out of the manifest tar currently published). Can you add an extra check to ensure is the MoM manifest?

src/signature.c
```diff
@@ -141,12 +139,18 @@ bool signature_verify(const char *data_filename, const char *sig_filename)
...
 
 error:
-    fclose(fp_data);
-    fclose(fp_sig);
+	if (fp_data) {
+		fclose(fp_data);
+	}
+    if (fp_sig) {
+    	fclose(fp_sig);
+    }
```
phmccarty: tabs+spaces here too

phmccarty:
This PR has been superceded by #83

PR #55, opened 2016-04-26
fix error checking/retry handling on pack download

fandeli:
In the update loop when downloading packs a return
status is checked with an unreacheable error number
-ENONET, this is because the function is called
download_subscribed_packs() without forcing the
pack to download, so the only chance to this to
return an error is at checking network aviability
with a return code -ENOSWUPDSERVER so far. This
was causing the retry mechanism on this point to
be completely ignored and skipped right away.

Beyond checking for -ENOSWUPDSERVER, this fix
checks whether *any* error ocurred in the packs
download function, granting that if some code
changes inside the caller is checking accorrdingly.

### DIFF
--- a/src/update.c
+++ b/src/update.c
@@ -356,7 +356,7 @@ load_server_manifests:
 download_packs:
 	/* Step 5: get the packs and untar */
 	ret = download_subscribed_packs(false);
-	if (ret == -ENONET) {
+	if (ret) {
 		// packs don't always exist, tolerate that but not ENONET
 		if (retries < MAX_TRIES) {
 			increment_retries(&retries, &timeout);


bryteise:
I am for this change. Having packs hold up progress for any reason is sensible since it would still allow much smaller/faster updates if a network blip happens.

We will need to improve the download_subscribed_packs code to make the most of this change though as any tar issues aren't handled very well making this more trouble in the short term.

phmccarty:
@bryteise I agree that the error handling in the pack handling logic needs improvement, but because this change is only for the 'update' path, where packs are always optional, I think we can address that pack untar errors, etc. later.

phmccarty:
This change looks fine. I will merge shortly.

phmccarty:
Merged

PR #56, opened 2016-04-26
Fix check-update content url message

chuyd:
Fix "Use the -c option instead" message in check-update since it does
not require the -c option, this occurs when the default contenturl
can not be foud in the config files, this patch sets the content url
with the version url value in check-update.

Signed-off-by: Jesus Ornelas Aguayo <jesus.ornelas.aguayo@intel.com>

### DIFF
--- a/src/check_update.c
+++ b/src/check_update.c
@@ -81,6 +81,7 @@ static bool parse_options(int argc, char **argv)
 				printf("Invalid --versionurl argument\n\n");
 				goto err;
 			}
+			set_content_url(optarg);
 			set_version_url(optarg);
 			break;
 		case 'P':


phmccarty:
Thanks for catching this issue! It is a side effect of running init_globals() for 'check-update', even though 'check-update' doesn't require set_content_url() being called at all... I'm not sure of the best approach to take here...

bryteise:
I'm actually in favor of having it setting the content url to something static that passes init_globals for check-update and then removing the -u option from the command.

But... if we want to have unified option handling between more swupd commands I'd just reject the -u option entirely and if only a content or version url is passed print a warning but attempt to use the same url for both.

fandeli:
+1 removing -u from command and set both to whatever -c is, is harder to
handle decently from init_globals()

phmccarty:
I'm hesitant to remove the -u option right now, or else scripts will break. When the unified getopt handling is implemented, the options will likely break anyway, so I would prefer to not break compatibility before that time when we can announce the breaking option-related changes and how to migrate scripts.

I think we should call set_content_url("") for now, and if check-update ever needs to download update content in the future, we can add the -c option.

phmccarty:
@chuyd After reconsidering the various possible solutions here, given the current state of option parsing, I think your approach is cleanest.

The future unified getopt interface will completely overhaul the parsing, and we can revisit the check-update options at that time.

phmccarty:
Merged

PR #57, opened 2016-04-26
Add check for do_staging return code

bradTpeters:
Add check for do_staging return code, aborting bundle-add
operation when any file or directory fails to stage. This
addresses file system corruption seen when client runs out
of disk space

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -444,7 +444,8 @@ int install_bundles(struct list *bundles, int current_version, struct manifest *
 		if (ret == 0) {
 			rename_staged_file_to_final(file);
 		} else {
-			printf("Failed to stage file(s) (ret = %d). Cannot proceed with bundle-add \n", ret);
+			printf("Failed to stage file: %s (ret = %d). Aborting bundle-add\n",
+                    file->filename, ret);
 			goto out;
 		}
 	}
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -443,6 +443,9 @@ int install_bundles(struct list *bundles, int current_version, struct manifest *
 		ret = do_staging(file, mom);
 		if (ret == 0) {
 			rename_staged_file_to_final(file);
+		} else {
+			printf("Failed to stage file(s) (ret = %d). Cannot proceed with bundle-add \n", ret);
+			goto out;
 		}
 	}
 


bradTpeters:
A test run with this patch fails with:
----
# swupd bundle-add os-core-dev
swupd-client bundle adder 3.5.2
   Copyright (C) 2012-2016 Intel Corporation

Downloading required packs...
Downloading os-core-dev pack for version 7500
Extracting pack.
Installing bundle(s) files...
Failed to stage file(s) (ret = -2). Cannot proceed with bundle-add 
Possible filedescriptor leak: fd_number="4",fd_details="socket:[5418]"
root@clr-bf85f5c2c3964f7fb7eff8072655dde2 /home #

----

Much better than corrupting the file system by installing a bunch of 0 length files, as it was without this patch.

bradTpeters:
A follow-on patch is going to clean up the partially installed bundle(s) contents

src/bundle.c
```diff
@@ -443,6 +443,9 @@ int install_bundles(struct list *bundles, int current_version, struct manifest *
 		ret = do_staging(file, mom);
 		if (ret == 0) {
 			rename_staged_file_to_final(file);
+		} else {
+			printf("Failed to stage file(s) (ret = %d). Cannot proceed with bundle-add \n", ret);
```
phmccarty: Since only a single file is being processed at this point, can we print out the file->filename that failed to stage? This is similar to the do_staging() error handling in the update path.

bryteise:
+1

bradTpeters:
Msg updated. I'm a little concerned that outputting filename may imply the cause of the error is the file mentioned, when in fact it is likely to be a disk out of space error

phmccarty:
@bradTpeters I agree with that concern, but do_staging() error handling needs rework anyway, so at this point in the code we don't have enough information about what errors actually occurred.

But by printing the filename that failed to stage, swupd developers will at least have some information to work with, until we start to send this type of information over telemetry. The message is not actionable for end users, which is the downside, but if users file a bug report, when there may be enough information in the output to work with.

phmccarty:
Looks good. I'll fix up a little whitespace issue afterward and then merge.

phmccarty:
Merged. (I squashed the whitespace change into the commit instead.)

PR #58, opened 2016-04-26
Move functional tests to use the BATS framework

bryteise:
This change also removes duplicate test from search.


phmccarty:
I'm seeing three test failures when running the suite on Arch. They are:

- update/include-old-bundle
- update/include
- update/newest-deleted

There appear to be differences in ordering of pack downloads? Maybe we need to sort subs before downloading?

phmccarty:
I'll merge this now and we can debug the pack download ordering issue after.

PR #59, opened 2016-04-27
fix to check whether resume download is supported

fandeli:
When pack is going to be downloaded an option for
curl called resume is activated no matter what,
however although major http server support 'Range'
command that enables this functionality, not all
complies with this, when swupd client tries to
re-download a partial file from one of the server
that does not support 'Range' it will just throw
an error and stop to work until partial download
is deleted by hand.

This patch address that issue by checking first
whether server supports 'Range' command and if not
then disable 'resume' of packs.

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -50,6 +50,7 @@ static CURL *curl = NULL;
 
 static int curr_version = -1;
 static int req_version = -1;
+static bool resume_download_enabled = true;
 
 int swupd_curl_init(void)
 {
@@ -231,7 +232,7 @@ int swupd_curl_get_file(const char *url, char *filename, struct file *file,
 		local->staging = filename;
 
 		if (lstat(filename, &stat) == 0) {
-			if (pack) {
+			if (pack && resume_download_enabled) {
 				curl_ret = curl_easy_setopt(curl, CURLOPT_RESUME_FROM_LARGE, (curl_off_t)stat.st_size);
 			} else {
 				unlink(filename);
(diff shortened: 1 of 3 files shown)


src/curl.c
```diff
@@ -363,6 +364,62 @@ int swupd_curl_get_file(const char *url, char *filename, struct file *file,
...
+{
+#define RESUME_BYTE_RANGE 2
+
+	CURLcode curl_ret;
+	char *version_string = NULL;
+	char *url = NULL;
+
+	if (!curl) {
+		abort();
+	}
```
phmccarty: I would prefer to do "assert(curl)" here instead, since it indicates a programming error if curl is not initialized at this point in the code.

But, the code is not consistent in this regard, so maybe we should make this type of change throughout the code base at once in a separate commit.

src/curl.c
```diff
@@ -363,6 +364,62 @@ int swupd_curl_get_file(const char *url, char *filename, struct file *file,
...
+	curl_ret = curl_easy_setopt(curl, CURLOPT_RESUME_FROM_LARGE, (curl_off_t)RESUME_BYTE_RANGE);
+	if (curl_ret != CURLE_OK) {
+		goto exit;
+	}
+
+	curl_ret = curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, swupd_download_version_to_memory);
+	if (curl_ret != CURLE_OK) {
+		goto exit;
+	}
+	curl_ret = curl_easy_setopt(curl, CURLOPT_WRITEDATA, (void *)version_string);
```
phmccarty: Do we need to download a file to test for the range support?

src/curl.c
fandeli: I agree in both, we need to do the change in all code and we need to do it
in a separate change, I can follow up and do it in a separate patch.

src/curl.c
phmccarty: Maybe we could simply download the header? 'search' does similar with the swupd_query_url_content_size() function.

phmccarty:
@fandeli I like this approach. I have a couple of suggestions that I added inline.

src/curl.c
phmccarty: Okay, that's fine. I think a separate patch would be better.

src/curl.c
fandeli: it downloads couple of bytes when servers supports resume, otherwise it returns an error (telling Range is not supported.

fandeli: it download that couple of bytes in memory, btw. I can try to set the only header download and see if I can get the error, otherwise it could be  less simply to "parse" whether Range is supported than receiving that error number directly, if the later is a must, I rather prefer the approach on trying to set Range with a couple of bytes.

phmccarty: I think it would worth testing downloading the header only, and see if the CURLE_RANGE_ERROR error is raised for that case.

fandeli: Ok so I went ahead and did little investigation, according to RFC 7233
snipet:
<-----snipet---->

An origin server that supports byte-range requests for a given target
   resource MAY send

     Accept-Ranges: bytes

   to indicate what range units are supported.  A client MAY generate
   range requests without having received this header field for the
   resource involved.  Range units are defined in Section 2
<https://tools.ietf.org/html/rfc7233#section-2>.

   A server that does not support any kind of range request for the
   target resource MAY send

     Accept-Ranges: none

   to advise the client not to attempt a range request.


</---snipet----/>

This does not have a "must" advertise if Accept-Ranges: none in case no
Ranges is accepted, having said that I went ahead and inspected headers in
one server that advertises Accept-Ranges and other that doesn't:


   1. Accept-Ranges:
   bytes
   2. Connection:
   Keep-Alive
   3. Content-Length:
   5
   4. Date:
   Wed, 04 May 2016 17:24:35 GMT
   5. ETag:
   "5-531e5c90e9080"
   6. Keep-Alive:
   timeout=5, max=100
   7. Last-Modified:
   Tue, 03 May 2016 01:11:14 GMT
   8. Server:
   Apache/2.4.6 (CentOS)



   1. Content-Length:
   268
   2. Content-type:
   text/html; charset=ANSI_X3.4-1968
   3. Date:
   Wed, 04 May 2016 17:19:10 GMT
   4. Server:
   SimpleHTTP/0.6 Python/2.7.11

As you can notice, SimpleHTTPServer don't even advertise Accept-Ranges
support, which means basically more validation to check one for advertising
Accept-Ranges: none and another for not even having that header in place.

Having said that, the already proposed patch is cheap because only is
requesting a couple of bytes and we get an straight answer code which we
can validate pretty simple, also if I request only the headers when
CURLOPT_RESUME_FROM_LARGE is set, somehow is skipped and not returning the
expected error.

Conclusion, IMHO this patch is simpler as going to examine the headers
which will require to create a linked list and parsing/checking for
different cases.

bryteise: Ah good findings there, I'd agree the current solution is simpler and more effective since we can't make any definitive decision based on the header anyway.

phmccarty: Agreed. The solution implemented in this patch is simpler, and we should go with that.

phmccarty:
Merged to master

Issue #45, opened 2016-04-19
Delay in server response may lead to truncated data in client


acornea:
Delays in the server response writes can make the client receive only truncated data, which can cause the client think there is no new updates available.
Tested on swupd v2.87 with a custom server.

Below the HTTPRequestHandler for a python HTTP Server for reproduction:
```
class SlowResponse(BaseHTTPServer.BaseHTTPRequestHandler):
    """Handler that returns data with a set delay between writes"""
    def do_GET(self):
        self.send_response(200)
        self.send_header("Content-type", "text/html")
        self.end_headers()
        response = "99990"
        delay = 0.00001 # seconds
        for c in response:
            self.wfile.write(c)
            sleep(delay)
```
Below are the responses observed for curl command and swupd check-update:

```
$ curl $IP:$PORT/version/format3/latest
99990
$ swupd check-update --url=$IP:$PORT
swupd-client software update checker 2.87
   Copyright (C) 2012-2015 Intel Corporation
   bsdiff portions Copyright Colin Percival, see COPYING file for details

There are no updates available
```

barbieri:
I don't see how checking content-length will help this given that the content-length may be too big.

For some things, like the version, we can have a limit on the expected file size, But for others you can't know beforehand unless told from a secure source (like a previous Manifest). However the entry points can still get content-lenght: 1000000000000 and that will lead to OOM.

acornea:
@barbieri i think your response should be about issue #44, not this one

acornea:
Coming back to the issue at hand. Re-tested with the following parameters: delay= 0.00001 , version = 12345670

Strace on client showed that, sometimes, due to the delay, only the last part of the version gets read. Examples below
```
recvfrom(4, "345670", 16384, 0, NULL, NULL) = 6
recvfrom(4, "2345670", 16384, 0, NULL, NULL) = 7
recvfrom(4, "70", 16384, 0, NULL, NULL) = 2
recvfrom(4, "12345670", 16384, 0, NULL, NULL) = 8
recvfrom(4, "5670", 16384, 0, NULL, NULL) = 4
```
The client then exits based on the information got.

However, there is one case that drew my attention. Sometimes, when the client receives only 3 characters (670), the client reports as the corresponding version  6704. 
```
sendto(4, "GET /version/format3/latest HTTP"..., 105, MSG_NOSIGNAL, NULL, 0) = 105
[...]

PR #61, opened 2016-04-28
Fix version to memory download

bryteise:
Contains whites pace fix from #60.

In the case where the swupd_download_version_to_memory callback was run
multiple times, it would overwrite previous data instead of appending.

Correct this behavior by keeping track of data written so far in the
struct passed to the callback.

Resolves issue #45.


phmccarty:
Looks good to me.

PR #62, opened 2016-04-29
Always sort subs after read_subscriptions_alt

bryteise:
When creating #58, some tests were observed to fail due to the subs list having a different order between systems. This change should prevent the behavior but needs tested on a system which saw the failing tests in the first place.

In order to avoid non deterministic behavior when processing the subs
list, sort it after running read_subscription_alt as it may add items to
the list in a different order between runs.


phmccarty:
+1

PR #64, opened 2016-05-02
Ensure bundle-remove gets one bundle argument

bryteise:
Instead of only checking if bundle-remove has at least one bundle
argument and ignoring additional ones, check that there is exactly one
argument passed to bundle-remove instead.


phmccarty:
Looks good.

PR #65, opened 2016-05-03
Correct return check of get_latest_version

bryteise:
Update verify to correctly check the return value of get_latest_version
for errors. Since get_latest_version could return a variety of negative
error codes to signify errors compare against that range instead of -1
to determine success of the operation.


phmccarty:
Addressed in #66

PR #66, opened 2016-05-05
Bundle add fix path

bryteise:
Patch set adds support for verify_fix_path running as expected when using bundle add.

This involves some invasive changes to manifest.c's recurse_manifests and consolidate_submanifests (the latter effectively being renamed into consolidate_files). These changes are mostly to break away from modifying a given manifest in place and instead operating on independant lists of bundles or files that can then be moved into the owning manifest as needed.

There is also an extra unrelated patch correcting how the return of get_latest_version was being used and adding a test to verify correct behavior.


src/bundle.c
```diff
@@ -109,18 +109,15 @@ static int load_bundle_manifest(const char *bundle_name, int version, struct man
...
 
-	sub_list = list_head(mom->submanifests);
-	if (sub_list != NULL) {
-		*submanifest = sub_list->data;
-		sub_list->data = NULL;
-		ret = 0;
-	}
+        *submanifest = sub_list->data;
+        sub_list->data = NULL;
+        ret = 0;
```
phmccarty: This commit expands a bunch of tabs into spaces. This is the first instance.

bryteise: Gah forgot to clang-format prior to sending again =(.

phmccarty:
Looks good. Merged.

PR #68, opened 2016-05-11
Enforce includes hierarchy with bundle-remove

bryteise:
Don't allow bundle-remove to remove bundles that are included by other
currently installed bundles.


src/bundle.c
```diff
@@ -171,6 +171,33 @@ static int unload_tracked_bundle(const char *bundle_name)
...
 
+/* Check if bundle is included by any subscribed bundles */
+bool is_included(const char *bundle_name, struct manifest *mom)
+{
+	bool ret = false;
+	struct list *b, *i;
+
+	b = list_head(mom->submanifests);
+	while (b) {
+                struct manifest *bundle = b->data;
```
phmccarty: There is a tabs vs. spaces issue with this patch :(

src/bundle.c
```diff
@@ -171,6 +171,33 @@ static int unload_tracked_bundle(const char *bundle_name)
...
+		b = b->next;
+
+                i = list_head(bundle->includes);
+                while (i) {
+                        char *name = i->data;
+                        i = i->next;
+
+                        if (strcmp(name, bundle_name) == 0) {
```
phmccarty: Can we report the "name" here, so that users will know and decide if they want to attempt removing it as well?

bryteise: I wanted to avoid this because there could be many names that come up for a bundle that is widely included.

src/bundle.c
bryteise: I need to just make a hook. Force pushed fix.

src/bundle.c
phmccarty: Okay, I see your point there. The most useful thing to report is the entire bundle hierarchy (or hierarchies) needing to be deleted beforehand, but that is out of scope of this patch.

phmccarty:
Looks good. Merged.

PR #69, opened 2016-05-12
Fix coredump by empty args and go to error if set_path_prefix returns false

chuyd:
The commit  063f0f1 fixes the coredump generated by the abort()
from the string_or_die function for the following scenarios:
    --path=, --url= --versionurl= --contenturl= are empty string values.
   
The commit a0b8931 fixes the scenarioo when  set_path_prefix returns
 false, displaying "cannot continue" but swupd continues its execution. This
is fixed by going to err when a return false is received from set_path_prefix

### DIFF
--- a/src/check_update.c
+++ b/src/check_update.c
@@ -81,6 +81,7 @@ static bool parse_options(int argc, char **argv)
 				printf("Invalid --versionurl argument\n\n");
 				goto err;
 			}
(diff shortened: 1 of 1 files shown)


chuyd:
I'll modify the pull request, to remove the  commit 063f0f1 which issue was fixed with the merged commit 1800f1

PR #70, opened 2016-05-13
Fix the swupd resume when set_path_prefix returns false 

chuyd:
When set_path_prefix returns false, a message "cannot continue"
appears but swupd continues its execution, this patch goes to
err when a false is returned by set_path_prefix.

Signed-off-by: Jesus Ornelas Aguayo <jesus.ornelas.aguayo@intel.com>


phmccarty:
Looks good, thanks!

Issue #67, opened 2016-05-11
Paths should be optional and shouldn't assume Clear Linux conventions


incandescant:
To help ease adoption on non-Clear Linux distributions paths which are currently hard-coded (and defaulting to Clear Linux conventions) should be easily overridden.

The following hard-coded paths are hard-coded to match Clear Linux conventions in swupd-client 3.5.3:

* BUNDLES_DIR set to "/usr/share/clear/bundles" in configure.ac
* UPDATE_CA_CERTS_PATH set to "/usr/share/clear/update-ca" in configure.ac
* SIGNATURE_CA_CERT set to "test-do-not-ship-R0-0.pem" in configure.ac
* CURLOPT_PINNEDPUBLICKEY set to "/usr/share/clear/update-ca/425b0f6b.key" in curl.c

Several other paths are hardcoded, but with slightly less unusual defaults, including:

* STATE_DIR set to "/var/lib/swupd" in configure.ac
* LOG_DIR set to "/var/log/swupd" in configure.ac
* LOCK_DIR set to "/run/lock" in configure.ac
* MOTD_FILE set to "/usr/lib/motd.d/001-new-release" in configure.ac

barbieri:
would `--with-bundles-dir=PATH` to replace `BUNDLES_DIR` be enough? This is easy to do in configure.ac

phmccarty:
Yes, I think adding --with-* options would be best. For the defaults, I also want to change the "clear" path components to "swupd" instead.

barbieri:
if you're going to move to `swupd` then it doesn't make sense to change the defaults I'd say. The confusion at the moment is we have `clear` in a context that doesn't mean anything. If it was `/usr/share/swupd/bundles`, it would be immediate to associate that to `swupd`.

Do you have an idea when you want to move these defaults?

incandescant:
I agree that --with-* options would be best. I think there is value in both having sane defaults (/usr/share/swupd/bundles) and allowing the defaults to be overridden (it's entirely possible to build an OS where datadir isn't /usr/share).

I'm happy to patch configure.ac in-line with the discussion here and submit a PR.

phmccarty:
@incandescant Sure, feel free to submit a fix for this.

pohly:
/usr/share/clear/versionstamp as used by swupd's verifytime command is another convention from Clear which should be configurable.

phmccarty:
Yes, agreed.

PR #71, opened 2016-05-13
Make paths configurable

incandescant:
Change the follwing symbols to be confiurable via --with-\* options:
BUNDLES_DIR, UPDATE_CA_CERTS_PATH, SIGNATURE_CA_CERT, STATE_DIR,
LOG_DIR, LOCK_DIR, MOTD_FILE

Note: this changes the default value of:
- BUNDLES_DIR from /usr/share/clear/bundles to
  /usr/share/swupd/bundles
- UPDATE_CA_CERTS_PATH from /usr/share/clear/update-ca to
  /usr/share/swupd/update-ca

Defaults remain the same for all other variables which are now
configurable

Signed-off-by: Joshua Lock joshua.g.lock@intel.com

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -98,27 +98,62 @@ AH_TEMPLATE([SWUPD_WITH_BINDMNTS],[cope with bind mounts over rootfs])
 AH_TEMPLATE([SWUPD_WITH_SELINUX],[handle selinux attributes])
 ## (3) variant extra options
 AH_TEMPLATE([MOUNT_POINT],[The mount point])
-AH_TEMPLATE([STATE_DIR],[The state directory for swupd content])
-AH_TEMPLATE([LOG_DIR],[Directory for swupd log files])
-AH_TEMPLATE([LOCK_DIR],[Directory for lock file])
-AH_TEMPLATE([BUNDLES_DIR],[Directory to use for bundles])
-AH_TEMPLATE([UPDATE_CA_CERTS_PATH],[Location of CA certificates])
-AH_TEMPLATE([SIGNATURE_CA_CERT],[CA certificate to use])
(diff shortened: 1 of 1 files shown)


incandescant:
Partially fixes #67 (but doesn't address CURLOPT_PINNEDPUBLICKEY set to "/usr/share/clear/update-ca/425b0f6b.key" in curl.c).

Tested by both:
* passing custom values for every added --with-* option and observing the passed values in config.h after invoking configure
* not passing custom values for any of the added --with* options and observing defaults in config.h after invoking configure

bryteise:
+1 needs a format bump though

phmccarty:
Looks good. I'll queue this for a format bump.

tmarcu:
This needs to be rebased. We no longer have need for the 2 CA_CERT variables, are the others still needing to become configurable?

phmccarty:
@tmarcu Because the configure.ac has been heavily revised since this patch was submitted, a rewrite is required.

I would suggest that the following defaults be configurable:
* STATE_DIR
* LOCK_DIR
* BUNDLES_DIR
* UPDATE_CA_CERTS_PATH
* MOTD_FILE

Making the above configurable will aid cross-distro adoption, because the filesystem layout/conventions may be different. And the default BUNDLES_DIR and UPDATE_CA_CERTS_PATH are Clear Linux oriented, so it would be nice to configure those too.

tmarcu:
STATE_DIR and CERT_PATH are configurable now, leaving:

LOCK_DIR
BUNDLES_DIR
MOTD_FILE

tmarcu:
Closing and opened new issue for https://github.com/clearlinux/swupd-client/issues/208

PR #72, opened 2016-05-17
Update verify

bryteise:
Pull request for update path to validate file hashes after successful tar extraction.

This also contains a change for the python web server test to increase sleep time before running the test, resolving issues on my system where tests would fail when the system was under load.


test/functional/update/verify-fullfile-hash/test.bats
```diff
@@ -0,0 +1,55 @@
...
+  [ "${lines[10]}" = "Statistics for going from version 10 to version 100:" ]
+  [ "${lines[11]}" = "    changed manifests : 1" ]
+  [ "${lines[12]}" = "    new manifests     : 0" ]
+  [ "${lines[13]}" = "    deleted manifests : 0" ]
+  [ "${lines[14]}" = "    changed files     : 1" ]
+  [ "${lines[15]}" = "    new files         : 0" ]
+  [ "${lines[16]}" = "    deleted files     : 0" ]
+  [ "${lines[17]}" = "Starting download of remaining update content. This may take a while..." ]
+  [ "${lines[18]}" = "Finishing download of update content..." ]
+  [ ! -f "$DIR/target-dir/foo" ]
```
phmccarty: Initially, I was confused about what this test's purpose was, but then I saw the full $output contents and understand now...

Should we test that the "Error: File content hash mismatch ..." message is in fact printed, since it proves we detect this error condition?

bryteise: We'd have to help the string because it contains the full path with test
directory. I wanted to avoid that bit.

phmccarty: I would be happy with just testing with a bash regexp. I'll add that in a followup commit.

phmccarty:
+1, will merge shortly.

PR #73, opened 2016-05-18
Make directory hashes independent on filename

jrguzman-intel:
Calculate the hash for directories is desiderable to be independent
on the dirname due to the subsequent calculation on the staged/HASH
file. Here is used const "DIRECTORY" string for input name for
all folders.

->
There is a patch for server side that must be applied at same time than this
jrguzman-intel/swupd-server#23
<-

### DIFF
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -31,6 +31,8 @@ extern "C" {
 
 #define MAX_TRIES 3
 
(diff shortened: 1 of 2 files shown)


phmccarty:
+1

tmarcu:
Merged into master branch.

PR #74, opened 2016-05-19
Replace system calls by local system_argv functions.

jrguzman-intel:
There are new functions that replace system() calls: system_argv()
and system_argv_pipe(). They are based on execvp() function and they
can manage any argument including filenames with characters that could
be interpreted as meta-characters in a regular system() function.
So no escape for the arguments needed and is less prone to errors.

### DIFF
--- a/include/swupd-build-variant.h
+++ b/include/swupd-build-variant.h
@@ -10,19 +10,15 @@
 
 #ifdef SWUPD_WITH_BSDTAR
 #define TAR_COMMAND "bsdtar"
-#define TAR_XATTR_ARGS ""
 #define TAR_XATTR_ARGS_STRLIST
 #else
 #define TAR_COMMAND "tar"
-#define TAR_XATTR_ARGS "--xattrs --xattrs-include='*'"
 #define TAR_XATTR_ARGS_STRLIST "--xattrs", "--xattrs-include='*'",
 #endif
 
 #ifdef SWUPD_WITH_SELINUX
-#define TAR_PERM_ATTR_ARGS "--preserve-permissions --selinux " TAR_XATTR_ARGS
 #define TAR_PERM_ATTR_ARGS_STRLIST TAR_XATTR_ARGS_STRLIST "--preserve-permissions", "--selinux"
 #else /* SWUPD_WITHOUT_SELINUX */
-#define TAR_PERM_ATTR_ARGS "--preserve-permissions " TAR_XATTR_ARGS
 #define TAR_PERM_ATTR_ARGS_STRLIST TAR_XATTR_ARGS_STRLIST "--preserve-permissions"
 #endif
 
--- a/include/swupd-build-variant.h
+++ b/include/swupd-build-variant.h
@@ -11,15 +11,19 @@
 #ifdef SWUPD_WITH_BSDTAR
 #define TAR_COMMAND "bsdtar"
 #define TAR_XATTR_ARGS ""
+#define TAR_XATTR_ARGS_STRLIST
 #else
 #define TAR_COMMAND "tar"
 #define TAR_XATTR_ARGS "--xattrs --xattrs-include='*'"
+#define TAR_XATTR_ARGS_STRLIST "--xattrs", "--xattrs-include='*'",
 #endif
 
 #ifdef SWUPD_WITH_SELINUX
 #define TAR_PERM_ATTR_ARGS "--preserve-permissions --selinux " TAR_XATTR_ARGS
+#define TAR_PERM_ATTR_ARGS_STRLIST TAR_XATTR_ARGS_STRLIST "--preserve-permissions", "--selinux"
 #else /* SWUPD_WITHOUT_SELINUX */
 #define TAR_PERM_ATTR_ARGS "--preserve-permissions " TAR_XATTR_ARGS
+#define TAR_PERM_ATTR_ARGS_STRLIST TAR_XATTR_ARGS_STRLIST "--preserve-permissions"
 #endif
 
 #endif
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -263,6 +263,10 @@ extern int get_dirfd_path(const char *fullname);
 extern int verify_fix_path(char *targetpath, struct manifest *manifest);
 extern void set_local_download(void);
 extern struct list *files_from_bundles(struct list *bundles);
+extern int system_argv(char *const argv[]);
+extern int system_argv_fd(char *const argv[], int newstdin, int newstdout, int newstderr);
+extern int system_argv_pipe(char *const argvp1[], int stdinp1, int stderrp1,
+							char *const argvp2[], int stdoutp2, int stderrp2);
 
 /* subscription.c */
 struct list *free_list_file(struct list *item);
--- a/src/download.c
+++ b/src/download.c
@@ -33,6 +33,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <assert.h>
 
 #include "config.h"
 #include "swupd-build-variant.h"
@@ -249,8 +250,8 @@ int untar_full_download(void *data)
 	char *tar_dotfile;
 	char *targetfile;
 	struct stat stat;
-	int err;
-	char *tarcommand;
+	int err, stderrfd;
+	char *tardir;
 
 	string_or_die(&tar_dotfile, "%s/download/.%s.tar", state_dir, file->hash);
 	string_or_die(&tarfile, "%s/download/%s.tar", state_dir, file->hash);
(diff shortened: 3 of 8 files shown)


include/swupd-build-variant.h
```diff
@@ -11,15 +11,19 @@
 #ifdef SWUPD_WITH_BSDTAR
 #define TAR_COMMAND "bsdtar"
 #define TAR_XATTR_ARGS ""
+#define TAR_XATTR_ARGS_STRLIST
 #else
 #define TAR_COMMAND "tar"
 #define TAR_XATTR_ARGS "--xattrs --xattrs-include='*'"
```
barbieri: why not remove the old?

include/swupd-build-variant.h
```diff
@@ -11,15 +11,19 @@
...
+#define TAR_XATTR_ARGS_STRLIST "--xattrs", "--xattrs-include='*'",
 #endif
 
 #ifdef SWUPD_WITH_SELINUX
 #define TAR_PERM_ATTR_ARGS "--preserve-permissions --selinux " TAR_XATTR_ARGS
+#define TAR_PERM_ATTR_ARGS_STRLIST TAR_XATTR_ARGS_STRLIST "--preserve-permissions", "--selinux"
 #else /* SWUPD_WITHOUT_SELINUX */
 #define TAR_PERM_ATTR_ARGS "--preserve-permissions " TAR_XATTR_ARGS
+#define TAR_PERM_ATTR_ARGS_STRLIST TAR_XATTR_ARGS_STRLIST "--preserve-permissions"
```
barbieri: add the trailing `,` to be consistent with `TAR_XATTR_ARGS_STRLIST`?

include/swupd.h
```diff
@@ -258,6 +258,10 @@ extern int get_dirfd_path(const char *fullname);
 extern int verify_fix_path(char *targetpath, struct manifest *manifest);
 extern void set_local_download(void);
 extern struct list *files_from_bundles(struct list *bundles);
+extern int system_argv(char *const argv[]);
+extern int system_argv_fd(char *const argv[], int newstdin, int newstdout, int newstderr);
+extern int system_argv_pipe(char *const argvp1[], int stdinp1, int stderrp1,
```
barbieri: these signatures are easy to get wrong, what if you use a structure? Then at least it could use member initializers and be cleaner:
```c
struct system_argv_pipe_args {
    struct {
       char *const argv[];
       int stdin;
       int stderr;
   } p1;
 struct {
       char *const argv[];
       int stdout;
       int stderr;
   } p2;   
};

#define system_argv_fd(...) system_argv_fd_impl((const struct system_argv_pipe_args){__VA_ARGS});

system_argv_fd(.p1 = {.stdin = STDIN_FILENO, argv = ...}, .p2 = ...);
```

src/download.c
```diff
@@ -289,14 +290,16 @@ int untar_full_download(void *data)
...
-
-	err = system(tarcommand);
-	if (WIFEXITED(err)) {
-		err = WEXITSTATUS(err);
-	}
-	free(tarcommand);
+	string_or_die(&tardir, "%s/staged/", state_dir);
+	char *const tarcmd[] = { TAR_COMMAND, "-C", tardir, TAR_PERM_ATTR_ARGS_STRLIST, "-xf", tarfile, NULL };
+	stderrfd = open("/dev/null", O_WRONLY);
+	if (stderrfd == -1) {
```
barbieri: what if this is moved in `system_argv_fd()` for convenience, use something like `fd == -2` to say "use /dev/null"

src/helpers.c
```diff
@@ -164,16 +166,17 @@ int create_required_dirs(void)
...
 	if (missing) { // (re)create dirs
-		char *cmd;
+		char *param;
 
 		for (i = 0; i < STATE_DIR_COUNT; i++) {
-			string_or_die(&cmd, "mkdir -p %s/%s", state_dir, dirs[i]);
-			ret = system(cmd);
+			string_or_die(&param, "%s/%s", state_dir, dirs[i]);
+			char *const cmd[] = { "mkdir", "-p", param, NULL };
+			ret = system_argv(cmd);
```
barbieri: maybe time to review and use `mkdir_p` in C?

ikeydoherty: Feel free to take it: https://github.com/ikeydoherty/libnica/blob/master/src/files.c#L44

include/swupd-build-variant.h
jrguzman-intel: @barbieri I didn't add the trailing "," because TAR_XATTR_ARGS could be an empty list whereas TAR_PERM_ATTR_ARGS not. So to use it properly in the variable declaration, the definition may be different

include/swupd.h
jrguzman-intel: @barbieri Why are easy to get wrong? Also using structures could be prone to errors if not well used. The intention to provide file descriptors is only if you need to use a different one from the standard ones.

include/swupd-build-variant.h
jrguzman-intel: @barbieri I was thinking to keep them in order to be used in another kind of declaration but you're right, I can remove unused defines.

src/download.c
jrguzman-intel: It could work, I can add it

include/swupd.h
barbieri: that number of parameters with the same type leads to "is the stdin the first or second after the argv?"

If you go with the macros, AFAIR the compiler will use the last value, so you can even define the defaults and let the user override, like fds inited to `-1`.

src/scripts.c
```diff
@@ -112,6 +126,7 @@ void run_preupdate_scripts(struct manifest *manifest)
 	char *script;
 
 	string_or_die(&script, "/usr/bin/clr_pre_update.sh");
+	char *const scriptcmd = { script, NULL };
```
bryteise: Should be: char *const scriptcmd[]

jrguzman-intel: @bryteise  Good catch! Will fix, thanks!

include/swupd-build-variant.h
phmccarty: This define is used in TAR_PERM_ATTR_ARGS, so I don't think it can be removed, unless I'm missing something.

include/swupd-build-variant.h
phmccarty: @jrguzman-intel Perhaps you can reference `TAR_XATTR_ARGS_STRLIST` at the end of the definition instead, to not need the trailing comma at all?

```
#define TAR_PERM_ATTR_ARGS_STRLIST  "--preserve-permissions", TAR_XATTR_ARGS_STRLIST
```

include/swupd.h
phmccarty: @jrguzman-intel I'm okay with using the signatures from your patch, but can you add docstrings where they are defined in helpers.c? I think having good documentation of the expected use (and limitations) of these helper functions will suffice for now.

src/download.c
phmccarty: @jrguzman-intel Do you want to implement the above suggestion for this patch?

src/helpers.c
phmccarty: @ikeydoherty I would like to start using libnica at some point, but I think it's out of scope for this patch.

phmccarty: I opened #81 to track this.

src/helpers.c
```diff
@@ -872,3 +875,147 @@ struct list *files_from_bundles(struct list *bundles)
...
+				assert(0);
+			}
+			close(newstderr);
+		}
+
+		execvp(*argv, argv);
+		/* Next line must not be reached */
+		assert(0);
+	} else if (pid < 0) {
+	    fprintf(stderr, "Failed to fork a child process\n");
```
phmccarty: This line should be indented properly.

src/helpers.c
```diff
@@ -872,3 +875,147 @@ struct list *files_from_bundles(struct list *bundles)
...
+{
+	int size = 0;
+
+	for (int i = 0; argv[i]; i++) {
+		size += strlen(argv[i]) + 1;
+	}
+
+	*output = malloc(size + 1);
+	if (!*output) {
+		fprintf(stderr, "Failed to allocate %i bytes", size);
```
phmccarty: The string printed should end with a newline.

src/manifest.c
```diff
@@ -537,15 +538,18 @@ static int retrieve_manifests(int current, int version, char *component, struct
 		goto out;
 	}
 
-	string_or_die(&tar, TAR_COMMAND " -C %s/%i -xf %s/%i/Manifest.%s.tar 2> /dev/null",
-		      state_dir, version, state_dir, version, component);
-
-	/* this is is historically a point of odd errors */
```
phmccarty: As much as I *hope* we do not see errors here anymore, can you restore this comment, for reference?

src/helpers.c
ikeydoherty: Agreed - would just complicate things

src/scripts.c
```diff
@@ -34,45 +34,59 @@
 
 static void update_kernel(void)
 {
-	char *kernel_update_cmd = NULL;
-
 	if (strcmp("/", path_prefix) == 0) {
-		string_or_die(&kernel_update_cmd, "/usr/bin/kernel_updater.sh");
+		char *const kernel_update_cmd[] = { "/usr/bin/kernel_updater.sh", NULL } ;
+
+		system_argv(kernel_update_cmd);
```
phmccarty: Since the return value is discarded, please use the (void) cast for this function. Similarly for the other call sites below.

src/scripts.c
```diff
@@ -34,45 +34,59 @@
...
+	}
 }
 
 static void update_bootloader(void)
 {
-	char *bootloader_update_cmd = NULL;
-
 	if (strcmp("/", path_prefix) == 0) {
-		string_or_die(&bootloader_update_cmd, "/usr/bin/gummiboot_updaters.sh");
+		char *const bootloader_update_cmd[] = { "/usr/bin/gummiboot_updaters.sh", NULL };
```
phmccarty: Unrelated to your patch, but the gummiboot_updaters.sh call can be removed anytime, since it's not used anymore.

phmccarty:
+1 on the latest changes.

phmccarty:
Rebased to latest master, and resolved all merge conflicts.

The new PR is #144

PR #75, opened 2016-05-24
Allow bundle-add to fallback to full files

bryteise:
Instead of requiring a pack is used for bundle-add, allow the use of
verify_fix_path when staging fails to: download, verify and stage the
item using the full file.

This change also separates the staging from installation so that all
files are first staged before they are all then renamed.

Correct output of tests where the 'required' wording was removed.


test/functional/bundleadd/fix-missing-file/test.bats
```diff
@@ -0,0 +1,40 @@
...
+
+t1_hash="e6d85023c5e619eb43d5cfbfdbdec784afef5a82ffa54e8c93bda3e0883360a3"
+
+setup() {
+  clean_test_dir
+  tar -C "$DIR/web-dir/10" -cf "$DIR/web-dir/10/Manifest.MoM.tar" Manifest.MoM Manifest.MoM.signed
+  tar -C "$DIR/web-dir/10" -cf "$DIR/web-dir/10/Manifest.os-core.tar" Manifest.os-core Manifest.os-core.signed
+  tar -C "$DIR/web-dir/10" -cf "$DIR/web-dir/10/Manifest.test-bundle.tar" Manifest.test-bundle Manifest.test-bundle.signed
+  sudo chown root:root "$DIR/web-dir/10/files/$t1_hash"
+  tar -C "$DIR/web-dir/10/files" -cf "$DIR/web-dir/10/files/$t1_hash.tar" $t1_hash --exclude=$t1_hash/*
```
phmccarty: Since t1_hash is just a normal file, the --exclude is not needed.

Without looking at the other tests, do we always pass --exclude in this case?

src/bundle.c
```diff
@@ -466,15 +462,32 @@ int install_bundles(struct list *bundles, int current_version, struct manifest *
...
+		iter = iter->next;
+
+		if (file->is_deleted || file->do_not_update || ignore(file)) {
+			continue;
+		}
+
+		/* This was staged by verify_fix_path */
+		if (!file->staging) {
+			file = search_file_in_manifest(mom, file->filename);
+		}
```
phmccarty: I'm not following why this bit is needed. Can you elaborate more?

test/functional/bundleadd/fix-missing-file/test.bats
bryteise: Ah goodness, I just failed on my copy pasta since t1_hash was previously a directory in that test.

src/bundle.c
bryteise: So this confused me at first too but it goes like this:
The to_install_files list is populated with a call to recurse_manifest that is separate from the the mom->files list (populated from recurse_manifest and then consolidated).

The rename_staged_file_to_final is going to look at the file structure and verify if it was staged or not (since this is bundle-add everything would be staged) but that structure is updated in do_staging which in the verify_fix_path case failed and so did not populate the file->staging string. verify_fix_path updates the file->staging string in the mom (using search_file_in_manifest) which is what I end up using here.

This could definitely have been further detailed but I'm actually in favor of changing how all this works to only have one copy of the file struct eventually.

phmccarty: Okay thanks, that makes sense.

test/functional/bundleadd/fix-missing-file/test.bats
phmccarty: No worries, let's just leave it as-is for now.

PR #76, opened 2016-05-24
Set ret to indicate error in update failure path

bryteise:
When unable to allocate the submanifest list for either the current or
server manifest, set ret to indicate failure.


phmccarty:
Looks correct

phmccarty:
Rebased and merged

PR #77, opened 2016-05-25
Correctly free manifest data

bryteise:
Call free_manifest instead of plain free on allocated manifests.


phmccarty:
Good catch! Will merge shortly.

PR #78, opened 2016-05-25
Stop using out of scope stack memory

bryteise:
When running get_latest_version, the swupd_curl_get_file would allocate
a version_container struct on the stack that would then be used by the
curl callback to keep track of the buffer offset. This use was invalid
however because the callback was run after the stack had been popped
which lead to undefined behavior.

Instead change the swupd_curl_get_file function to take the struct
itself so it will refer to memory valid for the entire length of the
call.

The swupd_curl_get_file function should likely be restructured at some
point so that in memory downloads are less of a hack (only able to
download a LINE_MAX worth of data) at some point however and this will
need to be updated again.

### DIFF
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -55,6 +55,11 @@ struct manifest {
 	char *component;
 };
 
+struct version_container {
+	size_t offset;
+	char *version;
+};
+
 struct header;
 
 extern bool force;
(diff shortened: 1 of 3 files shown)


phmccarty:
Looks good, merged.

PR #79, opened 2016-06-02
Skip verified unchanged files in update

bryteise:
Instead of just checking if the versions have changed for a file to be
updated, compare the current file's hash to updated file's expected hash
and only queue files for update that are changed.


phmccarty:
+1, looks good

PR #80, opened 2016-06-03
Fix for completion when absolute path is given

jrguzman-intel:


phmccarty:
Looks good, thanks!

PR #82, opened 2016-06-10
Improvements for update list creation

phmccarty:
The recently added call to verify_file(), used to skip files that are
already installed and have correct hashes, is a nice optimization for
the case when an update crosses a build where a new minversion was
applied to the server content.

However, I ran into a case where calling verify_file() like this results
in an incomplete update: If the corresponding hash calculation of a file
in the latest version *follows a symlink" for the path in the current
version, the hashes will match (assuming the same file content).

This issue is avoidable by checking for the file's presence in the
current version's manifest first, because manifests only contain real
paths (no non-leaf path components are symlinks).

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -616,16 +616,20 @@ struct list *create_update_list(struct manifest *current, struct manifest *serve
 		file = list->data;
 		list = list->next;
 
-		/* Skip files that aren't updated */
-		fullname = mk_full_filename(path_prefix, file->filename);
-		if (fullname == NULL) {
-			abort();
-		}
-		if (verify_file(file, fullname)) {
+		if (file->peer && hash_equal(file->hash, file->peer->hash)) {
+			/* Skip new/changed files if they are already installed
+			 * and the hash check passes. This makes minversion
+			 * bumps cheaper. */
+			fullname = mk_full_filename(path_prefix, file->filename);
+			if (fullname == NULL) {
+				abort();
+			}
+			if (verify_file(file, fullname)) {
+				free(fullname);
+				continue;
+			}
 			free(fullname);
-			continue;
 		}
-		free(fullname);
 
 		if ((file->last_change > current->version) ||
 		    (file->is_rename && file_has_different_hash_in_manifest(current, file)) ||


phmccarty:
I plan to add a functional test for the situation described in the second paragraph. That will be a separate patch.

src/manifest.c
```diff
@@ -616,16 +616,27 @@ struct list *create_update_list(struct manifest *current, struct manifest *serve
 		file = list->data;
 		list = list->next;
 
-		/* Skip files that aren't updated */
-		fullname = mk_full_filename(path_prefix, file->filename);
-		if (fullname == NULL) {
-			abort();
-		}
-		if (verify_file(file, fullname)) {
+		if (file_found_in_manifest(current, file)) {
```
bryteise: Good solution but needs to use peers because this is an O(N^2) operation otherwise (file_has_different_hash_in_manifest loops over the full manifest file list).

phmccarty: Good point. I will make that change.

phmccarty:
uploaded version 2 on the patch

src/manifest.c
```diff
@@ -616,16 +616,27 @@ struct list *create_update_list(struct manifest *current, struct manifest *serve
...
-		fullname = mk_full_filename(path_prefix, file->filename);
-		if (fullname == NULL) {
-			abort();
-		}
-		if (verify_file(file, fullname)) {
+		if (file->peer) {
+			/* Skip files that aren't updated, but only if they are
+			 * installed in the current version (tracked). */
+			if (file->is_tracked &&
+			    hash_equal(file->hash, file->peer->hash)) {
```
bryteise: I don't think we want to continue here yet because we also need to verify_file on the installed file first.

The file->is_tracked and hash_equal should be prerequisites for entering the verify_file block however.

phmccarty: Right, so there are some cases that were overlooked here. I will revisit and take a closer look.

phmccarty: After revisiting this, I don't think the "tracked" status (or deleted status) of a file matters for this check after all.

It should be sufficient to check for the peer and matching hashes before running verify_file().

phmccarty:
uploaded v3 patch

bryteise:
+1

PR #83, opened 2016-06-15
Signature verification

tmarcu:
Rebasing previous patches onto master with very minor changes to style and removing deprecated API calls.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -29,7 +29,11 @@ AS_IF([test "x$enable_bzip2" != "xno" ],
 	 AC_CHECK_LIB([bz2], [BZ2_bzBuffToBuffCompress], [], [AC_MSG_ERROR([the libbz2 library is missing])])],
   [AC_DEFINE(SWUPD_WITHOUT_BZIP2,1,[Do not use bzip2 compression])]
 )
-
+AC_ARG_ENABLE(
+  [signature-verification],
+  [AS_HELP_STRING([--enable-signature-verification], [Enable signature check (disabled by default)])],
+  [AC_DEFINE([SIGNATURES], [], [Enable signature check as default])]
+)
 AC_ARG_ENABLE(
   [tests],
   [AS_HELP_STRING([--disable-tests], [Do not enable unit or functional test framework (enabled by default)])]
@@ -83,7 +87,7 @@ AS_IF([test "$enable_tests" != "no"], [
 AM_CONDITIONAL([ENABLE_TESTS], [test "$enable_tests" != "no"])
 
 PKG_CHECK_MODULES([curl], [libcurl])
-PKG_CHECK_MODULES([openssl], [libcrypto >= 0.9.8])
+PKG_CHECK_MODULES([openssl], [libcrypto >= 1.0.2])
 AC_CHECK_LIB([pthread], [pthread_create])
 AC_CHECK_PROGS(TAR, tar)
 
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -106,6 +106,7 @@ struct file {
 	CURL *curl;    /* curl handle if downloading */
 };
 
+extern bool verify_signatures;
 extern bool download_only;
 extern bool local_download;
 extern bool verify_esp_only;
--- a/src/clr_bundle_add.c
+++ b/src/clr_bundle_add.c
@@ -53,6 +53,7 @@ static void print_help(const char *name)
 	printf("   -l, --list              List all available bundles for the current version of Clear Linux\n");
 	printf("   -x, --force             Attempt to proceed even if non-critical errors found\n");
 	printf("   -S, --statedir          Specify alternate swupd state directory\n");
+	printf("   -V, --verify-signatures Enable signature verification, default value is disabled\n");
 	printf("\n");
 }
 
@@ -67,6 +68,7 @@ static const struct option prog_opts[] = {
 	{ "format", required_argument, 0, 'F' },
 	{ "force", no_argument, 0, 'x' },
 	{ "statedir", required_argument, 0, 'S' },
+	{ "verify-signatures", no_argument, 0, 'V' },
 	{ 0, 0, 0, 0 }
 };
 
@@ -74,7 +76,7 @@ static bool parse_options(int argc, char **argv)
 {
 	int opt;
 
-	while ((opt = getopt_long(argc, argv, "hxu:c:v:P:p:F:lS:", prog_opts, NULL)) != -1) {
+	while ((opt = getopt_long(argc, argv, "Vhxu:c:v:P:p:F:lS:", prog_opts, NULL)) != -1) {
 		switch (opt) {
 		case '?':
 		case 'h':
@@ -132,6 +134,9 @@ static bool parse_options(int argc, char **argv)
 			break;
 		case 'x':
 			force = true;
+			break;
+		case 'V':
+			verify_signatures = true;
 			break;
 		default:
 			printf("error: unrecognized option\n\n");
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -457,16 +457,9 @@ static int try_delta_manifest_download(int current, int new, char *component, st
 			unlink(deltafile);
 			goto out;
 		}
-
-		if (!signature_download_and_verify(url, deltafile)) {
-			ret = -1;
-			unlink(deltafile);
-			goto out;
-		}
 	}
 
 	/* Now apply the manifest delta */
-
 	string_or_die(&newfile, "%s/%i/Manifest.%s", state_dir, new, component);
 
 	ret = apply_bsdiff_delta(original, newfile, deltafile);
@@ -478,7 +471,7 @@ static int try_delta_manifest_download(int current, int new, char *component, st
 	}
 
 	unlink(deltafile);
-	signature_delete(deltafile);
+//	delete_signature(deltafile);
 
 out:
 	free(original);
@@ -495,8 +488,8 @@ static int retrieve_manifests(int current, int version, char *component, struct 
 	char *url;
 	char *filename;
 	char *dir;
-	int ret = 0;
 	char *tar;
+	int ret = 0;
 
 	*manifest = NULL;
 
@@ -530,11 +523,6 @@ static int retrieve_manifests(int current, int version, char *component, struct 
 		goto out;
 	}
 
-	if (!signature_download_and_verify(url, filename)) {
-		unlink(filename);
-		goto out;
-	}
-
 	string_or_die(&tar, TAR_COMMAND " -C %s/%i -xf %s/%i/Manifest.%s.tar 2> /dev/null",
 		      state_dir, version, state_dir, version, component);
 
@@ -548,8 +536,20 @@ static int retrieve_manifests(int current, int version, char *component, struct 
 		goto out;
 	}
 
+	/* Only signing the MoM manifest*/
+	if (strncmp(component, "MoM", 3) == 0) {
+		free(filename);
+		free(url);
+		string_or_die(&filename, "%s/%i/Manifest.MoM", state_dir, version);
+		string_or_die(&url, "%s/%i/Manifest.MoM", content_url, version);
+		if (verify_signatures && !download_and_verify_signature(url, filename)) {
+			goto out;
+		}
+	}
+
 	*manifest = manifest_from_file(version, component);
 out:
+	delete_signature(filename);
 	free(filename);
 	free(url);
 	return ret;
--- a/src/signature.c
+++ b/src/signature.c
@@ -24,198 +24,347 @@
 #include <stdbool.h>
 #include <stdio.h>
 #include <string.h>
+#include <sys/stat.h>
 #include <unistd.h>
 
 #include <openssl/bio.h>
 #include <openssl/err.h>
 #include <openssl/pem.h>
+#include <openssl/crypto.h>
 
 #include "config.h"
 #include "signature.h"
 #include "swupd.h"
 
-/*
- * Implementation flavors:
- *   FAKE ..... do nothing, always return success
- *   FORGIVE .. do everything, always return success
- *   REAL ..... do everything, return the real status
- */
-#define IMPL_FAKE 0
-#define IMPL_FORGIVE 1
-#define IMPL_REAL 2
-
-#warning "TODO pick signing scheme"
-#if defined(SWUPD_LINUX_ROOTFS)
-#define IMPL IMPL_FAKE
+#define BUFFER_SIZE   4096
+#define CERTNAME "/usr/share/clear/update-ca/ClearLinuxRoot.pem"
+#ifdef SIGNATURES
+       bool verify_signatures = true;
+#else
+       bool verify_signatures = false;
 #endif
 
-#if IMPL != IMPL_FAKE
-
-static X509_STORE *create_store(const char *, const char *, const char *);
-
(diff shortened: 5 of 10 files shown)


src/clr_bundle_add.c
```diff
@@ -53,6 +53,7 @@ static void print_help(const char *name)
 	printf("   -l, --list              List all available bundles for the current version of Clear Linux\n");
 	printf("   -x, --force             Attempt to proceed even if non-critical errors found\n");
 	printf("   -S, --statedir          Specify alternate swupd state directory\n");
+    printf("   -V, --verify-sign       Enable signature verification, default value is disabled\n");
```
bryteise: Whitespace issues in this patch

src/clr_bundle_add.c
```diff
@@ -67,14 +68,15 @@ static const struct option prog_opts[] = {
 	{ "format", required_argument, 0, 'F' },
 	{ "force", no_argument, 0, 'x' },
 	{ "statedir", required_argument, 0, 'S' },
+    { "verify-sign", no_argument, 0, 'V' },
```
bryteise: Whitespace issue here

src/clr_bundle_add.c
```diff
@@ -133,6 +135,9 @@ static bool parse_options(int argc, char **argv)
 		case 'x':
 			force = true;
 			break;
+        case 'V':
```
bryteise: Whitespace issue here

src/verify.c
```diff
@@ -171,6 +173,9 @@ static bool parse_options(int argc, char **argv)
 		case 'x':
 			force = true;
 			break;
+        case 'V':
```
bryteise: Whitespace issue here

bryteise:
First patch has bundle-add and verify but should also have the other commands as well (update, bundle-remove, search, check-update)

configure.ac
```diff
@@ -29,7 +29,11 @@ AS_IF([test "x$enable_bzip2" != "xno" ],
 	 AC_CHECK_LIB([bz2], [BZ2_bzBuffToBuffCompress], [], [AC_MSG_ERROR([the libbz2 library is missing])])],
   [AC_DEFINE(SWUPD_WITHOUT_BZIP2,1,[Do not use bzip2 compression])]
 )
-
+AC_ARG_ENABLE(
+  [signing],
+  [AS_HELP_STRING([--enable-signing], [Enable signature check (disabled by default)])],
+  [AC_DEFINE([SIGNING], [], [Enable signature check as default])]
+)
```
phmccarty: Since swupd-client is only doing signature verification, I would prefer --enable-signature-verification instead.

The define could be SIGNATURES.

include/swupd.h
```diff
@@ -106,6 +106,7 @@ struct file {
 	CURL *curl;    /* curl handle if downloading */
 };
 
+extern bool signing_enabled;
```
phmccarty: would prefer "signatures_enabled"

src/manifest.c
```diff
@@ -530,9 +530,12 @@ static int retrieve_manifests(int current, int version, char *component, struct
 		goto out;
 	}
 
-	if (!signature_download_and_verify(url, filename)) {
-		unlink(filename);
-		goto out;
+	if (signing_enabled && !signature_download_and_verify(url, filename)) {
```
phmccarty: I would prefer to have swupd-server ship the .signed file in the MoM.tar (which I think it already does), and then we can try reading the local copy first, before attempting an on-the-fly download.

phmccarty: Also, "url" here ends with .tar, but we're not going to be signing the .tar, only the Manifest.MoM.

phmccarty:
@tmarcu There are various calls to `signature_download_and_verify()` that should be removed, because swupd-server should only be signing the Manifest.MoM. We should remove those calls in this patch series.

tmarcu:
Will post a V3 soon with many more changes, ignore previous PR for now.

configure.ac
```diff
@@ -29,7 +29,11 @@ AS_IF([test "x$enable_bzip2" != "xno" ],
 	 AC_CHECK_LIB([bz2], [BZ2_bzBuffToBuffCompress], [], [AC_MSG_ERROR([the libbz2 library is missing])])],
   [AC_DEFINE(SWUPD_WITHOUT_BZIP2,1,[Do not use bzip2 compression])]
 )
-
+AC_ARG_ENABLE(
+  [signatures],
```
phmccarty: The documentation on the line below is correct now, but this argument needs updating too.

src/signature.c
```diff
@@ -24,140 +24,333 @@
...
+	fclose(sig);
+
+	return result;
 }
 
-static X509_STORE *create_store(const char *ca_filename, const char *ca_dirname,
-				const char *crl_filename)
+/* Make sure the certificate exists and extract the public key from it.
+ *
+ * cert_filename: certificate used to verify signature.
```
phmccarty: "certname" instead?

src/signature.c
```diff
@@ -24,140 +24,333 @@
...
+	size_t sig_len;
+	size_t data_size;
+	size_t len = 0;
+	size_t bytes_left;
+	size_t count;
+	struct stat st;
+	int fd;
+	EVP_MD_CTX md_ctx;
+
+	sig_len = fread(sig_buffer, 1, BUFFER_SIZE, fp_sig);
```
phmccarty: fread() error handling needed

src/signature.c
```diff
@@ -24,140 +24,333 @@
...
+
+	fd = fileno(fp_data);
+	fstat(fd, &st);
```
phmccarty: error handling here too

src/signature.c
```diff
@@ -24,140 +24,333 @@
...
+	data_size = st.st_size;
+
+	/* Initialize verify context and use sha256 algorithm internally */
+	if (!EVP_VerifyInit(&md_ctx, EVP_sha256())) {
```
phmccarty: So, looking at openssl/evp.h, EVP_VerifyInit() is really EVP_DigestInit(), which is a deprecated function, in favor of using EVP_DigestInit(). I don't see a similar deprecation notice for EVP_VerifyInit(), and I also don't see any documentation of its return value...

Perhaps EVP_VerifyInit_ex() should be used instead?

src/signature.c
```diff
@@ -24,140 +24,333 @@
...
+		goto error;
+	}
+
+	/* read all bytes from file to calculate digest using sha256 and then sign it */
+	bytes_left = data_size;
+	while (bytes_left > 0) {
+		count = (bytes_left > BUFFER_SIZE ? BUFFER_SIZE : bytes_left);
+		len = fread(buffer, 1, count, fp_data);
+		if (len != count) {
+			fprintf(stderr, "Failed len!= count\n");
```
phmccarty: This message is not very informative, so I would suggest removing it, or changing it to be more user-friendly.

src/manifest.c
```diff
@@ -551,6 +544,16 @@ static int retrieve_manifests(int current, int version, char *component, struct
 		goto out;
 	}
 
+	/* Only signing the MoM manifest*/
+	if (strstr(filename, "Manifest.MoM")) {
```
phmccarty: I would prefer to check component=="MoM" to make this more robust.

src/manifest.c
```diff
@@ -551,6 +544,16 @@ static int retrieve_manifests(int current, int version, char *component, struct
...
+		free(filename);
+		string_or_die(&filename, "%s/%i/Manifest.MoM", state_dir, version);
+		if (verify_signatures && !download_and_verify_signature(url, filename)) {
+				unlink(filename);
+				goto out;
```
phmccarty: 1-tab indent only :-)

src/manifest.c
```diff
@@ -457,12 +457,13 @@ static int try_delta_manifest_download(int current, int new, char *component, st
...
 			goto out;
 		}
-
+/*
 		if (!download_and_verify_signature(url, deltafile)) {
 			ret = -1;
 			unlink(deltafile);
 			goto out;
 		}
+*/
```
phmccarty: Unrelated to this patch? I would suggest to remove code entirely if it's not needed at present and won't be in the future.

src/signature.c
```diff
@@ -24,198 +24,364 @@
...
 
-bool signature_verify(const char *data_filename, const char *sig_filename)
+/* Get the paths and filenames of certificates */
+bool get_certificates(void)
 {
-	BIO *bio_data = NULL;
-	BIO *bio_sig = NULL;
-	PKCS7 *pkcs7 = NULL;
-	int ret;
+	certname = (char *) X509_get_default_cert_file();
```
bradTpeters: We're going to be using one and only one specific cert for swupd signing, so instead of loading the full system-wide compliment of certs, I suggest we remove this call and replace it a call loading just the cert we want.  

X509_load_cert_file() may do the trick

tmarcu: After more testing I was able to implement this, but that function does _not_ replace X509_get_default_cert, which just gets us the certname to use in load_file. X509_load_cert_file() will replace the call to X509_add_cert() and it also is able to read certs in either DER or PEM format, so I agree this is a good replacement for manually adding the cert(s).

configure.ac
phmccarty: This line should read [signature-verification]

src/signature.c
```diff
@@ -357,9 +357,9 @@ bool download_and_verify_signature(const char *data_url, const char *data_filena
 	int ret;
 	bool result;
 
-	string_or_die(&sig_url, "%s.signed", data_url);
+	string_or_die(&sig_url, "%s.sig", data_url);
```
phmccarty: Any thoughts about this?

phmccarty:
Your branch is 35 commits behind master, too. Please rebase.

src/manifest.c
```diff
@@ -548,6 +536,16 @@ static int retrieve_manifests(int current, int version, char *component, struct
 		goto out;
 	}
 
+	/* Only signing the MoM manifest*/
+	if (strncmp(component, "MoM", 3) == 0) {
+		free(filename);
+		string_or_die(&filename, "%s/%i/Manifest.MoM", state_dir, version);
+		if (verify_signatures && !download_and_verify_signature(url, filename)) {
```
phmccarty: Can you check to see if Manifest.MoM.sig exists, and use the local copy if possible? If it's not present, then we can `download_and_verify_signature()`.

src/signature.c
```diff
@@ -24,198 +24,342 @@
...
+       bool verify_signatures = false;
 #endif
 
-#if IMPL != IMPL_FAKE
-
-static X509_STORE *create_store(const char *, const char *, const char *);
-
-static char *VERIF_FAIL = "Signature verification failed";
-static char *XSTORE_FAIL = "XSTORE creation failed";
+static bool validate_signature(FILE *, FILE *);
```
phmccarty: Please give the formal parameters names for these declarations, in accordance with the rest of the code base.

src/signature.c
phmccarty: thoughts on this one?

src/signature.c
```diff
@@ -24,198 +24,342 @@
...
+	if (store) {
+		X509_STORE_free(store);
+	}
+	if (lookup) {
+		X509_LOOKUP_free(lookup);
+	}
+	return false;
+}
+
+int verify_callback(int ok, X509_STORE_CTX *stor)
```
phmccarty: Is this the main use case for this callback, to print an error message?

src/signature.c
```diff
@@ -24,198 +24,342 @@
...
+
+	X509_STORE_set_verify_cb_func(store, verify_callback);
+
+	/* Add the certificates to be verified to the store */
+	if (!(lookup = X509_STORE_add_lookup(store, X509_LOOKUP_file()))) {
+		goto error;
+	}
+
+	/*  Load the our Root cert, which can be in either DER or PEM format */
+	if (!X509_load_cert_file(lookup, CERTNAME, X509_FILETYPE_PEM)) {
```
phmccarty: Is the code comment up-to-date? Or does this function try DER if the cert is not in PEM format?

phmccarty:
We need calls to `initialize_signature()` for most other subcommands (anything that loads a MoM).

src/signature.c
```diff
@@ -24,198 +24,342 @@
...
+
+/* This function makes sure the certificate is still valid by not having any
+ * compromised certificates in the chain.
+ * If there is no Certificate Revocation List (CRL) it may be that the private
+ * keys have not been compromised or the CRL has not been generated by the
+ * Certificate Authority (CA)
+ *
+ * returns: true if certificate is valid, false otherwise */
+static bool validate_certificate(void)
+{
```
bradTpeters: Looking at this closely, it appears we're not actually doing anything here. This is one big, noisy no-op.

Effectively we're loading a trusted cert chain of 1 cert into the store, loading a NULL revocation list, then calling X509_verify_cert() with the same cert we loaded into the store. 

Until we add in support for revocation, as well as move to a trusted cert-chain with root up around Mount Olympus, we should not make a call into this function. Keeping the code is fine, since we'll be using is eventually, but some added comments regarding when it should be called would be useful

src/signature.c
tmarcu: It deals with this internally when identifying the format it seems.

src/signature.c
```diff
@@ -37,7 +37,7 @@
 #include "swupd.h"
 
 #define BUFFER_SIZE   4096
-
+#define CERTNAME "/usr/share/clear/update-ca/ClearLinuxRoot.pem"
```
phmccarty: I would love to see this value set by a configure option instead of hardcoding the path.

phmccarty: Also, --enable-signature-verification would require this new option (e.g. --with-certificate-path) being set.

configure.ac
```diff
@@ -83,7 +87,7 @@ AS_IF([test "$enable_tests" != "no"], [
 AM_CONDITIONAL([ENABLE_TESTS], [test "$enable_tests" != "no"])
 
 PKG_CHECK_MODULES([curl], [libcurl])
-PKG_CHECK_MODULES([openssl], [libcrypto >= 0.9.8])
+PKG_CHECK_MODULES([openssl], [libcrypto >= 1.0.2])
```
tpepper: You should note in the commit message which 1.0.2 feature (or bugfix?) is required here versus 0.9.8.

src/manifest.c
```diff
@@ -548,6 +536,16 @@ static int retrieve_manifests(int current, int version, char *component, struct
 		goto out;
 	}
 
+	/* Only signing the MoM manifest*/
```
tpepper: "Only the MoM manifest is signed, so only it needs verified"

phmccarty:
@tmarcu Running `make check` fails:

```
test/signature_verify_test.c:49:7: error: too many arguments to function ‘initialize_signature’
```

src/manifest.c
tpepper: Can the download if .sig not present be buried inside download_and_verify_signature()?  That matches how the manifest fetching happens...either fetch from disk else remote, but fetch either way.  Then it's maybe just verify_signature() with "download signature iff not already present" is abstracted inside the verifier?

phmccarty:
Superceded by #85

bradTpeters:
I suppose that's fair; I was targeting those commands described by the
swupd spec. I'll send out a v2 Monday

PR #84, opened 2016-06-21
Implement bundle manifest hash checks

phmccarty:
The second commit in this series is the functional change. More details in the commit description.


include/swupd.h
```diff
@@ -161,7 +161,8 @@ extern void apply_heuristics(struct file *file);
 
 extern int file_sort_filename(const void *a, const void *b);
 extern int file_sort_filename_reverse(const void *a, const void *b);
-extern int load_manifests(int current, int version, char *component, struct file *file, struct manifest **manifest);
+extern struct manifest *load_mom(int current, int version);
```
tpepper: "int current" here can be dropped, if...(see comment a little further down)

src/manifest.c
```diff
@@ -574,29 +574,82 @@ static void set_untracked_manifest_files(struct manifest *manifest)
...
+ * Note that if the manifest fails to download, or if the manifest fails to be
+ * loaded into memory, this function will return NULL.
+ */
+struct manifest *load_mom(int current, int version)
 {
+	struct manifest *manifest = NULL;
 	int ret = 0;
 
-	*manifest = manifest_from_file(version, component);
+	ret = retrieve_manifests(current, version, "MoM", NULL);
```
tpepper: - ret = retrieve_manifests(current, version, "MoM", NULL);
+ ret = retrieve_manifests(version, version, "MoM", NULL);

If called in this way, then the NULL and version not less than version result in retrieve_manifests() not attempting a delta application.

tpepper: Then all the call sites become:

some_manifest_variable = load_mom(version);

That is sooooo much more readable than the current mess.  Way more sustainable / supportable / maintainable going forward.

phmccarty: Good point. That is a further useful simplication. I'll make the change.

src/manifest.c
```diff
@@ -574,29 +574,82 @@ static void set_untracked_manifest_files(struct manifest *manifest)
...
+ * MOM manifest.
+ *
+ * Value ranges and restrictions:
+ *   - CURRENT < VERSION for 'update' subcommand
+ *   - CURRENT == VERSION for other subcommands
+ *
+ * Note that if the manifest fails to download, or if the manifest fails to be
+ * loaded into memory, this function will return NULL.
+ */
+struct manifest *load_mom(int current, int version)
```
tpepper: Add this descriptive bit here:

```
// MoM's are not huge so deltas don't give much benefit, versus bundle           
// manifests which can be huge and beg for deltas.  If there was a MoM          
// delta you'd want to validate the input MoM prior to applying a delta,        
// and then validate the output MoM.  All complicated compared to just          
// getting the whole MoM and verifying it.
```

src/manifest.c
tpepper: Modulo the above little change to incrementally clean a little more aggresively, this patch looks good to me.  Really good. :)

src/manifest.c
phmccarty: Okay, looks good. I'll amend the commit if there are no further comments.

phmccarty:
New patch set uploaded.

src/hash.c
```diff
@@ -252,3 +253,41 @@ bool verify_file(struct file *file, char *filename)
...
+		}
+
+		string_or_die(&local, "%s/%i/Manifest.%s", state_dir,
+			      current->last_change, current->filename);
+
+		if (!verify_file(bundle, local)) {
+			printf("Warning: hash check failed for Manifest.%s\n",
+			       current->filename);
+			ret = 0;
+			// free(local);
```
tpepper: meant to remove?

phmccarty: yes, will remove

src/manifest.c
```diff
@@ -574,29 +574,78 @@ static void set_untracked_manifest_files(struct manifest *manifest)
...
+ * Note that if the manifest fails to download, or if the manifest fails to be
+ * loaded into memory, this function will return NULL.
+ */
+struct manifest *load_mom(int version)
 {
+	struct manifest *manifest = NULL;
 	int ret = 0;
 
-	*manifest = manifest_from_file(version, component);
+	ret = retrieve_manifests(version, version, "MoM", NULL);
```
bradTpeters: Isn't this ALWAYS going to perform a curl_get_file on the MoM, regardless of whether or not it is already on the filesystem?

phmccarty:
Patch set no. 3 uploaded

phmccarty:
and... patch set no. 4

src/hash.c
```diff
@@ -252,3 +253,39 @@ bool verify_file(struct file *file, char *filename)
...
 	}
 }
+
+/* Compares the hash for BUNDLE with that listed in the Manifest.MoM.  If the
+ * hash check fails, for now, we print a warning and return success, but
+ * eventually will switch to being a fatal error and exit early.
+ */
+int verify_bundle_hash(struct manifest *manifest, struct file *bundle)
+{
+	assert(strcmp(manifest->component, "MoM") == 0);
```
tpepper: one more assert here.  And the #include.

phmccarty: will remove

phmccarty:
uploaded patch set 5

tpepper:
lgtm

src/hash.c
```diff
@@ -252,3 +252,37 @@ bool verify_file(struct file *file, char *filename)
...
+		}
+
+		string_or_die(&local, "%s/%i/Manifest.%s", state_dir,
+			      current->last_change, current->filename);
+
+		if (!verify_file(bundle, local)) {
+			printf("Warning: hash check failed for Manifest.%s\n",
+			       current->filename);
+			ret = 0;
+		}
```
bradTpeters: Do you want to return here? Else we're continuing to verify a known bad manifest

bradTpeters:
This looks great Patrick; nicely done

phmccarty:
Thanks, I will merge shortly.

PR #85, opened 2016-06-23
Implement signature verification

tmarcu:
This patch enables the swupd client to do signing verification on the MoM, which
ensures a root of trust for the rest of the update by guaranteeing the
authenticity of the MoM and content it includes.

### DIFF
--- a/include/signature.h
+++ b/include/signature.h
@@ -8,12 +8,12 @@
  * @param ca_cert_filename - the file containing the CA certificate
  * @return true <=> no error
  */
-bool signature_initialize(const char *ca_cert_filename);
+bool initialize_signature(void);
 
 /*
  * Terminate usage of this module, free resources.
  */
-void signature_terminate(void);
+void terminate_signature(void);
 
 /*
  * Verify data file contents against a purported signature.
@@ -21,7 +21,7 @@ void signature_terminate(void);
  * @param sig_filename - the file containing the signature
  * @return true <=> no error
  */
-bool signature_verify(const char *data_filename, const char *sig_filename);
+bool verify_signature(const char *data_filename, const char *sig_filename);
 
 /*
  * The given data file has already been downloaded from the given URL.
@@ -30,12 +30,12 @@ bool signature_verify(const char *data_filename, const char *sig_filename);
  * @param data_url - the URL from which the data came
  * @param data_filename - the file containing the data
  */
-bool signature_download_and_verify(const char *data_url, const char *data_filename);
+bool download_and_verify_signature(const char *data_url, const char *data_filename);
 
 /*
  * Delete the signature file corresponding to given data file.
  * @param data_filename - the file containing the data
  */
-void signature_delete(const char *data_filename);
+void delete_signature(const char *data_filename);
 
 #endif /* SIGNATURE_H_ */
--- a/include/signature.h
+++ b/include/signature.h
@@ -8,7 +8,7 @@
  * @param ca_cert_filename - the file containing the CA certificate
  * @return true <=> no error
  */
-bool initialize_signature(void);
+void initialize_signature(void);
 
 /*
  * Terminate usage of this module, free resources.
--- a/src/clr_bundle_add.c
+++ b/src/clr_bundle_add.c
@@ -53,6 +53,7 @@ static void print_help(const char *name)
 	printf("   -l, --list              List all available bundles for the current version of Clear Linux\n");
 	printf("   -x, --force             Attempt to proceed even if non-critical errors found\n");
 	printf("   -S, --statedir          Specify alternate swupd state directory\n");
+	printf("   -V, --verify-signatures Enable signature verification, default value is disabled\n");
 	printf("\n");
 }
 
@@ -67,6 +68,7 @@ static const struct option prog_opts[] = {
 	{ "format", required_argument, 0, 'F' },
 	{ "force", no_argument, 0, 'x' },
 	{ "statedir", required_argument, 0, 'S' },
+	{ "verify-signatures", no_argument, 0, 'V' },
 	{ 0, 0, 0, 0 }
 };
 
@@ -74,7 +76,7 @@ static bool parse_options(int argc, char **argv)
 {
 	int opt;
 
-	while ((opt = getopt_long(argc, argv, "hxu:c:v:P:p:F:lS:", prog_opts, NULL)) != -1) {
+	while ((opt = getopt_long(argc, argv, "Vhxu:c:v:P:p:F:lS:", prog_opts, NULL)) != -1) {
 		switch (opt) {
 		case '?':
 		case 'h':
@@ -131,6 +133,9 @@ static bool parse_options(int argc, char **argv)
 			break;
 		case 'x':
 			force = true;
+			break;
+		case 'V':
+			verify_signatures = true;
 			break;
 		default:
 			printf("error: unrecognized option\n\n");
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -459,16 +459,9 @@ static int try_delta_manifest_download(int current, int new, char *component, st
 			unlink(deltafile);
 			goto out;
 		}
-
-		if (!signature_download_and_verify(url, deltafile)) {
-			ret = -1;
-			unlink(deltafile);
-			goto out;
-		}
 	}
 
 	/* Now apply the manifest delta */
-
 	string_or_die(&newfile, "%s/%i/Manifest.%s", state_dir, new, component);
 
 	ret = apply_bsdiff_delta(original, newfile, deltafile);
@@ -480,7 +473,7 @@ static int try_delta_manifest_download(int current, int new, char *component, st
 	}
 
 	unlink(deltafile);
-	signature_delete(deltafile);
+//	delete_signature(deltafile);
 
 out:
 	free(original);
@@ -532,11 +525,6 @@ static int retrieve_manifests(int current, int version, char *component, struct 
 		goto out;
 	}
 
-	if (!signature_download_and_verify(url, filename)) {
-		unlink(filename);
-		goto out;
-	}
-
 	string_or_die(&tar, TAR_COMMAND " -C %s/%i -xf %s/%i/Manifest.%s.tar 2> /dev/null",
 		      state_dir, version, state_dir, version, component);
 
@@ -549,6 +537,19 @@ static int retrieve_manifests(int current, int version, char *component, struct 
 	if (ret != 0) {
 		goto out;
 	}
+
+	/* Only MoM manifests are signed, so only the MoM gets verified*/
+	if (strncmp(component, "MoM", 3) == 0) {
+		free(filename);
+		free(url);
+		string_or_die(&filename, "%s/%i/Manifest.MoM", state_dir, version);
+		string_or_die(&url, "%s/%i/Manifest.MoM", content_url, version);
+		if (verify_signatures && !download_and_verify_signature(url, filename)) {
+			delete_signature(filename);
+			goto out;
+		}
+	}
+
 
 out:
 	free(filename);
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -538,19 +538,6 @@ static int retrieve_manifests(int current, int version, char *component, struct 
 		goto out;
 	}
 
-	/* Only MoM manifests are signed, so only the MoM gets verified*/
-	if (strncmp(component, "MoM", 3) == 0) {
-		free(filename);
-		free(url);
-		string_or_die(&filename, "%s/%i/Manifest.MoM", state_dir, version);
-		string_or_die(&url, "%s/%i/Manifest.MoM", content_url, version);
-		if (verify_signatures && !download_and_verify_signature(url, filename)) {
-			delete_signature(filename);
-			goto out;
-		}
-	}
-
-
 out:
 	free(filename);
 	free(url);
@@ -589,6 +576,8 @@ struct manifest *load_mom(int version)
 {
 	struct manifest *manifest = NULL;
 	int ret = 0;
+	char *filename;
+	char *url;
 
 	ret = retrieve_manifests(version, version, "MoM", NULL);
 	if (ret != 0) {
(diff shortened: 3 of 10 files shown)


tpepper:
Can the commit say "signature verification" instead of "signing verification"

phmccarty:
@tmarcu It seems like a second 'swupd verify' run skips the signature verification. The verify should happen every time.

include/signature.h
```diff
@@ -8,20 +8,20 @@
  * @param ca_cert_filename - the file containing the CA certificate
  * @return true <=> no error
  */
-bool signature_initialize(const char *ca_cert_filename);
+bool initialize_signature(void);
```
phmccarty: The call site in test/signature_verify_test.c needs to be updated, or else one can't run `make check`.

src/clr_bundle_add.c
```diff
@@ -53,6 +53,7 @@ static void print_help(const char *name)
 	printf("   -l, --list              List all available bundles for the current version of Clear Linux\n");
 	printf("   -x, --force             Attempt to proceed even if non-critical errors found\n");
 	printf("   -S, --statedir          Specify alternate swupd state directory\n");
+	printf("   -V, --verify-signatures Enable signature verification, default value is disabled\n");
```
phmccarty: Probably not worth mentioning "default value is disabled", because enabling verification can be set with ./configure at build time.

src/manifest.c
```diff
@@ -480,7 +473,7 @@ static int try_delta_manifest_download(int current, int new, char *component, st
 	}
 
 	unlink(deltafile);
-	signature_delete(deltafile);
+//	delete_signature(deltafile);
```
phmccarty: needed in the future? if not, remove this line.

src/signature.c
```diff
@@ -24,198 +24,342 @@
...
- */
-#define IMPL_FAKE 0
-#define IMPL_FORGIVE 1
-#define IMPL_REAL 2
-
-#warning "TODO pick signing scheme"
-#if defined(SWUPD_LINUX_ROOTFS)
-#define IMPL IMPL_FAKE
+#define BUFFER_SIZE   4096
+#define CERTNAME "/usr/share/clear/update-ca/ClearLinuxRoot.pem"
```
phmccarty: Again, I would prefer that this value be configurable at build time, but if you want to delay adding a new option (or consolidating with --enable-signature-verification), we can do that too.

src/signature.c
```diff
@@ -24,198 +24,342 @@
...
-#if IMPL != IMPL_FAKE
-
-static X509_STORE *create_store(const char *, const char *, const char *);
-
-static char *VERIF_FAIL = "Signature verification failed";
-static char *XSTORE_FAIL = "XSTORE creation failed";
+static bool validate_signature(FILE *, FILE *);
+static bool validate_certificate(void);
+static int verify_callback(int, X509_STORE_CTX *);
+static bool get_pubkey();
```
phmccarty: should be `static bool get_pubkey(void);`

src/signature.c
```diff
@@ -24,198 +24,342 @@
...
+       bool verify_signatures = false;
 #endif
 
-#if IMPL != IMPL_FAKE
-
-static X509_STORE *create_store(const char *, const char *, const char *);
-
-static char *VERIF_FAIL = "Signature verification failed";
-static char *XSTORE_FAIL = "XSTORE creation failed";
+static bool validate_signature(FILE *, FILE *);
```
phmccarty: To keep consistent with other function declarations, add variable names for the formal parameters.

src/signature.c
```diff
@@ -24,198 +24,342 @@
...
+static bool validate_certificate(void);
+static int verify_callback(int, X509_STORE_CTX *);
```
phmccarty: similarly here

src/signature.c
```diff
@@ -24,198 +24,342 @@
...
+	size_t bytes_left;
+	size_t count;
+	struct stat st;
+	int fd;
+	EVP_MD_CTX md_ctx;
 
-	string_or_die(&sig_url, "%s.signed", data_url);
+	sig_len = fread(sig_buffer, 1, BUFFER_SIZE, fp_sig);
+	if (sig_len == 0) {
+		fprintf(stderr, "Failed to get signature length!\n");
```
phmccarty: "Failed to read signature contents"

src/signature.c
```diff
@@ -24,198 +24,342 @@
...
+		goto error;
+	}
+	fd = fileno(fp_data);
+	if (fstat(fd, &st) != 0) {
+		fprintf(stderr, "Failed to stat data file\n");
```
phmccarty: strerror(errno) would be useful to print as well

src/signature.c
```diff
@@ -24,198 +24,342 @@
...
-		result = false;
-	} else {
-		result = signature_verify(data_filename, sig_filename);
+	/* read all bytes from file to calculate digest using sha256 and then sign it */
+	bytes_left = data_size;
+	while (bytes_left > 0) {
+		count = (bytes_left > BUFFER_SIZE ? BUFFER_SIZE : bytes_left);
+		len = fread(buffer, 1, count, fp_data);
+		if (len != count) {
+			fprintf(stderr, "Signature check failed, read %lu bytes, expected %lu\n", len, count);
```
phmccarty: "len" and "count" are both size_t, so %zu specifier is preferred

src/verify.c
```diff
@@ -641,7 +646,7 @@ int verify_main(int argc, char **argv)
 	 * FIXME: We need a command line option to override this in case the
 	 * certificate is hosed and the admin knows it and wants to recover.
 	 */
-	if (!signature_initialize(UPDATE_CA_CERTS_PATH "/" SIGNATURE_CA_CERT)) {
+	if (!initialize_signature()) {
```
phmccarty: Instead of calling `initialize_signature()` for each subcommand, perhaps we should move this into `swupd_init()`?

src/signature.c
```diff
@@ -24,198 +24,354 @@
...
-	} else {
-		result = signature_verify(data_filename, sig_filename);
+error:
+	if (signature) {
+		free(signature);
 	}
-	if (!result) {
-		unlink(sig_filename);
+	if (data) {
+		free(data);
```
phmccarty: I'm seeing a compiler warning:

```
src/signature.c:237:3: warning: ‘data’ may be used uninitialized in this function [-Wmaybe-uninitialized]
```

That pointer should be initialized to NULL.

src/signature.c
```diff
@@ -24,198 +24,354 @@
...
- *   REAL ..... do everything, return the real status
- */
-#define IMPL_FAKE 0
-#define IMPL_FORGIVE 1
-#define IMPL_REAL 2
-
-#warning "TODO pick signing scheme"
-#if defined(SWUPD_LINUX_ROOTFS)
-#define IMPL IMPL_FAKE
+#define BUFFER_SIZE   4096
```
bradTpeters: Need to increase our BUFFER_SZIE; our signature files are over 5k, and we're only reading from it once.
...
5385 Jun 23 17:44 Manifest.MoM.sig

src/signature.c
```diff
@@ -24,198 +24,362 @@
...
 
-	string_or_die(&sig_url, "%s.signed", data_url);
+	/* Read in the .sig file */
+	fd = fileno(fp_sig);
+	if (fstat(fd, &st) != 0) {
+		fprintf(stderr, "Failed to stat .sig file\n");
+	}
+	sig_len = st.st_size;
+	signature = malloc(sig_len);
+	count = fread(signature, 1, sig_len, fp_sig);
```
tpepper: This would be better as an mmap() intead of malloc()/fread().

src/signature.c
```diff
@@ -24,198 +24,362 @@
...
-	string_or_die(&sig_filename, "%s.signed", data_filename);
+	/* Read in the corresponding data file for the .sig */
+	fd = fileno(fp_data);
+	if (fstat(fd, &st) != 0) {
+		fprintf(stderr, "Failed to stat data file\n");
+		goto error;
+	}
+	data_size = st.st_size;
+	data = malloc(data_size);
+	count = fread(data, 1, data_size, fp_data);
```
tpepper: This would be better as an mmap() intead of malloc()/fread().

src/signature.c
bradTpeters: Correction, it looks like this is not longer used as of a later patch than I was viewing. 

Should remove this #define entirely

src/verify.c
```diff
@@ -641,7 +646,7 @@ int verify_main(int argc, char **argv)
...
 		printf("Can't initialize the SSL certificates\n");
```
phmccarty: This line prints even when verification is disabled...

src/signature.c
```diff
@@ -24,198 +24,386 @@
...
 {
-	char *sig_filename;
+	FILE *fp = NULL;
+	X509_LOOKUP *lookup = NULL;
+	X509_STORE_CTX *verify_ctx = NULL;
+
+	/* CRL and Chains are not required for the current setup, but we may
+	 * implement them in the future */
+	if (!crl) {
+		printf("No certificate revocation list provided\n");
```
phmccarty: let's add // TODO comments in the code for this instead

src/signature.c
```diff
@@ -24,198 +24,386 @@
...
+	}
+	if (!chain) {
+		printf("No certificate chain provided\n");
```
phmccarty: let's add // TODO comments in the code for this instead

src/signature.c
```diff
@@ -24,198 +24,386 @@
...
+	if (!x509_stack) {
+		fprintf(stderr, "Failed to initialize X509 certificate stack\n");
+		goto error;
+	}
+	sk_X509_push(x509_stack, cert);
+
+	/* Verify the signature, outdata can be NULL because we don't use it */
+	ret = PKCS7_verify(p7, x509_stack, store, indata, NULL, 0);
+	if (ret == 1) {
+		printf("VERIFY SUCCESS!\n");
```
phmccarty: debug print to remove

src/signature.c
```diff
@@ -24,198 +24,386 @@
...
-void signature_delete(const char UNUSED_PARAM *data_filename)
+/* Delete the signature file downloaded with download_and_verify_signature() */
+void delete_signature(const char *data_filename)
 {
+	char *sig_filename;
+	struct stat st;
+
+	string_or_die(&sig_filename, "%s.sig", data_filename);
+	if (stat(sig_filename, &st) == 0) {
+		unlink(sig_filename);
```
bradTpeters: Why delete the sig file here? It makes it difficult to investigate the cause of a failure, if there is one, so unless there is a good reason to remove this, I'd probably leave it

src/signature.c
```diff
@@ -24,198 +24,386 @@
...
+	if (X509_STORE_CTX_init(verify_ctx, store, cert, NULL) != 1) {
+		goto error;
+	}
+
+	/* Specify which cert to validate in the verify context.
+	 * This is required because we may add multiple certs to the X509 store,
+	 * but we want to validate a specific one out of the group/chain. */
+	X509_STORE_CTX_set_cert(verify_ctx, cert);
+
+	/* verify the certificate */
```
bradTpeters: Since we're not actually verifying the cert chain, we should remove this call. As part of this, see my comment in the verify_cert function

src/signature.c
```diff
@@ -24,198 +24,386 @@
...
+	if (!(store = X509_STORE_new())) {
+		goto error;
+	}
 
-	unlink(sig_filename);
+	X509_STORE_set_verify_cb_func(store, verify_callback);
 
-	free(sig_filename);
-}
+	/* Add the certificates to be verified to the store */
```
bradTpeters: The store created here is being used in the validate_signature() call as well.  I'd move store creation to it's own function, especially considering we may remove the call to validate_certificate till some future date

src/signature.c
tmarcu: Removed in coming patch.

src/signature.c
```diff
@@ -24,198 +24,386 @@
...
+	EVP_add_digest(EVP_sha256());
 
-	string_or_die(&sig_filename, "%s.signed", data_filename);
+	/* Read in the .sig file */
+	fd = fileno(fp_sig);
+	if (fstat(fd, &st) != 0) {
+		fprintf(stderr, "Failed to stat .sig file\n");
+	}
+	sig_len = st.st_size;
+	signature = malloc(sig_len);
```
phmccarty: need a check for `malloc()` failure

src/signature.c
```diff
@@ -24,198 +24,386 @@
...
+ * compromised certificates in the chain.
+ * If there is no Certificate Revocation List (CRL) it may be that the private
+ * keys have not been compromised or the CRL has not been generated by the
+ * Certificate Authority (CA)
+ *
+ * returns: true if certificate is valid, false otherwise */
+static bool validate_certificate(void)
 {
-	char *sig_filename;
+	FILE *fp = NULL;
```
phmccarty: This variable is unused, aside from when you free the memory if non-NULL (never happens). See other comment below.

src/signature.c
```diff
@@ -24,198 +24,386 @@
...
-}
 
-void signature_terminate(void)
-{
+error:
+	ERR_print_errors_fp(stderr);
+
+	if (fp) {
+		fclose(fp);
+	}
```
phmccarty: This block can be removed, because "fp" is unused.

src/manifest.c
```diff
@@ -588,8 +587,7 @@ struct manifest *load_mom(int version)
 	string_or_die(&filename, "%s/%i/Manifest.MoM", state_dir, version);
 	string_or_die(&url, "%s/%i/Manifest.MoM", content_url, version);
 	if (verify_signatures && !download_and_verify_signature(url, filename)) {
-		delete_signature(filename);
-		goto out;
+		printf("WARNING!!! FAILED TO VERIFY SIGNATURE OF Manifest.MoM\n");
```
phmccarty: Can the warning print the version as well?

phmccarty:
Good enough for now. I will merge this.

phmccarty:
@tpepper is going to follow up with some additional changes in a few PR.

src/clr_bundle_add.c
tpepper: fixed

src/signature.c
tpepper: This one still needs done.

src/signature.c
tpepper: done

src/signature.c
tpepper: I don't see another comment?  I'm leaving validate_certificate() as implemented for now.  The call is expected to succeed, and if it does not succeed that is an informative sign even today.

src/signature.c
tpepper: done

src/signature.c
tpepper: Done in new PR now

PR #86, opened 2016-06-24
Signature verification

tpepper:
This PR represents the final working state of Tudor and I as we went through OpenSSL's poor documentation and addressed review comments from PR #85 .  The src/signature.c:verify_signature() function is an abomination, but we think we doesn't leak openssl state and valgrind is agreeing.  The code is focused now on just the MoM signature verification flow, which means some functions are now static.  The cert and cert path are now autoconf configurable.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -11,11 +11,7 @@ AM_SILENT_RULES([yes])
 AC_PROG_CC
 AM_PROG_CC_C_O
 AC_LANG(C)
-AC_ARG_WITH([swupdcertsdir],
-  [AS_HELP_STRING([--with-swupdcertsdir=DIR], [swupdcertsdir files])],
-  [swupdcertsdir=$withval],
-  [swupdcertsdir="/usr/share/clear/update-ca"])
-AC_SUBST([swupdcertsdir], [$swupdcertsdir])
+
 AC_CONFIG_HEADERS([config.h])
 PKG_CHECK_MODULES([bsdiff], [bsdiff])
 PKG_CHECK_MODULES([lzma], [liblzma])
@@ -29,6 +25,19 @@ AS_IF([test "x$enable_bzip2" != "xno" ],
 	 AC_CHECK_LIB([bz2], [BZ2_bzBuffToBuffCompress], [], [AC_MSG_ERROR([the libbz2 library is missing])])],
   [AC_DEFINE(SWUPD_WITHOUT_BZIP2,1,[Do not use bzip2 compression])]
 )
+AC_ARG_ENABLE(
+  [signature-verification],
+  [AS_HELP_STRING([--enable-signature-verification], [Enable signature check (disabled by default)])],
+  [AC_DEFINE([SIGNATURES], [], [Enable signature check as default])]
+
+)
+
+if test "$enable_signature_verification" = "yes" ; then
+	AC_ARG_WITH([swupdcert],
+		[AS_HELP_STRING([--with-swupdcert=FILE], [swupd verification cert])],
+		[AC_DEFINE_UNQUOTED([SWUPDCERT], ["$withval"], [swupd verification cert])],
+		[AC_DEFINE([SWUPDCERT], ["ClearLinuxRoot.pem"], [swupd verification cert])])
+fi
 
 AC_ARG_ENABLE(
   [tests],
@@ -83,12 +92,13 @@ AS_IF([test "$enable_tests" != "no"], [
 AM_CONDITIONAL([ENABLE_TESTS], [test "$enable_tests" != "no"])
 
 PKG_CHECK_MODULES([curl], [libcurl])
-PKG_CHECK_MODULES([openssl], [libcrypto >= 0.9.8])
+PKG_CHECK_MODULES([openssl], [libcrypto >= 1.0.2])
 AC_CHECK_LIB([pthread], [pthread_create])
 AC_CHECK_PROGS(TAR, tar)
 
 # default to Linux rootfs build
 enable_linux_rootfs_build="yes"
+certs_path="/usr/share/clear/update-ca"
 
 # document all options for build variants
 ## (1) build variants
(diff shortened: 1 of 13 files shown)


src/signature.c
```diff
@@ -24,198 +24,375 @@
...
-}
-
-void signature_terminate(void)
-{
+	return result;
 }
 
-bool signature_verify(const char UNUSED_PARAM *data_filename, const char UNUSED_PARAM *sig_filename)
+#else
+bool download_and_verify_signature(const char *data_url, const char *data_filename)
```
phmccarty: The UNUSED_PARAM macro can be used for the function args.

test/signature_verify_test.c
```diff
@@ -46,19 +46,12 @@ int main(int argc, char **argv)
 		usage(argv[0]);
 	}
 
-	if (!signature_initialize(argv[3])) {
-		fprintf(stderr, "Can't initialize!\n");
-		exit(-1);
-	}
-
-	if (signature_verify(argv[1], argv[2])) {
+	if (verify_signature(argv[1], argv[2])) {
```
phmccarty: This test should be calling `download_and_verify_signature()`.

src/signature.c
```diff
@@ -24,198 +24,375 @@
...
+	if (!validate_certificate()) {
+		goto fail;
+	}
 
-static bool initialized = false;
+	/* Push our trust cert(s) to the stack, which is a set of certificates
+	 * in which to search for the signer's cert. */
+	x509_stack = sk_X509_new_null();
+	if (!x509_stack) {
+		return false;
```
phmccarty: This is a void function, so the above line should be updated.

tpepper: fixed

src/signature.c
tpepper: fixed.

test/signature_verify_test.c
tpepper: fixed

src/manifest.c
```diff
@@ -494,7 +486,7 @@ static int try_delta_manifest_download(int current, int new, char *component, st
 /* TODO: This should deal with nested manifests better */
 static int retrieve_manifests(int current, int version, char *component, struct file *file)
 {
-	char *url;
```
tpepper: this appears unneeded

phmccarty:
+1 from me

bradTpeters:
Still reviewing but looking good thus far

src/manifest.c
```diff
@@ -503,17 +495,19 @@ static int retrieve_manifests(int current, int version, char *component, struct
 
 	string_or_die(&filename, "%s/%i/Manifest.%s.tar", state_dir, version, component);
 	if (stat(filename, &sb) == 0) {
-		return 0;
+		ret = 0;
+		goto out;
```
phmccarty: I'm running into a segfault with this change. The "goto out" will lead to freeing "url", which is not initialized at this point.

configure.ac
```diff
@@ -29,6 +25,19 @@ AS_IF([test "x$enable_bzip2" != "xno" ],
 	 AC_CHECK_LIB([bz2], [BZ2_bzBuffToBuffCompress], [], [AC_MSG_ERROR([the libbz2 library is missing])])],
   [AC_DEFINE(SWUPD_WITHOUT_BZIP2,1,[Do not use bzip2 compression])]
 )
+AC_ARG_ENABLE(
+  [signature-verification],
+  [AS_HELP_STRING([--enable-signature-verification], [Enable signature check (disabled by default)])],
+  [AC_DEFINE([SIGNATURES], [], [Enable signature check as default])]
```
bradTpeters: Is this enabled or disabled by default? The text here seems to be divided

src/signature.c
```diff
@@ -24,198 +24,379 @@
...
+ *
+ * returns: true if can initialize and validate certificates, otherwise false */
+static void initialize_signature(void)
+{
+	ERR_load_crypto_strings();
+	ERR_load_PKCS7_strings();
+	if (!get_pubkey()) {
+		goto fail;
+	}
+	if (!validate_certificate()) {
```
bradTpeters: This call should not be made - it's a no-op with the current self-signed cert

src/signature.c
```diff
@@ -24,198 +24,379 @@
...
-		return NULL;
+	/* TODO: CRL and Chains are not required for the current setup, but we may
+	 * implement them in the future 
+	if (!crl) {
+		printf("No certificate revocation list provided\n");
 	}
-	if (X509_STORE_load_locations(store, ca_filename, ca_dirname) != 1) {
-		goto err;
+	if (!chain) {
+		printf("No certificate chain provided\n");
```
bradTpeters: This seems fair, but we should describe the result of this, not simply the fact that we don't have a cert chain.

bradTpeters: Could reword this to: "Proceeding with signature validate using self-signed certificate"

phmccarty: This code is commented out right now...

configure.ac
phmccarty: disabled by default. Passing the configure option as --enable-signature-verification enables it.

src/signature.c
```diff
@@ -17,205 +17,426 @@
...
+	PKCS7 *p7;
+	BIO *verify_BIO;
+
+	if (!initialize_signature()) {
+		terminate_signature();
+		return false;
+	}
+
+	//this function has issues: EVP_add_digest(EVP_sha256());
+	OpenSSL_add_all_digests();
```
bradTpeters: Here's a more limited digest add which seems to do the trick:

-       //this function has issues: EVP_add_digest(EVP_sha256());
-       OpenSSL_add_all_digests();
+       EVP_add_digest(EVP_sha1());
+       EVP_add_digest_alias(SN_sha1, "ssl3-sha1");
+       EVP_add_digest_alias(SN_sha1WithRSAEncryption, SN_sha1WithRSA);

tpepper: Does that work on the second run through the function?

Our big issue this afternoon was appearing to be that we must only call EVP_add_digest() once at process start (not at each run through our verifier function) and must only call the clean up functions one on process termination (not at the end of each run through the verifier function).

tpepper:
I moved the terminate_signature() and initialize_signature() into a new swupd_deinit() and the existing swupd_init().  I'm now able to verify, verify/fix, bundle add and bundle remove.

The pre-existing load_mom() segfault in swupd search is still there...didn't have time to look into it.  Am hoping @bradTpeters and @phmccarty made progress on it at the end of the day separately?

And I observe that bundle add/remove both have a bunch of pre-existing memory leakage...need a issue opened, run through valgrind and some thought on how to add proper memory management?

src/manifest.c
phmccarty: "url" now initialized, so this issue is fixed.

phmccarty:
There is a minor build issue when building without signature verification enabled, so I'll push a fix for that as a followup.

+1

phmccarty:
@tpepper I'm not seeing the swupd search bug anymore. Let me know if you still see it.

PR #88, opened 2016-06-27
don't segfault on not found MoM

tpepper:
If the MoM is not found on the server to match the os version on which
you're running, swupd search's MoM downloader returns 0 instead of an
error.  EMOM_NOTFOUND is a natural value to return.

When 0 is returned, the null MoM pointer is later derferenced leading to a
segfault.


bradTpeters:
Good find, though it's strange we're only seeing this with the new load_mom call. 

+1

phmccarty:
Looks good

PR #90, opened 2016-06-27
Ignore sigverify errors in existing tests

phmccarty:
Because signature verification is a feature that we need separate testing for, and swupd's output may print a verification error (or not) depending on how swupd was built, add a helper function to remove the verification error message when swupd is built with verification enabled. If verification is not enabled, swupd will not print any message, and the function is a no-op.


tpepper:
lgtm

PR #92, opened 2016-06-28
Try local signature verification first before downloading

phmccarty:
In case the Manifest.MoM.sig already exists locally (e.g. it was shipped
inside the Manifest.MoM.tar), then try to verify the local copy first.
If the verify fails because the .sig is missing, or for any other
reason, then download a fresh copy from the server.

For the implementation, I refactored verify_signature() to not print
error messages if the verification fails the first time, because we
"retry" by downloading the fresh copy. If the verification fails the
second time, then error messages are printed. Also, I consolidated the
cleanup logic in this function to reduce duplication.

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -118,145 +118,132 @@ void terminate_signature(void)
 	CRYPTO_cleanup_all_ex_data();
 }
 
-/* Verifies that the file and the signature exists, and does a signature
- * check afterwards.
+/* Verifies that the file and the signature exists, and does a signature check
+ * afterwards. If any error is to be considered a verify failure, then
+ * ERRORS_FATAL should be set to true.
  *
  * returns: true if able to validate the signature, false otherwise */
-static bool verify_signature(const char *data_filename, const char *sig_filename)
+static bool verify_signature(const char *data_filename, const char *sig_filename, bool errors_fatal)
 {
 	int ret;
+	bool result = false;
 	struct stat st;
+	char *errorstr = NULL;
 
-	int data_fd;
+	int data_fd = -1;
 	size_t data_len;
 	unsigned char *data = NULL;
-	BIO *data_BIO;
+	BIO *data_BIO = NULL;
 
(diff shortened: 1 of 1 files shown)


src/signature.c
```diff
@@ -390,10 +377,11 @@ int verify_callback(int ok, X509_STORE_CTX *stor)
 	return ok;
 }
 
-/* Downloads the corresponding signature filename from the
- * swupd server.
+/* Verifies signature for the file DATA_FILENAME, or downloads the signature
+ * first from DATA_URL and verifies the signature afterward.
```
tmarcu: I think the comment on download would be better stated saying: "Verifies signature for the local file DATA_FILENAME first, and on failure downloads the signature based on DATA_URL and tries to verify again."

phmccarty: Sounds good, will change that

phmccarty:
Updated patch set 2

src/signature.c
```diff
@@ -118,145 +118,132 @@ void terminate_signature(void)
...
-	BIO *sig_BIO;
+	BIO *sig_BIO = NULL;
 
-	PKCS7 *p7;
-	BIO *verify_BIO;
+	PKCS7 *p7 = NULL;
+	BIO *verify_BIO = NULL;
 
 	/* get the signature */
 	sig_fd = open(sig_filename, O_RDONLY);
```
bradTpeters: Why open() before fstat()'ing the file? The opposite order would make more sense IMO

bradTpeters:
New flow and retry looks good.  +1

src/signature.c
phmccarty: The fstat() is done to get the file size actually... And calling open() first avoids TOCTOU issues.

bradTpeters: Ah, thanks for clarifying

phmccarty:
Cool, will merge shortly

phmccarty:
Rebased and merged

PR #97, opened 2016-07-19
Move sign/verify to abort on failure

bradTpeters:
Moves swupd signature verification to mandatory, where
the operation will not proceed unless the following flag
is passed in:

  --force-continue-with-unverified-signature

The signed MoM serves as the top level chain of trust as
far as swupd is concerned, and it is used to extend content
trust down to the individual file level.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -589,8 +589,18 @@ struct manifest *load_mom(int version)
 	string_or_die(&filename, "%s/%i/Manifest.MoM", state_dir, version);
 	string_or_die(&url, "%s/%i/Manifest.MoM", content_url, version);
 	if (!download_and_verify_signature(url, filename)) {
-		printf("WARNING!!! FAILED TO VERIFY SIGNATURE OF Manifest.MoM\n");
+		if (!force_ignore_unverified_signature) {
+			printf("FAILED TO VERIFY SIGNATURE OF Manifest.MoM. Can proceed if"
+				"  --force-continue-with-unverified-signature used, but system"
+				" security WILL BE COMPROMISED\n");
+			return NULL;
+		} else {
+			printf("FAILED TO VERIFY SIGNATURE OF Manifest.MoM. Operation proceeding due to"
+				"  --force-continue-with-unverified-signature, but system security WILL BE"
+				" COMPROMISED\n");
+		}
 	}
+
 	free(filename);
 	free(url);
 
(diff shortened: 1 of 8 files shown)


tpepper:
Suggest commit message begin with the following slight edit:

Move swupd signature verification to mandatory.  Failed
signature verification now aborts 'swupd update' and 'swupd verify'.
Their operation may be forced to proceed past the signature
verification if the following flag is passed in:
...

src/manifest.c
```diff
@@ -589,8 +589,18 @@ struct manifest *load_mom(int version)
...
+		if (!force_ignore_unverified_signature) {
+			printf("FAILED TO VERIFY SIGNATURE OF Manifest.MoM. Can proceed if"
+				"  --force-continue-with-unverified-signature used, but system"
+				" security WILL BE COMPROMISED\n");
+			return NULL;
+		} else {
+			printf("FAILED TO VERIFY SIGNATURE OF Manifest.MoM. Operation proceeding due to"
+				"  --force-continue-with-unverified-signature, but system security WILL BE"
+				" COMPROMISED\n");
+		}
```
tpepper: The WILL BE COMPROMISED statement strikes me as overly confident and assuming.

"FAILED TO VERIFY SIGNATURE OF Manifest.MoM." followed by the exit in the first condition strikes me as sufficient.

In the second case I can see reason for a "...proceeding with forced" message.

Generally I've come to prefer more terse printf's as we've experienced a lot of confusion with verbose output on the console.  It is a standard paradigm to see a failure and then call the command with --help to see if there's any sort of force override flag for folks who are inclined to go there.

tpepper:
Aside from the minor comments above, looks good to me.

src/manifest.c
bradTpeters: Ok, I'll switch back to the "MAY BE COMPROMISED"; David wanted something more severe, but I originally used 'may' as well

bradTpeters:
I'll close and open a new PR incorporating your commit msg change

pohly:
@bradTpeters not sure how you guys normally handle it, but usually it is better to force-push a rewritten commit to the same branch while the PR is open. Then the existing PR will get updated and all comments related to the change stay in one place, instead of getting scattered among different PRs.

One of the Github quirks is that it doesn't allow reopening the PR *after* a force-push. So it's too later for this PR here (bradTpeters:mandatory_sig_ver already updated). New PR is #98.

PR #99, opened 2016-07-25
TLS comment clarification

tpepper:
The issue with TLS versions is on the server side.  The client
must specify which version to use, but that version must be
supported by the server.  And sadly the server world is often
running old OS's with old libraries and daemons.

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -436,7 +436,7 @@ static CURLcode swupd_curl_set_security_opts(CURL *curl)
 		goto exit;
 	}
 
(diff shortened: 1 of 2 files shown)


phmccarty:
+1

tpepper:
An additional little patch for something I've seen this past week:  there's no version number in bundle manifest hash mismatch printf.

phmccarty:
Good idea! +1 on the second patch.

bradTpeters:
+1 for both changes

tmarcu:
Merged +1

PR #100, opened 2016-07-26
Print some build-time options in --version output

phmccarty:
To make compile-time options more easily discoverable at runtime, make
the --version output more verbose by printing a few of them. Other
options can be easily added in the future by adding an appropriate
AC_DEFINE call in configure.ac, and updating the new header added in
this commit (swupd-build-opts.h).

### DIFF
--- a/include/swupd-build-opts.h
+++ b/include/swupd-build-opts.h
@@ -1,0 +1,56 @@
+/*
+ *   Software Updater - client side
+ *
+ *      Copyright © 2016 Intel Corporation.
+ *
+ *   This program is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License as published by
+ *   the Free Software Foundation, version 2 or later of the License.
+ *
+ *   This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


bradTpeters:
This looks great Patrick! Ran it through a quick test run as well

+1

tpepper:
Nice simple way to get some additional useful info!

bradTpeters:
One comment - Here's sample output:

---------
swupd-client swupd 3.6.0
   Copyright (C) 2012-2016 Intel Corporation

Compile-time options: +BZIP2 +SIGVERIFY -COVERAGE -BSDTAR
---------

It would probably be preferable to have the copyright appear first, followed by version and compile time options

phmccarty:
@bradTpeters I had not considered reordering the lines. I do like that copyright statement being printed first.

phmccarty:
@bradTpeters On second thought, I just invoked a few GNU tools with --version, and they all print the program name and version info first, followed by the copyright statement. So, it's "common" to see the ordering swupd currently uses.

Maybe we could remove the blank line? It seems unnecessary.

tmarcu:
Merged +1

PR #102, opened 2016-08-01
Enable HTTP/2 protocol as option

fenrus75:
Allow curl to use the HTTP/2 protocol, with an automatic fallback to
HTTP/1.1.
Because of the optional nature, there's also no strict error handling
performed, if requesting HTTP/2 fails it fails.... and we get HTTP/1.1.

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -480,6 +480,8 @@ CURLcode swupd_curl_set_basic_options(CURL *curl, const char *url)
 		goto exit;
 	}
 
+	curl_easy_setopt(curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_2_0);
+
 	if (update_server_port > 0) {
 		curl_ret = curl_easy_setopt(curl, CURLOPT_PORT, update_server_port);
 		if (curl_ret != CURLE_OK) {


src/curl.c
```diff
@@ -480,6 +480,8 @@ CURLcode swupd_curl_set_basic_options(CURL *curl, const char *url)
 		goto exit;
 	}
 
+	curl_easy_setopt(curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_2_0);
+
```
bradTpeters: While I see no problem with this,  I'd propose we let curl decide which to use with:
   CURL_HTTP_VERSION_NONE   From curl.haxx.se docs: "We don't care about what version the library uses. libcurl will use whatever it thinks fit.", 

or even 

      CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE : 
          "Issue non-TLS HTTP requests using HTTP/2 without HTTP/1.1 Upgrade. It requires prior knowledge that the server supports HTTP/2 straight away. HTTPS requests will still do HTTP/2 the standard way with negotiated protocol version in the TLS handshake. "

fenrus75: That does not work; the NONE means HTTP/1.1
We need to explicitly give HTTP/2 permission.

2_PRIOR_KNOWLEDGE is also not a good option since that removed the potential for a fallback to HTTP/1.1
for when infrastructure (like proxies) are not HTTP/2 ready

bradTpeters: If NONE indeed falls back immediately to 1.1, then that is definitely a poor choice. Do you have any references (code or doc's) which clarify that?

bradTpeters:
I see you are correct -  curl/lib/http.c, use_http_1_1plus():  does indeed default to 1.1 if NONE is set

Change looks good to me

+1

tmarcu:
+1 lgtm, do you have any thoughts on using CURL_HTTP_VERSION_2TLS in the future?

tmarcu:
Merged into master.

PR #103, opened 2016-08-02
Enospace

bradTpeters:
Series of 3 patches which 1) refactor a minimalist download list prior to acting, 2) add a pre-bundle(s)-add test for sufficient space, and 3) pay attention to download return codes to ensure a bundle-add does not proceed if download fails


tmarcu:
Closing and tracking for future work when logic can be reconsidered to cover the system constraints better.

PR #106, opened 2016-08-09
Add configure.ac support for state dir options

bradTpeters:
Add configuration support for contenturl, versionurl, and formatid.

The order or preference for state config options is now:
  1. swupd cmdline: -u or -v options passed to the specific command
  2. ./configure: --with-contenturl=, --with-versionurl=, or --with-formatid=
  3. Fall back to the defaults provided in /usr/share/defaults/swupd

While adding these new configure options, this patch also adds a report
out from the configure script, showing selected options

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -18,18 +18,30 @@ PKG_CHECK_MODULES([lzma], [liblzma])
 PKG_CHECK_MODULES([zlib], [zlib])
 AC_ARG_ENABLE(
 	[bzip2],
-	AS_HELP_STRING([--disable-bzip2],[Do not use bzip2 compression (uses bzip2 by default)])
+	AS_HELP_STRING([--disable-bzip2],[Do not use bzip2 compression (uses bzip2 by default)]),
 )
+
+BZIP="yes"
+AS_IF([test -n "$enable_bzip2" -a "$enable_bzip2" = "yes"],
+		[BZIP="$enable_bzip2"]
+)
+
 AS_IF([test "x$enable_bzip2" != "xno" ],
   [AC_DEFINE(SWUPD_WITH_BZIP2,1,[Use bzip2 compression])
-	 AC_CHECK_LIB([bz2], [BZ2_bzBuffToBuffCompress], [], [AC_MSG_ERROR([the libbz2 library is missing])])],
-  [AC_DEFINE(SWUPD_WITHOUT_BZIP2,1,[Do not use bzip2 compression])]
+	 AC_CHECK_LIB([bz2], [BZ2_bzBuffToBuffCompress], [], [AC_MSG_ERROR([the libbz2 library is missing])])
+	 BZIP="yes"],
+  [AC_DEFINE(SWUPD_WITHOUT_BZIP2,1,[Do not use bzip2 compression])
+	 BZIP="no"]
 )
 AC_ARG_ENABLE(
   [signature-verification],
   [AS_HELP_STRING([--enable-signature-verification], [Enable signature check (disabled by default)])],
   [AC_DEFINE([SIGNATURES], 1, [Enable signature check as default])]
+)
 
+SIGVERIFICATION="no"
+AS_IF([test -n "$enable_signature_verification" -a "$enable_signature_verification" = "yes" ],
+		[SIGVERIFICATION="$enable_signature_verification"]
 )
 
 if test "$enable_signature_verification" = "yes" ; then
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -18,8 +18,11 @@ PKG_CHECK_MODULES([lzma], [liblzma])
 PKG_CHECK_MODULES([zlib], [zlib])
 AC_ARG_ENABLE(
 	[bzip2],
-	AS_HELP_STRING([--disable-bzip2],[Do not use bzip2 compression (uses bzip2 by default)])
+	AS_HELP_STRING([--disable-bzip2],[Do not use bzip2 compression (uses bzip2 by default)]),
+	BZIP='no',
+	BZIP='yes'
```
phmccarty: The semantics of AC_ARG_ENABLE are confusing, so these simple assignments won't work as expected. The action in the third argument is run if either "--enable-bzip2" or "--disable-bzip2" is passed. Or, if neither option is passed, the action in the fourth argument is run.

I would recommend not setting either third or fourth arguments (the current state), and checking the "enable_bzip2" variable instead, which is automatically defined. See the AS_IF macro use below for an example.

configure.ac
```diff
@@ -29,19 +32,45 @@ AC_ARG_ENABLE(
   [signature-verification],
   [AS_HELP_STRING([--enable-signature-verification], [Enable signature check (disabled by default)])],
   [AC_DEFINE([SIGNATURES], 1, [Enable signature check as default])]
-
+    SIGVERIFICATION=${enableval}
```
phmccarty: I would do the same thing here, just read the "enable_signature_verification" variable to decide what to do later in the script.

configure.ac
```diff
@@ -29,19 +32,45 @@ AC_ARG_ENABLE(
...
 )
 
 if test "$enable_signature_verification" = "yes" ; then
 	AC_ARG_WITH([swupdcert],
 		[AS_HELP_STRING([--with-swupdcert=FILE], [swupd verification cert])],
-		[AC_DEFINE_UNQUOTED([SWUPDCERT], ["$withval"], [swupd verification cert])],
+		[AC_DEFINE_UNQUOTED([SWUPDCERT], ["$withval"], [swupd verification cert])]
+		  SWUPDCERT=${withval},
```
phmccarty: The configure is a little inconsistent with shell variable reference syntax, but we should standardize on one style, for readability.

Since the line above uses "$withval", I would use that here too.

configure.ac
```diff
@@ -29,19 +32,45 @@ AC_ARG_ENABLE(
...
+	  VERSIONURL=${withval},
+	VERSIONURL='!! Warning !! VersionURL not specified!'
+)
+
+AC_ARG_WITH([formatid],
+	[AS_HELP_STRING([--with-formatid=NUM], [Default format identifier])],
+	[AC_DEFINE_UNQUOTED([FORMATID], ["$withval"], [Default format identifier])]
+	  FORMATID=${withval},
+	FORMATID='!! Warning !! FormatID not specified!'
+)
```
phmccarty: For these three new options, I would prefer keeping with the existing convention for the AC_ARG_ENABLE options, using AS_IF to check values of variables that are already defined with AC_ARG_WITH. An example:

```
AS_IF([test "$with_formatid" = "yes"], [FORMATID="yes"],
        [test "$with_formatid" = "no"], [AC_MSG_ERROR([--without-formatid not supported])],
        [FORMATID='!! Warning !! FormatID not specified!']
)
```

Or, if you want to support --without-formatid, use:

```
AS_IF([test "$with_formatid" = "yes"], [FORMATID="yes"],
        [FORMATID='!! Warning !! FormatID not specified or set to "no"!']
)
```

bradTpeters:
Modified to use the AS_IF macro, avoiding the --without-X issue Patrick mentioned

configure.ac
phmccarty: @bradTpeters Can you make the same change for the AC_ARG_ENABLE uses you modified (i.e. set the shell variables within the appropriate AS_IF instead)?

configure.ac
```diff
@@ -29,7 +32,8 @@ AC_ARG_ENABLE(
   [signature-verification],
   [AS_HELP_STRING([--enable-signature-verification], [Enable signature check (disabled by default)])],
   [AC_DEFINE([SIGNATURES], 1, [Enable signature check as default])]
-
+    SIGVERIFICATION='yes',
+	SIGVERIFICATION='no'
```
phmccarty: Similarly for this change, throw it in an AS_IF.

configure.ac
```diff
@@ -39,9 +43,52 @@ if test "$enable_signature_verification" = "yes" ; then
...
+	[test "$with_formatid" = "yes"],	[FORMATID='!! Warning !! --with-formatid not specified!'],
+	[FORMATID='!! Warning !! --with-formatid not specified!']
+)
+
 AC_ARG_ENABLE(
   [tests],
-  [AS_HELP_STRING([--disable-tests], [Do not enable unit or functional test framework (enabled by default)])]
+  [AS_HELP_STRING([--disable-tests], [Do not enable unit or functional test framework (enabled by default)])],
+  TESTS='no',
+  TESTS='yes'
```
phmccarty: Similarly for this change, throw it in an AS_IF.

bradTpeters:
Latest adds support for the AC_ARG_ENABLE macro's as well

configure.ac
```diff
@@ -39,9 +49,55 @@ if test "$enable_signature_verification" = "yes" ; then
...
+	[FORMATID='!! Warning !! --with-formatid not specified!']
+)
+
 AC_ARG_ENABLE(
   [tests],
-  [AS_HELP_STRING([--disable-tests], [Do not enable unit or functional test framework (enabled by default)])]
+  [AS_HELP_STRING([--disable-tests], [Do not enable unit or functional test framework (enabled by default)])],
+)
+
+TESTS="no"
```
phmccarty: tests are enabled by default, so this variable should default to "yes"

configure.ac
```diff
@@ -18,8 +18,14 @@ PKG_CHECK_MODULES([lzma], [liblzma])
 PKG_CHECK_MODULES([zlib], [zlib])
 AC_ARG_ENABLE(
 	[bzip2],
-	AS_HELP_STRING([--disable-bzip2],[Do not use bzip2 compression (uses bzip2 by default)])
+	AS_HELP_STRING([--disable-bzip2],[Do not use bzip2 compression (uses bzip2 by default)]),
 )
+
+BZIP="no"
```
phmccarty: bzip2 support is enabled by default, so this variable should default to "yes"

bradTpeters:
Third try's a charm?

bradTpeters:
Patrick no +1? Did you have more feedback?

phmccarty:
No, I forgot to comment.

+1 from me on this latest revision.

tmarcu:
Merged 38839b8

PR #107, opened 2016-08-11
Improve output for miscellaneous curl errors

phmccarty:
Because libcurl has many different error codes, referring the end user
to curl.h is not ideal for the errors swupd does not handle explicitly.
Instead, call curl_easy_strerror() to print a generic error string as
indicated by the error code.


tmarcu:
+1

tmarcu:
Merged into master de2ef70

PR #108, opened 2016-08-12
Cleanups for certificate verification failure path

phmccarty:
This path series addresses several issues: two double free issues that lead to segfaults, two memory leaks, and an issue with error handling in swupd_init().

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -371,7 +371,7 @@ error:
 int verify_callback(int ok, X509_STORE_CTX *stor)
 {
 	if (!ok) {
-		fprintf(stderr, "Error: %s\n",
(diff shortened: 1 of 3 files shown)


src/signature.c
```diff
@@ -112,6 +112,7 @@ void terminate_signature(void)
 		fclose(fp_pubkey);
 	}
 	if (cert) {
+		X509_free(cert);
```
phmccarty: Running into a segfault now with this line added... My previous testing showed that this plugged a memory leak, but now I'm seeing invalid frees being reported.

I will remove this line.

phmccarty:
All but the last patch has been merged. I'll open a new PR with the corrected patch.

PR #110, opened 2016-08-16
Make pinned pubkey configurable

rojkov:
The server may move to a new location where a different pubkey
needs to be used.

Signed-off-by: Dmitry Rozhkov dmitry.rozhkov@linux.intel.com


tmarcu:
Removing certificate pinning from swupd in order to support the move to a CDN.

PR #111, opened 2016-08-18
Fix fp_pubkey double free

bradTpeters:
Fix double-free which happens when loading a key fails. This patch
ensures fp_pubkey cleanup only happens in terminate_signature().


bradTpeters:
This should be applied after Patrick's double free patch set

phmccarty:
This solution would work okay. Or, setting fp_pubkey to NULL after fclose() is called in both of these spots would also work.

tmarcu:
+1 will merge after following format bump.

tmarcu:
Merged into master 4a303e3

PR #112, opened 2016-08-22
Simplify branch code in hash_needs_work.

sofar:
These simple branches can be expressed in simply 1 line instead
of 5, without comprimising legibility.


phmccarty:
+1

tmarcu:
Merged into master 7ae10e2

PR #113, opened 2016-08-22
Text book implementation of `remove()`.

sofar:
The definition of `remove(2)` is exactly this implementation,
so this can be shortened.


phmccarty:
+1

tmarcu:
Merged into master 3cde6ff

PR #114, opened 2016-08-22
Statistics: inline counter increments.

sofar:
This is a negligable performance increase and more of a cleanup
of symbol space. In order for these to inline, we need the stats
to be exported instead, so we go from 8+1 exports to 1+1.

### DIFF
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -170,14 +170,47 @@ extern void link_manifests(struct manifest *m1, struct manifest *m2);
 extern void link_submanifests(struct manifest *m1, struct manifest *m2);
 extern void free_manifest(struct manifest *manifest);
 
-extern void account_new_file(void);
-extern void account_deleted_file(void);
-extern void account_changed_file(void);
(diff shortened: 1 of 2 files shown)


include/swupd.h
```diff
@@ -170,14 +170,48 @@ extern void link_manifests(struct manifest *m1, struct manifest *m2);
...
+{
+	swupd_stats[6]++;
+};
+
+static inline void account_delta_miss(void)
+{
+	swupd_stats[7]++;
+};
+
+extern int swupd_stats[];
```
phmccarty: swupd_stats was already declared on line 173, so this one can be removed.

sofar: Ahh yes, I moved it and forgot to drop the copy.

sofar:
Updated with Patrick's comment addressed.

phmccarty:
+1

tmarcu:
Merged into master 3cde6ff

PR #115, opened 2016-08-22
Namespace cleanup: make functions/vars static.

sofar:
Remaining hits by findstatic.pl are public API functions and
dangling statistic increments (but those last are fixed by the
PR cleaning up the stats, so they can be ignored for this PR).


phmccarty:
+1

tmarcu:
Merged into master 76c5686

PR #116, opened 2016-08-22
Unconditionally remove file after download if errors.

sofar:
We've created the convention that zero-sized packs are "already
processed" earlier and don't need to be re-downloaded. However,
if we have an unlikely failure that Curl created a zero-length
file, but was unable to write any bytes output to it (due to
e.g. network connection failure), then we'd leave a zero-sized
pack file on the system, potentially causing a redownload attempt
to skip all further re-downloads.

Ergo, we should always remove zero-sized pack files if any error
happened.


sofar:
Ignore this, I need to think this over some more.

PR #117, opened 2016-08-23
Fix a memory leak

phmccarty:
X509_STORE_CTX_free() is being called in the "verify success" path, but
not in the error path.


tmarcu:
+1 Merged

PR #118, opened 2016-08-23
Do not mark a pack file as "downloaded and extracted" on tar error.

sofar:
Because this code did not ever look at the return value of
`tar`, it is possible that a download succeeded to download
something (possibly a bad file on the server, etc.) properly,
and tar then failed to extract it (possible local problems
with disk space, a random memory failure, etc.).

When this condition happens, we should never leave an empty
pack file on the system, since we're sure that there was an
error extracting it, and the best course of action is to
either re-download it or download individual files. Of course,
the latter is likely going to happen anyway.

### DIFF
--- a/src/packs.c
+++ b/src/packs.c
@@ -78,11 +78,14 @@ static int download_pack(int oldversion, int newversion, char *module)
 	}
 	free(tar);
 	unlink(filename);
-	/* make a zero sized file to prevent redownload */
-	tarfile = fopen(filename, "w");
-	free(filename);
-	if (tarfile) {
-		fclose(tarfile);
+
+	if (err == 0) {
+		/* make a zero sized file to prevent redownload */
+		tarfile = fopen(filename, "w");
+		free(filename);
+		if (tarfile) {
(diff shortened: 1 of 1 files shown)


bryteise:
This change would cause the client to try and redownload bad server content (which shouldn't exist but given that this is the curl didn't get an error path...) and since these are packs, likely large amounts server content. The previous behavior would short circuit the redownload because even though there was an error returned and the file is queued to redownload, the download function would early return due to the zero length tar file being present.

After the pack download section the individual files would be downloaded to work around the bad packs.

I'd prefer to keep that behavior but add comments to make it clear the reason behind the current logic.

bryteise:
It will retry in the next pass but I don't think that helps since the fix would need to be done in a very small window (a few minutes at the longest) of first getting the tar error for it to make a difference for the end user running into a bad pack. The user is just as likely to have a faster experience skipping the pack redownload entirely and going to full files than hoping we fix the pack just as they are seeing the issue.

sofar:
Alright then, closing this.

PR #123, opened 2016-08-31
Move sign/verify to abort on failure

bradTpeters:
Moves swupd signature verification to mandatory, though --force
is supported in some cases, described here:

content trust down to the individual file level. When a client-side
files hash is found to differ from that expected given by the MoM,
we will abort or fix, as described:

 - During Update: Any sig or hash failure: Report and abort. Can be forced.
 - During Verify: Any sig or hash failure: Warn
 - During Verify fix: Any sig or hash failure: Autofix

### DIFF
--- a/include/swupd-error.h
+++ b/include/swupd-error.h
@@ -22,5 +22,6 @@
 #define EREQUIRED_DIRS 19       /* Cannot create required dirs */
 #define ECURRENT_VERSION 20     /* Cannot determine current OS version */
 #define ESIGNATURE 21           /* Cannot initialize signature verification */
+#define EHASH 22           /* Expected hash value does not match actual  */
 
 #endif
--- a/src/download.c
+++ b/src/download.c
@@ -251,6 +251,7 @@ int untar_full_download(void *data)
 	struct stat stat;
 	int err;
 	char *tarcommand;
+	char *log_cmd = NULL;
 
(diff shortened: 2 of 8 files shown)


bradTpeters:
This new pull request is rebased on recent patches, and modified from the previous PR.

As discussed face to face, I am seeing a strange failure case in which, on verify with a invalid cert, I see a double free with the following. Currently investigating, insights welcome:
----

Failed PEM_read_X509() for /usr/share/clear/update-ca/ClearLinuxRoot.pem
139941999204120:error:0D07207B:asn1 encoding routines:ASN1_get_object:header too long:asn1_lib.c:157:
139941999204120:error:0D068066:asn1 encoding routines:ASN1_CHECK_TLEN:bad object header:tasn_dec.c:1185:
139941999204120:error:0D07803A:asn1 encoding routines:ASN1_ITEM_EX_D2I:nested asn1 error:tasn_dec.c:374:Type=X509
139941999204120:error:0906700D:PEM routines:PEM_ASN1_read_bio:ASN1 lib:pem_oth.c:83:
Failed to load valid certificate. Attempting to proceed
Verifying version 10170
Curl error: Out of memory
Error: Network issue, unable to download manifest
*** Error in `/home/swupd-client.1496.move2mandatory/.libs/lt-swupd': double free or corruption (top): 0x00000000009f1e20 ***

Issue #94, opened 2016-07-18
verify_fix_path not called for manifests


tpepper:
This function is meant to be a generic helper to insure integrity and survive-ability.  It is not applied to manifests that don't hash check and actually should be.

fenrus75:
As a general design rule it is a MUST that each object for which the hash or signature check fails, gets deleted from the OS filesystem before the client application exits. It is NEVER ok to leave known-bad content on the disk, both from general security principles angle as well as from a recoverability angle (one bad download should never be able to brick the machine)

tpepper:
to elaborate on this, there are three cases of verification/failure/recovery:

# MoM signature fails to verify: recover is reinstall system or run with --force (NOTE: a PR for the --force feature is soon to be coming from @bradTpeters)
# bundle Manifest hash does not match expected from the MoM: auto recover by unlinking file and redownloading it
# file hash does not match expected from the bundle Manifest: auto recover by unlinking file and redownloading it

The first item there is already in the works.  The third item there is already implemented.  This issue is about the middle item.

fenrus75:
I think we also should at least try to redownload the MoM once or twice before giving up.

tpepper:
Yes.../me just edited to clarify.  My intent was to call out that we have a basic pattern of "attempt auto recover by unlinking file and redownloading it".  Beyond that, there are additional options, depending on the case.  The above comment should capture that better now.

matthewrsj:
This issue was addressed in https://github.com/clearlinux/swupd-client/commit/ba4f80034d9bef9ba7c6bd832268d9f661cc1727

PR #125, opened 2016-09-01
Recover from missing or bad bundle manifests files.

sofar:
This addresses #94.

We can't verify that bundles are correct if a bundle manifest we need is missign or corrupted.

We first check if these files are missing, and download them before verifying them.

If they are present or have just been downloaded, we can verify the hash checksum. If this fails to verify, we immediately erase the bad file and return an error code.

Because the overlying routines properly do retries, we'll get back to the start of this routine again and download the proper bundle manifest file, completing the recovery.

### DIFF
--- a/src/hash.c
+++ b/src/hash.c
@@ -32,6 +32,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#include "swupd-build-variant.h"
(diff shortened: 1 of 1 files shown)


phmccarty:
This PR was merged a few days ago (commit ba4f800).

PR #126, opened 2016-09-01
Remove unused function

phmccarty:
A recent commit removed the declaration for this function in swupd.h,
and now the compiler warns about it being unused, so the function can be
removed entirely now.


sofar:
lol

PR #129, opened 2016-09-06
Move signature verification to Mandatory

bradTpeters:
Moves swupd signature verification to mandatory, though --force
is supported in some cases, described here:

content trust down to the individual file level. When the signature
of the top-level MoM is invalid or cannot be verified for any reason,
we warn and abort the operation.
  --force will allow this to proceed, explicitly accepting the
	unverified MoM and outputting a log entry to the Journal to this
	effect. This may introduce a security hole

For files being downloaded from a Clear server, the downloaded file
must match the MoM hash, or a warning is emitted and the operation is
aborted.
  --force has no effect

### DIFF
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -64,6 +64,7 @@ struct version_container {
 
 struct header;
 
+extern bool badcert;
 extern bool force;
 extern int verbose;
 extern int update_count;
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -579,6 +579,7 @@ struct manifest *load_mom(int version)
 	int ret = 0;
 	char *filename;
 	char *url;
+	char *log_cmd = NULL;
 
 	ret = retrieve_manifests(version, version, "MoM", NULL);
 	if (ret != 0) {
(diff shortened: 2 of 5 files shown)


bradTpeters:
This v3 implementation is actually far smaller than the prior patch; retaining the same behavior for all subcommands really simplified the logic.

bradTpeters:
A follow-on patch, which I'm working on, will add retry logic, but that will be an independent patch

src/manifest.c
```diff
@@ -589,7 +590,17 @@ struct manifest *load_mom(int version)
 	string_or_die(&filename, "%s/%i/Manifest.MoM", state_dir, version);
 	string_or_die(&url, "%s/%i/Manifest.MoM", content_url, version);
 	if (!download_and_verify_signature(url, filename)) {
-		printf("WARNING!!! FAILED TO VERIFY SIGNATURE OF Manifest.MoM\n");
+		if (!force) {
+			printf("WARNING!!! FAILED TO VERIFY SIGNATURE OF Manifest.MoM\n");
```
phmccarty: "filename" and "url" will leak memory here

src/manifest.c
```diff
@@ -589,7 +590,17 @@ struct manifest *load_mom(int version)
...
+			return NULL;
+		} else {
+			printf("FAILED TO VERIFY SIGNATURE OF Manifest.MoM. Operation proceeding due to\n"
+				"  --force, but system security may be compromised\n");
+			string_or_die(&log_cmd, "echo \"swupd security notice:"
+				" --force used to bypass MoM signature verification failure\" | systemd-cat");
+			(void) system(log_cmd);
```
phmccarty: code style: no intervening space between (void) and system().

bradTpeters:
Thanks Patrick - updated accordingly

tpepper:
Definitely like the simplification, even if getting there meant some churn.  We're in a better place.

Can you run the code through valgrind for a positive and negative signature?

include/swupd.h
```diff
@@ -64,6 +64,7 @@ struct version_container {
 
 struct header;
 
+extern bool badcert;
```
tpepper: A minor nit, cut this is more self-descriptive if named "bad_mom_sig".

bradTpeters: This isn't meant to indicate a bad signature though; if you look at initialize_signature, all it does is load the cert and check for any revocations in the cert chain. Thus a failure is just an indication of a "badcert".  The follow-on logic, and the way I use this flag, is to indicate that when we have a bad cert, do not bother with a signature verification

phmccarty:
Brad's v4 patch is tracked in #140

PR #130, opened 2016-09-07
Fix swupd operations when curl is not HTTP2-enabled

phmccarty:
To support running swupd on distros without an HTTP2-enabled curl
package, explicitly check for the "unsupported protocol" error, returned
by libcurl in this event.


phmccarty:
This fixes running the test suite on Arch Linux, and also works okay on Clear Linux.

tmarcu:
+1 Merged f442666

PR #132, opened 2016-09-07
Update fallback preference for global configuration

bradTpeters:
Updates the order of preference for content_url, version_url
and formate_string, from:
  1. Runtime flags
  2. Configure time options
  3. State dir files

to:
  1. Runtime flags
  2. State dir files
  3. Configure time options

This patch also changes the logic of the setter functions to allow
multiple calls. Once the respective global is set, that function will
return true or 0, ie "success", (depending on the function).

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -127,35 +127,31 @@ static int set_url(char **global, char *url, const char *path)
 		} else {
 			/* no option passed; use the default value */
 			ret = set_default_value(global, path);
-			if (ret < 0) {
-				return ret;
-			}
+			return ret;
 		}
 	}
 
 	return ret;
 }
 
-void set_content_url(char *url)
+int set_content_url(char *url)
 {
-	int ret;
-
-	ret = set_url(&content_url, url, default_content_url_path);
-	if (ret < 0) {
-		printf("\nDefault content URL cannot be read. Use the -c option instead.\n");
-		exit(EXIT_FAILURE);
+	if (content_url) {
+		/* Only set once; we assume the first successful set is the best choice */
+		return 0;
 	}
+
+	return set_url(&content_url, url, default_content_url_path);
 }
 
-void set_version_url(char *url)
+int set_version_url(char *url)
 {
-	int ret;
-
-	ret = set_url(&version_url, url, default_version_url_path);
-	if (ret < 0) {
-		printf("\nDefault version URL cannot be read. Use the -v option instead.\n");
-		exit(EXIT_FAILURE);
+	if (version_url) {
+		/* Only set once; we assume the first successful set is the best choice */
+		return 0;
 	}
+
+	return set_url(&version_url, url, default_version_url_path);
 }
 
 static bool is_valid_integer_format(char *str)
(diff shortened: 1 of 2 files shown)


src/globals.c
```diff
@@ -127,35 +127,31 @@ static int set_url(char **global, char *url, const char *path)
 		} else {
 			/* no option passed; use the default value */
 			ret = set_default_value(global, path);
```
phmccarty: Can you rename this function to `set_default_value_from_path` instead, to better match the new semantics?

src/globals.c
```diff
@@ -198,6 +194,10 @@ bool set_format_string(char *userinput)
 {
 	int ret;
 
+	if (format_string) {
+		return true;
+	}
```
phmccarty: This should not be needed; see lines 220-222.

src/globals.c
```diff
@@ -318,25 +318,47 @@ bool init_globals(void)
...
+	3. Configure time settings
+*/
+	if (!set_format_string(NULL)) {
 #ifdef FORMATID
-	set_format_string(FORMATID);
-#else
-	(void)set_format_string(NULL);
-#endif /* FORMATID */
+		/* Fallback to configure time format_string if other sources fail */
+		set_format_string(FORMATID);
```
phmccarty: Perhaps we should add error handling here, and move the `printf("\nDefault format ...)` message to this function instead?

src/globals.c
```diff
@@ -318,25 +318,47 @@ bool init_globals(void)
...
+		ret = set_version_url(VERSIONURL);
+		if (ret) {
+			printf("\nDefault version URL not found. Use the -v option instead.\n");
+			exit(EXIT_FAILURE);
+		}
 #else
-	set_version_url(NULL);
-#endif /* VERSIONURL */
+		/* We have no choice but to fail */
+		printf("\nDefault version URL not found. Use the -v option instead.\n");
```
phmccarty: I would prefer not to see duplicate identical printf's like this... Can you move the printf and exit outside the ifdef block?

src/globals.c
```diff
@@ -318,25 +318,47 @@ bool init_globals(void)
...
+		ret = set_content_url(CONTENTURL);
+		if (ret) {
+			printf("\nDefault content URL not found. Use the -c option instead.\n");
+			exit(EXIT_FAILURE);
+		}
 #else
-	set_content_url(NULL);
-#endif /* CONTENTURL */
+		/* We have no choice but to fail */
+		printf("\nDefault content URL not found. Use the -c option instead.\n");
```
phmccarty: Similar comment here (duplicate printf's and exit).

phmccarty:
In the commit description, "formate_string" should be "format_string".

src/globals.c
bradTpeters: I did miss that, but it looks like the check on 220 only catches the NULL case. Given the new approach, which allows for multiple calls with a non-NULL arg, I need to move that check up to cover userinput being non-NULL as well.

src/globals.c
bradTpeters: Agreed, good idea

bradTpeters:
New PR at https://github.com/clearlinux/swupd-client/pull/138

PR #133, opened 2016-09-08
Fix hysteresis logic for fullfile downloads

phmccarty:
This is a first pass at properly enforcing the hysteresis for the fullfile downloads. Some initial testing looks promising, so I'm opening this PR now for early feedback.

### DIFF
--- a/src/download.c
+++ b/src/download.c
@@ -46,6 +46,7 @@
  */
 
 static CURLM *mcurl = NULL;
+static size_t mcurl_size = 0;
 static struct list *failed = NULL;
 
 /*
@@ -126,8 +127,8 @@ static int swupd_curl_hashmap_insert(struct file *file)
 }
 
 /* hysteresis thresholds */
-static int MAX_XFER = 25;
-static int MAX_XFER_BOTTOM = 15;
+static size_t MAX_XFER = 25;
+static size_t MAX_XFER_BOTTOM = 15;
 
 int start_full_download(bool pipelining)
 {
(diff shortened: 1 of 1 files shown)


phmccarty:
For reference, I have a `hysteresis-debug` branch for debugging the hysteresis behavior.

src/download.c
```diff
@@ -434,23 +438,56 @@ static int perform_curl_io_and_complete(int *left)
...
+		int numfds = 0;
+
+		// Wait for activity on the multi-stack
+		curlm_ret = curl_multi_wait(mcurl, NULL, 0, 500, &numfds);
+		if (curlm_ret != CURLM_OK) {
+			return -1;
+		}
+
+		// Either a timeout was hit, or no events to process
+		if (numfds <= 0) {
```
bradTpeters: Can numfds ever be < 0? Should probably be == 0

src/download.c
```diff
@@ -434,23 +438,56 @@ static int perform_curl_io_and_complete(int *left)
...
+			curlm_ret = curl_multi_perform(mcurl, &running);
+			if (curlm_ret != CURLM_OK) {
+				return -1;
+			}
+
+			/* If there are no actively running transfers, we can
+			 * be sure there is no work left to do. */
+			if (running == 0) {
```
bradTpeters: This logic seems valid, though the comment may be more clear if you say "there are no longer any transfers in progress". 'actively running' rather implies a bits are being downloaded, and if not, that transfer is not counted as running

src/download.c
```diff
@@ -420,11 +419,16 @@ static int perform_curl_io_and_complete(int *left)
...
-	}
 
-	curlm_ret = curl_multi_perform(mcurl, left);
-	if (curlm_ret != CURLM_OK) {
-		return -1;
+		/* "bounded" is false when the remainder of the multi-stack is
+		 * to be processed, ignoring the hysteresis bound. */
+		if (bounded) {
+			left--;
+		}
```
bradTpeters: Would be helpful to comment on the difference between left and mcurl_size

phmccarty: I added a docstring for the function that describes the "numfds" parameter, and "left" is initialized to its value. Maybe I could decrement "numfds" instead. Would that be more clear?

src/download.c
phmccarty: Since the last argument to `curl_multi_wait()` is the address of an `int` (signed), I wanted to cover the < 0 case for this reason. The < 0 case makes no sense to me, but I wanted to inspect the curl source to verify it will return >= 0 in all conditions.

src/download.c
phmccarty: Thanks, I like that rewording. Will fix.

src/download.c
bradTpeters: The curl documentation certainly specifies a >=0 return, but probably good to double check. If it somehow were < 0, we could not even be sure our behaviour here is correct

src/download.c
bradTpeters: Yes I think that would be more clear, if it is equivalent

src/download.c
phmccarty: After inspecting the source, it looks like `numfds` can never be < 0, so I'll modify the check.

src/download.c
phmccarty: Okay, I'll clarify it.

phmccarty:
Uploaded patch set 2, addressing Brad's comments.

phmccarty:
I tested this patch a few more times in the past couple of days, and everything looks okay to me. In fact, I'm seeing some performance gain for fullfile downloads with this patch compared to current master.

phmccarty:
@tmarcu merged this change (commit 981e053)

PR #134, opened 2016-09-13
Make remove() error messages correct and useful.

sofar:
We do not want to print this message if there was no file to begin
with, so check that errno isn't ENOENT.

We also want to print strerror(errno) instead of a code, and last,
we want to print the right filename here as well.

This is due to commit f06bcdba6d5b994a426cf23c76b97c79989f6427.

We should seriously consider also reverting that commit, as there
is almost no value in these error messages.

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -102,6 +102,7 @@ int rm_staging_dir_contents(const char *rel_path)
 void unlink_all_staged_content(struct file *file)
 {
 	char *filename;
+	int ret;
 
 	/* downloaded tar file */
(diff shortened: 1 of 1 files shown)


phmccarty:
I am in favor of reverting f06bcdb, mainly because `unlink_all_staged_content()` was intended to be called speculatively, ignoring any errors that may arise from removing the content.

sofar:
@tmarcu I'm OK with reverting as well. If you end up doing the revert, just close this one.

tmarcu:
Reverting and voiding the call so we can ignore the output in this function.

tmarcu:
07fc972

PR #135, opened 2016-09-13
Stop using readdir_r, switch to recommended readdir() instead.

sofar:
Man 3 readdir_r states: "This function is deprecated; use readdir(3)
instead."

Further, it also reads:

```
*  In the current POSIX.1 specification (POSIX.1-2008), readdir(3) is
   not required to be thread-safe.  However, in modern
   implementations (including the glibc implementation), concurrent
   calls to readdir(3) that specify different directory streams are
   thread-safe.  Therefore, the use of readdir_r() is generally
   unnecessary in multithreaded programs.  In cases where multiple
   threads must read from the same directory stream, using readdir(3)
   with external synchronization is still preferable to the use of
   readdir_r(), for the reasons given in the points above.
```

I don't expect swupd-client to be used with anything other than
glibc at this moment, and I don't see that the current code is
unsafely being used in multithreaded code, so we should avoid
using the deprecated interface.

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -58,11 +58,10 @@ void check_root(void)
 int rm_staging_dir_contents(const char *rel_path)
 {
 	DIR *dir;
-	struct dirent entry;
-	struct dirent *result;
+	struct dirent *entry;
 	char *filename;
 	char *abs_path;
-	int ret;
+	int ret = -1;
 
 	string_or_die(&abs_path, "%s/%s", state_dir, rel_path);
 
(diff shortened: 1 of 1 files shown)


sofar:
Needs testing.

src/helpers.c
```diff
@@ -72,17 +71,22 @@ int rm_staging_dir_contents(const char *rel_path)
...
 	}
 
-	while ((ret = readdir_r(dir, &entry, &result)) == 0) {
-		if (result == NULL) {
+	errno = 0;
+	while(true) {
+		entry = readdir(dir);
+		if (!entry) {
+			if (errno) {
+				ret = errno;
```
tmarcu: If errno (for some reason) isn't set then we don't set ret and return an uninitialized int. We should initialize it to -1 and also set it to 0 after the last free(filename)  - in case no files were removed or it's an empty directory.

src/helpers.c
```diff
@@ -402,18 +405,23 @@ static int swupd_rm_dir(const char *path)
...
 	}
 
-	while ((ret = readdir_r(dir, &entry, &result)) == 0) {
-		if (result == NULL) {
+	errno = 0;
+	while (true) {
+		entry = readdir(dir);
+		if (!entry) {
+			if (!errno) {
+				ret = errno;
```
tmarcu: Same comment as above.

tmarcu:
Aside from the comments I agree with the change. Even in a multithreaded mode swupd shouldn't have threads accessing the same dir.

sofar:
Updated - made the changes to `ret` as requested.

PR #137, opened 2016-09-15
Don't print signature check result if OK.

sofar:
This adds almost no value, since any failure results in a
thorough error message and possibly exit or retries.


bradTpeters:
I disagree; positive affirmation of a secure operation has a lot of value IMO.  Certainly worth 1 line of output

sofar:
There are hundreds of things that succeed in an update, possibly thousands. The only success message that matters is the one at the end that indicates that everything succeeded.

Also, this is printed twice, not just once.

bradTpeters:
I do see your point. If we're that worried about minimizing output, +1

phmccarty:
+1

tmarcu:
+1 Merged 026de00, this was a helpful message when debugging correctness.

PR #138, opened 2016-09-27
Update fallback preference for global configuration

bradTpeters:
Updates the order of preference for content_url, version_url
and format_string, from:
  1. Runtime flags
  2. Configure time options
  3. State dir files

to:
  1. Runtime flags
  2. State dir files
  3. Configure time options

This patch also changes the logic of the setter functions to allow
multiple calls. Once the respective global is set, that function will
return true or 0, ie "success", (depending on the function).

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -69,7 +69,7 @@ static const char *default_version_url_path = "/usr/share/defaults/swupd/version
 static const char *default_content_url_path = "/usr/share/defaults/swupd/contenturl";
 static const char *default_format_path = "/usr/share/defaults/swupd/format";
 
-static int set_default_value(char **global, const char *path)
+static int set_default_value_from_path(char **global, const char *path)
 {
 	char line[LINE_MAX];
 	FILE *file;
@@ -81,7 +81,6 @@ static int set_default_value(char **global, const char *path)
 
 	file = fopen(rel_path, "r");
 	if (!file) {
-		printf("Error: Unable to open %s\n", rel_path);
 		free(rel_path);
 		return ret;
 	}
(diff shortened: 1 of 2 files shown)


src/globals.c
```diff
@@ -318,25 +310,48 @@ bool init_globals(void)
...
 	}
 
-/* Set configuration defaults based on options provided
-	at configure time.  */
+/* Set defaults with following order of preference:
+	1. Runtime flags
+	2. State dir configuration files
+	3. Configure time settings
+*/
+	if (!set_format_string(NULL)) {
```
tmarcu: Without looking through the code, it's hard to tell why we're doing a set_format_string(NULL) and then dropping down into the baked in configure time setting. Maybe a comment stating that the NULL param call is actually saying "First try to apply the default config file path and do not pass any values in." Otherwise the code looks ok with the new priority changes.

phmccarty: Using NULL to mean "use the default value" was a convention I adopted when I refactored the globals code a while back. But I think I only documented it in commit descriptions.

src/globals.c
```diff
@@ -318,25 +310,48 @@ bool init_globals(void)
...
 #ifdef FORMATID
-	set_format_string(FORMATID);
+		/* Fallback to configure time format_string if other sources fail */
+		set_format_string(FORMATID);
```
phmccarty: `set_format_string()` can return false if the provided build-time value is not a positive integer. I would recommend in a future patch to cast the function call to (void) here, as it was previously, and add a configure-time check to ensure a positive integer value.

phmccarty:
+1

src/globals.c
bradTpeters: Sure Ill add a comment

bradTpeters:
Updated

src/helpers.c
```diff
@@ -566,7 +566,9 @@ void swupd_deinit(int lock_fd)
 	free_subscriptions();
 	free_globals();
 	v_lockfile(lock_fd);
+#if 0
```
phmccarty: debug code?

bradTpeters: Ugh yeah, good eye Patrick

bradTpeters:
Debug code removed. Blasted file descriptor leak ;)

phmccarty:
+1

src/globals.c
tmarcu: This should be noted as an issue maybe so we can track it?

phmccarty: Good point. I opened an issue (#141)

tmarcu:
+1 merged e8ae92a

PR #139, opened 2016-09-28
Do not report leaks in sockets.

sofar:
The only sockets held are irrelevant for our file descriptor
leak check, which focuses on leaks in opening/closing regular
files. Some of the sockets are things like pacrunner that
curl opens, and we should just ignore leaks in them.


sofar:
@bradTpeters this should do the trick.

phmccarty:
+1

bradTpeters:
Yep +1

tmarcu:
+1 and merged f071915

PR #140, opened 2016-09-30
Move signature verification to Mandatory

bradTpeters:
Moves swupd signature verification to mandatory, though --nosigcheck
enables user to bypass all signature-based security checks:

content trust down to the individual file level. When the signature
of the top-level MoM is invalid or cannot be verified for any reason,
we warn and abort the operation.
  --nosigcheck will allow this to proceed, explicitly accepting the
    unverified MoM and outputting a log entry to the Journal to this
    effect. This may introduce a security hole, since theoretically it
    becomes possible for update content to be installed which did not
    originate with the Clear Linux team.

For files being downloaded from a Clear server, the downloaded file
must match the MoM hash, or a warning is emitted and the operation is
aborted.

Signed-off-by: Brad T. Peters brad.t.peters@intel.com


src/signature.c
```diff
@@ -361,13 +387,13 @@ static bool validate_certificate(void)
...
 
-	return false;
+	return verify_ctx->error;
 }
 
 int verify_callback(int ok, X509_STORE_CTX *stor)
 {
 	if (!ok) {
-		fprintf(stderr, "Certificate verification error: %s\n",
+		printf("Certificate verification error: %s\n",
```
tmarcu: Why not print to stderr?

src/signature.c
```diff
@@ -361,13 +387,13 @@ static bool validate_certificate(void)
 		X509_STORE_CTX_free(verify_ctx);
 	}
 
-	return false;
+	return verify_ctx->error;
```
tmarcu: +1 on this, we print out the error in the callback so we should return it too for better debugging.

src/signature.c
```diff
@@ -260,12 +283,14 @@ static bool get_pubkey(void)
 	cert = PEM_read_X509(fp_pubkey, NULL, NULL, NULL);
 	if (!cert) {
 		fprintf(stderr, "Failed PEM_read_X509() for %s\n", CERTNAME);
+		fp_pubkey = NULL;
```
tmarcu: This was removed earlier to prevent a double free... https://github.com/clearlinux/swupd-client/pull/111/commits/e4f1e110cd0203984e39c708496ec045921e18df

src/signature.c
```diff
@@ -281,11 +306,12 @@ static bool get_pubkey(void)
...
  * Certificate Authority (CA)
  *
- * returns: true if certificate is valid, false otherwise */
-static bool validate_certificate(void)
+ * returns: 0 if certificate is valid, X509 store error code otherwise */
+static int validate_certificate(void)
 {
 	X509_LOOKUP *lookup = NULL;
 	X509_STORE_CTX *verify_ctx = NULL;
+	X509_VERIFY_PARAM *param;
```
phmccarty: This variable is unused

src/signature.c
```diff
@@ -64,14 +65,36 @@ static char *crl = NULL;
...
+	ret = validate_certificate();
+	if (ret) {
+		if (ret == X509_V_ERR_CERT_NOT_YET_VALID) {
+			/* If we can retrieve an approx. good system time, report out to user */
+			if (stat("/usr/lib/os-release", &statt) != -1) {
+				mod_sec = statt.st_mtim.tv_sec;
+				char timebuf[30];
+				alttime = localtime(&mod_sec);
+				strftime(timebuf, sizeof(timebuf), "%b %d, %Y", alttime);
+				printf("Failure: Cerficiate not yet valid. Check the system clock. Should be at least %s\n",
```
phmccarty: s/Cerficiate/Certificate

src/signature.c
```diff
@@ -64,14 +65,36 @@ static char *crl = NULL;
...
+
+	ret = validate_certificate();
+	if (ret) {
+		if (ret == X509_V_ERR_CERT_NOT_YET_VALID) {
+			/* If we can retrieve an approx. good system time, report out to user */
+			if (stat("/usr/lib/os-release", &statt) != -1) {
+				mod_sec = statt.st_mtim.tv_sec;
+				char timebuf[30];
+				alttime = localtime(&mod_sec);
+				strftime(timebuf, sizeof(timebuf), "%b %d, %Y", alttime);
```
phmccarty: I would prefer to use `%F`, which is a more neutral, standardized time/date format. For example, today would be 2016-11-18.

src/signature.c
phmccarty: swupd is not very consistent with printing errors to stderr versus not

src/signature.c
phmccarty: Don't we hit a file descriptor leak by setting these to NULL? `terminate_signature` will only call `fclose` if `fp_pubkey` is not NULL.

src/manifest.c
```diff
@@ -589,7 +590,19 @@ struct manifest *load_mom(int version)
...
+		if (sigcheck) {
+			printf("WARNING!!! FAILED TO VERIFY SIGNATURE OF Manifest.MoM\n");
+			free(filename);
+			free(url);
+			return NULL;
+		} else {
+			printf("FAILED TO VERIFY SIGNATURE OF Manifest.MoM. Operation proceeding due to\n"
+				"  --nosigcheck, but system security may be compromised\n");
+			string_or_die(&log_cmd, "echo \"swupd security notice:"
+				" --nosigcheck used to bypass MoM signature verification failure\" | systemd-cat");
```
phmccarty: My comment from the previous PR still holds: It would be nice to pass extra args to `systemd-cat` to provide more context. For example:

```
systemd-cat --priority="err" --identifier="swupd"
```

src/signature.c
tmarcu: We can keep this for consistency for now then. A revamp of the printing/error output seems needed overall.

PR #143, opened 2016-10-07
Set O_CLOEXEC when opening the certificate file

phmccarty:
Because swupd-client on Clear Linux may fork/exec clr-boot-manager near
the end of its operation, which itself checks for file descriptor leaks
before exiting, it will report leaks for files it did not open (i.e.
inherited from the parent process, swupd).

To resolve this, set O_CLOEXEC when opening the cert file using the "e"
flag (glibc extension).

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -250,7 +250,7 @@ error:
  * returns: true if it can get the public key, false otherwise */
 static bool get_pubkey(void)
 {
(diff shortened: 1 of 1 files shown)


phmccarty:
Tested on Clear build 10820 with this patch applied to v3.6.5.

PR #144, opened 2016-10-07
Replace system calls by local system_argv functions.

phmccarty:
There are new functions that replace system() calls: system_argv()
and system_argv_pipe(). They are based on execvp() function and they
can manage any argument including filenames with characters that could
be interpreted as meta-characters in a regular system() function.
So no escape for the arguments needed and is less prone to errors.

Signed-off-by: Jose R Guzman jose.r.guzman.mosqueda@intel.com
Signed-off-by: Patrick McCarty patrick.mccarty@intel.com

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -35,6 +35,7 @@
 #include <sys/statvfs.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <sys/wait.h>
 
 #include "config.h"
 #include "signature.h"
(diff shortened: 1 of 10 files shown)


phmccarty:
There are 10 failing functional tests with this patch, so those issues need looking at.

src/helpers.c
```diff
@@ -874,3 +877,167 @@ struct list *files_from_bundles(struct list *bundles)
...
+		if (ws == -1) {
+			fprintf(stderr, "Failed to wait for child process\n");
+			assert(0);
+		}
+
+		if (WIFEXITED(child_exit_status)) {
+			status = WEXITSTATUS(child_exit_status);
+		} else {
+			fprintf(stderr, "Child process didn't exit\n");
+			assert(0);
```
sofar: Uhh, this is just for testing, right? Whatever happens to subprocesses, we should really be sure that we can stop updating and not throw an error and abort.

phmccarty: Yes, this issue needs to be addressed. In particular, the child process may exit due to a signal, and swupd would incorrectly reach this code path in that situation...

phmccarty:
I rebased this branch, removed asserts, and resolved the issues with the functional tests not passing. It turns out that we needed to handle `execvp` failures, because the helper scripts often do not exist for tests.

phmccarty:
I just recalled there is a call to `popen()` as well, so we need a helper function for it.

tmarcu:
After further discussions this will not be implemented due to uncertainty about alternatives to our shell and tar usage.

PR #145, opened 2016-10-10
Initialize tmprenamedir in the state directory

phmccarty:
Initializing this directory here fixes an issue with creating custom
Clear Linux Docker images with swupd; when "tmprenamedir" doesn't exist
in the image layer and is later created in the container layer, a
rename() call will fail with EXDEV, due to the layers existing on
different overlayed filesystems. Ensuring that "tmprenamedir" is always
present in the state directory resolves the issue.

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -130,8 +130,8 @@ static int create_required_dirs(void)
 	int ret = 0;
 	int i;
 	char *dir;
-#define STATE_DIR_COUNT 3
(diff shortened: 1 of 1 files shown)


bryteise:
+1

phmccarty:
After further testing and understanding of how the Docker layering system works, this change doesn't fix the issue. Since the target file for the rename() will always be new, it will exist in the container layer, so EXDEV is still raised due to "tmprenamedir" being in the image layer.

PR #147, opened 2016-10-11
Remove staging content during verify_fix_path

bryteise:
When a file doesn't verify as part of verify_fix_path, prior to
downloading and staging the new file, first remove the old file staging
content.

This change was required due to a failure with swupd-client when run
inside Docker due to /var/lib/swupd being layered under multiple mount
points because of separate Docker Run commands. This caused rename(2) to
fail when the expected layout of /var/lib/swupd is to be the same mount
point.

With this change in place, the rename source content that would be from
a different mount is removed and redownloaded into the same mount as the
rename target.

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -782,11 +782,15 @@ int verify_fix_path(char *targetpath, struct manifest *target_MoM)
 		} else {
 			goto end;
 		}
+		/* In some circumstances (Docker using layers between updates/bundle adds,
+                 * corrupt staging content) we could have content which fails to stage.
+		 * In order to avoid this causing failure in verify_fix_path, remove the
+		 * staging content before proceeding. This also cleans up in case any prior
+		 * download failed in a partial state.
+		 */
+		unlink_all_staged_content(file);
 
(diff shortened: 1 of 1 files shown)


src/helpers.c
```diff
@@ -782,6 +782,12 @@ int verify_fix_path(char *targetpath, struct manifest *target_MoM)
 		} else {
 			goto end;
 		}
+		/* In some circumstances (Docker using layers between updates/bundle adds,
+                 * corrupt staging content) we could have content which fails to stage.
+                 * In order to avoid this causing failure in verify_fix_path, remove the
+                 * staging content before proceeding.
+                 */
+		unlink_all_staged_content(file);
```
phmccarty: `unlink_all_staged_content(file)` also unlinks the tar dotfile in /download/, so `unlink(tar_dotfile)` below can be removed.

bryteise: Updated to remove unlink and reworked code comment.

phmccarty:
Looks good. +1

PR #149, opened 2016-10-25
Refactoring subscription management, accounting, and update list creation

phmccarty:
This patch series addresses a number of issues:
- Subscription management for the 'update' subcommand. To make the consolidation/link steps simpler, there should the "from" subscription list and the "to" subscription list, instead of just the "to" list.
- Reporting "bundle" changes for updates instead of "manifest" changes
- Other accounting issues for update stats
- Reworking the update list creation logic. There is a major performance bug in the current code: hash checks are performed for every file that did not change.


src/manifest.c
```diff
@@ -808,10 +811,17 @@ void link_submanifests(struct manifest *m1, struct manifest *m2)
 			list2 = list2->next;
 
 			if (file2->last_change > m1->version && !file2->is_deleted) {
-				account_changed_manifest();
+				if (subbed1 && subbed2) {
+					account_changed_bundle();
+				}
```
sofar: `else if`?  these 2 conditions can never occur at the same time.

sofar:
Looks OK to me.

phmccarty:
Pushed patch set 2, addressing Auke's comment.

src/manifest.c
```diff
@@ -649,14 +649,34 @@ struct manifest *load_manifest(int current, int version, struct file *file, stru
...
+{
+	/* Not safe to perform the hash check if there was a type change
+	 * involving symlinks. */
+	if (file->is_link != file->peer->is_link) {
+		return false;
+	}
+
+	char *fullname = mk_full_filename(path_prefix, file->filename);
+	if (fullname == NULL) {
+		abort();
```
fenrus75: swupd shall not abort().

fenrus75: it still does not need to abort, it can just return FALSE in the function.

bryteise: I think just removing the check entirely makes sense here.

phmccarty: Sounds good. Will remove that check.

phmccarty:
Uploaded patch set 3.

PR #150, opened 2016-11-02
Add options to handle xattr at compile time.

icarus-sparry:
Swupd used to use a sorted blob of extended file attributes as part of
the data that was used to verify the contents. This caused problems if
extra attributes were added. In particular if a clearlinux system was
being run in a container that had selinux enabled in the base OS.

Add a configure option to allow the attributes to be considered or
not. Based on advice, the default is to ignore them, as this is what
everyone expects.

Tidied up passing a --selinux flag to tar, this is also handled by a new
config option.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -147,7 +147,7 @@ AC_ARG_ENABLE([tar-selinux],
 )
 AS_IF([test "x$enable_tar_selinux" = "xyes"],
 	    [AC_DEFINE(SWUPD_TAR_SELINUX,1,[give --selinux option to tar])
-	    AS_IF(test "x$enable_xattr" = "xyes",,
+	    AS_IF(test "x$XATTR" = "xno",
 	    echo "Must have --enable-xattr to have --enable-tar-selinux" >&2
 	    AS_EXIT(1))],
 	    [TARSELINUX=no]
--- a/configure.ac
+++ b/configure.ac
@@ -126,8 +126,31 @@ AC_ARG_ENABLE(
 	AS_HELP_STRING([--enable-bsdtar], [Use alternative bsdtar command (uses tar by default)])
 )
 AS_IF([test "x$enable_bsdtar" = "xyes" ],
-	[AC_DEFINE(SWUPD_WITH_BSDTAR, 1, [Use bsdtar])],
-	[AC_DEFINE(SWUPD_WITHOUT_BSDTAR, 1, [Use default tar])]
+	[AC_DEFINE(SWUPD_WITH_BSDTAR, 1, [Use bsdtar])]
+)
+dnl Enable extended attribute support
+XATTR="yes"
+AC_ARG_ENABLE(
+	[xattr],
+	AS_HELP_STRING([--enable-xattr],[Use extended file attributes (unused by default)])
+)
+AS_IF([test "x$enable_xattr" = "xyes"],
+	    [AC_DEFINE(SWUPD_WITH_XATTRS,1,[Use extended file attributes])
+	    AS_IF(test "x$enable_bsdtar" = "xyes",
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -126,8 +126,31 @@ AC_ARG_ENABLE(
...
+	    AS_EXIT(1))],
+	    [XATTR=no]
+)
+TARSELINUX="yes"
+AC_ARG_ENABLE([tar-selinux],
+	AS_HELP_STRING([--enable-tar-selinux],[give --selinux option to tar])
+)
+AS_IF([test "x$enable_tar_selinux" = "xyes"],
+	    [AC_DEFINE(SWUPD_TAR_SELINUX,1,[give --selinux option to tar])
+	    AS_IF(test "x$enable_xattr" = "xyes",,
```
phmccarty: I think this line would be clearer if written as:

```
AS_IF(test "x$enable_xattr" != "xno",
```
so that the empty second argument to AS_IF is not needed.

src/xattrs.c
```diff
@@ -37,6 +37,14 @@ enum xattrs_action_type_t_ {
 
 typedef enum xattrs_action_type_t_ xattrs_action_type_t;
 
+/* If SWUPD_WITH_XATTRS is not defined, create functions to override
+ * glibc's interface to the system calls
+ */
+#ifndef SWUPD_WITH_XATTRS
+#define lgetxattr(p,n,b,l) (-1)
+#define llistxattr(p,b,l) (0)
```
phmccarty: Is there a reason for expanding to `0` instead of `-1` here?

configure.ac
icarus-sparry: The logic is trying to say that it doesn't make sense to ask for --selinux to be passed to tar unless you are also asking for xattr to be recorded. I can try and make it clearer.

src/xattrs.c
icarus-sparry: If I am going to return -1, I should also set errno. So I am saying there are zero attributes, rather than saying there is an error trying to get them (perhaps ENOTSUP). As you can tell I am in two minds about which is the correct approach.

phmccarty: Thanks for clarifying. I think returning `0` would make more sense in this context. But now I'm wondering if `lgetxattr` should also return `0`? Unless zero-length xattr values ever make sense?

configure.ac
phmccarty: That makes sense. Your existing test here is correct, but I thought it would slightly more clear passing two args to `AS_IF` (test, action), rather than three (test, do nothing, action).

src/xattrs.c
tmarcu: While I think ENOTSUP makes sense in the error case, I'm more in favor of returning 0 (no xattrs to report) when it's not defined, as we don't care at that point and want to continue with the update.

icarus-sparry: The problem is that extended attributes are essentially a key/value store, with the value being a binary blob. Therefore you can have a zero length value,  (This is a space efficient way of storing binary data, just see if the key is defined, but don't store any data).

The lgetxattr call takes a filename and an attribute name, fills in a user provided buffer and returns the length of the attribute value.

Having said that, currently the only way that lgetxattr is called are the two calls inside xattr_get_value (which incidentally has a race condition if the attributes are being changed at the same time as this thing tries to read them). xattr_get_value is only called from xattrs_do_action, and only then if llistxattr gave it some names of attributes to retrieve. Since we say there are never any names, we never will call lgetxattr. The (-1) is just in there for defensive programming, and because I didn't want to start adding in more headers (errno.h).

icarus-sparry:
Additional comments in xattr.c.
Logic simplified in configure.ac

phmccarty:
Looks good to me. +1

PR #151, opened 2016-11-03
Change interface for functional test output checking

phmccarty:
Instead of checking entries in the "lines" array that BATS populates with swupd-client output, I decided to implement a new interface for checking output.

The motivation for this change is to ease maintenance of the swupd-client functional tests, especially when its output changes.

This interface requires each functional test to create a file named `lines-checked` that declares a list of checked lines of output, and when the test in run, the swupd-client output is inspecting to make sure every entry in `lines-checked` is present, and in order.

Additionally, the `lines-checked` file supports matching with regular expressions (the ERE supported by AWK) by prefixing the line with "REGEXP:", since several of the current tests that output paths with the full path_prefix need this treatment.


phmccarty:
One gap with this new interface is that it will not detect unexpected error messages... The checked lines could appear anywhere in the output as long as they are in the correct order.

I will experiment with adding a global ignore list containing regexps for messages that can be skipped for tests.

phmccarty:
Added the global ignore list support with enforcement, and I fixed up the remaining functional test issues, primarily with unchecked lines.

PR #152, opened 2016-11-03
Swupd performance

tmarcu:
Rebased patches from package to merge into main code base.

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -484,6 +484,11 @@ CURLcode swupd_curl_set_basic_options(CURL *curl, const char *url)
 		goto exit;
 	}
 
+	curl_ret = curl_easy_setopt(curl, CURLOPT_PIPEWAIT, 1);
+	if (curl_ret != CURLE_OK) {
+		goto exit;
+	}
+
 	if (update_server_port > 0) {
 		curl_ret = curl_easy_setopt(curl, CURLOPT_PORT, update_server_port);
 		if (curl_ret != CURLE_OK) {
--- a/src/download.c
+++ b/src/download.c
@@ -506,6 +506,8 @@ static int poll_fewer_than(size_t xfer_queue_high, size_t xfer_queue_low)
 	return 0;
 }
 
+extern int nonpack;
+
 /* full_download() attempts to enqueue a file for later asynchronous download
    - NOTE: See swupd_curl_get_file() for single file synchronous downloads. */
 void full_download(struct file *file)
@@ -525,6 +527,13 @@ void full_download(struct file *file)
 	} else if (ret < 0) { /* error */
 		goto out_bad;
 	} /* else (ret == 0)	   download needed */
+
+	/* Only print the first file so we don't spam on large misses */
+	if (nonpack == 0) {
+		printf("File %s was not in a pack\n", file->filename);
+	}
+	/* If we get here the pack is missing a file so we have to download it */
+	nonpack++;
 
 	curl = curl_easy_init();
 	if (curl == NULL) {
--- a/src/download.c
+++ b/src/download.c
@@ -149,7 +149,7 @@ int start_full_download(bool pipelining)
 	 * is even more broken than average. This at least will allow the user to update, albeit slowly.
 	 */
 	if (pipelining) {
-		curl_multi_setopt(mcurl, CURLMOPT_PIPELINING, 1);
+		curl_multi_setopt(mcurl, CURLMOPT_PIPELINING, CURLPIPE_MULTIPLEX | CURLPIPE_HTTP1);
 	} else {
 		/* survival: don't go too parallel in verify/fix loop */
 		MAX_XFER = 1;
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -162,7 +162,7 @@ static struct manifest *alloc_manifest(int version, char *component)
 	return manifest;
 }
 
-static struct manifest *manifest_from_file(int version, char *component, int header_only)
+static struct manifest *manifest_from_file(int version, char *component, bool header_only)
 {
 	FILE *infile;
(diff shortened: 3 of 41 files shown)


src/version.c
```diff
@@ -63,6 +68,7 @@ int get_latest_version(void)
 	free(path);
 	free(url);
 	free(tmp_version.version);
+	cached_version = ret;
```
phmccarty: It might be better to set `cached_version` outside the error path (i.e. immediately before the `out` label).

src/manifest.c
```diff
@@ -177,7 +177,7 @@ static struct manifest *manifest_from_file(int version, char *component)
 
 	string_or_die(&filename, "%s/%i/Manifest.%s", state_dir, version, component);
 
-	infile = fopen(filename, "rb");
+	infile = fopen(filename, "rbm");
```
phmccarty: It would nice to see this optimization mentioned in the commit message.

src/manifest.c
```diff
@@ -632,12 +637,13 @@ struct manifest *load_manifest(int current, int version, struct file *file, stru
 		return NULL;
 	}
 
-	ret = verify_bundle_hash(mom, file);
+	if (!header_only)
```
phmccarty: code style (needs curly brackets)

phmccarty: also, because `header_only` is not a bool or pointer, it would be more clear to use an explicit integer comparison.

src/curl.c
```diff
@@ -484,6 +484,9 @@ CURLcode swupd_curl_set_basic_options(CURL *curl, const char *url)
 		goto exit;
 	}
 
+	curl_easy_setopt(curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_2_0);
```
phmccarty: This option is already set 4 lines above (and with error handling).

src/curl.c
```diff
@@ -484,6 +484,9 @@ CURLcode swupd_curl_set_basic_options(CURL *curl, const char *url)
...
+	curl_easy_setopt(curl, CURLOPT_PIPEWAIT, 1);
```
phmccarty: needs error handling

src/download.c
```diff
@@ -149,7 +149,7 @@ int start_full_download(bool pipelining)
 	 * is even more broken than average. This at least will allow the user to update, albeit slowly.
 	 */
 	if (pipelining) {
-		curl_multi_setopt(mcurl, CURLMOPT_PIPELINING, 1);
+		curl_multi_setopt(mcurl, CURLMOPT_PIPELINING, CURLPIPE_MULTIPLEX | CURLPIPE_HTTP1);
```
phmccarty: this should be called out as a change for libcurl compatibility (1 and 0 were used in versions prior to 7.43.0, and the bitmask is used in newer versions).

src/download.c
```diff
@@ -525,6 +527,8 @@ void full_download(struct file *file)
 	} else if (ret < 0) { /* error */
 		goto out_bad;
 	} /* else (ret == 0)	   download needed */
+	if (!nonpack) printf("File %s was not in a pack\n", file->filename);
```
phmccarty: code style (curly brackets needed, with the `printf` on a separate line)

phmccarty: also, integer comparison would be preferable to using `!` here.

src/update.c
```diff
@@ -406,6 +408,8 @@ int main_update()
 	} else if (ret == 0) {
 		printf("Update complete. System already up-to-date at version %d\n", current_version);
 	}
+	if (nonpack)
+		printf("%i files were not in a pack\n", nonpack);
```
phmccarty: needs curly brackets

src/manifest.c
tmarcu: Why don't we just switch to bool?

phmccarty: I'm okay with switching to bool.

src/version.c
phmccarty: Looking at this again, I think setting cached_version to a negative value for the error case makes sense here; if the download is retried, then `swupd_curl_get_file` will be called again due to the negative cached_version.

tmarcu: +1 it propagates through correctly with a negative value.

tmarcu:
V2, updated code based on comments and updated several commit messages to be more verbose about the changes.

src/manifest.c
```diff
@@ -162,7 +162,7 @@ static struct manifest *alloc_manifest(int version, char *component)
 	return manifest;
 }
 
-static struct manifest *manifest_from_file(int version, char *component)
+static struct manifest *manifest_from_file(int version, char *component, int header_only)
```
phmccarty: `header_only` should be a bool

src/search.c
```diff
@@ -276,7 +276,7 @@ static void do_search(struct manifest *MoM, char search_type, char *search_term)
 		done_with_bundle = false;
 
 		/* Load sub-manifest */
-		subman = load_manifest(file->last_change, file->last_change, file, MoM);
+		subman = load_manifest(file->last_change, file->last_change, file, MoM, 0);
```
phmccarty: pass true/false as needed

src/search.c
```diff
@@ -439,9 +439,7 @@ static int download_manifests(struct manifest **MoM, struct list **subs)
 
 		if (access(untard_file, F_OK) == -1) {
 			/* Do download */
-			printf(" '%s' manifest...\n", file->filename);
-
-			subMan = load_manifest(current_version, file->last_change, file, *MoM);
+			subMan = load_manifest(current_version, file->last_change, file, *MoM, 0);
```
phmccarty: same here

phmccarty:
+1

PR #153, opened 2016-11-04
Remove certificate pinning

tmarcu:


phmccarty:
A commit message about why this is necessary would be helpful.

PR #154, opened 2016-11-07
Print error message for CURLE_COULDNT_CONNECT

phmccarty:
Certain types of network-connectivity problems hit this curl error, so
print a diagnostic message when it occurs.

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -319,6 +319,7 @@ exit:
 			err = -1;
 			break;
 		case CURLE_COULDNT_CONNECT:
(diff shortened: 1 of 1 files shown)


phmccarty:
The functional test suite emits many of these two curl errors, so we either need to ignore them or add special handling when using the file:// protocol.

phmccarty:
I decided to only print a message for the CURLE_COULDNT_CONNECT error, which seems to occur frequently for network connectivity issues.

PR #156, opened 2016-11-15
tar: extract failure handling

pohly:
tar errors when unpacking individual files used to be ignored in swupd
client < 3.0 and current 3.6.0 still prints a message about "ignoring
tar extract failure for fullfile ..." but the error is not really
getting ignored. Instead, downloading is tried again a few times
before swupd gives up.

This is relevant in Ostro OS where IMA prevents bsdtar from setting a
security.ima xattr when the value is redundant (see "[Linux-ima-user]
copying files with security.ima xattr" - the consensus is that the
kernel is too strict and needs to be fixed), leading to errors that
can be ignored.

The same can happen when extracting a pack file, and there errors also
should be ignored.

In both cases it is useful to see the actual error messages from the
tar command, so the redirection to /dev/null gets removed.

### DIFF
--- a/src/download.c
+++ b/src/download.c
@@ -290,7 +290,7 @@ int untar_full_download(void *data)
 	}
 
 	/* modern tar will automatically determine the compression type used */
-	string_or_die(&tarcommand, TAR_COMMAND " -C %s/staged/ " TAR_PERM_ATTR_ARGS " -xf %s 2> /dev/null",
+	string_or_die(&tarcommand, TAR_COMMAND " -C %s/staged/ " TAR_PERM_ATTR_ARGS " -xf %s",
 		      state_dir, tarfile);
 
 	err = system(tarcommand);
@@ -301,6 +301,7 @@ int untar_full_download(void *data)
 	if (err) {
 		printf("ignoring tar extract failure for fullfile %s.tar (ret %d)\n",
 		       file->hash, err);
+		err = 0;
 		goto exit;
 		/* FIXME: can we respond meaningfully to tar error codes?
 		 * symlink untars may have perm/xattr complaints and non-zero
--- a/src/packs.c
+++ b/src/packs.c
@@ -69,12 +69,17 @@ static int download_pack(int oldversion, int newversion, char *module)
 	free(url);
 
 	printf("Extracting pack.\n");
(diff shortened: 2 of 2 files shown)


src/download.c
```diff
@@ -301,6 +301,7 @@ int untar_full_download(void *data)
 	if (err) {
 		printf("ignoring tar extract failure for fullfile %s.tar (ret %d)\n",
 		       file->hash, err);
+		err = 0;
```
phmccarty: I agree it's helpful to see error messages for untar failures, but the intention _is_ to retry downloads in the failure case, and retries will not occur if the error code is reset to 0.

So I would prefer keeping the existing `err` value, and removing the `printf` on the line above. @tmarcu Thoughts?

src/packs.c
```diff
@@ -69,13 +69,18 @@ static int download_pack(int oldversion, int newversion, char *module)
...
 
 	err = system(tar);
 	if (WIFEXITED(err)) {
 		err = WEXITSTATUS(err);
 	}
+	if (err) {
+		printf("ignoring tar extract failure for '%s' (ret %d)\n",
+		       tar, err);
+		err = 0;
+	}
```
phmccarty: Same comment here as above.

pohly:
Treating tar errors as fatal would also be okay. I have backported the kernel patch which avoids the security.ima warnings and that seems to work okay.

src/download.c
tmarcu: I agree, we add to a failed list that is retried on multiple types of failures so err should be left alone. If we're still ignoring tar extract failures we should remove the printf, if that's something we don't want to ignore I'd change the printf to remove "ignoring" and make it a hard statement.

tmarcu:
Also in the spirit of survivability, we should try to recover and retry when possible before declaring it's fatal.

src/download.c
bryteise: Could go with "Problem extracting tar file, attempting redownload" and make sure we are cleaning the tar file (it looked like download.c didn't do this for full files just packs).

pohly:
I no longer need to ignore tar errors, so this change isn't needed anymore.

Can someone else take care of the broken printf()?

PR #157, opened 2016-11-15
swupd-client check-update return 1 if none available

icarus-sparry:
make the return code for swupd-client be 1 if there is no update
available. This follows in the tradition of grep, which returns 0 if
there are matches, 1 if there are not, and 2 for errors.

### DIFF
--- a/src/check_update.c
+++ b/src/check_update.c
@@ -123,6 +123,7 @@ err:
 	return false;
 }
 
+/* Return 0 if there is an update available, nonzero if not */
 static int check_update()
 {
 	int current_version, server_version;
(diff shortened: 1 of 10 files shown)


phmccarty:
+1

Can you implement the same change for the `update --status` command as well?

icarus-sparry:
Added the code for update --status.

icarus-sparry:
Also added the print of the current OS, which was talked about on IRC

icarus-sparry:
I need to figure out the github workflow so I can squash these 5 commits (including 2 merges) to 1.

icarus-sparry:
OK, that is better, only one commit to merge now (although I would accept a request to turn it into 2, one to add the printing of the current OS version in the check-update path and one to do the two return code changes).

phmccarty:
@icarus-sparry I'm okay with the single commit. +1 from me.

phmccarty:
One other thing: there are some functional test updates needed:

* 3 tests fail because the lines from `lines-checked` do not exactly match printed output. This is due to the added `printf()`. I would recommend updating `lines-checked` for the 3 tests.
* Now that the exit status for `update --status` and `check-update` tests will be either 0 or 1, the tests should explicitly check for the expected exit status.

phmccarty:
Looks good to me. +1 (again)

PR #160, opened 2016-11-17
fix --enable-xattr

pohly:
Commit fc0f570d added a check that prevents using --enable-xattr
together with --enable-bsdtar, perhaps because it was assumed that this
wouldn't work because there is no special tar option as in the GNU tar
case.

But bsdtar worked fined in combination with IMA and Smack xattrs in
Ostro OS. Xattrs are set accordingly there on the server side, so the
check needs to be removed to allow the combination.

Besides that, enabling xattrs also had no effect because xattrs.c
never got to see the SWUPD_WITH_XATTRS define due to not including
config.h.

@icarus-sparry this is the fix I mentioned on IRC.

### DIFF
--- a/include/swupd-build-variant.h
+++ b/include/swupd-build-variant.h
@@ -9,17 +9,27 @@
 #include "swupd.h"
 
 #ifdef SWUPD_WITH_BSDTAR
+
+/* bsdtar always includes xattrs, without needing special parameters
+ * for it. This has been tested in Ostro OS in combination with IMA
+ * and Smack, but not with SELinux.
+ */
 #define TAR_COMMAND "bsdtar"
 #define TAR_XATTR_ARGS ""
-#else
+
+#else /* SWUPD_WITH_BSDTAR */
+
+/* GNU tar requires special options, depending on how the OS uses xattrs.
+ * This has been tested in Clear Linux OS in combination with SELinux.
+ */
 #define TAR_COMMAND "tar"
-/* configure.ac ensures a sensible configuration of bsdtar/selinux/xattr */
 #ifdef SWUPD_TAR_SELINUX
 #define TAR_XATTR_ARGS "--xattrs --xattrs-include='*' --selinux"
 #else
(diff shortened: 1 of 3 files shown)


icarus-sparry:
Probably should delete the misleading comment in swupd-build-variant.h about configure.ac ensures a sensible configuration.

pohly:
@icarus-sparry I've replaced the comment in a separate commit. It wasn't that misleading (after all, configure still picks sensible combinations when using GNU tar), but there was indeed a bit more that could have been said about this.

include/swupd-build-variant.h
```diff
@@ -27,4 +36,4 @@
 #define TAR_PERM_ATTR_ARGS "--preserve-permissions "
 #endif
 
-#endif
+#endif /* SWUPD_WITH_BSDTAR */
```
phmccarty: This `#endif` closes the include guard, not SWUPD_WITH_BSDTAR. Can you move the comment to the correct spot?

pohly:
@phmccarty good catch, fixed commit force-pushed.

include/swupd-build-variant.h
```diff
@@ -25,6 +34,7 @@
 #define TAR_PERM_ATTR_ARGS "--preserve-permissions " TAR_XATTR_ARGS
 #else /* SWUPD_WITHOUT_XATTRS */
 #define TAR_PERM_ATTR_ARGS "--preserve-permissions "
-#endif
+
+#endif /* SWUPD_WITH_BSDTAR */
```
phmccarty: Sorry, I should have been more specific... The comment should be on line 31, correct?

pohly: Duh. Still not right. I was expecting the #ifdef SWUPD_WITH_XATTRS to be inside the section for GNU tar and hence placed my endif comment where it is now, which (after actually reading what it does) of course isn't right.

Third time is the charm - force-pushed.

phmccarty:
+1

tmarcu:
Thanks @pohly :+1:

PR #161, opened 2016-11-17
Add configure option to re-enable updating of config files

pohly:
This is required by distros which ship configuration files under
/etc. Those files then are considered read-only and must be updated as
part of OS updates.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -153,6 +153,15 @@ AS_IF([test "x$enable_tar_selinux" = "xyes"],
 	    [TARSELINUX=no]
 )
 
+AC_ARG_ENABLE(
+        [stateless],
+        AS_HELP_STRING([--disable-stateless],[OS is not stateless, do not ignore configuration files (stateless by default)])
+)
+AS_IF([test "x$enable_stateless" = "xno"],
+            [AC_DEFINE(OS_IS_STATELESS,0,[OS is not stateless])],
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -153,6 +153,13 @@ AS_IF([test "x$enable_tar_selinux" = "xyes"],
 	    [TARSELINUX=no]
 )
 
+AC_ARG_ENABLE(
+        [stateless],
+        AS_HELP_STRING([--disable-stateless],[OS is not stateless, do not ignore configuration files (stateless by default)]),
+        AC_DEFINE(OS_IS_STATELESS,0,[OS is not stateless]),
+        AC_DEFINE(OS_IS_STATELESS,1,[OS is stateless])
+)
```
phmccarty: There is one case not captured by the above macro: If `--enable-stateless` is passed, `OS_IS_STATELESS` will be set to 0.

Granted, one could hope that users will not pass options undocumented in `./configure --help`, but some people may want to always explicitly enable/disable all package-specific options.

pohly: @phmccarty you are right, the enable_val needs to be checked in the first case. I'll fix that.

pohly:
@phmccarty revised commit force-pushed.

phmccarty:
+1, thanks.

tmarcu:
+1

PR #162, opened 2016-11-17
ignore xattrs when verifying Manifest files

pohly:
When IMA or Smack are active on the client, the downloaded Manifest
files will be assigned certain xattrs (security.ima
resp. security.SMACK64). Those xattrs did not exist on the server side
(because it is most likely not having those kernel features enabled)
and besides, the swupd-server code wouldn't include them in the
Manifest hashes even if they existed (see write_manifest_plain() in
src/manifest.c).

Therefore the client must ignore xattrs when verifying Manifest files.
This is the only place where verification gets relaxed. All other locations
still use xattrs, just as before.

This is the solution that was agree upon when discussing https://github.com/clearlinux/swupd-server/pull/34

### DIFF
--- a/src/hash.c
+++ b/src/hash.c
@@ -236,7 +236,14 @@ bool verify_file(struct file *file, char *filename)
 	}
 
 	local->filename = file->filename;
-	local->use_xattrs = true;
+	/*
+	 * xattrs are currently not supported for manifest files.
+	 * They are data files produced by the swupd-server and
+	 * therefore do not have any of the xattrs normally
+	 * set for the actual system files (like security.ima
+	 * when using IMA or security.SMACK64 when using Smack).
+	 */
+	local->use_xattrs = !file->is_manifest;
 
 	populate_file_struct(local, filename);
 	if (compute_hash(local, filename) != 0) {


icarus-sparry:
First I have heard that ignoring xattrs was the agreed solution for verifying Manifest files. Where was it discussed?

pohly:
@icarus-sparry see https://github.com/clearlinux/swupd-server/pull/34#discussion_r85227494

phmccarty:
I like this change. The xattr support is meant for actual update *content*, not the manifest files that describe the update content.

+1

tmarcu:
I like the concept of this change, and it's the correct thing to do.
However, what do you think about not changing the function signature since the manifest is the -only- place we pass in false, and changing it such that verify_file() checks if (file->is_manifest): don't use xattrs? The filetype is populated so we always know what kind of file it is, and then we wouldn't have to sprinkle in a new param everywhere.

icarus-sparry:
@tmarcu, sorry, but I don't think this is the right thing to do. The point about xattrs is that they are extensible metadata. One suggested use for them is to assign an Author field to a document, another is for libraries such as libcurl to record where something was downloaded from. If xattrs every get used in this manner then we will need to provide a list of which attributes are to be included into the verification, so we can know to ignore the Download and Author fields, but keep the security.ima.

Once you have this need for every file, then providing an empty list for Manifests is easy.

pohly:
@icarus-sparry: there is no security.ima xattr on the Manifest files at the moment. What you describe (per xattr control) can be a future improvement, although using it for security.ima probably would need a good justification given also the complexity on the server side. In the meantime, can we please focus on a solution that has a chance of getting implemented soon?

@tmarcu yes, checking file->is_manifest also would work for me. Do you want me to provide (and test) a patch for that?

tmarcu:
@icarus-sparry I don't believe the current patch or the proposal solve the xattr list problem still. I agree supporting xattrs fully will require a more complex change to differentiate out of a list, but that might be out of scope for this patch since we ignore then fully as is. 
@pohly Could you do the file->is_manifest change? It's kind-of a rewrite of this patch, but the resulting patch should only be a couple lines in the end.

pohly:
@tmarcu using "is_manifest" indeed made the patch a lot simpler. Revised commit force-pushed.

I verified that this avoids the hash mismatch errors for the manifests in Ostro while still finding hash mismatches for system files.

PR #164, opened 2016-11-22
Move signature verification to mandatory

tmarcu:
This sets swupd signature verification to mandatory, failing and exiting
for any verify error, though --nosigcheck enables user to bypass all
signature-based security checks.

The signed MoM serves as the top level chain of trust, and it is used to
extend content trust down to the individual file level. When the signature
of the top-level MoM is invalid or cannot be verified for any reason, we
warn and abort the operation.

Passing --nosigcheck will allow this to proceed, explicitly accepting the
unverifiable MoM and outputting a log entry to the Journal. This is not
recommended and unsupported by upstream once the chain of security is broken,
because it may imply update content was or may be installed that was not
generated by the official upstream.


src/signature.c
```diff
@@ -64,14 +65,36 @@ static char *crl = NULL;
...
+				char timebuf[30];
+				alttime = localtime(&mod_sec);
+				strftime(timebuf, sizeof(timebuf), "%F", alttime);
+				printf("Failure: Certificate not yet valid. System clock should be at least %s\n",
+						timebuf);
+			} else {
+				printf("Failure: Certificate not yet valid, check system time\n");
+			}
+		} else {
+			printf("Failed to verify certificate: %s\n", X509_verify_cert_error_string(ret));
```
phmccarty: In the interest of making error strings discoverable, it would be nice for this `printf` to live on line 82. So the code would look like:

```
if (ret) {
	printf("Failed to verify certificate: %s\n", X509_verify_cert_error_string(ret));
	if (ret == X509_V_ERR_CERT_NOT_YET_VALID) {
		/* If we can retrieve an approx. good system time, report out to user */
		if (stat("/usr/lib/os-release", &statt) != -1) {
			mod_sec = statt.st_mtim.tv_sec;
			char timebuf[30];
			alttime = localtime(&mod_sec);
			strftime(timebuf, sizeof(timebuf), "%F", alttime);
			printf("System clock should be at least %s\n", timebuf);
		}
	}

	goto fail;
}
```

phmccarty:
Also, for the record, the functional test suite fails with this patch. @tmarcu will be working on resolving those failures.

phmccarty:
+1

PR #166, opened 2016-11-22
Do not block on systemctl post-update triggers

tmarcu:
We call systemctl to update triggers once an update finishes, which enqueues a
task and then waits to finish by default thus blocking the system() call. This
adds unnecessary time to the update, so set --no-block and move on since we
already reached the end of the update.


phmccarty:
+1

PR #167, opened 2016-11-23
Telemetry - re-introduce telemetry.

sofar:
This is a new telemetry framework for swupd-client. The patch is split up in a few changes:

- clean out some old telemetry stuff that was left around and is unused.
- Separate patch to always create new telemetry drop folder
- introduce new telemetry() function and use it in 3 locations (check, update, and verify paths) to create a telemetry record at completion of those actions.

This won't actually deliver records to the telemetry service. For that, a separate telemetry agent will be created that runs as a separate process and does not affect swupd operation. That "probe" will have its own repository.

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -135,8 +135,8 @@ static int create_required_dirs(void)
 	int ret = 0;
 	int i;
 	char *dir;
-#define STATE_DIR_COUNT 3
-	const char *state_dirs[] = { "delta", "staged", "download" };
+#define STATE_DIR_COUNT 4
+	const char *state_dirs[] = { "delta", "staged", "download", "telemetry" };
 	struct stat buf;
 	bool missing = false;
 
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -136,7 +136,7 @@ static int create_required_dirs(void)
 	int i;
 	char *dir;
 #define STATE_DIR_COUNT 3
-	const char *dirs[] = { "delta", "staged", "download" };
+	const char *state_dirs[] = { "delta", "staged", "download" };
 	struct stat buf;
 	bool missing = false;
 
@@ -146,7 +146,7 @@ static int create_required_dirs(void)
 		missing = true;
 	}
 	for (i = 0; i < STATE_DIR_COUNT; i++) {
-		string_or_die(&dir, "%s/%s", state_dir, dirs[i]);
+		string_or_die(&dir, "%s/%s", state_dir, state_dirs[i]);
 		ret = stat(dir, &buf);
 		if (ret) {
 			missing = true;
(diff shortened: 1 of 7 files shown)


sofar:
Fixed up all the telemetry messages starting with `action=`.

src/helpers.c
```diff
@@ -135,8 +135,8 @@ static int create_required_dirs(void)
 	int ret = 0;
 	int i;
 	char *dir;
-#define STATE_DIR_COUNT 3
-	const char *dirs[] = { "delta", "staged", "download" };
+#define STATE_DIR_COUNT 4
+	const char *dirs[] = { "delta", "staged", "download", "telemetry" };
```
tmarcu: This should maybe be renamed to state_dirs so it's obvious it's referencing the above define.

phmccarty: That would be an improvement, though it could be incorporated in a separate cleanup patch.

src/telemetry.c
```diff
@@ -0,0 +1,61 @@
...
+
+void telemetry(telem_prio_t level, const char *fmt, ...)
+{
+	va_list args;
+	char *filename;
+	int fd;
+
+	string_or_die(&filename, "%s/telemetry/%d.%d.XXXXXX", state_dir, RECORD_VERSION, level);
+
+	if (!filename) {
```
phmccarty: `filename` must be non-null at this point, because `string_or_die` aborts if memory allocation fails.

src/update.c
```diff
@@ -400,6 +404,22 @@ int main_update()
...
+		"current_version=%d\n"
+		"server_version=%d\n"
+		"result=%d\n"
+		"time=%5.1f\n",
+		current_version,
+		server_version,
+		ret,
+		delta);
+
+	if (server_version > current_version)
```
phmccarty: This `if` block needs surrounding curly brackets (code style)

src/telemetry.c
```diff
@@ -0,0 +1,61 @@
+
+/*
+ *   Software Updater - telemetry collection
+ *
+ *      Copyright © 2012-2016 Intel Corporation.
```
phmccarty: just 2016?

src/telemetry.c
sofar: ok, thanks, I'll remove this check.

src/update.c
sofar: yup!

sofar:
I've fixed up all the comments, and added 1 more commit to rename the `dirs` to `state_dirs`. I've inserted it before the series since that makes more sense.

src/helpers.c
sofar: Done, separate commit 2nd in this series.

phmccarty:
+1

PR #168, opened 2016-11-29
staging: better handle file type change from directory

andred:
If the file type changes from directory to something else,
we have to make sure to end up in the right place when
extracting it.

So far, we end up the directory case, but of course we
need to get into the case of non-directory.

Note that at least bsdtar handles extracting a new non-
directory over an existing directory with the same name
fine, i.e. the existing directory is correctly removed,
and the new file (or symlink) created instead.

Signed-off-by: André Draszik <git@andred.net>

### DIFF
--- a/src/staging.c
+++ b/src/staging.c
@@ -122,6 +122,13 @@ int do_staging(struct file *file, struct manifest *MoM)
 				ret = -ETYPE_CHANGED_FILE_RM;
 				goto out;
 			}
+			/* Since we have removed the old file, it doesn't
+			 * matter what type it used to be. This is in
+			 * particular important if file has changed from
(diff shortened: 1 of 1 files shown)


tmarcu:
We discussed this a bit yesterday to come up with a more complete solution to the problem of directory -> symlink. For the trivial case, empty dir -> symlink the code should work to just remove the dir and create a new link. However, considering the case of a directory that's populated, we need to do better:
- If dir is empty: rmdir & create link
- If dir is not empty: all files should be marked deleted, delete all marked files first & try to rmdir
- If leftover content remains: mv dir -> dir.timestamp, create the new link/file & continue.

PR #169, opened 2016-12-01
Cleanups for the configure.ac; make the certificate location configurable

phmccarty:
To better support running the functional test suite with signature
verification enabled, make the certificate location configurable. Note
that the basename of the certificate used for verification can be
configured separately with the --with-swupdcert=NAME option.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -91,7 +91,7 @@ swupd_sig_verifytest_LDADD = $(SWUPD_CORE_LIBS)
 
 noinst_HEADERS = $(top_srcdir)/include/*
 
-swupdcertsdir = @update_ca_certs_path@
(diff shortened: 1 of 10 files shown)


phmccarty:
I added a few additional cleanup patches for configure.ac before adding the new option.

bryteise:
Could we also get this path configurable through the command line?

tmarcu:
@bryteise The command line is getting bloated and ugly, what are the flags you propose to add?

bryteise:
@tmarcu -C --certificate-path

phmccarty:
Currently untested, but I implemented a -C/--certpath option for the subcommands to override the builtin default value.

phmccarty:
Rebased to latest master.

PR #170, opened 2016-12-06
Telemetry: new API, add bundle add/remove telemetry.

sofar:
- Change telemetry API - add class component. 
- Telemetry: add telemetry for bundle add, remove.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -316,7 +316,14 @@ int remove_bundle(const char *bundle_name)
 out_free_mom:
 	free_manifest(current_mom);
 out_free_curl:
-
+	telemetry(ret ? TELEMETRY_CRIT : TELEMETRY_INFO,
+		"bundleremove",
+		"bundle=%s\n"
+		"current_version=%d\n"
+		"result=%d\n",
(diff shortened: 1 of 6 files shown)


src/bundle.c
```diff
@@ -552,13 +560,29 @@ int install_bundles_frontend(char **bundles)
 
 	for (; *bundles; ++bundles) {
 		bundles_list = list_prepend_data(bundles_list, *bundles);
+		if (*bundles) {
+			if (bundles_list_str) {
+				string_or_die(&bundles_list_str, "%s, %s", bundles_list_str, *bundles);
```
phmccarty: You will want to `free(bundles_list_str)` before this line to avoid a memory leak.

phmccarty: Or rather, store the old `bundles_list_str` in a temp variable before freeing, and use the temp variable in the `string_or_die()`.

ahkok: Ok, I'll fix that.

ahkok:
Updated to fix a minor leak.

phmccarty:
+1

PR #173, opened 2017-01-05
Some error handling improvements; miscellaneous cleanups

phmccarty:
This series is a first pass at some error handling improvements, with the goal of using the same error codes for certain classes of errors, and to make sure the main program always exits with a positive value for errors.

I have not yet addressed the more complex error handling scenarios (e.g. in `src/staging.c`). That will come later.

One behavior change should be noted: for swupd operations that require zero packs, the operation will now fail if the packs fail to extract with `tar`. Previously, pack extraction issues were ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,9 +45,13 @@ swupd_*
 tap-driver.sh
 test-suite.log
 test/*.gcno
+test/functional/ClearLinuxRoot.pem
+test/functional/private.pem
+test/functional/*.log
 test/functional/*/*/web-dir/*/*.tar
 test/functional/*/*/web-dir/*/staged/*.tar
 test/functional/*/*/web-dir/*/files/*.tar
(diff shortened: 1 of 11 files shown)


phmccarty:
One other thing to note: `bundle-add` and `verify --install` will fail for tar extract failures. If that behavior is not desired, then I will need to address this issue.

bryteise:
@phmccarty Is that failure going to happen if there was a download problem (corrupt tar)? Assuming it fails if there was no download issue then that is correct behavior imo.

phmccarty:
@bryteise The new failure case is when the download succeeds (from libcurl's perspective), but the tar extract fails.

Also, my previous comment was incorrect. `bundle-add` doesn't check for pack download errors, so only `verify --install` would be affected by the new failure case.

PR #175, opened 2017-01-13
Add new bundle-list sub-command

MarioCarrilloA:
This group of patches add a new sub-command to swupd "bundle-list", this sub-command allows to show which bundles are installed in the local system.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -13,6 +13,7 @@ swupd_SOURCES = \
 	$(swupd_verify_SOURCES) \
 	$(clr_bundle_add_SOURCES) \
 	$(clr_bundle_rm_SOURCES) \
+	$(clr_bundle_ls_SOURCES) \
 	$(swupd_search_SOURCES)
 
 check_PROGRAMS = \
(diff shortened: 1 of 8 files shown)


sofar:
Please also update the man page to add the new subcommand:

https://github.com/clearlinux/swupd-client/blob/master/docs/swupd.1.md

bryteise:
It would be nice if the new subcommand got the bundle-add --list-bundles feature added to it as well (as an option) to keep the very similar features in a single place. Note this causes the need for the bundle-list subcommand to support a lot of the other common swupd options.

Should have tests for this change too.

phmccarty:
@bryteise I like that idea. `bundle-list` lists the installed bundles, and `bundle-list --all` could list all available bundles, with identical output to `bundle-add --list`.

MarioCarrilloA:
Thank you so much for feedback, I am going to work in a new commit about man page, and once these changes are merged, I am going to work in to move _bundle-add --list_ to _bundle-list_ sub-command :)

MarioCarrilloA:
@phmccarty @sofar @bryteise documentation commit done :)

sofar:
I'm OK with this as-is, but the question remains who will add the requested additional `--all` command.

MarioCarrilloA:
@sofar I will add it, once this request has been merged :)

PR #177, opened 2017-01-30
Remove .signed files in functional tests

phmccarty:
These files have never been used by the functional tests, so they can be safely removed from the tree.


bryteise:
Yaay!

tmarcu:
Thank you +10

PR #178, opened 2017-01-31
Fix swupd options for bundle-list --all unit test

MarioCarrilloA:
The swupd unit tests need a group of
options for execution environment, this
var was missed.

Signed-off-by: Mario Alfredo Carrillo Arevalo <mario.alfredo.c.arevalo@intel.com>


phmccarty:
It looks like bundle-list doesn't support the `-S/--statedir` option, according to test results.

MarioCarrilloA:
@phmccarty yes, this var uses some options that `swupd bundle-list` does not support, I will send a new set of patches for that

tmarcu:
+1, thanks for the quick fix!

PR #179, opened 2017-02-01
clr-boot-manager fixes

ikeydoherty:
This change adds a minor code-styling fix, but more importantly it moves `swupd-client` away from using the legacy compatibility scripts inside `clr-boot-manager`, enforcing direct usage of it.

Note that a potential issue with the previous behaviour is the compatibility scripts do not contain a bin prefix, i.e. `$root/usr/bin/*.sh` invoke `clr-boot-manager update $*`, meaning the host side tool is
employed rather than the rootfs one.

### DIFF
--- a/src/scripts.c
+++ b/src/scripts.c
@@ -49,8 +49,10 @@ static void update_boot(void)
 
 static void update_triggers(void)
 {
-	system("/usr/bin/systemctl --no-block daemon-reload");
-	system("/usr/bin/systemctl --no-block restart update-triggers.target");
+	__attribute__((unused)) int ret = 0;
+
+	ret = system("/usr/bin/systemctl --no-block daemon-reload");
+	ret = system("/usr/bin/systemctl --no-block restart update-triggers.target");
 }
 
 void run_scripts(void)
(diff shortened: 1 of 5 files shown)


ikeydoherty:
Well then.

ikeydoherty:
Redoing.

src/scripts.c
```diff
@@ -32,41 +32,18 @@
...
-
-	if (strcmp("/", path_prefix) == 0) {
-		string_or_die(&bootloader_update_cmd, "/usr/bin/systemdboot_updater.sh");
-	} else {
-		string_or_die(&bootloader_update_cmd, "%s/usr/bin/systemdboot_updater.sh --path %s", path_prefix, path_prefix);
-	}
-
-	system(bootloader_update_cmd);
-	free(bootloader_update_cmd);
+	system(boot_update_cmd);
```
phmccarty: The previous code didn't have this, but can you cast to (void) to make it clear we're not checking the return value?

ikeydoherty: void isn't sufficient enough with my local build so I'll add an unused int

phmccarty: Yeah, it doesn't silence GCC, but at least reading through the code it becomes obvious.

ikeydoherty:
Force pushed new set

ikeydoherty:
Oooo I know why it failed for me before - I suffer the `umask` issue on my host.. :)

ikeydoherty:
Merci

PR #181, opened 2017-02-03
Fix swupd_lock file descriptor leaking

fandeli:
The swupd_lock is not properly closed
when the work is done with bundle-list
subcommand.


fandeli:
Was having a break and found this little sucker so I went and fix it ;)

phmccarty:
+1

tmarcu:
Nice catch :) :+1:

PR #182, opened 2017-02-07
Change certpath to be full path of certificate

tmarcu:
The mixer and image creator treat the certpath as the full path of the
certificate filename, and swupd should too. If someone is overriding the
certificate with the cert path option, use the supplied string and don't
append a pre-defined name to it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -161,7 +161,7 @@ AC_ARG_WITH(
   [certpath],
   [AS_HELP_STRING([--with-certpath=PATH], [Location of CA certificates])]
 )
-default_cert_path="/usr/share/clear/update-ca"
+default_cert_path="/usr/share/clear/update-ca/Swupd_Root.pem"
 cert_path=
 # Makes sure --with-certpath receives an argument that is not "yes" or "no",
 # and uses the default path if only --enable-signature-verification is passed.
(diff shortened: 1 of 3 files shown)


src/signature.c
```diff
@@ -70,7 +70,7 @@ bool initialize_signature(void)
 	struct tm *alttime;
 	struct stat statt;
 
-	string_or_die(&CERTNAME, "%s/%s", cert_path, SWUPDCERT);
+	string_or_die(&CERTNAME, "%s", cert_path);
```
phmccarty: I'm okay with this approach.

However, the configure option `--with-swupdcert` should be removed as well, since it would become a no-op.

phmccarty: And also the default cert path setting would need to be changed.

tmarcu: I think the configure option should stay, because it provides the default fallback support builtin.

phmccarty: I'm not following.

SWUPDCERT is set by `--with-swupdcert`, so by removing the one place the macro is used, the option is no longer useful.

tmarcu: Confused it with the --with-certpath option lol.

configure.ac
```diff
@@ -179,18 +179,6 @@ AS_IF(
 )
 
 AC_ARG_WITH(
-  [swupdcert],
-  [AS_HELP_STRING([--with-swupdcert=FILE], [swupd verification cert])],
-  [AC_DEFINE_UNQUOTED([SWUPDCERT], ["$withval"], [swupd verification cert])],
-  [AC_DEFINE([SWUPDCERT], ["Swupd_Root.pem"], [swupd verification cert])]
```
phmccarty: The line with SWUPDCERT in `AC_MSG_NOTICE` should also be removed.

PR #184, opened 2017-02-24
Refresh bash completion script

phmccarty:
The scripts/swupd-completion.sh script can be run from the top source
directory to regenerate the completion script (swupd.bash).


bryteise:
Should this be something done as part of make dist perhaps?

phmccarty:
I added it to the release process steps.

PR #186, opened 2017-02-27
Add package names to completion

icarus-sparry:
Remove the script which creates the completion script based on the
output of swupd --help. It wasn't being used (e.g. the completion
didn't have bundle-list in it).

Restructure the completion script to use a case statement to list the
valid completion options. IMHO this makes the code easier to
understand.

Add in package name completion for bundle-add. This requires
/var/lib/swupd/XXXXX/Manifest.MoM to exist (where XXXXX is the
contents of /var/lib/swupd/version), be in the correct format
etc.

Add in package name completion for bundle-remove. This uses the
contents of /usr/share/clear/bundles to get the list of installed
bundles. It would be nice to use $(swupd bundle-list) but it aborts if
it is not being run as root, so this means you can't have completion
for "sudo swupd bundle-remove".

TODO: Fix bundle completion if --path is specified.


phmccarty:
I haven't looked at this patch yet, but I did update the completion script in #184. swupd.bash is intended to be updated as part of the release process, but in recent releases this was not done.

icarus-sparry:
The patch removes the script for automatically generating the completion function (based on the help output) and replaces the function with a manually written one. I can certainly alter the automatic generation script to work in 2 parts, get the flags automatically from the --help output and then add extra completions based on the name, but I am not sure of the value now that swupd-client is fairly stable.

tmarcu:
@icarus-sparry I'm inclined to keep the script that just auto-generates it, so it can be automated at some point and not a user required action.

icarus-sparry:
Whilst there are advantages to doing this, there are also downsides.

1)      The script which does the generation has to be updated. This negates some of the advantage of automatic generation.

2)      You get completion of options which you no longer want, e.g. ‘-l’ to ‘bundle-add’.

If you don’t generate the script, there is no need to add code to do the automation of the generation.

tmarcu:
+1

tmarcu:
Merged in PR #192 to make Makefile happy.

PR #187, opened 2017-02-28
Accept multiple bundles at bundle-remove

fandeli:
This is a wrapper around remove_bundle()
to add consistency for bundle-remove
subcommand respect bundle-add that accepts
one or more bundles to be removed.

This must not be the optimal implementation
but it works fine and gives a better user
usage experience.


fandeli:
failures I'm seeing are expected according to log, behavior changed little bit on bundle-remove so need to patch those tests accordingly. I'll take a look and send patches according new behavior.

tmarcu:
Thank you! You may add them to this PR and force push over it when you get the tests patched if you want.

fandeli:
ready to go! all green ;)

src/clr_bundle_rm.c
```diff
@@ -121,56 +142,150 @@ static bool parse_options(int argc, char **argv)
...
 	if (!parse_options(argc, argv)) {
 		return EINVALID_OPTION;
 	}
 
-	ret = remove_bundle(bundle_name);
-	free(bundle_name);
+	for (; *bundles; ++bundles, total++) {
+		printf("Removing bundle: %s\n", *bundles);
+
+		if (remove_bundle(*bundles) != 0) {
```
tpepper: This patch would introduce less code (and be more performant) if remove_bundle() was refactored to include a core helper function, wrapped by setup/cleanup calls.  Then here you could call the core helper instead of working around the setup/cleanup done instead the current remove_bundle().

fandeli: I agree and in fact this is just a dumb wrapper around bundle-remove to align consistency of use along with bundle-add, I know there is an optimal solution and doesn't even need any wrappers, however, I decided to go this path for two reasons:
1) the less the core of remove_bundle() core function changes the more the chance to no break anything else (overall with includes stuff and the so), then having it working on multiple removes we can add more tests over that functionality, the test coverages will ensure that in the future "optimized" function does not break anything.
2) Wanted to get this as soon as possible (without a broken swupd-remove of course) for consistency and usability, then we can go from this to implement optimal solution.

Having said that, I don't want to try optimizing anymore this wrapper, what I'd expect is to have it merged, rise test coverage and start developing the optimized solution that would for sure change the core heart of remove_bundle().

fandeli:
If no further comments or concerns on this patch can we have it merged please?

tmarcu:
Merging and setting a new roadmap for refactoring the bundle code so this does not remain the implementation for bundle removes.

PR #188, opened 2017-02-28
Fix certificate issues with makefile

tmarcu:
The client does not do certificate pinning anymore, and it should not provide
certificates. This patch removes the unused certificates and clarifies the
certpath option.


phmccarty:
+1

PR #189, opened 2017-03-05
Add verbose timing to swupd operation

tmarcu:
This introduces a -t/--time option to update, verify, and bundle-add commands, displaying raw elapsed time and CPU process time for integral swupd operations within the given code paths.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -418,8 +418,12 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
 	struct list *iter;
 	struct list *to_install_bundles = NULL;
 	struct list *to_install_files = NULL;
+	timelist times;
+
+	times = init_timelist();
 
 	/* step 1: check bundle args are valid if so populate subs struct */
+	grabtime_start(&times, "Add bundles and recurse");
 	ret = add_subscriptions(bundles, subs, current_version, mom, 0);
 
 	switch (ret) {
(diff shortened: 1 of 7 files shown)


src/bundle.c
```diff
@@ -420,8 +420,10 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
 	struct list *iter;
 	struct list *to_install_bundles = NULL;
 	struct list *to_install_files = NULL;
+	struct time times[5] = {0};
 
 	/* step 1: check bundle args are valid if so populate subs struct */
+	grab_time(START, times, 0, "Add bundles and recurse");
```
phmccarty: For ease of use, it would be nice to omit the third argument if possible... So, `grab_time(START, times, "description")`.

src/bundle.c
```diff
@@ -440,18 +442,23 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
 		ret = ERECURSE_MANIFEST;
 		goto out;
 	}
+	grab_time(STOP, times, 0, NULL);
```
phmccarty: Similarly here, it would be nice to simply call `grab_time(STOP, times)`.

src/clr_bundle_add.c
```diff
@@ -138,6 +140,9 @@ static bool parse_options(int argc, char **argv)
 		case 'n':
 			sigcheck = false;
 			break;
+		case 't':
```
phmccarty: You'll want to add `t` to the `getopt_long()` string on line 81.

src/bundle.c
tmarcu: Yeah I had the same idea only recently after posting this to make it more self contained and managed...will try to post a v2 if I can get it done today.

src/bundle.c
tmarcu: /me reinvents wheel due to language limitations.

PR #191, opened 2017-03-21
Try to recover from invalid certificate date

tmarcu:
The system clock may be terribly off, especially on new hardware that has not
yet been calibrated. Updates rely on the certificate and system time being
sane to verify validity, so if a mismatch is found the certificate will
be deemed invalid and the update stopped. This patch attempts to fix the
system time to something sane using the time from the swupd binary itself,
and then the os-release file as backup, neither of which should have been
touched by any user except root.

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -42,6 +42,7 @@
 #include "swupd.h"
 
 #ifdef SIGNATURES
+#define WEEKLEN 604800 // length of week in seconds to validate sys time
 
 static char *CERTNAME;
 
@@ -57,6 +58,30 @@ STACK_OF(X509) *x509_stack = NULL;
 //TODO: static char *chain = NULL;
 static char *crl = NULL;
 
+bool get_filetime(const char *filename, time_t *mod_sec, struct tm *alttime)
+{
+	struct stat statt;
+
+	if (stat(filename, &statt) != -1) {
+		*mod_sec = statt.st_mtime;
+		char timebuf[30];
+		alttime = localtime(mod_sec);
+		strftime(timebuf, sizeof(timebuf), "%F", alttime);
+		return true;
+	}
+	return false;
+}
+
+bool set_time(time_t mtime, char *time)
+{
+	if (stime(&mtime) != 0) {
+		printf("Failed to set system time");
+		return false;
+	}
+	printf("Set system time to %s\n", time);
+	return true;
+}
+
 /* This function must be called before trying to sign any file.
  * It loads string for errors, and ciphers are auto-loaded by OpenSSL now.
  * If this function fails it may be because the certificate cannot
@@ -67,8 +92,7 @@ bool initialize_signature(void)
 {
 	int ret = -1;
 	time_t mod_sec = 0;
-	struct tm *alttime;
-	struct stat statt;
+	struct tm *alttime = NULL;
 
 	string_or_die(&CERTNAME, "%s", cert_path);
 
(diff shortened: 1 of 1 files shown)


src/signature.c
```diff
@@ -82,20 +100,38 @@ bool initialize_signature(void)
...
+
+			time (&currtime);
+			timeinfo = localtime (&currtime);
+			printf ("Warning: Current time %s is incorrect\nAttempting to fix...", asctime(timeinfo));
+
+			/* If we can retrieve an approx. good system time, use it and try again */
+			if (fix_time("/usr/bin/swupd", mod_sec, alttime) == false) {
+				goto fail;
+			}
+			ret = validate_certificate();
```
bryteise: If this succeeds it will goto fail still.

src/signature.c
```diff
@@ -82,20 +100,38 @@ bool initialize_signature(void)
...
+				if (fix_time("/usr/lib/os-release", mod_sec, alttime) == false) {
+					goto fail;
+				}
+				ret = validate_certificate();
+				if (ret) {
+					goto fail;
+				}
+				goto success;
+			} else if (ret) {
+				goto fail;
```
bryteise: This logic is pretty twisted. Maybe break this block out into another function to avoid all the gotos?

src/signature.c
```diff
@@ -82,20 +100,38 @@ bool initialize_signature(void)
...
-		printf("Failed to verify certificate: %s\n", X509_verify_cert_error_string(ret));
 		if (ret == X509_V_ERR_CERT_NOT_YET_VALID) {
-			/* If we can retrieve an approx. good system time, report out to user */
-			if (stat("/usr/lib/os-release", &statt) != -1) {
-				mod_sec = statt.st_mtim.tv_sec;
-				char timebuf[30];
-				alttime = localtime(&mod_sec);
-				strftime(timebuf, sizeof(timebuf), "%F", alttime);
-				printf("System clock should be at least %s\n", timebuf);
+			time_t currtime;
```
bryteise: I'm not sure about doing this every time we get that error. This will reset the system time in the case of a valid system time and bad cert. Could we instead check to see if our system time is set to before the file's date instead?

src/signature.c
tmarcu: +1

src/signature.c
tmarcu: +1

src/signature.c
tmarcu: Then we should also check if the system time is ahead of the file's date as well. This check is a bit annoying, and at worst we'd be setting a sane system time to a still sane system time. If the cert is bad, is there any case where it should not fail anyway?

bryteise: Yea this part is the most ugly/difficult.

So the matrix:
if system time is good and cert is good -> everything works
if system time is good and cert is bad -> need to figure out how to change this code do the rightish thing
if system time is bad and cert is good -> this code as is works
if system time is bad and cert is bad -> this code as is works

So while not breaking the bad system time case we need to handle the good system time case with bad cert.

We can do some kind of time delta between system time and file time and see if that is some amount off. Possibly could account for the cert expiration dates but I think figuring out the cert being used is really hard here and maybe makes this more fragile than the simple date delta from file and system time.

tmarcu: Sure, then I think the bad cert case is really still deciding if the system time is good or not first, and if it's good and still fails, the cert must be bad so quit.If the system time seems weird then we try to fix it and check again like normal. I'll make some changes and push again.

src/signature.c
```diff
@@ -82,20 +106,40 @@ bool initialize_signature(void)
...
+			struct tm *timeinfo;
+			double timediff;
+
+			time(&currtime);
+			timeinfo = localtime(&currtime);
+			printf("Warning: Current time %s is \nAttempting to fix...", asctime(timeinfo));
```
tmarcu: Will move this to a more appropriate spot in V3.

phmccarty:
I don't like the idea of SWUPD setting the system time. Its purpose is to update files, not change system-wide settings... I would prefer that a warning be printed regarding a potential time skew, and then the user can fix their system time.

tpepper:
I agree with Patrick.  I don't like setting the time.  I worry about the one case William calls out.  If a bad cert input causes the automation to incorrectly set the time, can you recover?

tmarcu:
If the cert is bad I would assume the system has already been compromised at that point.

tmarcu:
Closing and opening new branch based on previous discussions to separate this functionality out into its own binary.

PR #193, opened 2017-03-23
clarify printf

tpepper:
I noticed this printf says something about getwd(), but that's not
actually being called directly there and that's not super useful to a
non-developer human anyway.  So I translated the function name to an
english description.


tmarcu:
+1 ty.

PR #194, opened 2017-03-23
Improved error reporting for adding bad bundle

icarus-sparry:
Currently if you try and add a non existent bundle, e.g. "foo" you get
two lines of output

  foo bundle name is invalid, skipping it...
  bundle(s) already installed, exiting now

This is caused by the add_subscriptions function calling itself
recursivly but failing to pass up results in a meaningful way. This
change makes the return value of add_subscriptions be a bitmask so it
can signal errors and packages added distinctly.

I did think about changing this function to return a struct but
decided this was a step too far.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -333,12 +333,16 @@ out_free_curl:
 	return ret;
 }
 
-/* tristate return, -1 for errors, 1 for no errors but no new subscriptions, 0 for no errors and new subscriptions */
+/* bitmapped return
+   1 error happened
+   2 new subscriptions
+*/
+#define add_sub_ERR 1
+#define add_sub_NEW 2
 int add_subscriptions(struct list *bundles, struct list **subs, int current_version, struct manifest *mom, int recursion)
 {
-	bool new_bundles = false;
 	char *bundle;
-	int ret;
+	int ret = 0;
 	int retries = 0;
 	int timeout = 10;
 	struct file *file;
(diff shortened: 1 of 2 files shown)


src/bundle.c
```diff
@@ -424,12 +422,16 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+	switch (ret)
+	{
+	case 0: /* no errors, but nothing added */
+		printf("bundle(s) already installed, exiting now\n");
+		/* fall through */
+	case add_sub_ERR: /* error of some kind */
+	case add_sub_ERR|add_sub_NEW: /* error of some kind but we did add something */
 		ret = EBUNDLE_INSTALL;
 		goto out;
+	default: break;
```
bryteise: Isn't this causing add_sub_NEW (even by itself) to not cause bundle installations like it did previously?

icarus-sparry: I hope not! The "default: break;" is in there to be sure that the switch has a case for every value that can be returned from add_subscriptions. Currently there are 4 possible bit patterns 00, 01, 10, 11. Of these 01 and 11 indicate that some error in the processing happened, e.g. failed to download a manifest. these will cause it to not install bundles.

00 is what you get if nothing was added, again this cause it to not install anything, but that is OK as there is nothing to install.

10 is the case handled by the "default:", it will break out of the switch statement and  continue to the set_subscription_versions call after it.

However you have drawn my attention to me not updating the '00' case message which was the point of the patch! Thanks.

bryteise: Ah goodness yes, I was something thinking of ret | (add_sub_ERR|add_sub_NEW) not a case statement matching exactly add_sub_ERR|add_sub_NEW X(.

phmccarty:
Interesting approach!  +1 from me.

tmarcu:
Great work +1

PR #195, opened 2017-03-24
Try to recover from invalid certificate date

tmarcu:
The system clock may be terribly off, especially on new hardware that has not
yet been calibrated. Updates rely on the certificate and system time being
sane to verify validity, so if a mismatch is found the certificate will
be deemed invalid and the update stopped. This patch attempts to fix the
system time to something sane using the time from the versionstamp,
which should not have been touched by any user except root. If the time is
normal and verification fails, the cert cannot be trusted.

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -66,9 +66,6 @@ static char *crl = NULL;
 bool initialize_signature(void)
 {
 	int ret = -1;
-	time_t mod_sec = 0;
-	struct tm *alttime;
-	struct stat statt;
 
 	string_or_die(&CERTNAME, "%s", cert_path);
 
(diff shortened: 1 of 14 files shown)


src/verifytime.c
```diff
@@ -0,0 +1,96 @@
...
+{
+	if (stime(&mtime) != 0) {
+		printf("Failed to set system time");
+		return false;
+	}
+	printf("Set system time to %s\n", time);
+	return true;
+}
+
+int  main()
```
phmccarty: extraneous space before `main`.

src/verifytime.c
```diff
@@ -0,0 +1,96 @@
...
+	timeinfo = localtime(&currtime);
+
+	versionstamp = get_versionstamp();
+	if (versionstamp == 0) {
+		return 1;
+	}
+	time_t versiontime = versionstamp;
+
+	/* Give it a day's worth of tolerance */
+	if ((unsigned long int) currtime < (versionstamp - DAY_SECONDS)) {
```
phmccarty: should remove space after cast (clang-format will catch this)

src/verifytime.c
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	if (fgets(data, 11, fp) == NULL) {
+		printf("Failed to read %s\n", filename);
+		fclose(fp);
+		return 0;
+	}
+
+	fclose(fp);
+	return strtoul(data, NULL, 10);
```
phmccarty: Since a 0 return value indicates an error condition, if the versionstamp value is 0, we should handle that case.

src/verifytime.c
```diff
@@ -0,0 +1,96 @@
...
+	unsigned long int versionstamp;
+
+	time(&currtime);
+	timeinfo = localtime(&currtime);
+
+	versionstamp = get_versionstamp();
+	if (versionstamp == 0) {
+		return 1;
+	}
+	time_t versiontime = versionstamp;
```
phmccarty: cast to `time_t` here?

src/signature.c
```diff
@@ -82,16 +79,22 @@ bool initialize_signature(void)
...
+			struct tm *timeinfo;
+			ASN1_TIME *certstart = X509_get_notBefore(cert);
+			ASN1_TIME *certexpire = X509_get_notAfter(cert);
+
+			/* The system time wasn't sane, print out what it is and the cert validity range */
+			time(&currtime);
+			timeinfo = localtime(&currtime);
+			printf("Warning: Current time is %s\n", asctime(timeinfo));
+			printf("Certificate validity is: ");
+			b = BIO_new_fp(stdout, BIO_NOCLOSE);
```
phmccarty: If this function returns NULL, how should that be handled?

src/verifytime.c
```diff
@@ -0,0 +1,96 @@
+/*
+ *   Software Updater - client side
+ *
+ *      Copyright © 2012-2016 Intel Corporation.
```
phmccarty: should be 2017 only

src/signature.c
tmarcu: Adding checks for all the openssl calls.

PR #196, opened 2017-03-27
Make the backward compatability check work

icarus-sparry:
Use the correct variable BASH_VERSINFO, rather than BASH_VERSION

### DIFF
--- a/scripts/swupd-completion.sh
+++ b/scripts/swupd-completion.sh
@@ -150,7 +150,7 @@ cat >> $SCRIPTNAME << 'EOM'
     COMPREPLY=($(compgen -W "${opts}" -- ${2}));
     return 0
 }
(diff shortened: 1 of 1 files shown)


scripts/swupd-completion.sh
```diff
@@ -150,7 +150,7 @@ cat >> $SCRIPTNAME << 'EOM'
     COMPREPLY=($(compgen -W "${opts}" -- ${2}));
     return 0
 }
-if [ ${BASH_VERSION[0]} -gt 4 ] || ( [ ${BASH_VERSION[0]} -eq 4 ] && [ $BASH_VERSION[1] -ge 4] )
+if [ ${BASH_VERSINFO[0]} -gt 4 ] || ( [ ${BASH_VERSINFO[0]} -eq 4 ] && [ ${BASH_VERSINFO[1]} -ge 4 ] )
```
phmccarty: Oh, we missed the "space before `]`" issue too. Thanks for catching that.

+1

icarus-sparry: Yep. Moral don't add extra stuff at the last minute "just in case someone is running an old version".

PR #197, opened 2017-03-29
Error handling fixes for fullfile downloads

phmccarty:
This series addresses a couple of double frees encountered in error paths and also a "hysteresis escape" that was hitting these error paths.

### DIFF
--- a/src/download.c
+++ b/src/download.c
@@ -452,6 +452,9 @@ static int perform_curl_io_and_complete(int count, bool bounded)
 	return 0;
 }
 
+/* Fixed window added to hysteresis upper bound before it is enforced. */
+#define BUFFER 5
+
 /* 2 limits, so that we can have hysteresis in behavior. We let the caller
  * add new transfer up until the queue reaches the high threshold. At this point
  * we don't return to the caller and instead process the queue until its len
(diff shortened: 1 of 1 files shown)


phmccarty:
After further testing, I discovered that the hysteresis upper bound may often be exceeded, just not enough to run into any error conditions in most cases.

To improve behavior, I added a small buffer to the upper bound to reduce the possibility of retries kicking in, and I revised how the multi stack is emptied which results in a more balanced add/remove sequence from the stack.

I uploaded a revised patch series with these changes.

src/download.c
```diff
@@ -452,6 +452,9 @@ static int perform_curl_io_and_complete(int count, bool bounded)
 	return 0;
 }
 
+/* Fixed window added to hysteresis upper bound before it is enforced. */
+#define BUFFER 5
```
tmarcu: If the upper bound is bypassed, are we guaranteed to run out of file descriptors? Or is this 5 a best guess at how much room we have before we run out?

src/download.c
```diff
@@ -500,6 +500,12 @@ static int poll_fewer_than(size_t xfer_queue_high, size_t xfer_queue_low)
 			}
 		}
 
+		// Do more work before processing the queue
+		curlm_ret = curl_multi_perform(mcurl, &running);
```
tmarcu: This looks very similar to how the initial version was coded, with multiple curl_multi_perform()s happening. Looking thru the docs, would it make sense to check for CURLM_CALL_MULTI_PERFORM in these calls? There's also nothing wrong with just calling it multiple times "just in case" per the docs... not sure which is a better path.

src/download.c
phmccarty: There is no chance for an upper bound bypass to result in running out of fds; this is purpose behind the conditional check below. The code comment above that check describes the reason for adding a bit extra, though "5" is an arbitrary number, it could be larger or smaller.

src/download.c
phmccarty: Good catch in the docs, but that error is only returned in curl < 7.20.0. Clear Linux runs a much newer version.

src/download.c
phmccarty: Alternatively, we could decide to change the hysteresis upper/lower bounds entirely instead of using a "buffer" like this. But I would want to do a bunch of experimentation first before considering that.

src/download.c
tmarcu: Ugh, not good enough catch apparently. Thanks for pointing out :+1:

PR #198, opened 2017-03-31
Remove bad or unusable Manifest content and retry

tmarcu:
For various reasons - partial downloads, power failure, user exiting -
the /var/lib/swupd state directory may become corrupt. When this happens,
manifests can be rendered unusable, causing various errors in swupd such as
signature failures, or errors reading the manifest. This patch attempts to
avoid such things causing updates or verifies to fail by cleaning up the
Manifests artifacts of what it could not use, and re-downloading them to
try again.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -567,6 +567,29 @@ static void set_untracked_manifest_files(struct manifest *manifest)
 	}
 }
 
+/* Removes the extracted Manifest.<bundle> and accompanying tar file, cache file, and
+ * the signature file */
+void remove_manifest_files(char *filename, int version, char *hash)
+{
+	char *file;
+
+	printf("Warning: Removing corrupt Manifest.%s artifacts and re-downloading...\n", filename);
+	string_or_die(&file, "%s/%i/Manifest.%s", state_dir, version, filename);
+	unlink(file);
+	free(file);
+	string_or_die(&file, "%s/%i/Manifest.%s.tar", state_dir, version, filename);
+	unlink(file);
+	free(file);
+	string_or_die(&file, "%s/%i/Manifest.%s.sig", state_dir, version, filename);
+	unlink(file);
+	free(file);
+	if (hash != NULL) {
+		string_or_die(&file, "%s/%i/Manifest.%s.%s", state_dir, version, filename, hash);
+		unlink(file);
+		free(file);
+	}
+}
+
 /* Loads the MoM (Manifest of Manifests) for VERSION.
  *
  * Implementation note: MoMs are not huge so deltas do not give much benefit,
@@ -585,17 +608,38 @@ struct manifest *load_mom(int version)
 	char *filename;
 	char *url;
 	char *log_cmd = NULL;
+	bool retried = false;
 
+verify_mom:
 	ret = retrieve_manifests(version, version, "MoM", NULL);
 	if (ret != 0) {
 		printf("Failed to retrieve %d MoM manifest\n", version);
 		return NULL;
 	}
 
+	manifest = manifest_from_file(version, "MoM", false);
+
+	if (manifest == NULL) {
+		if (retried == false) {
+			remove_manifest_files("MoM", version, NULL);
+			retried = true;
+			goto verify_mom;
+		}
+		printf("Failed to load %d MoM manifest\n", version);
+		goto out;
+	}
+	retried = false;
+
 	string_or_die(&filename, "%s/%i/Manifest.MoM", state_dir, version);
 	string_or_die(&url, "%s/%i/Manifest.MoM", content_url, version);
 	if (!download_and_verify_signature(url, filename)) {
 		if (sigcheck) {
+			/* cleanup and try one more time, statedir could have got corrupt/stale */
+			if (retried == false) {
+				remove_manifest_files("MoM", version, NULL);
+				retried = true;
+				goto verify_mom;
+			}
 			printf("WARNING!!! FAILED TO VERIFY SIGNATURE OF Manifest.MoM\n");
 			free(filename);
 			free(url);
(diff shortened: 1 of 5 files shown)


tmarcu:
*Note that the manifest removal/retry only happens once because it doesn't signify a network failure. Those retry paths are handled separately in update and verify.

src/manifest.c
```diff
@@ -567,6 +567,28 @@ static void set_untracked_manifest_files(struct manifest *manifest)
 	}
 }
 
+/* Removes the extracted Manifest.<bundle> and accompanying tar file, as well as
+ * the signature file it exists */
+void remove_manifest_files(char *filename, int version)
+{
+	char *file;
+
+	printf("Warning: Removing corrput Manifest.%s and re-downloading...\n", filename);
```
phmccarty: s/corrput/corrupt

src/manifest.c
```diff
@@ -567,6 +567,28 @@ static void set_untracked_manifest_files(struct manifest *manifest)
...
+	string_or_die(&file, "%s/%i/Manifest.%s", state_dir, version, filename);
+	unlink(file);
+	free(file);
+	string_or_die(&file, "%s/%i/Manifest.%s.tar", state_dir, version, filename);
+	unlink(file);
+	free(file);
+
+	/* Remove signature if it exists */
+	string_or_die(&file, "%s/%i/Manifest.%s.sig", state_dir, version, filename);
+	if (access(file, F_OK) != -1) {
```
phmccarty: Alternatively, you could just call `unlink` without the `access` check and not check for `unlink` errors. That's a common pattern in other parts of the code (to "speculatively unlink").

src/manifest.c
```diff
@@ -594,8 +617,15 @@ struct manifest *load_mom(int version)
...
 	string_or_die(&url, "%s/%i/Manifest.MoM", content_url, version);
+verify_mom:
 	if (!download_and_verify_signature(url, filename)) {
 		if (sigcheck) {
+			/* cleanup and try one more time, statedir could have got corrupt/stale */
+			if (retried == false) {
+				remove_manifest_files("MoM", version);
+				retried = true;
+				goto verify_mom;
+			}
```
phmccarty: same logic needed if `sigcheck` is false?

src/manifest.c
```diff
@@ -594,8 +617,15 @@ struct manifest *load_mom(int version)
 
 	string_or_die(&filename, "%s/%i/Manifest.MoM", state_dir, version);
 	string_or_die(&url, "%s/%i/Manifest.MoM", content_url, version);
+verify_mom:
```
phmccarty: Does this label need to be moved before `retrieve_manifests` instead? Else, I'm not seeing how the Manifest.MoM would be redownloaded, only the Manifest.MoM.sig.

src/manifest.c
```diff
@@ -661,6 +692,11 @@ struct manifest *load_manifest(int current, int version, struct file *file, stru
 	manifest = manifest_from_file(version, file->filename, header_only);
 
 	if (manifest == NULL) {
+		if (retried == false) {
+			remove_manifest_files(file->filename, version);
+			retried = true;
+			goto retry_load;
+		}
```
phmccarty: What if the bundle manifest hash check fails in `verify_bundle_hash`? Should that also retry?

src/verify.c
```diff
@@ -667,13 +669,19 @@ int verify_main(int argc, char **argv)
...
 	if (!official_manifest) {
 		/* This is hit when or if an OS version is specified for --fix which
 		 * is not available, or if there is a server error and a manifest is
 		 * not provided.
 		 */
+		if (retries < MAX_TRIES) {
+			increment_retries(&retries, &timeout);
+			printf("Retry #%d downloading from/to MoM Manifests\n", retries);
+			goto load_manifests;
+		}
```
phmccarty: I'm not seeing why verify.c requires these changes, since you added retries to `load_mom` already.

src/manifest.c
phmccarty: Also, what if `manifest_from_file` returns an error (see code a few lines below this one)? Retries for that case?

src/manifest.c
tmarcu: Yeah I figured but didn't want to not check for errors. If you want that to stay consistent I can change it.

src/manifest.c
tmarcu: I'd rather not put it in the sigcheck == false path, but it makes sense to retry if manifest_from_file fails, and to move manifest_from_file above the sigcheck. If we don't care about sigs and loaded the manifest fine, then don't retry.

src/manifest.c
phmccarty: My preference is for consistency :-)

src/manifest.c
phmccarty: Okay, I see that you're right about the sigcheck == false path; signature check failures are not enforced, so any number of additional errors might appear and not worth adding special handling.

I agree with moving `manifest_from_file` before the signature check.

src/verify.c
tmarcu: Sure, then I think load_mom should have these multiple retries when downloading and loading it up, and then a single cleanup/retry if signature fails.

tmarcu:
Tested Manifest corruption in header, body, and signature (MoM only) for both verify and update paths, and confirmed both operations succeeded in removing the bad files and replaced with correct ones, finishing their procedures fully.

src/verify.c
```diff
@@ -606,6 +606,8 @@ int verify_main(int argc, char **argv)
 	struct manifest *official_manifest = NULL;
 	int ret;
 	int lock_fd;
+	int retries = 0;
+	int timeout = 5;
```
phmccarty: The retries in update.c use a timeout of 10 seconds. Did you want to use 10 here too?

phmccarty:
+1 from me

src/verify.c
tmarcu: I think not, and I'm a little tempted to lower the ones in update (in a different patch) to be consistent.

phmccarty: Okay. I was asking because you're adding "10" and "5" in the same commit but don't explain why.

tmarcu: Fixing for consistency, will update them all again in a later PR.

phmccarty: Okay, thanks!

PR #199, opened 2017-04-04
Improved return code for bundleadd + test

icarus-sparry:
Arrange for "bad names" to give non-zero exit code
Arrange for "do nothing" in particular when packages are already
installed to give a zero return code.

This means that typos in names result in failures, but that trying to
add an existing package is fine.


test/functional/bundleadd/add-rc/test.bats
```diff
@@ -0,0 +1,65 @@
...
+  sudo rm "$DIR/target-dir/usr/foo"
+}
+
+# 8 combinations of rc to check, but skip all off
+# bad name on/off (1)
+# existing name on/off (2)
+# new name on/off (4)
+# Do it in order 1 4 3 2 5 7 6
+
+@test "bundle-add returncodes part 1" {
```
tmarcu: Is there a part 2? :) These might be better split up into separate @test "<>" sections, but this is ok for now.

icarus-sparry: There was (and 3, 4, 5, 6, 7), but I removed them and put into a single test.

PR #201, opened 2017-04-13
Need to reexec systemd if it changed

tmarcu:
We must not run these with --no-block because order must be preserved,
and critical components may need to be reexec'd, making this a core piece
of the swupd update process.

### DIFF
--- a/src/scripts.c
+++ b/src/scripts.c
@@ -51,7 +51,13 @@ static void update_triggers(void)
 {
 	__attribute__((unused)) int ret = 0;
 
-	ret = system("/usr/bin/systemctl --no-block daemon-reload");
+	/* These must block so that new update triggers are executed after */
+	if (need_systemd_reexec) {
+		ret = system("/usr/bin/systemctl daemon-reexec");
+	} else {
+		ret = system("/usr/bin/systemctl daemon-reload");
+	}
+
 	ret = system("/usr/bin/systemctl --no-block restart update-triggers.target");
 }
 
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -75,6 +75,7 @@ extern int update_skip;
 extern bool update_complete;
 extern bool need_update_boot;
 extern bool need_update_bootloader;
+extern bool need_systemd_reexec;
 
 struct update_stat {
 	uint64_t st_mode;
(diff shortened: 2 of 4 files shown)


src/scripts.c
```diff
@@ -51,7 +51,10 @@ static void update_triggers(void)
 {
 	__attribute__((unused)) int ret = 0;
 
-	ret = system("/usr/bin/systemctl --no-block daemon-reload");
+	if (need_systemd_reexec)
```
phmccarty: need curly brackets for this if block

src/scripts.c
```diff
@@ -51,7 +51,11 @@ static void update_triggers(void)
 {
 	__attribute__((unused)) int ret = 0;
 
-	ret = system("/usr/bin/systemctl --no-block daemon-reload");
+	if (need_systemd_reexec) {
+		ret = system("/usr/bin/systemctl daemon-reexec");
+	}
```
phmccarty: Since `need_systemd_reexec` will generally be false, maybe we could optimize for that case? Like so:

```
if (need_systemd_reexec) {
        ret = system("/usr/bin/systemctl daemon-reexec");
        ret = system("/usr/bin/systemctl daemon-reload");
} else {
        ret = system("/usr/bin/systemctl --no-block daemon-reload");
}
```

phmccarty: I may have misunderstood the intent of removing --no-block here. Tudor, can you clarify that?

fenrus75: the reload must not have --no-block, because the execution of the triggers really needs to be the NEW triggers, not the old version triggers.....

sofar: agreed. The reload is also a fairly fast and deterministic thing (it can only take so long to re-parse the changed units).

Adding --no-block here could mess things up, especially for `reload`

sofar: However, doing a `reload` after a `reexec` is just duplicate, so the code should be:

```
if (need_systemd_reexec) {
        ret = system("/usr/bin/systemctl daemon-reexec");
} else {
        ret = system("/usr/bin/systemctl daemon-reload");
}
```

tmarcu: @sofar noted, thanks for mentioning that.

phmccarty:
+1

PR #202, opened 2017-04-17
Delta manifests

tmarcu:
This patch-set enables delta file application for manifests, decreasing manifest download size by a factor of 150x or more in some cases.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -431,29 +431,21 @@ static int try_delta_manifest_download(int current, int new, char *component, st
 	char *url = NULL;
 	int ret = 0;
 	struct stat buf;
-
 	if (strcmp(component, "MoM") == 0) {
 		// We don't do MoM deltas.
 		return -1;
 	}
 
 	if (!file->peer) {
+		return -1;
 		goto out;
 	}
 
 	string_or_die(&original, "%s/%i/Manifest.%s", state_dir, current, component);
 
-	populate_file_struct(file, original);
-	ret = compute_hash(file, original);
-	if (ret != 0) {
-		goto out;
-	}
-	if (!hash_equal(file->peer->hash, file->hash)) {
-		goto out;
-	}
-
 	string_or_die(&deltafile, "%s/Manifest-%s-delta-from-%i-to-%i", state_dir, component, current, new);
-
+	/* If we cannot get a delta, quit and don't mess with the file struct.
+	 * Populating it early and then exiting early corrupts the manifest list */
 	memset(&buf, 0, sizeof(struct stat));
 	ret = stat(deltafile, &buf);
 	if (ret || buf.st_size == 0) {
@@ -464,6 +456,17 @@ static int try_delta_manifest_download(int current, int new, char *component, st
 			unlink(deltafile);
 			goto out;
 		}
+	}
+
+	populate_file_struct(file, original);
+	ret = compute_hash(file, original);
+	if (ret != 0) {
+		ret = -1;
+		goto out;
+	}
+	if (!hash_equal(file->peer->hash, file->hash)) {
+		ret = -1;
+		goto out;
 	}
 
 	/* Now apply the manifest delta */
@@ -478,13 +481,15 @@ static int try_delta_manifest_download(int current, int new, char *component, st
 	}
 
 	unlink(deltafile);
-
+	ret = compute_hash(file, newfile); // MUST save new hash!
+	if (ret != 0) {
+		ret = -1;
+	}
 out:
 	free(original);
 	free(url);
 	free(newfile);
 	free(deltafile);
-
 	return ret;
 }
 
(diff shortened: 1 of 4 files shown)


src/manifest.c
```diff
@@ -502,11 +502,15 @@ static int retrieve_manifests(int current, int version, char *component, struct
...
-	string_or_die(&filename, "%s/%i/Manifest.%s.tar", state_dir, version, component);
+	/* Check for fullfile only, we will not be keeping the .tar around */
+	string_or_die(&filename, "%s/%i/Manifest.%s", state_dir, version, component);
 	if (stat(filename, &sb) == 0) {
+
 		ret = 0;
 		goto out;
 	}
+	free(filename);
+	string_or_die(&filename, "%s/%i/Manifest.%s.tar", state_dir, version, component);
```
phmccarty: Did you want to unlink() the tar file after it successfully extracts? (near the end of this function)

src/manifest.c
```diff
@@ -688,7 +688,8 @@ struct manifest *load_manifest(int current, int version, struct file *file, stru
 {
 	struct manifest *manifest = NULL;
 	int ret = 0;
-	bool retried = false;
+	int retried = 0;
```
phmccarty: Can you explain the retry changes in the commit description? Is it a bug fix? etc. It seems to increase the complexity of the code, and it's not obvious to me what benefit it has.

src/manifest.c
```diff
@@ -878,17 +878,16 @@ void link_submanifests(struct manifest *m1, struct manifest *m2, struct list *su
...
-				if (subbed1 && subbed2) {
+				if (!server && subbed1 && subbed2) {
 					account_changed_bundle();
 				} else if (!subbed1 && subbed2) {
 					account_new_bundle();
 				}
 			}
-			if (file2->last_change > m1->version && file2->is_deleted) {
-				if (subbed1) {
+			if (!server && (file2->last_change > m1->version && file2->is_deleted && subbed1)) {
```
phmccarty: I liked having the inner conditional check (`if (subbed1) ...`), because it mirrors the conditional block above, making it easier (for me) to understand the code.

src/manifest.c
```diff
@@ -878,17 +887,16 @@ void link_submanifests(struct manifest *m1, struct manifest *m2, struct list *su
...
 					account_changed_bundle();
 				} else if (!subbed1 && subbed2) {
 					account_new_bundle();
 				}
 			}
-			if (file2->last_change > m1->version && file2->is_deleted) {
-				if (subbed1) {
-					account_deleted_bundle();
-				}
+			if (!server && (file2->last_change > m1->version && file2->is_deleted && subbed1)) {
```
phmccarty: I liked having the inner conditional check (if (subbed1) ...), because it mirrors the conditional block above, making it easier (for me) to understand the code.

src/manifest.c
tmarcu: Artifact of debugging, retested with sane code and things are all in order still.

src/manifest.c
tmarcu: +1

src/manifest.c
tmarcu: Fixed for consistency.

src/update.c
```diff
@@ -190,7 +190,7 @@ static int update_loop(struct list *updates, struct manifest *server_manifest)
 	return ret;
 }
 
-int add_included_manifests(struct manifest *mom, struct list **subs)
```
bryteise: So you updated the function but not all the callers here, I assume this patch didn't even compile.

bryteise: Did this get split after making the changes? Do you recommend me reviewing just the final diff?

tmarcu: The patch set isn't meant to be applied partially...I can change it if needed, but for reviewer's sake I broke it up a lot more so it was easier to look through.

src/manifest.c
```diff
@@ -431,29 +431,21 @@ static int try_delta_manifest_download(int current, int new, char *component, st
...
 	struct stat buf;
-
 	if (strcmp(component, "MoM") == 0) {
 		// We don't do MoM deltas.
 		return -1;
 	}
 
 	if (!file->peer) {
+		ret = -1;
 		goto out;
```
bryteise: Can just return -1 here since there isn't any allocation going on yet.

src/manifest.c
```diff
@@ -466,25 +458,37 @@ static int try_delta_manifest_download(int current, int new, char *component, st
...
 	xattrs_copy(original, newfile);
 	if (ret != 0) {
+
 		unlink(newfile);
 	} else if ((ret = xattrs_compare(original, newfile)) != 0) {
 		unlink(newfile);
 	}
 
 	unlink(deltafile);
+	compute_hash(file, newfile); // MUST save new hash!
```
bryteise: Shouldn't this be checked?

bryteise:
Review completed.

src/manifest.c
tmarcu: Absolutely!

PR #203, opened 2017-04-18
Fix download/extract pack message

tmarcu:
Swupd incorrectly says "Downloading pack <name>..." even if it does not
really download the pack. Fix this by saying downloading packs and only
printing if we downloaded a pack to extract.

### DIFF
--- a/src/packs.c
+++ b/src/packs.c
@@ -52,8 +52,6 @@ static int download_pack(int oldversion, int newversion, char *module)
 		return 0;
 	}
 
-	printf("Downloading %s pack for version %i\n", module, newversion);
-
 	string_or_die(&url, "%s/%i/pack-%s-from-%i.tar", content_url, newversion, module, oldversion);
 
 	err = swupd_curl_get_file(url, filename, NULL, NULL, true);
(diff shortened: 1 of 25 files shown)


tmarcu:
The tests need to be updated to support the new printf's that are happening, ignore travis for the moment!

phmccarty:
+1 for the code changes

src/packs.c
```diff
@@ -104,6 +102,7 @@ int download_subscribed_packs(struct list *subs, bool required)
 		return -ENOSWUPDSERVER;
 	}
 
+	printf("Downloading packs...\n");
```
phmccarty: I noticed there is a "Downloading packs" printf in bundle.c. Can that one be removed?

tmarcu: @phmccarty good catch, yes I think the one in bundle.c should be removed.

test/functional/ignore-list
```diff
@@ -5,6 +5,8 @@
 ^swupd-client software verify [0-9.]+$
 ^.*Copyright \(C\) .*$
 ^$
+^Downloading packs...
+^Extracting.*
```
phmccarty: IMO, any future changes to the pack download code would benefit from *not* ignoring the "Extracting ..." lines, because they print the version now. We always want to make sure we're downloading the correct packs for the correct version (or not downloading them at all).

tmarcu: Wanted the discussion ;) Fixed in V2 though.

phmccarty:
+1

PR #205, opened 2017-04-20
Swupd fix

tmarcu:
Swupd incorrectly handled the file structs when applying deltas to manifests, creating an incorrect hash that then cause updates to retry infinitely. This patch fixes the retry code, and also addresses the corruption in try_manifest_delta so file structs are properly created.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -438,7 +438,6 @@ static int try_delta_manifest_download(int current, int new, char *component, st
 
 	if (!file->peer) {
 		return -1;
-		goto out;
 	}
 
 	string_or_die(&original, "%s/%i/Manifest.%s", state_dir, current, component);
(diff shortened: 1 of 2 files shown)


src/manifest.c
```diff
@@ -481,6 +469,7 @@ static int try_delta_manifest_download(int current, int new, char *component, st
 	}
 
 	unlink(deltafile);
+	populate_file_struct(file, newfile);
 	ret = compute_hash(file, newfile); // MUST save new hash!
```
bryteise: Could we pass in the file struct as a pointer to a pointer so that if this function fails it won't have modified the file struct passed in and if it succeeds can modify the owning pointer to point to the new file struct?

phmccarty: I like this suggestion.

tmarcu: Per testing and attempted implementation...this is not a feasible solution at this time, and we confirmed that calling populate_file_struct() resets the entire struct contents that could have been corrupt, so it is not needed to use a secondary struct.

PR #206, opened 2017-04-24
swupd verify --picky option

icarus-sparry:
Two changes. The first changes printf( to fprintf(stderr, almost everywhere so diagnostics, warnings and errors can be separated from data (e.g. hash values for files).

Second change adds a "swupd verify --picky" to list to stdout all the files/directories under /usr which are not in the current manifest, ignoring /usr/local and /usr/lib/modules.

### DIFF
--- a/src/extra_files.c
+++ b/src/extra_files.c
@@ -55,7 +55,7 @@ static struct fileskip {
 	const char *shortname; /* Without prefix */
 	char *name;	    /* full name, including any path_prefix */
 	int len;
-} skippers[] =
+} skip_dirs[] =
     { { shortname : "/lib/modules" },
       { shortname : "/local" } };
 static int path_prefix_len;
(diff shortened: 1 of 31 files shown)


src/extra_files.c
```diff
@@ -0,0 +1,165 @@
...
+			found->in_manifest = true;
+		}
+	}
+	/* list files/directories which are extra */
+	for (int i = 0; i < nF; i++) {
+		int skip_len; /* Length of directory name we are skipping */
+		if (!F[i].in_manifest) {
+			if (skip_dir) {
+				/* This is under the current directory to be skipped */
+				if ((strncmp(F[i].filename, skip_dir, skip_len) == 0) &&
```
icarus-sparry: It is not used before it is set in line 147. If skip_dir is NULL then skip_len is not used. I could set it to be 0 or 42 or whatever but it wouldn't every be used.

PR #207, opened 2017-04-28
Retry pack downloads on bundle-add

tmarcu:
The download_subscribed_packs() could fail for network issues or other related
problems, and because the only option for getting new packs is by downloading
zero packs or fullfiles, swupd should retry to get the packs before falling
into the verify_fix_path flow which signifies pack downloads errored out.


bryteise:
Looks like an improvement to me, unfortunate we went straight to full files before.

PR #210, opened 2017-05-08
Fix infinite signature failure loop

tmarcu:
Unlike the regular manifest loading code, loading a MoM also requires signature
verification to succeed. We cannot reset the retries after a signature
failed because the manifest may load fine, but still fail verification,
resulting in endlessly loading and verifying the Manifest.MoM

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -633,7 +633,6 @@ verify_mom:
 		printf("Failed to load %d MoM manifest\n", version);
 		goto out;
 	}
-	retried = false;
 
 	string_or_die(&filename, "%s/%i/Manifest.MoM", state_dir, version);
 	string_or_die(&url, "%s/%i/Manifest.MoM", content_url, version);


src/manifest.c
```diff
@@ -633,7 +633,6 @@ struct manifest *load_mom(int version)
 		printf("Failed to load %d MoM manifest\n", version);
 		goto out;
 	}
-	retried = false;
```
phmccarty: So, just to be clear, are these the intended failure patterns? These assume the -n option is not passed.
1) if retreive_manifests() fails, return an error
2) if manifest_from_file() fails, retry once; second failure returns an error
3) if download_and_verify_signature() fails, retry once; second failure returns an error

If the above is correct, maybe the failure for step 3 should goto a new label that such that only download_and_verify_signature() is called again? My reasoning is that manifest_from_file() returning success means the MoM was already downloaded and has been parsed correctly.

phmccarty: So, it looks like my analysis may be off... Can you clarify when retries need to happen and when they do not?

tmarcu: The retry code in load_mom and such functions is different then the top level code since it assumes it is dealing with IO errors or system errors, which will not benefit from timeouts. If retrieving a manifest fails - i.e due to curl/network errors - then timeouts will be hit and allow for more time for the network to come up if it went down.
This code can be refactored to support both cases more fluidly and reduce these kinds of errors, but that will be a significant rework. Therefore, I opened an issue to track the general problem of retries https://github.com/clearlinux/swupd-client/issues/211 and we can plan that out separately.

phmccarty: Okay, thanks. Then I think your change looks good for now.

phmccarty:
+1

PR #212, opened 2017-05-08
Add support for new manifest header line

tmarcu:
The swupd client should support parsing extra data in the case that a format
bump occurred, or other possibly breaking change, so that it can at least
notify the user of more action needing to be taken post update.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -162,7 +162,7 @@ static struct manifest *alloc_manifest(int version, char *component)
 	return manifest;
 }
 
-static struct manifest *manifest_from_file(int version, char *component, bool header_only)
+static struct manifest *manifest_from_file(int version, char *component, bool header_only, bool latest)
 {
 	FILE *infile;
 	char line[MANIFEST_LINE_MAXLEN], *c, *c2;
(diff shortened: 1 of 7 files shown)


src/manifest.c
```diff
@@ -233,6 +233,9 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
 		if (strncmp(line, "contentsize:", 12) == 0) {
 			contentsize = strtoull(c, NULL, 10);
 		}
+		if (strncmp(line, "actions:", 8) == 0) {
+			post_update_action = strdup(c);
+		}
```
phmccarty: Is this header field only intended for the MoM? If so, can you enforce that here?

Also, for 'update', more than one MoM is parsed, so if both MoMs define "actions:", then the post_update_action value for the current MoM will be overwritten by the value from the latest MoM. Is that behavior acceptable? Or do you need two globals?

tmarcu: Solid questions! This is intended for the MoM only, so we can change that to check that only if it's in the MoM. 
The latest MoM should be the one it gets the action from, so in this case it will always be doing that - implicitly if both happen to define an action. I don't think that needs extra code, so I'll update with a comment clarifying that.

phmccarty:
+1

src/manifest.c
```diff
@@ -233,6 +233,9 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
 		if (strncmp(line, "contentsize:", 12) == 0) {
 			contentsize = strtoull(c, NULL, 10);
 		}
+		if (strncmp(line, "actions:", 8) == 0) {
+			post_update_action = strdup(c);
```
tpepper: Can you do some input validation on the "actions:" data?

tmarcu: We could explore that path when/if enabling swupd to follow the actions itself instead of just printing a warning. I am inclined to say the actions: are automatically validated if the signature verification passes, and we should also not ever publish corrupt manifests to begin with, so perhaps the validation would happen before it ever gets to clients?

PR #216, opened 2017-05-09
PoC of swupd autoupdate command

icarus-sparry:
"swupd autoupdate" returns 0 if autoupdates are enabled.
"sudo autoupdate --enable" will unmask swupd-update.service
"sudo autoupdate --disable" will mask swupd-update.service

### DIFF
--- a/src/autoupdate.c
+++ b/src/autoupdate.c
@@ -1,0 +1,131 @@
+/*
+ *   Software Updater - client side
+ *
+ *      Copyright (c) 2012-2016 Intel Corporation.
+ *
+ *   This program is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License as published by
+ *   the Free Software Foundation, version 2 or later of the License.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


icarus-sparry:
Seems to work. I still need to think about tests. Enabling/disabling on a test system seems a bad idea. Perhaps run in a docker container?

tmarcu:
+1 on the code, docker would be an option to test this if you can set up a simple dockerfile and run systemd in it.

src/autoupdate.c
```diff
@@ -0,0 +1,131 @@
...
+	if (enable && disable) {
+		fprintf(stderr, "Can not enable and disable at the same time\n");
+		exit(EXIT_FAILURE);
+	}
+	if (enable) {
+		int rc;
+		check_root();
+		fprintf(stderr, "Running systemctl to enable updates\n");
+		rc = system("/usr/bin/systemctl unmask --now swupd-update.service swupd-update.timer"
+			    " && /usr/bin/systemctl start swupd-update.timer swupd-update.service > /dev/null");
```
tmarcu: Upon further testing, this will block if swupd is already running in the background doing an update. For large/slow updates it may look like it's frozen or hanging. What kind of extra checks can we put in before to keep it from blocking?

PR #219, opened 2017-05-23
Add verify --fix --picky functionality

tmarcu:
This is the basic implementation for a "factory reset" type of command.
It will take the functionality of verify --picky and act on the files found
as verify --fix does with files found mismatching in the manifest.

### DIFF
--- a/src/extra_files.c
+++ b/src/extra_files.c
@@ -95,10 +95,11 @@ static int bsearch_helper(const void *A, const void *B)
 }
 
 /* expect the start to end in /usr and be the absolute path to the root */
-int walk_tree(struct manifest *manifest, const char *start)
+int walk_tree(struct manifest *manifest, const char *start, bool fix)
 {
 	/* Walk the tree, */
 	int rc;
+	char *temp;
 	path_prefix_len = strlen(path_prefix);
 	/* Set up the directories to skip */
 	for (size_t i = 0; i < sizeof(skip_dirs) / sizeof(skip_dirs[0]); i++) {
@@ -131,8 +132,9 @@ int walk_tree(struct manifest *manifest, const char *start)
 			found->in_manifest = true;
 		}
 	}
-	/* list files/directories which are extra */
-	for (int i = 0; i < nF; i++) {
+	/* list files/directories which are extra.
+	 * This is reverse so that files are removed before their parent dirs */
+	for (int i = nF - 1; i >= 0; i--) {
 		int skip_len; /* Length of directory name we are skipping
 			       * could have used strlen(skip_dir), but speed! */
 		if (!F[i].in_manifest) {
@@ -148,9 +150,23 @@ int walk_tree(struct manifest *manifest, const char *start)
 			if (F[i].dir) { /* Start of new dir to skip */
 				skip_dir = F[i].filename;
 				skip_len = strlen(skip_dir);
-				printf("%s/\n", F[i].filename);
+				if (fix) {
+					string_or_die(&temp, "%s%s", path_prefix, F[i].filename);
+					fprintf(stderr, "REMOVING DIR %s/\n", F[i].filename);
+					remove(temp);
+					free(temp);
+				} else {
+					printf("%s/\n", F[i].filename);
+				}
 			} else {
-				printf("%s\n", F[i].filename);
+				if (fix) {
+					string_or_die(&temp, "%s%s", path_prefix, F[i].filename);
+					fprintf(stderr, "REMOVING %s\n", F[i].filename);
+					remove(temp);
+					free(temp);
+				} else {
+					printf("%s\n", F[i].filename);
+				}
 			}
 		} else {
 			skip_dir = NULL;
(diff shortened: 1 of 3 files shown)


src/extra_files.c
```diff
@@ -131,8 +131,9 @@ int walk_tree(struct manifest *manifest, const char *start)
 			found->in_manifest = true;
 		}
 	}
-	/* list files/directories which are extra */
-	for (int i = 0; i < nF; i++) {
+	/* list files/directories which are extra.
+	 * This is reverse so that files are removed before their parent dirs */
+	for (int i = nF-1; i > 0; i--) {
```
phmccarty: code style nit: whitespaces are needed surrounding the subtraction operator

src/extra_files.c
```diff
@@ -148,9 +149,19 @@ int walk_tree(struct manifest *manifest, const char *start)
...
+					printf("REMOVING DIR %s/\n", F[i].filename);
+					remove(F[i].filename);
+				} else {
+					printf("%s/\n", F[i].filename);
+				}
 			} else {
-				printf("%s\n", F[i].filename);
+				if (fix) {
+					printf("REMOVING %s\n", F[i].filename);
+					remove(F[i].filename);
```
icarus-sparry: *VERY IMPORTANT.* F[x].filename holds the name relative to the root. If path_prefix_len is not zero you need to prepend the prefix in the calls to remove.

src/verify.c
```diff
@@ -774,11 +774,21 @@ int verify_main(int argc, char **argv)
 		if ((file_not_fixed_count == 0) && (file_not_replaced_count == 0)) {
 			remove_orphaned_files(official_manifest);
 		}
+		if (cmdline_option_picky) {
+			char *start = mk_full_filename(path_prefix, "/usr");
+			fprintf(stderr, "--picky removing extra files under %s\n", start);
+			ret = walk_tree(official_manifest, start, true);
+			if (ret >= 0) {
+				file_checked_count = ret;
```
icarus-sparry: Not sure this statistic adds much value in the context of a 'fix'.

src/extra_files.c
```diff
@@ -148,9 +149,19 @@ int walk_tree(struct manifest *manifest, const char *start)
...
+				if (fix) {
+					printf("REMOVING DIR %s/\n", F[i].filename);
+					remove(F[i].filename);
+				} else {
+					printf("%s/\n", F[i].filename);
+				}
 			} else {
-				printf("%s\n", F[i].filename);
+				if (fix) {
+					printf("REMOVING %s\n", F[i].filename);
```
tmarcu: Yep should be stderr.

src/extra_files.c
tmarcu: Thanks for pointing this out, this has caused problems in mixer before when not properly handled.

src/verify.c
tmarcu: I think it helps understanding there's files that are just bad on your system that should be something else, + there are files on your system swupd didn't add. If anyone else has strong opinions on this though I'm open to them.

src/extra_files.c
```diff
@@ -131,8 +132,9 @@ int walk_tree(struct manifest *manifest, const char *start)
 			found->in_manifest = true;
 		}
 	}
-	/* list files/directories which are extra */
-	for (int i = 0; i < nF; i++) {
+	/* list files/directories which are extra.
+	 * This is reverse so that files are removed before their parent dirs */
+	for (int i = nF-1; i >= 0; i--) {
```
phmccarty: there's still a whitespace issue on this line...

src/extra_files.c
```diff
@@ -148,9 +150,23 @@ int walk_tree(struct manifest *manifest, const char *start)
 			if (F[i].dir) { /* Start of new dir to skip */
 				skip_dir = F[i].filename;
 				skip_len = strlen(skip_dir);
-				printf("%s/\n", F[i].filename);
+				if (fix) {
+					string_or_die(&temp, "%s%s", path_prefix, F[i].filename);
+					printf("REMOVING DIR %s/\n", F[i].filename);
```
phmccarty: should this message be printed to stdout or stderr?

src/extra_files.c
```diff
@@ -148,9 +150,23 @@ int walk_tree(struct manifest *manifest, const char *start)
...
+					remove(temp);
+					free(temp);
+				} else {
+					printf("%s/\n", F[i].filename);
+				}
 			} else {
-				printf("%s\n", F[i].filename);
+				if (fix) {
+					string_or_die(&temp, "%s%s", path_prefix, F[i].filename);
+					printf("REMOVING %s\n", F[i].filename);
```
phmccarty: similar comment for this line regarding stdout vs. stderr.

src/verify.c
phmccarty: I like seeing the progress output being paralleled between `--picky` and `--fix --picky`, so I think this change is appropriate.

src/extra_files.c
tmarcu: Is this an error? It's not an error in execution of the code/swupd, but it can be considered an error of the system state. Is that appropriate to print to stderr?

phmccarty: It's not an error, but there is currently a mixture of printf/fprintf(stderr) in the code, and I'm trying to better understand the convention we established recently. I would just ignore my comments for now and we can revisit later.

PR #220, opened 2017-05-25
verify: enforce format transitions

phmccarty:
Add error checking to enforce format boundaries when using `swupd verify -m VER`. Also, enable the -x/--force option to bypass the error checking, and add two new functional tests.


tmarcu:
+1 for the code, have you tested this at all yet?

phmccarty:
@tmarcu Yes, I did some manual tests, and they passed. I would add new functional tests, but we always pass `--format=staging` right now, so the test library will need some rethink first.

tpepper:
Do you see value in allowing a --force-fix-to-incompatible-format type override flag, even if the default sensibly becomes only moving within a given format?  It might trash the system.  But it might also fix an otherwise trashed system.  Arguably once fix repairs to a version within a given format, an update is expected to work though so, maybe no need unless that two step update becomes too onerous.

tpepper:
Another thought on the --force ability:  If I'm running a mix, I might have a different format (eg: default is 1).  If there's no override, I can't verify/fix/force myself between an "official" (non-mix) build and a "mixer" build.  That's a potentially handy thing to allow.

phmccarty:
I agree that we should add a --force option to override the default enforcement. Maybe we can just reuse the existing --force option?

@tmarcu Can you comment about the mixer consideration @tpepper mentioned above? I'm not sure of the details, especially with your new mixer integration changes.

tmarcu:
@phmccarty I think this should be an override-able feature with the general --force, at most. If you're on a mix you will not be forcing over it 'cause swupd will see the format it's in and be happy. However, there are cases where it would be nice not to be hard limited on ignoring the barrier.

matthewrsj:
@phmccarty regarding your comment about always passing `--format=staging` there now exists a `$SWUPD_OPTS_NO_FMT` that allows you to pass all the opts except the format flag, which you can then set yourself.

phmccarty:
Uploaded a revised patch set to enable the existing -x/--force option for overriding the format error checking, and adding two new functional tests for the error check and the override capability.

tmarcu:
+1

matthewrsj:
+1

PR #222, opened 2017-06-05
Add -Y option to parseargs argument

albertomurillo:
--picky option was being recognized by swupd verify but not
its equivalent -Y

Signed-off-by: Alberto Murillo Silva <alberto.murillo.silva@intel.com>


tmarcu:
+1

PR #224, opened 2017-06-22
Mixer integration

tmarcu:
This patch set introduces functionality for swupd to handle content that is locally built, and allows it to multiplex between upstream content and local user additive data. In the curent state of swupd, the only method of adding custom user content is by using the Mixer tool to create a custom image, and then acting as the OSV over it. Adding content to an official upstream image is not officially supported.
This patch enables users to effortlessly add their own bundles to an already running, official Clear Linux image, without needing to mix their own version from scratch.
The client will automatically detect if user content exists in /usr/share/mix, and appropriately attempt to load it for an update, verify, or bundle operation. It takes the content in /usr/share/mix, and merges the mix Manifest.MoM with the official Manifest.MoM, such that swupd only has to consume a single MoM per usual. The merged MoM contains all upstream data, and all custom data the client will update, and is signed by a custom user cert only if the upstream MoM verifies succesfully, ensuring continued security.
Should the client detect that there is a newer upstream release, it will automatically regenerate an updated Manifest.MoM to update itself even if the user has not updated their custom bundles or OS version.


tmarcu:
Please ignore travis for this first iteration...

src/update.c
```diff
@@ -272,6 +272,17 @@ static bool system_on_mix(void)
 	return !access("/usr/share/defaults/swupd/mixed", R_OK);
 }
 
+static bool need_new_upstream(int server)
+{
+	if (!access("/usr/share/mix/.clearversion", R_OK)) {
```
phmccarty: use the MIX_DIR macro

src/update.c
```diff
@@ -272,6 +272,17 @@ static bool system_on_mix(void)
...
+		int version = read_mix_version_file("/usr/share/mix/.clearversion", path_prefix);
```
phmccarty: use the MIX_DIR macro

src/version.c
```diff
@@ -96,13 +96,12 @@ int get_current_version(char *path_prefix)
 	FILE *file;
 	int v = -1;
 	char *buildstamp;
-	char *src, *dest;
 
-	string_or_die(&buildstamp, "%s/usr/lib/os-release", path_prefix);
```
phmccarty: The changes to this function seem unrelated... You already added a new function to read a mix version file. That seems appropriate, but I think Clear Linux's version should still be read from /usr/lib/os-release, since that is the distro-agnostic location to include such metadata.

src/update.c
```diff
@@ -223,20 +262,28 @@ static void run_postupdate_action(void)
 	free(post_update_action);
 }
 
+static bool system_on_mix(void)
+{
+	return !access("/usr/share/defaults/swupd/mixed", R_OK);
```
phmccarty: `access()` does not return a bool, so I would prefer to see this code expanded for clarity.

src/subscriptions.c
```diff
@@ -111,6 +111,38 @@ void read_subscriptions_alt(struct list **subs)
...
+				create_and_append_subscription(subs, ent->d_name);
+			}
+		}
+
+		closedir(dir);
+	}
+
+	free(path);
+	*subs = list_sort(*subs, subscription_sort_component);
+}
```
phmccarty: This function looks very, very similar to the one above. Looks like a good opportunity for refactoring.

src/update.c
```diff
@@ -555,6 +555,15 @@ int main_update()
 	run_scripts();
 	grabtime_stop(&times);
 
+	/* Create the state file that will tell swupd it's on a mix on future runs */
+	if (mix_exists && !system_on_mix()) {
+		int fd = open("/usr/share/defaults/swupd/mixed", O_RDWR | O_CREAT, S_IRWXU | S_IRGRP | S_IROTH);
```
phmccarty: the executable bit should not be set, so you can change `S_IRWXU` to `S_IRUSR | S_IWUSR`

src/update.c
```diff
@@ -555,6 +555,15 @@ int main_update()
...
+		if (fd == -1) {
+			fprintf(stderr, "ERROR: Failed to create 'mixed' statefile\n");
```
phmccarty: if the file fails to be created, what is the fallout? is it safe to print the error and continue?

src/update.c
```diff
@@ -223,9 +267,26 @@ static void run_postupdate_action(void)
...
 }
 
+static bool system_on_mix(void)
+{
+	return !access("/usr/share/defaults/swupd/mixed", R_OK);
+}
+
+static bool need_new_upstream(int server)
+{
+	if (!access("/usr/share/mix/.clearversion", R_OK)) {
```
phmccarty: use the MIX_DIR macro

src/update.c
```diff
@@ -223,9 +267,26 @@ static void run_postupdate_action(void)
...
+		int version = read_mix_version_file("/usr/share/mix/.clearversion", path_prefix);
```
phmccarty: use the MIX_DIR macro

src/subscriptions.c
tmarcu: I can add logic to ignore os-core and pick between bundle dirs depending on the mix_state?

src/version.c
tmarcu: Is there a reason we even have a version file then? They are always in sync with one another, and it's easier to read this simple file vs parsing the os-release file in C.

src/update.c
tmarcu: *Mumbles about type promotion in C...*  ;) 
I'll change the definition for clarity.

src/update.c
```diff
@@ -317,36 +328,50 @@ int main_update()
...
+			printf("NEW CLEAR AVAILABLE %d\n", server_version);
+			ret = check_manifests_uniqueness(server_version, mix_server_version);
+			if (ret) {
+				printf("\n\t!! %i collisions were found between mix and upstream, please re-create mix !!\n", ret);
+			}
+
+			/* Update the clearversion that will be used to generate the new mix content */
+			FILE *verfile = fopen(MIX_DIR ".clearversion", "w+");
+			if (!verfile) {
+				fprintf(stderr, "ERROR: fopen() returned %s\n", strerror(errno));
```
tmarcu: That's not guaranteed, that'd be undefined in this case depending on why fopen failed on that file.What do you recommend instead?

src/version.c
phmccarty: I don't know why we have that version file :-) After considering this some more, I'm inclined to read the version file instead, since it's "formatless" and thus trivial to consume, and we already do this for the conf files under /usr/share/defaults/swupd.

But we should rename the version file to `/usr/share/swupd/version`, or similar. This naming change can be introduced for a format bump, of course. The Clear Linux distro "version" can still live in /usr/lib/os-release, since that's the standard location for people to discover this information, and that file will always be in sync with swupd's version file. I'm also not bothered with the duplicate information, since it is already duplicate.

How does this sound?

phmccarty: If you want to go this route, I would prefer seeing this change pushed in later, especially because it would tied to a format bump release.

src/signature.c
```diff
@@ -414,8 +414,9 @@ int verify_callback(int ok, X509_STORE_CTX *stor)
  * returns: true if signature verification succeeded, false if verification
  * failed, or the signature download failed
  */
```
phmccarty: docstring needs an update

src/signature.c
```diff
@@ -414,8 +414,9 @@ int verify_callback(int ok, X509_STORE_CTX *stor)
...
-bool download_and_verify_signature(const char *data_url, const char *data_filename)
+bool download_and_verify_signature(const char *data_url, const char *data_filename, int version, bool mix_exists)
```
phmccarty: You could simplify this change by only adding one extra parameter instead of two. Maybe

```
bool download_and_verify_signature(const char *data_url, const char *data_filename, const char *mixer_sig)
```

The caller would initialize mixer_sig to NULL for the non-mix case, or set the full path for the mix case. Then the conditional code below can check if mixer_sig is NULL or non-NULL.

src/curl.c
```diff
@@ -219,6 +219,7 @@ int swupd_curl_get_file(const char *url, char *filename, struct file *file,
 				abort();
 			}
 		}
+
```
phmccarty: stray extra line added

src/manifest.c
```diff
@@ -171,11 +171,18 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
...
 	uint64_t contentsize = 0;
 	int manifest_hdr_version;
 	int manifest_enc_version;
 
-	string_or_die(&filename, "%s/%i/Manifest.%s", state_dir, version, component);
+	if (!statepath) {
+		basedir = state_dir;
+	} else {
+		basedir = statepath;
+	}
```
phmccarty: Is your "statepath" always the same for a mix? If so, I would remove the statepath parameter and set basedir according to the value of is_mix (i.e. the global value, or a hardcoded value for the mix).

src/manifest.c
```diff
@@ -494,14 +514,25 @@ static int try_delta_manifest_download(int current, int new, char *component, st
...
+	bool is_mix = false;
+
+	if (!statepath) {
+		basedir = state_dir;
+	} else {
+		basedir = statepath;
+	}
+
+	/* if they differ then we are on a mix */
+	is_mix = strcmp(basedir, state_dir);
```
phmccarty: strcmp does not return a bool....

src/manifest.c
```diff
@@ -525,12 +555,26 @@ static int retrieve_manifests(int current, int version, char *component, struct
...
 		if (try_delta_manifest_download(current, version, component, file) == 0) {
 			return 0;
 		}
 	}
 
+	/* If it's mix content just hardlink instead of curl download */
+	if (is_mix) {
+		string_or_die(&filename, "%s/%i/Manifest.%s.tar", state_dir, version, component);
+		string_or_die(&url, "%s/%i/Manifest.%s.tar", basedir, version, component);
+		ret = link(url, filename);
```
phmccarty: if the hardlink fails, what is the desired behavior?

src/verify.c
```diff
@@ -308,6 +308,20 @@ static struct list *download_loop(struct list *files, int isfailed)
...
 
 		if (hash_needs_work(file, local.hash)) {
+			/* Mix content is local, so don't queue files up for curl downloads */
+			if (file->is_mix) {
+				char *filename;
+				char *url;
+				string_or_die(&url, "%s/%i/files/%s.tar", MIX_STATE_DIR, file->last_change, file->hash);
+				string_or_die(&filename, "%s/download/.%s.tar", state_dir, file->hash);
+				file->staging = filename;
+				link(url, filename);
```
phmccarty: error handling for hardlink failure?

include/swupd.h
```diff
@@ -45,17 +50,20 @@ struct sub {
 
 	/* oldversion: set to 0 by calloc(), possibly overridden by MoM read */
 	int oldversion;
+	int is_mix : 1;
```
phmccarty: this new field for `struct sub` does not appear to be used, so you can remove it

src/manifest.c
```diff
@@ -494,14 +514,25 @@ static int try_delta_manifest_download(int current, int new, char *component, st
...
 	int ret = 0;
 	char *tar;
 	struct stat sb;
+	bool is_mix = false;
+
+	if (!statepath) {
+		basedir = state_dir;
+	} else {
+		basedir = statepath;
+	}
```
phmccarty: Similar to my question from the code in `manifest_from_file`, is statepath always the same value for the mix case? If so, then I would prefer the API to be consistent and add is_mix as a new parameter to the function instead.

src/subscriptions.c
phmccarty: Sure, that sounds okay to me.

src/update.c
```diff
@@ -76,7 +78,16 @@ static struct list *full_download_loop(struct list *updates, int isfailed)
 		if (file->is_deleted) {
 			continue;
 		}
-
+		/* Mix content is local, so don't queue files up for curl downloads */
+		if (file->is_mix) {
+			string_or_die(&url, "%s/%i/files/%s.tar", MIX_STATE_DIR, file->last_change, file->hash);
+			string_or_die(&filename, "%s/download/.%s.tar", state_dir, file->hash);
+			file->staging = filename;
+			link(url, filename);
```
phmccarty: no check for hardlink failure?

src/update.c
phmccarty: So, I'm mainly concerned about calling fprintf(...), and nothing else. The error doesn't appear very actionable to me, so is even worth printing?

src/update.c
```diff
@@ -258,20 +320,52 @@ int main_update()
...
+
+			/* Update the clearversion that will be used to generate the new mix content */
+			FILE *verfile = fopen(MIX_DIR ".clearversion", "w+");
+			if (!verfile) {
+				fprintf(stderr, "ERROR: fopen() returned %s\n", strerror(errno));
+			}
+			fprintf(verfile, "%d", server_version);
+			fclose(verfile);
+
+			system("/usr/bin/add-pkg.sh");
```
phmccarty: Do you care about the return value of this script?

tpepper:
Some of the commits could use commit messages.  Like "Fix compliant manifest check"...what was broken?  I can see it based on the patch, but review is easier if the break+fix are described and the code clearly matches the description.

tpepper:
For local mix content, what's the trust requirement?  If /usr/share/mix is the place, who creates that and enforces non-administrative users cannot mutate its contents?

tpepper:
The overall PR message could also use more description of the architecture...what's a mix, the sub-use cases this is enabling for how a mix and non-mix content are merged together at runtime on a swupd client system.

src/update.c
tpepper: Probably need to add /usr/share/defaults/swupd to create_required_dirs().  And/or enforce in swupd-server and mixer that it exists.

tmarcu:
Squashed into more manageable commits, 8 total now. Still addressing comments above in the code and will push final version.

tmarcu:
Rebased onto latest master.

src/manifest.c
```diff
@@ -713,9 +786,11 @@ struct manifest *load_manifest(int current, int version, struct file *file, stru
 		}
 		return NULL;
 	}
-	retried = false;
+	if (!retried) {
+		retried = false;
```
matthewrsj: `if (retried == false) { retried = false; }` ?

src/manifest.c
```diff
@@ -1395,3 +1492,73 @@ struct file *search_file_in_manifest(struct manifest *manifest, const char *file
...
+	}
+	return array;
+}
+
+void print_manifest_array(struct file **array, int filecount)
+{
+	for (int i = 0; i < filecount; i++) {
+		printf("[%d]/%d %s\n", i, filecount, array[i]->filename);
+	}
+	printf("SIZE: %d\n", filecount);
```
matthewrsj: I think it would make more sense to print something like `NUMBER OF FILES: filecount` since this isn't the actual size of all the files in the manifest, just the number of them.

src/packs.c
```diff
@@ -52,19 +52,31 @@ static int download_pack(int oldversion, int newversion, char *module)
...
 	}
 
-	string_or_die(&url, "%s/%i/pack-%s-from-%i.tar", content_url, newversion, module, oldversion);
-
-	err = swupd_curl_get_file(url, filename, NULL, NULL, true);
-	if (err) {
+	if (is_mix) {
+		string_or_die(&url, "%s/%i/pack-%s-from-%i.tar", MIX_STATE_DIR, newversion, module, oldversion);
+		err = link(url, filename);
+		printf("Linked %s to %s\n", url, filename);
```
matthewrsj: This should be printed after the `if (err) {...}` statement below, so it doesn't look like the link succeeded when it didn't.

src/packs.c
```diff
@@ -52,19 +52,31 @@ static int download_pack(int oldversion, int newversion, char *module)
...
 		free(url);
-		if ((lstat(filename, &stat) == 0) && (stat.st_size == 0)) {
-			unlink(filename);
+	} else {
+		string_or_die(&url, "%s/%i/pack-%s-from-%i.tar", content_url, newversion, module, oldversion);
+
+		err = swupd_curl_get_file(url, filename, NULL, NULL, true);
+		if (err) {
+			free(url);
+			if ((lstat(filename, &stat) == 0) && (stat.st_size == 0)) {
```
matthewrsj: Why only remove empty files on error?

src/update.c
```diff
@@ -87,6 +106,39 @@ static struct list *full_download_loop(struct list *updates, int isfailed)
...
+	free_manifest_array(mixerfull);
+	free_manifest(clear);
+	free_manifest(mixer);
+
+	return ret;
+}
+
+void setup_mix_update()
+{
+}
```
matthewrsj: Is this a TODO?

src/verify.c
```diff
@@ -308,6 +308,26 @@ static struct list *download_loop(struct list *files, int isfailed)
...
+					if (rename(url, filename) != 0) {
+						fprintf(stderr, "Failed to copy local mix file: %s\n", filename);
+						continue;
+					}
+				}
+				untar_full_download(file);
+				free(filename);
+				free(url);
+				continue;
+			}
```
matthewrsj: I've seen the above hunk in two other places in the diff. Could this be turned into a helper function with a little modification?

src/manifest.c
icarus-sparry: +1 to the query. I also think the variable needs to be renamed, perhaps to "partial_cleanup_needed"?
The whole function could do with being rewritten to remove all the retry logic and putting the retry logic in a new function.

src/manifest.c
```diff
@@ -1391,3 +1476,73 @@ struct file *search_file_in_manifest(struct manifest *manifest, const char *file
...
+{
+	struct file **found;
+	int ret = 0;
+
+	qsort(a, searchsize, sizeof(struct file *), cmpnames);
+	qsort(b, size, sizeof(struct file *), cmpnames);
+	printf("Checking manifest uniqueness...\n");
+	for (int i = 0; i < searchsize; i++) {
+		found = bsearch(a[i], b, size, sizeof(struct file *), bsearch_file_helper);
+		if (found) {
```
icarus-sparry: Do the manifests only have existing files in them, and not for example deleted files? In other words do we have a case where the current manifest has a file, e.g. /etc/oldissue marked as deleted but our new package wants to create this file.

src/manifest.c
```diff
@@ -1391,3 +1476,73 @@ struct file *search_file_in_manifest(struct manifest *manifest, const char *file
...
+				continue;
+			}
+			if (hash_equal(a[i]->hash, (*found)->hash)) {
+				continue;
+			}
+			fprintf(stderr, "ERROR: Conflict found for file: %s\n", a[i]->filename);
+			ret++;
+		}
+	}
+	return ret; // No collisions were found, so manifest is purely additive
```
icarus-sparry: Poor comment, I suggest "num collisions found, 0 => purely additive"

src/packs.c
icarus-sparry: My guess is so the next time it will try getting the file again. see comment on line 91

src/manifest.c
```diff
@@ -525,12 +551,30 @@ static int retrieve_manifests(int current, int version, char *component, struct
...
 			return 0;
 		}
 	}
 
+	/* If it's mix content just hardlink instead of curl download */
+	if (is_mix) {
+		string_or_die(&filename, "%s/%i/Manifest.%s.tar", state_dir, version, component);
+		string_or_die(&url, "%s/%i/Manifest.%s.tar", basedir, version, component);
+		ret = link(url, filename);
+		/* Try doing a regular rename if hardlink fails */
```
icarus-sparry: This link, if it fails call rename, ought to be its own function. It probably ought to have a "make a copy" as another fallback.

src/update.c
icarus-sparry: At least thow in the filename (MIX_DIR/.clearversion)

src/packs.c
matthewrsj: You're right, makes sense.

src/manifest.c
```diff
@@ -543,6 +587,19 @@ static int retrieve_manifests(int current, int version, char *component, struct
 		goto out;
 	}
 
+	if (ret) {
```
matthewrsj: Yeah, line number changed with rebase.

matthewrsj:
This will also need usage documentation somewhere

include/swupd.h
```diff
@@ -120,6 +130,12 @@ struct file {
 	FILE *fh;      /* file written into during downloading */
 };
 
+struct filerecord {
+	char *filename;
+	bool dir;
```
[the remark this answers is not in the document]
tmarcu: Perhaps, but this isn't new code it's simply moved. Let's keep that in another PR.

src/manifest.c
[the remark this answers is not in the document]
tmarcu: /me wonders if this is an artifact from previous rebase...

src/manifest.c
```diff
@@ -1391,3 +1476,73 @@ struct file *search_file_in_manifest(struct manifest *manifest, const char *file
...
+	struct file *file;
+	int i = 0;
+	int numfiles = manifest->filecount;
+
+	if (!manifest) {
+		fprintf(stderr, "Manifest is NULL!\n");
+		return NULL;
+	}
+
+	array = malloc(sizeof(struct file *) * numfiles);
```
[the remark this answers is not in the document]
tmarcu: If we can't count on the headers being correct then there's no point in having them. That implies we have bad code and are OK leaving it to create our updates...

src/manifest.c
```diff
@@ -1391,3 +1476,73 @@ struct file *search_file_in_manifest(struct manifest *manifest, const char *file
...
+
+	iter = list_head(manifest->files);
```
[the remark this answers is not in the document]
tmarcu: I don't disagree :)

src/manifest.c
```diff
@@ -1391,3 +1476,73 @@ struct file *search_file_in_manifest(struct manifest *manifest, const char *file
...
+{
+	return strcmp((*(struct file **)a)->filename, (*(struct file **)b)->filename);
+}
+
+int enforce_compliant_manifest(struct file **a, struct file **b, int searchsize, int size)
+{
+	struct file **found;
+	int ret = 0;
+
+	qsort(a, searchsize, sizeof(struct file *), cmpnames);
```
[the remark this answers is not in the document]
tmarcu: It won't work with the way it's sorted to begin with. The sort function in swupd sorts based on multiple heuristics and bsearch does not play well with that because it's not _really_ sorted.

src/manifest.c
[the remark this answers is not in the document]
phmccarty: @tmarcu +1 to addressing code consistency later, considering the number of functional changes in this PR.

src/manifest.c
tmarcu: If there's any collision the type/action on the file doesn't matter, because it's a conflict between the official manifest & user manifest. This of course can just be ignored, but it's a warning that you're possibly breaking something by overriding it.

src/manifest.c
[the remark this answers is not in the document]
icarus-sparry: The code says

sort the array 'a', using the filename as the key
sort the array 'b', using the filename as the key.
loop through the elements of 'a' in order, searching for each of them in 'b'.
     search, using bsearch, the 'b' array for this element.
     if found, then ensure that the hash is the same....
There is nothing here that seems to depend on processing the elements in 'a' in alphabetical order. You would get the same return value if you processed them in reverse order, or a random order. Hence my suggestion to not bother sorting 'a'.

Now if you choosing to step through the two arrays at the same time, rather than stepping through 'a' and searching for matching elements in 'b' with bsearch, you would want both sorted. Your costs would then be n.log(n) for each of the 2 sorts, and n for the compare, rather than n.log(n) for the sort of 'b' and n (length of a) log(n) (bsearch of b).

tmarcu:
Fixed a double free bug and tested this again from scratch on my actual system -> add package, migrate to mix, bundle add, run new package, remove bundle. Will throw out some documentation for this, but what format would you like it in? Man page? Readme?

phmccarty:
@tmarcu Maybe a `swupd-mixer` man page in section 7. The man pages under docs/ are written in rst and automatically converted to man pages during the build.

tmarcu:
@phmccarty Sounds good, I can do that and I will do a page for the mixer docs page on clearlinux.org as well I think.

tmarcu:
Will push this https://github.com/tmarcu/clear-linux-documentation/pull/1/commits/5c9e9f3c4a16b8b2f994644d5106d32f964b4986 out as well once this PR is merged.

src/update.c
```diff
@@ -314,7 +412,11 @@ int main_update()
 	/* Step 3: setup manifests */
 
 	/* get the from/to MoM manifests */
-	current_manifest = load_mom(current_version, false);
+	if (system_on_mix()) {
+		current_manifest = load_mom(current_version, false, mix_exists);
```
matthewrsj: How would you get into a state where you system is on a mix (`system_on_mix()` - `/usr/share/defaults/swupd/mixed` exists) but a mix doesn't exist (`/usr/share/mix` doesn't have any bundles in it)?

Can you be on an empty mix? If not, this could just be

`current_manifest = load_mom(current_version, false, system_on_mix());`

src/update.c
```diff
@@ -483,6 +586,16 @@ int main_update()
 	run_scripts(re_update);
 	grabtime_stop(&times);
 
+	/* Create the state file that will tell swupd it's on a mix on future runs */
+	if (mix_exists && !system_on_mix()) {
+		int fd = open("/usr/share/defaults/swupd/mixed", O_RDWR | O_CREAT, S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH);
```
matthewrsj: Again, can we avoid hardcoding this?

src/verify.c
```diff
@@ -704,7 +724,14 @@ int verify_main(int argc, char **argv)
 	times = init_timelist();
 
 	grabtime_start(&times, "Load and recurse Manifests");
-	official_manifest = load_mom(version, false);
+
+	/* If upstream URL is passed, user is likely trying to get back to the official stream,
+	 * so ignore any custom mixer related state */
+	if (strncmp(content_url, "https://download.clearlinux.org/update", 38) == 0) {
```
matthewrsj: Instead of hardcoding the upstream, which we are trying to get away from, why not detect that it is non-empty? if they specify a content_url you are probably right in assuming they want to get back to upstream, but we definitely shouldn't only make that assumption when it points to `https://specific.clear.url`.

Anyways, this ignores the perfectly valid `https://cdn.download.clearlinux.org/update`

src/update.c
```diff
@@ -278,20 +339,57 @@ int main_update()
...
+			ret = EXIT_FAILURE;
+			goto clean_curl;
+		}
+		/* Check if a new upstream version is available so we can update to it still */
+		if (need_new_upstream(server_version)) {
+			printf("NEW CLEAR AVAILABLE %d\n", server_version);
+			ret = check_manifests_uniqueness(server_version, mix_server_version);
+			if (ret) {
+				printf("\n\t!! %i collisions were found between mix and upstream, please re-create mix !!\n", ret);
+			}
```
matthewrsj: So, this seems like an error, but you don't actually error here. This should also go to stderr.

tmarcu: It's not so much an error as it is us saying what we explicitly support right now. I can make it overridable with a config, but I want people to be able to still add conflicting content and just be told if that is the case.

src/verify.c
tmarcu: Probably right and assuming are not good enough IMO. Another option would be to compare it to the URLs listed in the /usr/share/defaults/swupd/* and compare those, then it wouldn't have to be maintained here.

src/update.c
tmarcu: Absolutely :)

src/packs.c
```diff
@@ -52,20 +52,30 @@ static int download_pack(int oldversion, int newversion, char *module)
...
-	if (err) {
+	if (is_mix) {
+		string_or_die(&url, "%s/%i/pack-%s-from-%i.tar", MIX_STATE_DIR, newversion, module, oldversion);
+		err = link(url, filename);
+		printf("Linked %s to %s\n", url, filename);
+		if (err) {
+			free(filename);
+			free(url);
+			return err;
+		}
```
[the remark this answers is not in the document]
tmarcu: Debug, will remove.

docs/swupd-add-pkg.4.rst
```diff
@@ -0,0 +1,48 @@
...
+===========
+
+Adds a given local package to the specified <bundle_name> file, and
+automatically generates all content with mixer for swupd to use. It creates
+a very minimalist mix, unlike regular mixer usage where one composes the
+entire OS. This instead outputs only the new bundle content, and os-core -
+to signify the version changes - in order to augment the regular upstream
+update content. Swupd is aware of this new user added content, and if it
+exists on the system, swupd will attempt to use it for any of its operations,
+e.g update, verify, bundle add, etc.
```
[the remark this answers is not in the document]
tmarcu: Packages cannot ever conflict or mismatch between bundles, otherwise update creation (chroots actually) will fail. Adding the package would error out and the user would have to fix the conflict before they would ever get update content for it, so this is OK.

include/swupd.h
```diff
@@ -195,6 +212,18 @@ extern void grabtime_stop(timelist *list);
 extern void print_time_stats(timelist *list);
 extern timelist init_timelist(void);
 
+static inline int bsearch_helper(const void *A, const void *B)
```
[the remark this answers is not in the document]
phmccarty: That is a good point. It is `bsearch_file_helper()` that is used elsewhere now (in a different commit), and `bsearch_helper()` is still only used in extra_files.c, so it doesn't need to live in swupd.h.

src/manifest.c
```diff
@@ -1391,3 +1476,73 @@ struct file *search_file_in_manifest(struct manifest *manifest, const char *file
...
 }
+
+/* Ideally have the manifest sorted already by this point */
+struct file **manifest_files_to_array(struct manifest *manifest)
+{
+	struct file **array;
+	struct list *iter;
+	struct file *file;
+	int i = 0;
+	int numfiles = manifest->filecount;
```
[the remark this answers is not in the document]
phmccarty: This comment is still relevant... The check for manifest is NULL (line 1426) needs to happen before this line (line 1424), else you'll get a segfault if that condition holds.

src/manifest.c
```diff
@@ -1413,3 +1413,67 @@ struct file *search_file_in_manifest(struct manifest *manifest, const char *file
...
+
+	if (!manifest) {
+		fprintf(stderr, "Manifest is NULL!\n");
```
phmccarty: This message is not very helpful for the end user. Do we need to print anything in this case?

src/manifest.c
```diff
@@ -1413,3 +1413,67 @@ struct file *search_file_in_manifest(struct manifest *manifest, const char *file
...
+static int cmpnames(const void *a, const void *b)
+{
+	return strcmp((*(struct file **)a)->filename, (*(struct file **)b)->filename);
+}
+
+int enforce_compliant_manifest(struct file **a, struct file **b, int size)
+{
+	struct file **found;
+
+	//qsort(a, size, sizeof(char *), cmpnames);
```
phmccarty: Can you remove this commented code?

src/manifest.c
```diff
@@ -1413,3 +1413,67 @@ struct file *search_file_in_manifest(struct manifest *manifest, const char *file
...
+	print_manifest_array(a, size);
```
phmccarty: Is this function called for debugging purposes?

src/manifest.c
```diff
@@ -1413,3 +1413,67 @@ struct file *search_file_in_manifest(struct manifest *manifest, const char *file
...
+	printf("Checking manifest uniqueness...\n");
+	for (int i = 0; i < size; i++) {
+		found = bsearch(a[i], b, size, sizeof(struct file *), bsearch_file_helper);
+		if (found) {
+			if (hash_equal(a[i]->hash, (*found)->hash)) {
+				continue;
+			}
+			fprintf(stderr, "ERROR: Conflict found for file: %s & %s\n", a[i]->filename, (*found)->filename);
```
phmccarty: This error message could be very long. Maybe split it into multiple lines:

```
fprintf(stderr, "ERROR: Conflict found for file:\n  %s\n  %s\n", a[i]->filename, (*found)->filename);
```

include/swupd.h
```diff
@@ -21,6 +21,8 @@ extern "C" {
 #define SWUPD_VERSION_IS_DEVEL(v) (((v) % SWUPD_VERSION_INCR) == 8)
 #define SWUPD_VERSION_IS_RESVD(v) (((v) % SWUPD_VERSION_INCR) == 9)
 
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
```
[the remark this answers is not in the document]
phmccarty: I don't see `MIN` used anywhere either...

src/bundle.c
```diff
@@ -601,6 +601,7 @@ int add_subscriptions(struct list *bundles, struct list **subs, int current_vers
 		}
 
 	retry_manifest_download:
+		printf("add_subscriptions: load_manifest(): %s %d\n", file->filename, file->is_mix);
```
phmccarty: debug print?

src/manifest.c
```diff
@@ -538,12 +559,30 @@ static int retrieve_manifests(int current, int version, char *component, struct
...
+	if (file && current < version && strcmp(component, "full") != 0) {
 		if (try_delta_manifest_download(current, version, component, file) == 0) {
 			return 0;
 		}
 	}
 
+	/* If it's mix content just hardlink instead of curl download */
+	if (is_mix) {
+		string_or_die(&filename, "%s/%i/Manifest.%s.tar", state_dir, version, component);
+		string_or_die(&url, "%s/%i/Manifest.%s.tar", basedir, version, component);
```
phmccarty: This variable does not store a URL. Can you rename it to something better?

src/manifest.c
phmccarty: I agree, it would be nice to this logic wrapped in a separate function. Should there be a copy fallback, @tmarcu ?

src/manifest.c
```diff
@@ -538,12 +559,30 @@ static int retrieve_manifests(int current, int version, char *component, struct
...
+		ret = link(url, filename);
+		/* Try doing a regular rename if hardlink fails */
+		if (ret) {
+			ret = rename(url, filename);
```
phmccarty: There is no error handling for a `rename()` failure, so I think that should at least be addressed. If errors are handled by fallthrough, I would add a code comment.

src/manifest.c
```diff
@@ -678,12 +739,14 @@ struct manifest *load_mom(int version, bool latest)
 			free(log_cmd);
 		}
 	}
+	printf("MOM IS GOODDDD\n");
```
phmccarty: debug print

src/manifest.c
```diff
@@ -678,12 +739,14 @@ struct manifest *load_mom(int version, bool latest)
...
 	free(filename);
 	free(url);
-
+	printf("manifest is %s\n", manifest->component);
```
phmccarty: debug print

src/manifest.c
```diff
@@ -678,12 +739,14 @@ struct manifest *load_mom(int version, bool latest)
...
 	return manifest;
 
 out:
+	printf("why\n");
```
phmccarty: debug print

src/update.c
```diff
@@ -79,7 +82,23 @@ static struct list *full_download_loop(struct list *updates, int isfailed)
 		if (file->is_deleted) {
 			continue;
 		}
-
+		/* Mix content is local, so don't queue files up for curl downloads */
+		if (file->is_mix) {
+			string_or_die(&url, "%s/%i/files/%s.tar", MIX_STATE_DIR, file->last_change, file->hash);
```
phmccarty: This is not a URL. Needs a better variable name.

src/update.c
phmccarty: Looks like it should be removed.

src/update.c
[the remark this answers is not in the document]
phmccarty: Right, this would be more helpful:

```
fprintf(stderr, "ERROR: fopen() failed for %s (%s)\n", MIX_DIR ".clearversion", strerror(errno));

```

src/update.c
```diff
@@ -278,20 +349,57 @@ int main_update()
...
+
+	if (mix_exists) {
+		ret = check_mix_versions(&mix_current_version, &mix_server_version, path_prefix);
+		if (ret < 0) {
+			ret = EXIT_FAILURE;
+			goto clean_curl;
+		}
+		/* Check if a new upstream version is available so we can update to it still */
+		if (need_new_upstream(server_version)) {
+			printf("NEW CLEAR AVAILABLE %d\n", server_version);
```
phmccarty: debug print

src/verify.c
```diff
@@ -316,6 +316,26 @@ static struct list *download_loop(struct list *files, int isfailed)
 		}
 
 		if (hash_needs_work(file, local.hash)) {
+			/* Mix content is local, so don't queue files up for curl downloads */
+			if (file->is_mix) {
+				char *filename;
+				char *url;
+				string_or_die(&url, "%s/%i/files/%s.tar", MIX_STATE_DIR, file->last_change, file->hash);
```
phmccarty: Not a URL. Needs a better variable name.

src/verify.c
phmccarty: Yep, it is very similar to the code added to `full_download_loop()`. Not a dealbreaker for now, but a FIXME would be helpful to inform future refactoring.

src/version.c
```diff
@@ -97,25 +97,23 @@ int get_current_version(char *path_prefix)
...
+				/* Drop quotes and newline in value */
+				dest = src;
+				while (*src) {
+					if (*src == '\'' || *src == '"' || *src == '\n') {
+						++src;
+					} else {
+						*dest = *src;
+						++dest;
+						++src;
+					}
```
phmccarty: Weird whitespace changes appear here, and no functional changes.

src/version.c
```diff
@@ -97,25 +97,23 @@ int get_current_version(char *path_prefix)
...
 				}
-			}
-			*dest = 0;
+				*dest = 0;
```
phmccarty: more weird whitespace change

src/version.c
```diff
@@ -97,25 +97,23 @@ int get_current_version(char *path_prefix)
...
 
-			v = strtoull(&line[11], NULL, 10);
-			break;
+				v = strtoull(&line[11], NULL, 10);
+				break;
```
phmccarty: more weird whitespace change

src/version.c
```diff
@@ -170,6 +168,44 @@ int check_versions(int *current_version,
...
+	}
+	free(buildstamp);
+	fclose(file);
+	return v;
+}
+
+int check_mix_versions(int *current_version, int *server_version, char *path_prefix)
+{
+	*current_version = read_mix_version_file("/usr/share/clear/version", path_prefix);
+	*server_version = read_mix_version_file(MIX_STATE_DIR "version/format1/latest", path_prefix);
```
phmccarty: Is the hardcoded `format1` an issue?

src/main.c
```diff
@@ -52,6 +52,7 @@ static const struct option prog_opts[] = {
 	{ "certpath", required_argument, 0, 'C' },
 	{ "time", no_argument, 0, 't' },
 	{ "no-boot-update", no_argument, 0, 'b' },
+	{ "migrate", no_argument, 0, 'b' },
```
phmccarty: The subject of this commit is "Add better error handling throughout and cleanup", so I did not expect to see a new command line option added here. I would prefer to see it added in a separate commit so that I can make better sense of the changes.

add-pkg.sh
```diff
@@ -0,0 +1,125 @@
...
+
+if [[ -n "$pkg" && -n "$bundle" ]]; then
+	echo "Adding $pkg to $bundle..."
+
+	# Write package to bundle definition
+	echo "$pkg" | sudo tee -a "$BUNDLE_DIR/$bundle" > /dev/null
+fi
+
+# Copy os-core from upstream
+sudo -E curl -O "https://raw.githubusercontent.com/clearlinux/clr-bundles/$clearver/bundles/os-core"
```
phmccarty: This curl command could hit a server error, so it could (for example) download a 404 response. You can avoid this issue by passing the `-f` option and handling the special curl error exit status 22.

The same comment applies to all other curl commands in this script.

src/update.c
tmarcu: This was an intended print, so does everyone agree to remove it?

include/swupd.h
```diff
@@ -199,6 +224,18 @@ extern int get_value_from_path(char **contents, const char *path, bool is_abs_pa
 extern int get_version_from_path(const char *abs_path);
 extern timelist init_timelist(void);
 
+static inline int bsearch_helper(const void *A, const void *B)
+{
+	return strcmp(*(const char **)A, ((struct filerecord *)B)->filename);
+}
```
phmccarty: This function is not used anywhere aside from extra_files.c, so probably does not need to live in this header.

src/manifest.c
```diff
@@ -1408,3 +1485,66 @@ struct file *search_file_in_manifest(struct manifest *manifest, const char *file
...
+struct file **manifest_files_to_array(struct manifest *manifest)
+{
+	struct file **array;
+	struct list *iter;
+	struct file *file;
+	int i = 0;
+	int numfiles;
+
+	if (!manifest) {
+		fprintf(stderr, "Manifest is NULL!\n");
```
phmccarty: this error message doesn't look very useful

src/update.c
phmccarty: Because the message is in all-caps, it caught my attention. If that's intended, then okay.

docs/swupd-add-pkg.4.rst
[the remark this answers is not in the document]
matthewrsj: I think we determined swupd should error out in the case of collisions unless an override is passed. That should be added to the documentation.

docs/swupd.1.rst
```diff
@@ -228,6 +228,11 @@ SUBCOMMANDS
         Do not perform an update, instead download all resources needed
         to perform the update, and exit.
 
+    - `-m, --migrate`
```
matthewrsj: We should actually reserve `-m` for specifying the version to update to (#257)

src/update.c
matthewrsj: This seems useful to me.

docs/swupd.1.rst
tmarcu: Are we expecting people to do swupd update -m 18850?

phmccarty: Yes, according to discussion in #257, there is demand for that functionality.

src/manifest.c
tmarcu: That's arguable, it's technically storing a local file url, vs a regular http URL a few lines down if it fails. We have stack to spare, but I don't think it's too confusing to just use the same var. @matthewrsj thoughts?

src/manifest.c
tmarcu: Yes they are handled by fallthrough and there's a comment on https://github.com/clearlinux/swupd-client/blob/mixer-integration/src/manifest.c#L584 about it, but that may not be the best place for it. I added another in the earlier location as well to be more clear.

phmccarty: Ah, okay, I see that now. I didn't look too closely at the surrounding code block for context. Maybe the comment on line 568 could be expanded so that reading the function top to bottom will be more clear?

src/manifest.c
matthewrsj: Honestly if it is just a file url I don't really care about the naming scheme. `url = "file://"` seems reasonable to me

phmccarty: Maybe I'm misunderstanding the code, but `url` does not appear to have a `file://` prefix in this case. Are you suggesting that it should have a `file://` prefix?

src/main.c
```diff
@@ -160,7 +161,7 @@ static bool parse_options(int argc, char **argv)
 		case 'x':
 			force = true;
 			break;
-		case 'm':
+		case 'T':
```
phmccarty: +1 on the new short option, but you'll want to update the short option list and also the prog_opts array.

tmarcu: Ugh, I looked at it like 3 times before agreeing that I updated all the parts. Thanks @phmccarty

src/update.c
phmccarty: Okay, +1 for this line.

docs/swupd.1.rst
phmccarty: No update on #257 yet, but I think the team decided to move forward with implementing a `-m` option for swupd update. In this PR, @tmarcu added a `-T` option instead.

src/manifest.c
phmccarty: I will ignore my concern here for now. To me, storing a string representing a file path in a variable that implies it is a URL is confusing, but it is cosmetic when thinking about this PR as a whole.

add-pkg.sh
phmccarty: Any thoughts on this, @tmarcu ?

src/verify.c
phmccarty: Similarly for this comment, I'm going to ignore my concern about variable-name appropriateness for now.

src/update.c
phmccarty: Similarly for this comment, I'm going to ignore my concern about variable-name appropriateness for now.

src/manifest.c
phmccarty: Any comments from @tmarcu or @matthewrsj here? Do we need a copy fallback for this case?

src/manifest.c
```diff
@@ -1391,3 +1476,73 @@ struct file *search_file_in_manifest(struct manifest *manifest, const char *file
...
+	struct file **found;
+	int ret = 0;
+
+	qsort(a, searchsize, sizeof(struct file *), cmpnames);
+	qsort(b, size, sizeof(struct file *), cmpnames);
+	printf("Checking manifest uniqueness...\n");
+	for (int i = 0; i < searchsize; i++) {
+		found = bsearch(a[i], b, size, sizeof(struct file *), bsearch_file_helper);
+		if (found) {
+			if (strcmp(a[i]->filename, "/usr/lib/os-release") == 0 ||
```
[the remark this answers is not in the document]
phmccarty: I agree too. The code will be easier to read if this extended conditional check was separated out into a new function.

src/packs.c
phmccarty: This is a valid concern; printing the progress message is not helpful when `link()` fails, so printing it after the error check below is better.

src/update.c
```diff
@@ -79,7 +82,23 @@ static struct list *full_download_loop(struct list *updates, int isfailed)
...
-
+		/* Mix content is local, so don't queue files up for curl downloads */
+		if (file->is_mix) {
+			string_or_die(&url, "%s/%i/files/%s.tar", MIX_STATE_DIR, file->last_change, file->hash);
+			string_or_die(&filename, "%s/download/.%s.tar", state_dir, file->hash);
+			file->staging = filename;
+			ret = link(url, filename);
+			/* Try doing a regular rename if hardlink fails */
+			if (ret) {
+				if (rename(url, filename) != 0) {
```
phmccarty: If the rename fails, should there a copy fallback here too?

tmarcu: So now I understand the copy fallback would be the tar dance we do, and I don't think that fallback is needed here because of how it's set up. The rename fallback is enough in this case, but if we change up the locations for the local mix content maybe we can revisit this.

phmccarty: Okay, sounds fine to me.

add-pkg.sh
tmarcu: Yes I need to add -f :)

src/update.c
tmarcu: You can technically have an empty mix with no bundles added, and you just want to have moved your system to the mix state, or you simply didn't add anything yet and are just incrementing the version/upstream content only.

src/manifest.c
tmarcu: See comment (way) below, I don't think this feature needs a copy fallback in the current state that it's in.

src/manifest.c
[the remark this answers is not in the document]
tmarcu: Done.

src/packs.c
tmarcu: Done.

src/update.c
matthewrsj: I see, thanks for explaining.

docs/swupd.1.rst
matthewrsj: Looks like the documentation still needs to be updated though.

src/verify.c
matthewrsj: Agree.

docs/swupd-add-pkg.4.rst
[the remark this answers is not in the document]
matthewrsj: the `--allow-mix-collisions` should still be documented.

docs/swupd.1.rst
phmccarty: /me looks at the source line we are commenting on... Yes, this .rst file should be updated.

add-pkg.sh
```diff
@@ -0,0 +1,139 @@
...
+BUILD_STATE_DIR="/var/lib/mix"
+MIX_CERT="$MIX_DIR/Swupd_Root.pem"
+PRIVKEY="$MIX_DIR/private.pem"
+FORMAT="$(awk -F '=' '/FORMAT/ {print $2}' $BUILDCONF  | tr -d ' ')"
+
+pkg=$1
+bundle=$2
+
+type "mixer"
+if [ $? -ne 0]; then
```
phmccarty: You need a space before the `]`, or else the shell will report a syntax error.

add-pkg.sh
```diff
@@ -0,0 +1,139 @@
...
+	echo -e "Please install the mixer bundle before proceeding:\n  swupd bundle-add mixer"
```
phmccarty: Should the script `exit 1` here?

add-pkg.sh
```diff
@@ -0,0 +1,139 @@
...
+	echo "Adding $pkg to $bundle..."
+
+	# Write package to bundle definition
+	echo "$pkg" | sudo tee -a "$BUNDLE_DIR/$bundle" > /dev/null
+fi
+
+# Copy os-core from upstream
+sudo -E curl -f -O "https://raw.githubusercontent.com/clearlinux/clr-bundles/$clearver/bundles/os-core"
+if [ $? -eq 22 ]; then
+	sudo rm -rf "os-core"
```
phmccarty: Should the script `exit 1` here too? Similar comment applies to all the new error handling for the curl 22 exit status.

add-pkg.sh
tmarcu: 🤕

icarus-sparry: And whilst you are changing this, can you change it to

if type mixer >/dev/null ; then

to avoid a random "/usr/bin/mixer" message being output.

Makefile.am
```diff
@@ -286,7 +286,8 @@ MANPAGES = \
 	docs/check-update.timer.4 \
 	docs/swupd-update.service.4 \
 	docs/swupd-update.timer.4 \
-	docs/update-triggers.target.4
+	docs/update-triggers.target.4 \
+	docs/swupd-add-pkg.4
```
icarus-sparry: If you had this in alphabetical order it would be even less of a change (one line added vs one line removed and 2 added)

matthewrsj:
+1 from me except for you still need to add a blurb about `--allow-mix-collisions` to the man page.

tmarcu:
Added

matthewrsj:
+1

tmarcu:
:checkered_flag:

PR #225, opened 2017-06-26
Differentiate between various early network failures.

ahkok:
We draw a hard line with error codes, such that:

Error code 16 is reserved for failures resulting from the basic
network check. Any call to check_network() that fails will
return this error code. No other error path returns this
error code.

Additionally, check_network() is called in every normal code path.

All other, possibly network related issues, return a different
error code. If the basic network check succeeds, but e.g. pack
downloads fail, we return a new (23) error code so that we
can better establish the conditions through telemetry and
determine whether the error is on the client or the server,
which is highly likely with this new error code 23.

Subsequently, for automated updates, we can choose to ignore
update failures if the automated periodical updates fail
due to no network being available. Calling the program
from a shell will still produce an error code, as normal.

### DIFF
--- a/src/update.c
+++ b/src/update.c
@@ -415,8 +415,8 @@ download_packs:
 			printf("Retry #%d downloading packs\n", retries);
 			goto download_packs;
 		}
-		printf("No network, or server unavailable for pack downloads\n");
-		ret = ENOSWUPDSERVER;
+		printf("Pack download failed\n");
+		ret = EDOWNLOADPACKS;
 		goto clean_exit;
 	}
 	grabtime_stop(&times);
--- a/test/functional/checkupdate/no-server-content/lines-checked
+++ b/test/functional/checkupdate/no-server-content/lines-checked
@@ -1,2 +1,3 @@
 Attempting to download version string to memory
-Cannot reach update server
+Current OS version: 10
+There are no updates available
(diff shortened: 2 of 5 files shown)


ahkok:
s/18/16/

tmarcu:
@ahkok  Can you check why test/functional/checkupdate/no-server-content/test.bats is failing? Does make check run on your system?

ahkok:
Removed the service unit change, for now, from this PR.

ahkok:
I don't understand why make check fails (it does for me, locally, as well).

The check that fails is `run sudo sh -c "$SWUPD check-update $SWUPD_OPTS_NO_CERT"`

ahkok:
Updated make check with the new expected output. That fIxes that.

src/update.c
```diff
@@ -415,8 +415,8 @@ int main_update()
 			printf("Retry #%d downloading packs\n", retries);
 			goto download_packs;
 		}
-		printf("No network, or server unavailable for pack downloads\n");
-		ret = ENOSWUPDSERVER;
+		printf("Pack download failed");
```
phmccarty: missing `\n` for this new error message

test/functional/checkupdate/no-server-content/lines-checked
```diff
@@ -1,2 +1,6 @@
+swupd-client software update checker 3.9.4
+   Copyright (C) 2012-2016 Intel Corporation
+
```
phmccarty: You should remove the above three lines, because they match the ignored patterns (see test/functional/ignore-list for the pattern list). In particular, the first line would need modification for every new release of swupd-client so would create a maintenance problem.

ahkok:
Thanks @phmccarty, I fixed up those 2 comments.

phmccarty:
+1

PR #227, opened 2017-07-19
Rerun swupd until there are no new version available

albertomurillo:
swupd needs to be run multiple times when there are format bumps.

This PR reruns swupd after a successful update until there are no
new updates available on the server.

Signed-off-by: Alberto Murillo Silva <alberto.murillo.silva@intel.com>

### DIFF
--- a/src/update.c
+++ b/src/update.c
@@ -237,6 +237,7 @@ int main_update()
 	struct timespec ts_start, ts_stop; // For main swupd update time
 	timelist times;
 	double delta;
+	bool updated = false;
 
 	srand(time(NULL));
 
@@ -485,6 +486,7 @@ clean_curl:
 	swupd_deinit(lock_fd, &latest_subs);
 
 	if ((current_version < server_version) && (ret == 0)) {
+		updated = true;
 		printf("Update successful. System updated from version %d to version %d\n",
 		       current_version, server_version);
 	} else if (ret == 0) {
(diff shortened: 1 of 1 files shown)


tmarcu:
-1
@albertomurillo Thanks for the PR, but this won't work for various reasons.
Specifically, we HAVE to re-exec swupd instead of just recursing or using the same process. If we go through a format bump, swupd could have undergone a breaking change and the current running process may not be able to read manifests part of the new format. The certificate could have also changed in which case the running swupd will not be able to update and verify the new MoM.
I think this approach could work if instead of re-cursing, we re-exec'd at the end point so the newly installed swupd will be used,and it will be able to handle all of the above problems. Otherwise we would be running swupd from a version older than the one we updated to.
We can also *only* re-exec if the update was successful, otherwise we cannot guarantee the new swupd was installed correctly.

phmccarty:
I agree with @tmarcu that this can only work if swupd is re-exec'd, because (a) swupd may have updated and we want to run the newer version, and/or (b) swupd introduced incompatible changes after the update as part of a format bump. These incompatible changes can be any of the items Tudor mentioned.

tmarcu:
Proper implementation found in https://github.com/clearlinux/swupd-client/pull/228

albertomurillo:
@tmarcu awesome!

PR #228, opened 2017-07-26
Re-execute swupd update for format bumps

matthewrsj:
When a device is out of date by more than a format bump re-exec swupd
update until the device version matches the server version. This is
achieved by reading the "actions" field in the Manifest.MoM, which is
yet to be implemented in swupd-server. When "update" is specified in the
actions of the current manifest and the current update is successful,
swupd will re-exec itself with the same flags it was originally called
with.

A functional test has also been added.

Notes:
* The actions field still needs to be implemented in swupd-server
* This PR does ~~not~~ check for consistency between /usr/lib/os-release and the version file in swupd's statedir. ~~so errors can occur if a build is released that does not update that file (unlikely). That can be changed in this PR or the future~~


matthewrsj:
Travis can't copy the files I want copied. I'll look at this tomorrow.

matthewrsj:
It was some `.gitignore` strangeness brought on by excluding `swupd` but needing `/usr/share/defaults/swupd/format`. Unfortunately I couldn't negate that rule with `!` as explained in this [stackoverflow answer](https://stackoverflow.com/a/5534865). I just copy it into place at test time to get around this.

matthewrsj:
Of course the test won't run if I don't add it to the Makefile. Fixed.

src/globals.c
```diff
@@ -543,3 +545,10 @@ void free_globals(void)
 		bundle_to_add = NULL;
 	}
 }
+
+void save_cmd(int argc, char **argv)
+{
+	for (int i = 0; i < argc; i++) {
+		string_or_die(&swupd_cmd, "%s %s", swupd_cmd, argv[i]);
```
tmarcu: This will just overwrite previous flags from argv as it loops if there is more than 1 flag, and leak all the previous ones since it sets the pointer *swupd_cmd to a new one with string_or_die(). So when this loop finishes, you only have saved argv[n] where n is the last argument passed, and none of the others. Need to append/concat to the &swupd_cmd as you're iterating through argv insead. You can also create a swupd_cmds list of size argc-1 and do something like:
`string_or_die(&swupd_cmd[i]...` to save each flag into the list.

src/scripts.c
```diff
@@ -58,10 +60,15 @@ static void update_triggers(void)
 		ret = system("/usr/bin/systemctl daemon-reload");
 	}
 
-	ret = system("/usr/bin/systemctl --no-block restart update-triggers.target");
+	if (!block) {
+		string_or_die(&block_flag, "--no-block");
+	}
+
+	string_or_die(&restart_cmd, "/usr/bin/systemctl restart %s update-triggers.target", block_flag);
```
tmarcu: --no-block must come _before_ the command, aka
`+	string_or_die(&restart_cmd, "/usr/bin/systemctl %s restart update-triggers.target", block_flag);`

matthewrsj:
Last commit treats post update actions as a list so we aren't restricted to only one possible post update action in the future.

src/globals.c
```diff
@@ -206,14 +210,39 @@ static int set_default_value_from_path(char **global, const char *path)
...
-	string_or_die(global, "%s", line);
+	string_or_die(contents, line);
 	ret = 0;
 fail:
 	fclose(file);
 	free(rel_path);
 	return ret;
 }
 
+int get_int_from_path(const char *abs_path) {
```
tmarcu: What is the int we are getting with this function? Could be renamed to be a bit more clear.

src/helpers.c
```diff
@@ -898,3 +898,26 @@ struct list *files_from_bundles(struct list *bundles)
...
+	 * Compare version numbers in state_dir/version and /usr/lib/os-release.
+	 * The version in /usr/lib/os-release must match or be greater than
+	 * state_dir/version for successful updates (greater than occurs during
+	 * format bumps).
+	 */
+	int os_release_v = -1;
+	int state_v = -1;
+	char *state_v_path;
+
+	string_or_die(&state_v_path, "%s/version", state_dir);
```
tmarcu: Need to free state_v_path?

src/globals.c
```diff
@@ -173,15 +173,19 @@ void print_time_stats(timelist *head)
...
+int get_value_from_path(char **contents, const char *path, bool is_abs_path)
 {
 	char line[LINE_MAX];
 	FILE *file;
 	char *c;
 	char *rel_path;
 	int ret = -1;
 
-	string_or_die(&rel_path, "%s%s", path_prefix, path);
+	if (is_abs_path) {
```
tmarcu: Can you write a comment on why abs_path is needed here now?

src/globals.c
matthewrsj: I see what you mean by leaking the previous swupd_cmds and I'll fix that - but I'm not seeing it only save the last arg. When I test this it saves the whole command the way I expect (with an extra space in front from the first iteration where swupd_cmd is empty).

When I print out suwpd_cmd after saving it I see this:

```
  /path/to/lt-swupd update -S /path/to/state -p /path/to/swupd-client/test/functional/update/re-update-required/target-dir -u file:///path/to/web-dir -C /path/to/Swupd_Root.pem -I
```

tmarcu: Forgive my selective reading, you're right. The leaks do need to be fixed still, but it indeed saves the whole string. I just ignored the swupd_cmd being passed to the first formatter %s and assumed you were treating argv[1] as being --path /blah/blah. Thanks for clearing it up :)

src/manifest.c
```diff
@@ -235,7 +235,10 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
 		}
 		if (latest && strncmp(component, "MoM", 3) == 0) {
 			if (strncmp(line, "actions:", 8) == 0) {
-				post_update_action = strdup(c);
+				post_update_actions = list_prepend_data(post_update_actions, strdup(c));
+				if (!post_update_actions->data) {
+					abort();
```
tmarcu: If this was part of the main update code path I'd agree with abort, but I think it would be better in this case to just print a loud warning saying we couldn't parse/read the action, and that an update or verify should be run, instead of aborting. We want the update to always proceed since it's critical, but executing the action is arguably not critical imho - considering the current state of things :)

matthewrsj:
Comments addressed and squashed.

PR #229, opened 2017-08-03
Add basic progress status for long-running tasks

matthewrsj:
Add a basic progress indication for a couple potentially long-running
swupd tasks (download_pack and add_missing_files). The progress
indication is displayed as a percentage:
```
...85%
```
It overwrites its own line so the progress will update in place. It may
be interrupted by other swupd output:
```
...45%
Extracting os-core pack for version 10000
...50%
Extracting os-core-update pack for version 9000
...55%
```
The progress indication will then resume updating in place, on the same
line.

This progress is only displayed when outputting to a TTY, and will not
output anything when the output is being redirected. This is nice for
testing and logging, but means that these progress updates will not be
caught by processes that capture swupd output, such as ister.

## RFC
I'd like to hear thoughts on this approach. A possible work-around for hiding the progress updates from calling processes is to log the progress to a file when not in a tty.

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -40,6 +40,8 @@
 #include "signature.h"
 #include "swupd.h"
 
+static int last_percentage = -1;
+
 void check_root(void)
 {
 	if (getuid() != 0) {
@@ -943,8 +945,13 @@ bool string_in_list(char *string_to_check, struct list *list_to_check)
 void print_progress(unsigned int count, unsigned int max)
 {
 	if (isatty(fileno(stdout))) {
-		printf("\r\t...%d%%", (int)(100 * ((float)count / (float)max)));
-		fflush(stdout);
+		/* Only print when the percentage changes, so a maximum of 100 times per run */
+		int percentage = (int)(100 * ((float)count / (float)max));
+		if (percentage != last_percentage) {
+			printf("\r\t...%d%%", percentage);
+			fflush(stdout);
+			last_percentage = percentage;
+		}
 	} else {
 		// Print the first one, then every 10 after that
 		if (count % 10 == 1) {
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -945,5 +945,10 @@ void print_progress(unsigned int count, unsigned int max)
 	if (isatty(fileno(stdout))) {
 		printf("\r\t...%d%%", (int)(100 * ((float)count / (float)max)));
 		fflush(stdout);
+	} else {
+		// Print the first one, then every 10 after that
+		if (count % 10 == 1) {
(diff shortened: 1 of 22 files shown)


matthewrsj:
I initially had a more complex solution that displayed an actual progress bar, but thought the percentage was simpler and gave more information.

tmarcu:
An even easier solution would be to print something along the lines of: 

**.......................... <continually append dots to this line every ~N seconds>**
or append whatever character you choose for the progress bar. The chroot builder and server do this because it just sits there doing tons of file IO, but at least you see that some output is getting updated and it is not frozen. 
¯\_(ツ)_/¯

test/functional/ignore-list
```diff
@@ -13,3 +13,4 @@
 ^sh: .*/usr/bin/clr-boot-manager: No such file or directory$
 ^sh: .*/usr/bin/clr-boot-manager: not found$
 ^Possible filedescriptor leak: .*$
+^\.*$
```
phmccarty: Oh sorry, I didn't notice the anchors... It's still not specific enough, since it will also match a blank line, and there is already an ignore entry for that. Maybe `^\.+$` instead?

matthewrsj: right it ignores lines containing only dots or no dots. I'll make it explicit.

matthewrsj:
@tmarcu @phmccarty thoughts? This now prints a dot `.` for each 10 files when output is redirected so that control characters are not printed to log files, tests, etc.

src/helpers.c
```diff
@@ -939,3 +939,16 @@ bool string_in_list(char *string_to_check, struct list *list_to_check)
 
 	return false;
 }
+
+void print_progress(unsigned int count, unsigned int max)
+{
+	if (isatty(fileno(stdout))) {
+		printf("\r\t...%d%%", (int)(100 * ((float)count / (float)max)));
```
phmccarty: Even though this `printf` is overwriting the line, due to the carriage return, it seems wasteful. For example, if there are 100,000 files to full_download, 100,000 prints are issued (followed by a flush).

Can this be made smarter to only print when the percentage value changes?

src/update.c
```diff
@@ -173,6 +177,7 @@ static int update_loop(struct list *updates, struct manifest *server_manifest)
 	sync();
 
 	/* rename to apply update */
+	printf("Applying update\n");
```
phmccarty: Can you add this in a separate commit (and associated functional test updates)?

matthewrsj:
Comments addressed

phmccarty:
+1

PR #230, opened 2017-08-07
Print bundle dependencies when failing to remove

matthewrsj:
A bundle will fail to be removed from the filesystem when other bundles
require that bundle as a dependency. Print a list of each of these
bundles along with the error message.

Also  add --has-dep=BUNDLE argument will display a tree representing all
installed bundles that recursively include BUNDLE.

One may pass the --all argument as well to list the dependency tree for
BUNDLE including all installable bundles available on the server.

**Example output without --all:**
```
Installed bundles that have os-installer as a dependency:
format:
 # * is-required-by
 #   |-- is-required-by
 # * is-also-required-by
 # ...

  * mixer
    |-- os-clr-on-clr
```
**Example output with --all:**
```
All installable and installed bundles that have os-installer as a dependency:
format:
 # * is-required-by
 #   |-- is-required-by
 # * is-also-required-by
 # ...

  * clr-devops
  * mixer
    |-- os-clr-on-clr
        |-- os-clr-on-clr-dev
    |-- clr-devops
    |-- os-clr-on-clr-dev
```


matthewrsj:
Rebased to latest master

matthewrsj:
Two latest commits add the `--deps BUNDLE` subcommand to bundle-list. It allows users to list the nested included bundles for BUNDLE and can give them an idea of what they are about to bundle-add.

Example:
```

**Comment by matthewrsj** [comment]
Fixes #183

File: src/bundle.c [hunk 1]
```diff
@@ -171,30 +175,242 @@ static int unload_tracked_bundle(const char *bundle_name, struct list **subs)
...
+	// add_subscriptions takes a list, so construct one with only bundle_name
+	struct list *bundles = NULL;
+	bundles = list_prepend_data(bundles, bundle_name);
+	ret = add_subscriptions(bundles, &subs, current_version, mom, true, 0);
+	list_free_list(bundles);
+	if (ret != add_sub_NEW) {
+		// something went wrong or there were no includes, print a message and exit
+		const char *m;
+		if (ret & add_sub_ERR) {
+			m = "Processing error";
```
phmccarty: ... it is assigned to. I don't recall the C language semantics for this one, but I would think assigning to a `const char *` in a later statement would be forbidden. Testing this with an int type (`const int a; a = 1;`) results in a compiler error.

If someone can chime in as to why the const is apparently ignored in this case, please do so.

Regardless, it's more common in the swupd code base to dynamically allocate strings with `string_or_die()`, so I would take that route.

test/functional/bundlelist/all/test.bats
```diff
@@ -15,9 +15,6 @@ teardown() {
 
 @test "bundle-list all bundles" {
   run sudo sh -c "$SWUPD bundle-list $SWUPD_OPTS --all"
-
-  [ "$status" -eq 0 ]
```
phmccarty: keep the exit status check?

matthewrsj: oops..

src/bundle.c
matthewrsj: I took this logic from elsewhere in bundle.c, emphasizing the point that the bundle code needs rework. I'll fix this up.

src/bundle.c
```diff
@@ -171,30 +175,242 @@ static int unload_tracked_bundle(const char *bundle_name, struct list **subs)
...
+				}
+
+				*reqd_by = list_append_data(*reqd_by, bundle_str);
+				required_by(reqd_by, bundle->component, mom, rl);
 			}
 		}
 	}
+}
+/* Return recursive list of included bundles */
+int show_included_bundles(char *bundle_name)
```
matthewrsj: For now, I can move the swupd_init/deinit out of the two options introduced here.

matthewrsj:
Comments addressed.

src/manifest.c
```diff
@@ -934,7 +934,7 @@ void link_submanifests(struct manifest *m1, struct manifest *m2, struct list *su
 
 /* if component is specified explicitly, pull in submanifest only for that
  * if component is not specified, pull in any tracked component submanifest */
-struct list *recurse_manifest(struct manifest *manifest, struct list *subs, const char *component)
+struct list *recurse_manifest(struct manifest *manifest, struct list *subs, const char *component, bool server)
```
phmccarty: I'm not very happy with the current state of this function. For example, if `component` is non-NULL, it will call `load_manifest` once, and that's it. Simpler would be to wrap `load_manifest` differently so that the API is less confusing. At which point, we could remove the `component` parameter entirely, and the code would become much more clear (IMO).

That said, leave it as-is for now, but I would strongly consider refactoring this function (and possibly renaming it) for conciseness and clarity.

test/functional/bundlelist/has-dep-nested-not-installed/test.bats
```diff
@@ -0,0 +1,27 @@
...
+  create_manifest_tar 10 test-bundle
+  create_manifest_tar 10 test-bundle2
+  create_manifest_tar 10 test-bundle3
+}
+
+teardown() {
+  clean_tars 10
+}
+
+@test "bundle-list list bundle has-deps with nested dependent bundles" {
```
phmccarty: test description needs a refresh

test/functional/bundlelist/has-dep-nested-server/test.bats
```diff
@@ -0,0 +1,26 @@
...
+  create_manifest_tar 10 test-bundle
+  create_manifest_tar 10 test-bundle2
+  create_manifest_tar 10 test-bundle3
+}
+
+teardown() {
+  clean_tars 10
+}
+
+@test "bundle-list list bundle has-deps with nested dependent bundles" {
```
phmccarty: test description needs a refresh

test/functional/bundlelist/none-has-deps/test.bats
```diff
@@ -0,0 +1,25 @@
...
+  create_manifest_tar 10 os-core
+  create_manifest_tar 10 test-bundle
+  create_manifest_tar 10 test-bundle2
+}
+
+teardown() {
+  clean_tars 10
+}
+
+@test "bundle-list list bundle has-deps with nested dependent bundles" {
```
phmccarty: test description needs a refresh

src/bundle.c
```diff
@@ -171,30 +175,236 @@ static int unload_tracked_bundle(const char *bundle_name, struct list **subs)
...
+	struct list *reqd_by = NULL;
+
+	if (!server && !is_tracked_bundle(bundle_name)) {
+		fprintf(stderr, "Error: Bundle \"%s\" does not seem to be installed\n", bundle_name);
+		fprintf(stderr, "       try passing --all to check uninstalled bundles\n");
+		ret = EBUNDLE_NOT_TRACKED;
+		goto out;
+	}
+
+	version = server ? get_latest_version() : get_current_version(path_prefix);
```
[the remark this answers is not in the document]
matthewrsj: My thought was that when `--all` is specified, we should show the dependency tree for the latest available version since they are asking for server dependencies. The use case for this is running bundle tests against the all dependent bundles in the latest version.

You are right though, for the more general case it makes more sense to report dependencies for the user's current version. The use case above is not that important since the testing will happen on the latest/staged version anyways.

src/bundle.c
```diff
@@ -171,30 +175,236 @@ static int unload_tracked_bundle(const char *bundle_name, struct list **subs)
...
+	if (version < 0) {
+		fprintf(stderr, "Error: Unable to determine current OS version\n");
```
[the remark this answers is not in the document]
matthewrsj: There doesn't seem to be any issue if version should be the current version since get_current_version returns -1 it can't determine the current version.

src/manifest.c
matthewrsj: I agree, this should be fixed up in the future.

matthewrsj:
Comments addressed

src/bundle.c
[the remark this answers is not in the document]
phmccarty: Thanks for clarifying. Yes, I think the bundle commands should only care about the current version, at least for now. In newer builds, there may be several new bundles available, and the user won't know about those bundles until they have updated anyway.

src/bundle.c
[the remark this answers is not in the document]
phmccarty: I was referring to the case when `get_latest_version()` is called, but you have amended this line. Ignore the comment :)

src/manifest.c
phmccarty: Okay, sounds good.

src/clr_bundle_ls.c
```diff
@@ -133,7 +138,20 @@ static bool parse_options(int argc, char **argv)
 			}
 			set_cert_path(optarg);
 			break;
-
+		case 'D':
+			if (!optarg) {
+				fprintf(stderr, "Invalid --has-dep argument\n\n");
+				goto err;
+			}
+			cmdline_option_has_dep = optarg;
```
phmccarty: Use `string_or_die()` here.

src/clr_bundle_ls.c
```diff
@@ -133,7 +138,20 @@ static bool parse_options(int argc, char **argv)
...
+			break;
+		case 'd':
+			if (!optarg) {
+				fprintf(stderr, "Invalid --deps argument\n\n");
+				goto err;
+			}
+			cmdline_option_deps = optarg;
```
phmccarty: Also use `string_or_die()` here.

src/clr_bundle_ls.c
phmccarty: I am suggesting this primarily for consistency; the rest of the code base only assigns dynamically allocated strings to the globals.

matthewrsj: I agree! Fixed.

src/clr_bundle_ls.c
```diff
@@ -133,7 +138,20 @@ static bool parse_options(int argc, char **argv)
 			}
 			set_cert_path(optarg);
 			break;
-
+		case 'D':
+			if (!optarg) {
+				fprintf(stderr, "Invalid --has-dep argument\n\n");
+				goto err;
+			}
+			string_or_die(&cmdline_option_has_dep, optarg);
```
phmccarty: For safety, you should do `string_or_die(&cmdline_option_has_dep, "%s", optarg)`, since `optarg` is a variable set from user input.

src/clr_bundle_ls.c
```diff
@@ -133,7 +138,20 @@ static bool parse_options(int argc, char **argv)
...
+			break;
+		case 'd':
+			if (!optarg) {
+				fprintf(stderr, "Invalid --deps argument\n\n");
+				goto err;
+			}
+			string_or_die(&cmdline_option_deps, optarg);
```
phmccarty: Similar comment for this call as above.

src/clr_bundle_ls.c
matthewrsj: Fixed.

src/bundle.c
```diff
@@ -172,28 +172,37 @@ static int unload_tracked_bundle(const char *bundle_name, struct list **subs)
 }
 
 /* Return list of bundles that include bundle_name */
-struct list *required_by(const char *bundle_name, struct manifest *mom)
+void required_by(struct list **reqd_by, const char *bundle_name, struct manifest *mom, int rl)
```
tmarcu: I'd personally prefer a different, more descriptive name than r1, even though there is a comment below. Up to you on that :)

tmarcu:
+1 from me, you could smash the deps commits into one if you'd like as well.

src/bundle.c
matthewrsj: yeah yeah, done. ;)

PR #232, opened 2017-08-08
Wait for test server to become available before testing

matthewrsj:
Check that the simple test server in the checkupdate/slow-server/ test
is available before actually running the swupd command. This check is
done by testing the return status of a curl command on the server up to
ten times until successful.

Since this is an historically touchy test, output an additional
debug.log in the test directory with much more verbose logging (using
set -x).


phmccarty:
+1

phmccarty:
Ah, one thing I noticed. It would also be nice to add `debug.log` to the .gitignore.

matthewrsj:
You are right @phmccarty, added to gitignore

phmccarty:
+1

PR #233, opened 2017-08-09
Add status checks for test swupd commands

matthewrsj:
Adds status checks for each swupd command called in the functional
tests. At this point several of these tests fail due to some successful
commands returning error statuses.

First commit should have test failures. I will push the fixes for those failures shortly.

Four following commits fix one previously-failing test each.

### DIFF
--- a/src/clr_bundle_ls.c
+++ b/src/clr_bundle_ls.c
@@ -75,7 +75,10 @@ static bool parse_options(int argc, char **argv)
 			print_help(argv[0]);
 			exit(EXIT_SUCCESS);
 		case 'a':
-			return list_installable_bundles();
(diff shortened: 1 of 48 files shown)


.gitignore
```diff
@@ -45,6 +45,8 @@ swupd_*
 swupd.bash
 tap-driver.sh
 test-driver
+test-bundle*
+test-file*
```
phmccarty: Some of the tests use files with these names, and they are committed to the repo. So, you have the option of making these patterns more specific to certain paths, or you can add a .gitignore instead the test directories that need it.

matthewrsj: You are correct. I moved the gitignore bits to the specific test.

phmccarty:
+1 from me

PR #235, opened 2017-08-17
Ensure state_dir is a directory and owned by root

icarus-sparry:
As the state_dir (normally /var/lib/swupd) contains things like the
downloaded bundles, it is important that there is no access to
non-root users who could potentially explot races to replace files
between them being checked and them being installed.

Ensure that the state_dir is root owned, and mode 0700, as well as the
critical directories below it.

Note that a proper fix would involve using chdir to move into the
directory and only use relative paths. This would prevent faulty
permissions higher up the directory tree being exploited. It *might*
be possible to use openat(2) to prevent this exploitation.

Add a test to ensure that the enforcement is taking place.


src/helpers.c
```diff
@@ -130,60 +130,66 @@ void unlink_all_staged_content(struct file *file)
...
+	}
+	if ((sb.st_uid == 0) &&
+	    (S_ISDIR(sb.st_mode)) &&
+	    ((sb.st_mode & 0777) == 0700)) {
+		return false;
+	}
+	/* Oops, not owned by root or
+	 * not a directory or wrong perms
+	 */
+	ret=swupd_rm(state_dir);
```
matthewrsj: I know users shouldn't be putting things in /var/lib/swupd, but should we at least print a warning that we are removing content before doing so? Or no since that should be swupd-managed anyways?

matthewrsj:
If we decide to use this elsewhere for other directories (other than state dir) in the future, we can add a log print at that time.

+1

PR #239, opened 2017-08-31
F/createdirs

icarus-sparry:
Added a better test than just seeing that swupd_rm didn't return an error.


matthewrsj:
The change looks good, but you have the original commit in here as well.

PR #240, opened 2017-08-31
Test a statedir is removed directly

icarus-sparry:
Don't use the return value from swupd_rm, which is poorly defined, but
test directly that the file no longer exists.

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -152,12 +152,15 @@ static int ensure_root_owned_dir(const char *dirname)
 	/* Oops, not owned by root or
 	 * not a directory or wrong perms
 	 */
-	ret = swupd_rm(state_dir);
-	if (ret) {
+	swupd_rm(dirname);
+	errno = 0;
+	ret = stat(dirname, &sb);
+	if ((ret != -1) || (errno != ENOENT)) {
 		fprintf(stderr,
-			"Error \"%s\" not owned by root"
-			"but couldn't be deleted\n",
-			state_dir);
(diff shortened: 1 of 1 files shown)


src/helpers.c
```diff
@@ -152,12 +152,15 @@ static int ensure_root_owned_dir(const char *dirname)
 	/* Oops, not owned by root or
 	 * not a directory or wrong perms
 	 */
-	ret = swupd_rm(state_dir);
-	if (ret) {
+	swupd_rm(state_dir);
```
matthewrsj: Something I didn't notice before: Shouldn't this be `dirname` instead of assuming the function is being called on `state_dir`? Otherwise why pass anything in at all?

src/helpers.c
```diff
@@ -152,12 +152,15 @@ static int ensure_root_owned_dir(const char *dirname)
...
+	ret = stat(dirname, &sb);
+	if ((ret != -1) || (errno != ENOENT)) {
 		fprintf(stderr,
-			"Error \"%s\" not owned by root"
-			"but couldn't be deleted\n",
-			state_dir);
+			"Error \"%s\" not owned by root, is not a directory, "
+			"or has the wrong permissions.\n"
+			"However it couldn't be deleted. stat gives error '%s'\n",
+			state_dir, strerror(errno));
```
matthewrsj: Again, you check `dirname`, yet report the error for `state_dir`

matthewrsj:
+1

Issue #238, opened 2017-08-22
swupd is unable to downgrade from version


dborquez:
In order to downgrade CLR from version: 17230 to version 16860 I executed:
* swupd verify -f -m 16860

* As result swupd reported this error:

```
...
Hash mismatch for file: /usr/share/vulkan/icd.d/intel_icd.x86_64.json
        fixed
Inspected 232925 files
  8231 files were missing
    8231 of 8231 missing files were replaced
    0 of 8231 missing files were not replaced
  7915 files did not match
    7915 of 7915 files were fixed
    0 of 7915 files were not fixed
  0 files found which should be deleted
Calling post-update helper scripts.
sh: relocation error: /usr/lib64/haswell/libc.so.6: symbol __tunable_get_val, version GLIBC_PRIVATE not defined in file ld-linux-x86-64.so.2 with link time reference
sh: relocation error: /usr/lib64/haswell/libc.so.6: symbol __tunable_get_val, version GLIBC_PRIVATE not defined in file ld-linux-x86-64.so.2 with link time reference
sh: relocation error: /usr/lib64/haswell/libc.so.6: symbol __tunable_get_val, version GLIBC_PRIVATE not defined in file ld-linux-x86-64.so.2 with link time reference
Fix successful
```

* After that Clear Linux was broken:

```
# cat /usr/lib/os-release
cat: relocation error: /usr/lib64/haswell/libc.so.6: symbol __tunable_get_val, version GLIBC_PRIVATE not defined in file ld-linux-x86-64.so.2 with link time reference
```

phmccarty:
Please post the list of bundles you have installed.

dborquez:
It just have this set of bundles:
- openssh-server
- clr-devops

matthewrsj:
ok, I was able to reproduce this using the 17230 live image from https://download.clearlinux.org/releases/17230/ 

I created a gist of the full output here: https://gist.github.com/matthewrsj/b25f15da14768b97c63d4706ebaee3e0

I will be checking if this error exists when going the other direction, 16860 -> 17230.

@dborquez to fix your current system, take the following steps:
- Download the [clear linux installer image](https://download.clearlinux.org/current/clear-17260-installer.img.xz) to a usb drive.
- Boot the installer image on your borked system
  * In "Step 1 of 6 (Network Requirements)" verify that a connection the the update server has been `established`.
[...]

PR #242, opened 2017-09-01
Require --force or --picky when verifying to a different version

matthewrsj:
When running verify --fix with -m specifying a different version than
the current OS version, require --picky to enforce management of files
under /usr or require --force as an override. Adding --picky can solve
issues such as issue #238 where fixing backward caused glibc library
confusion.

Add functional tests as well.

Addresses #238


src/helpers.c
```diff
@@ -995,3 +995,13 @@ bool is_compatible_format(int format_num)
...
 }
+
+/* Check if the version argument is the same version as the current OS */
+bool is_current_version(int version)
+{
+	if (version < 0) {
+		return false;
+	}
+
+	return (version == get_current_version(path_prefix));
```
phmccarty: Can `get_current_version()` return an error at this point in the code? If so, I would handle its error before deciding how to return from this function.

src/verify.c
```diff
@@ -719,6 +719,22 @@ int verify_main(int argc, char **argv)
 		}
 	}
 
+	/* If --fix was specified but --force and --picky are not present, check
+	 * that the user is only fixing to their current version. If they are
+	 * fixing to a different version, print an error message that they need to
+	 * specify --force or --picky. */
+	if (cmdline_option_fix && !is_current_version(version)) {
```
phmccarty: Do you want to add special handling for `verify --install` as well?

src/helpers.c
matthewrsj: well at this point get_current_version can return a -1, but version cannot be -1. So if get_current_version encounters an error (version == get_current_version(...)) will evaluate to false.

src/verify.c
matthewrsj: yes.

matthewrsj: Actually, per our meatspace convo, the --install option should be fixed up outside this PR, and should not require --force or --picky to be run, and possible made into its own sub-command.

src/helpers.c
phmccarty: Okay, sounds good. I was lacking context of the surrounding code, so I think you can keep it as is.

src/verify.c
phmccarty: Agreed. Let's not special-case --install in this case for now.

matthewrsj:
@phmccarty any other comments?

phmccarty:
Nope, +1 from me.

PR #244, opened 2017-09-07
Add newline to informational print

matthewrsj:
Add newline so progress updates are printed correctly. Instead of this:
```
...10%File /file was not in a pack
...11%
```
Print a newline before the message:
```
...10%
File /file was not in a pack
...11%
```


phmccarty:
+1

PR #246, opened 2017-09-11
Add --no-boot-update flag for container usage

matthewrsj:
Since clr-boot-manager does not work in a container add the
--no-boot-update option to allow installation into a container to skip
the boot partition management.

Also add functional tests.


matthewrsj:
Rebased to the latest master to insure bundleadd tests still pass with latest bundleadd changes.

test/functional/bundleadd/boot-skip/lines-checked
```diff
@@ -0,0 +1,4 @@
+Attempting to download version string to memory
+Extracting test-bundle pack for version 10
+Installing bundle(s) files...
+Bundle(s) installation done.
```
phmccarty: I think the reason why the new warning message is not printed here is because `apply_heuristics()` is not called in the `bundle-add` code path. Also note that that function is not called for `verify --fix` or `verify --install` either, but the `need_update_boot` variable is set unconditionally for their respective code paths, resulting in the warning being printed.

Perhaps `apply_heuristics()` should be run for `bundle-add`?

matthewrsj:
`apply_heuristics()` now called for bundle-add.

phmccarty:
+1

PR #247, opened 2017-09-12
Do hash checks for bundle add

tmarcu:
Hash check every file before adding it to the file system.


src/bundle.c
```diff
@@ -721,6 +721,22 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+	iter = list_head(to_install_files);
+	while (iter) {
+		file = iter->data;
+		iter = iter->next;
+
+		string_or_die(&fullname, "%s/staged/%s", state_dir, file->hash);
+
+		ret = verify_file(file, fullname);
+		if (!ret) {
+			free(fullname);
```
phmccarty: you'll want to free() after the fprintf() below, not before it

src/bundle.c
```diff
@@ -721,6 +721,22 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+			fprintf(stderr, "FAILED HASH CHECK %s\n", fullname);
```
phmccarty: I would prefer this message to be more consistent with other errors (i.e. "Error: hash check failed for %s\n")

matthewrsj:
FYI @tmarcu and @phmccarty I'm working on fixing the tests. It looks like a few code changes need to be made. I can make those if you don't mind @tmarcu .

phmccarty:
+1

tmarcu:
+1 from the past

PR #248, opened 2017-09-12
Special handling for pack download/extract progress

matthewrsj:
Printing a percentage status for pack download and extraction is
somewhat meaningless, since the list will not be very long and output is
already sufficiently verbose.

Instead, update the pack fprintfs to include the current and total
number of packs. Print the pack name for every pack, not just those that
are being extracted so that the "(pack n of N)" message is meaningful.


src/packs.c
```diff
@@ -111,11 +110,15 @@ int download_subscribed_packs(struct list *subs, bool required)
 		iter = iter->next;
 
 		if (sub->oldversion == sub->version) { // pack didn't change in this release
+			fprintf(stderr, "no changes for %s pack for current version (pack %d of %d)\n",
+				sub->component, ++complete, list_length);
```
phmccarty: I'm not convinced we should print a progress message for this case. If a user has many bundles installed and is updating all the time, not many bundles will change (on average) during an update, but all of these extra progress messages will be printed.

src/packs.c
```diff
@@ -111,11 +110,15 @@ int download_subscribed_packs(struct list *subs, bool required)
...
 			continue;
 		}
 
+		fprintf(stderr, "%s pack for version %i (pack %d of %d)\n",
```
phmccarty: Hmm. The reason why we made the last change to pack progress messages was to avoid printing *anything* when a delta pack does not exist, because this was confusing to users.

phmccarty:
I agree that the current progress message state for packs is subpar, but I would rather see progress messages only for packs that have been downloaded.

matthewrsj:
I see what you mean. The reason I added in the progress messages for *all* packs is because the `(pack n of N)` got confusing when the `n` would jump over unchanged packs and `N` isn't representative of the number of packs actually being processed.

Any suggestions?

phmccarty:
You could check for pack existence first. Then you could iterate over the list of packs detected and output something like:

```
Downloading 5 packs
[1/5] Downloading c-basic pack for build 12000...
[1/5] Extracting pack...
[2/5] Downloading os-core pack for build 12010...
[2/5] Extracting pack...
...
```

matthewrsj:
@phmccarty that looks much nicer. Updated to that.

phmccarty:
Thanks @matthewrsj. I like the new formatting of the messages better. However, "Downloading ..." is still printed whenever packs do not exist, and this is what has been confusing to end users.

Do you agree that the message should not be printed when delta/zero packs do not exist?

A possible way to implement this, accounting for missing packs, is described in my previous comment.

matthewrsj:
I do agree.

I don't see the difference in your previous comment between the first example and the example in your edit. Is that not the same case?

phmccarty:
There is no difference, but I just wanted to clarify that "N" in "Downloading N packs ..." is the number of packs that exist. The original example did not mention it.

matthewrsj:
Right.  OK I updated it so that the message is not printed at all when there are zero packs to download. Other than that, this will only print the specific `Downloading...` and `Extracting...` messages for the packs that need to be downloaded.

matthewrsj:
Closing this for now until I figure out my errors. Opened #255 to track.

PR #268, opened 2017-09-22
Fix for issue 260, make portability

icarus-sparry:
Use = rather than := on the line. It is only an efficiency gain.

replace patsubst and filter with a new variable and standard
substitution.

Replace %-style pattern with suffix rule.


phmccarty:
+1, thanks!

matthewrsj:
+1

PR #269, opened 2017-09-22
fix for issue 261, compiler warning

icarus-sparry:
glibc marks system() as a function whose return value should not be
ignored, which is generally reasonable. In this case we do want to
ignore the return value. Just casting to vooid is not enough to stop
the gcc warning. So instead we use the value in an if statement with
an empty body, which we trust the optimizer to remove so there is no
run time penalty.

The alternative solution would be to use
    #pragma GCC diagnostic push
    #pragma GCC diagnostic ignored "-Wwarn_unused_result"
    system(log_cmd);
    #pragma GCC diagnostic pop
but that is rather gcc specific.


matthewrsj:
I don't really like it but it's better than assigning to an `__attribute__((unused))` that I've seen suggested elsewhere. Also better than the macros to silence it there.

+1

Issue #258, opened 2017-09-21
Implement better functional test creation workflow


phmccarty:
Currently, in order to write a new functional test, the BKM is to copy an existing functional test *similar* to the new one you want to write, and modify it accordingly. Pain points include several manual steps: (a) calculating swupd hashes of files, (b) updating the manifests according to the test's needs, and (c) calculating the swupd hash of each bundle manifest to include in the Manifest.MoM as a final step.

A workflow should be implemented to facilitate automating test creation, in particular the three items listed above.

matthewrsj:
Closed by @castulo in several PRs. The only remaining work is porting the actual tests, which is not part of this issue (and as I understand it is pretty much done anyways).

PR #272, opened 2017-09-26
Fix + test for renames

icarus-sparry:
Not sure what people think of putting in binary files for the bsdiff.

Factored as 3 patches, updated swupdlib.bash functionality, a test for rename, and the fix for rename.


matthewrsj:
You seem to be checking in a couple binary delta files, but it looks like those are automatically added during the test setup?

icarus-sparry:
I don't believe they are generated during the test setup, they are copied from web-dir to staged. We definitely could do with some better tooling around creating tests.

matthewrsj:
> We definitely could do with some better tooling around creating tests.

Agree, see #258 

Instead of adding the delta packs, you can just add the files to the `web-dir/20/staged` directory.

I also noticed you aren't checking the output with a lines-checked file.

icarus-sparry:
I think that checking the output with a "lines-checked" adds very little value, and just adds to future maintenance issues. The test checks that the files which should be renamed are removed, and the new ones are present, which is the thing we are trying to test, not that we are outputting the correct copyright message.
OK, will rework the test to not use the delta files but have staged ascii files instead.

icarus-sparry:
If we don't have the delta files, how do we know that we are actually doing the rename, rather than just doing normal processing of adding files from staging and removing them from target-dir?

matthewrsj:
@icarus-sparry let's stick with deltas

matthewrsj:
lgtm and the test gives me confidence that this is actually working. I would like one more +1 on this before merging though.

matthewrsj:
@phmccarty review ping

test/functional/swupdlib.bash
```diff
@@ -107,4 +121,30 @@ sign_manifest_mom() {
...
+  while read -u 9 line
+  do  [ -z "$line" ] && break
+  done
+  # Process lines
+  mkdir -p "$DIR/target-dir"
+  while read -u 9 code hash ver path
+  do
+      case "$code" in
+	  ("F"*) sudo cp "$DIR/web-dir/$REL/staged/$hash" "$DIR/target-dir/$path" ;;
+	  ("D"*) sudo mkdir -p "$DIR/target-dir/$path" ;;
```
phmccarty: Do you want to add symlink ("L") support right now, or defer to later?

icarus-sparry: I propose to defer it to later.

phmccarty: Okay, I opened #302 to track this.

phmccarty: +1 to defer for now

Issue #260, opened 2017-09-21
autoreconf warning


phmccarty:
When running `./autogen.sh` or `autoreconf -fi`, the following warning is emitted:

```
Makefile.am:220: warning: ':='-style assignments are not portable
Makefile.am:220: warning: patsubst %.bats,%.log,$(filter %.bats,$(dist_check_SCRIPTS: non-POSIX variable name
Makefile.am:220: (probably a GNU make extension)
Makefile.am:286: warning: '%'-style pattern rules are a GNU make extension
```

matthewrsj:
Fixed in #268

PR #274, opened 2017-09-26
Fix (doc) build failure for separate src and build dirs.

klihub:
Without this patch, if the docs are not up to date generating them fails if separate src and build dirs are used.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -287,7 +287,8 @@ clean-local:
 man: $(dist_man_MANS)
 
 .rst:
-	rst2man.py $< > $@
+	mkdir -p "$(dir $@)"
+	rst2man.py "$<" > "$@.tmp" && mv -f "$@.tmp" "$@"
 
 swupd.bash: $(top_srcdir)/scripts/swupd-completion.sh $(top_srcdir)/swupd
 	bash $<


matthewrsj:
Looks like sourceforge is down https://twitter.com/sfnet_ops/status/912673956712341514 which explains the Travis error.

Makefile.am
```diff
@@ -287,6 +287,7 @@ clean-local:
 man: $(dist_man_MANS)
 
 .rst:
+	mkdir -p $$(dirname $@)
```
icarus-sparry: Good idea, but should be
        mkdir -p "$(dir $@)"
to let make do the calculation,

Makefile.am
```diff
@@ -287,6 +287,7 @@ clean-local:
...
 	rst2man.py $< > $@
```
icarus-sparry: Whilst one is mucking around with this, it would be better to sau

     rst2man.py "<$" > "$@.tmp" && mv -f "$@.tmp" "$@"

to protect against signals partway through processing the manual source.

klihub:
Updated accordingly. Then again, now that I stopped also thinking about what I'm doing, I guess you'd like more something like this (since you seem to keep pre-generated man pages in the repo):

    .rst:
    	rst2man.py "$<" > "$@.tmp" && \
    	        mv -f "$@.tmp" "$(dir $<)/$(notdir $@)"

icarus-sparry:
No actually. Icarus's rule of Makefiles is that a rule should generate exactly $@, nothing more (and hopefully nothing less, with obvious exceptions for .PHONY targets). If you don't then make will get confused, and will either do too much or too little work.

This kind of rule is easy to specify in GNUmake with % pattern rules, but I just submitted a fix for #260 so I am reluctant to add them back in.

klihub:
Let's just then make sure the output dir exists.

matthewrsj:
This lgtm unless @icarus-sparry has any other comments?

icarus-sparry:
lgtm

Issue #275, opened 2017-09-26
Progress for content download ends before 100%


matthewrsj:
When updating sometimes the progress indicator for content download will stop before it reaches 100%
```
Starting download of remaining update content. This may take a while...
        ...69%
Finishing download of update content...
```

matthewrsj:
Fixed by #279

Issue #256, opened 2017-09-19
Improve progress for redirects


matthewrsj:
Redirected output prints a `.` on the same line for every 10 files it installs on the system (or processes or downloads etc.). User-facing programs that display this output to the user, such as Ister, only display the output when a newline is printed. Because of this the user still will not see swupd progress until the task is complete, at which point it will print out a single extremely long line of dots:
```
................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
```
This is not very useful to the user as they still could think the process is hanging. Once it finally completes they get a very meaningless block of dots (since the line is wrapped).

It would be better to print redirected progress as a percentage of the total, each percentage on a new line:
```
5%
10%
15%
...
100%
```

matthewrsj:
Fixed by #279

PR #279, opened 2017-09-26
F/275

icarus-sparry:
2 progress meter related changes, as 3 patches.

1) Fix the fix-missing-file so it doesn't get upset with extra dots coming out partway through other output. This will happen when we flush the progress meter.

2) Make the progress meter more accurate, and show 100% at the end of each step.  Issue #275 

3) Flush the non interactive progress meter, Issue #256 - adding this breaks the fix-missing-file.

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -40,8 +40,6 @@
 #include "signature.h"
 #include "swupd.h"
 
-static int last_percentage = -1;
-
 void check_root(void)
 {
 	if (getuid() != 0) {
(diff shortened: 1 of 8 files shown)


matthewrsj:
Please rebase to latest master to fix CI error.

src/helpers.c
```diff
@@ -40,7 +40,6 @@
 #include "signature.h"
 #include "swupd.h"
 
-static int last_percentage = -1;
```
matthewrsj: `make compliant` complains about an extra line here.

matthewrsj:
+1, I'll merge this after the code style fix is made.

icarus-sparry:
Style updated. It does strike me that we should be sending progress meters to stderr (i.e. the terminal) rather than stdout to maybe be consumed by another program.

matthewrsj:
I agree. #294 open.

Issue #285, opened 2017-09-27
use libarchive directly


pohly:
This is related to https://github.com/clearlinux/swupd-server/pull/48. Once the server uses libarchive, the client should also use bsdtar, which is already possible.

But calling libarchive directly would have several advantages:
- when unpacking Manifest files, restoring file attributes can be skipped, thus avoiding a problem with "ioctl(4, FS_IOC_GETFLAGS, 0x7ffe5c46d874) = -1 ENOTTY (Inappropriate ioctl for device)" when the statedir is on a tmpfs
- tar archives can be unpacked directly, without storing the entire file on disk first (GNUtar needs to seek in the file to detect compression methods, libarchive doesn't)
- file name or link name encoding errors in a locale which does not support UTF-8 can be ignored because the fallback is to skip re-encoding, which is exactly what swupd needs (would solve https://github.com/clearlinux/swupd-client/issues/280)

Right now, https://github.com/clearlinux/swupd-client/issues/249 is implemented with statedir on the target partition (because of https://github.com/clearlinux/swupd-client/issues/273), but then the target partition must have at least enough space for "unpacked rootfs + pack-os-core-from-0.tar".

pohly:
The goal is to do "swupd verify --install -p /mnt/some-partition" and directly stream the OS content onto the target partition.

Right now we put the state directory onto the target partition (partly because of issue #273, partly because unpacking Manifests into /tmp fails due to the ioctl() problem), but that means that additional space is needed on the OS partition.

matthewrsj:
Fixed by #327

PR #287, opened 2017-09-27
show tar errors

pohly:
Sending all tar stderr to /dev/null makes it hard to determine why
something failed.

This is particularly relevant for the issues we still do not fully
understand in https://github.com/clearlinux/swupd-server/pull/48.
I also needed this to see that I was simply running out of disk space
during a test.

### DIFF
--- a/src/download.c
+++ b/src/download.c
@@ -198,7 +198,7 @@ static int check_tarfile_content(struct file *file, const char *tarfilename)
 	FILE *tar;
 	int count = 0;
 
-	string_or_die(&tarcommand, TAR_COMMAND " -tf %s/download/%s.tar 2> /dev/null", state_dir, file->hash);
+	string_or_die(&tarcommand, TAR_COMMAND " -tf %s/download/%s.tar", state_dir, file->hash);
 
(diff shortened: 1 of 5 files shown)


matthewrsj:
@tmarcu @phmccarty tar errors were originally redirected to /dev/null to silence the slew of `time stamp is in the future` errors correct?

Maybe this could be made optional with a `--verbose` flag? That flag could also absorb the `--verbose-time` flag.

phmccarty:
@matthewrsj Yes, that's the primary reason for silencing tar. Many upstream projects ship files with timestamps that are skewed...

IMO, a solution here needs to silence tar warnings but not tar errors.

pohly:
Looks like the right solution will be to call libarchive directly (issue #285) and then control which errors need to be printed inside swupd code.

If that is taking too long to implement, then showing stderr for a "--verbose" flag in the meantime would be useful. It would also be useful to show all GET requests when such a flag is given.

pohly:
@tmarcu As far as I am concerned, you can close this issue and implement something else instead. I don't have plans to change on this PR.

Issue #257, opened 2017-09-19
Add -m flag for swupd update


matthewrsj:
Add flag to allow user to update to a particular version as long as that version is > the current version. Context surrounding this feature can be found in #249

pohly:
When -m is given, the version URL parameter is not necessary and if given, must not be used.

matthewrsj:
Good point, I agree that it should not be supported to switch between content streams like that, since update doesn't check all the files on the filesystem. swupd-client should print a warning that the versionurl parameter is being ignored.

bryteise:
@pohly The version URL is needed because -m could give "latest" which needs to be validated against the version URL.

bryteise:
Oh nevermind I see what you are saying =P.

bryteise:
I'm still a little unsure about saying it should not be used though. In the case where somebody sets up a mirror, an alternate version and version url would be acceptable as they should be stopped from going to a different content stream due to the content being signed.

pohly:
Let's take an example:

```
swupd update -m 200 -c http://example.org/swupd-content -v http://example.org/swupd-versions
```

Why would it need to access example.org/swupd-versions? What additional checks depend on data found there?

My expectation is that "swupd update" will only need information from example.org/swupd-content/200/ and example.org/swupd-content/<current version>.

bryteise:
In this case if you give
`swupd update -m latest` you'd want to give the version url as it needs to look up what the latest version is on the mirror.

matthewrsj:
Maybe `-m latest` can be treated differently? Or not allowed at all? It is default behavior to update to latest, so that flag would essentially be ignored anyways would it not?

pohly:
My expectation was that -m would always take an integer version number.

If someone wants "-m latest", they should simply not use "-m".

bryteise:
Yes, you can remove latest handling in update and that would resolve my worry.

pohly:
Exactly :-) "-m 100" is meant to override the "update to latest" behavior.

bryteise:
I always like being able to specify the default behavior with options as well but it isn't a requirement and makes this case harder so the latest handling can be dropped here for sure.

cmarcelo:
> When -m is given, the version URL parameter is not necessary and if given, must not be used.

It is still needed if the update to a version in another format. Could be avoided if update is between same format.

Issue #249, opened 2017-09-13
update inactive partition


pohly:
We'd like to do A/B partition updates with swupd. "swupd verify --install/--fix" almost work, but not quite.

The expectation is:
* Existing files are verified (attributes + content).
* This full check should be on by default, with an option to disable it when the existing file content is trusted and the hash for it is known (i.e. there's a manifest for it) - not crucial.
* Delta packs should be used when possible.
* This is needs to be a "first class citizen" in the features supported by swupd:
  * top-level command would be better than hiding it behind "verify" (not crucial)
  * automated testing for the feature under various circumstances

"swupd verify --install" does not actually replace modified files. To reproduce with 3.12.0:
```
# rm -rf /tmp/swupd-state/ /tmp/swupd-root/ && mkdir /tmp/swupd-root
# swupd verify --install --picky -c https://download.clearlinux.org/update/ -v https://download.clearlinux.org/update/ -F 19 -m 17700 --time -S /tmp/swupd-state -p /tmp/swupd-root
...
# swupd verify --install --picky -c https://download.clearlinux.org/update/ -v https://download.clearlinux.org/update/ -F 19 -m 17730 --time -S /tmp/swupd-state -p /tmp/swupd-root
...
Adding any missing files

Inspected 2093 files
  0 files were missing
Calling post-update helper scripts.
...
# swupd verify --fix --picky -c https://download.clearlinux.org/update/ -v https://download.clearlinux.org/update/ -F 19 -m 17730 --time -S /tmp/swupd-state -p /tmp/swupd-root
...
Adding any missing files

Fixing modified files
Hash mismatch for file: /tmp/swupd-root/usr/lib/os-release
	fixed
Hash mismatch for file: /tmp/swupd-root/usr/lib/systemd/systemd-bootchart
	fixed
Hash mismatch for file: /tmp/swupd-root/usr/share/clear/version
	fixed
Hash mismatch for file: /tmp/swupd-root/usr/share/clear/versionstamp
	fixed
--picky removing extra files under /tmp/swupd-root/usr
Inspected 2067 files
  0 files were missing
  4 files did not match
    4 of 4 files were fixed
    0 of 4 files were not fixed
  0 files found which should be deleted
```

Only the "verify --fix" really changed the files.

Running the "verify --install -m 17700" followed by "verify --fix -m 17730" under strace and inspecting /tmp/swupd-state shows that it downloaded individual files instead of the pack-os-core-from-17700.tar.

[...]

PR #288, opened 2017-09-27
support mixing http and file URLs

pohly:
I need this in the context of the script that implements
https://github.com/clearlinux/swupd-client/issues/249: it works around
the lack of the "-m" parameter for "swupd
update" (https://github.com/clearlinux/swupd-client/issues/257) by
creationg a local directory with "version/format<xxx>/latest" and
using that via file://, while the actual content comes via http://.

### DIFF
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -134,7 +134,6 @@ struct time {
 typedef TAILQ_HEAD(timelist, time) timelist;
 
 extern bool download_only;
-extern bool local_download;
 extern bool verify_esp_only;
 extern bool have_manifest_diskspace;
 extern bool have_network;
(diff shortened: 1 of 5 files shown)


matthewrsj:
So will this still be necessary after adding a -m parameter for update, or is this just a temporary workaround?

matthewrsj:
I'm inclined to +1 this anyways, since it seems silly to require both URLs use the same protocol just for the existing logic.

@phmccarty do you recall if there was another reason for requiring each to be the same protocol?

phmccarty:
@matthewrsj There is no technical reason for forbidding the protocol mix.

matthewrsj:
+1

PR #291, opened 2017-09-28
DO NOT MERGE: Rough draft of fallback capath support.

busykai:
@tmarcu, @phmccarty, @matthewrsj 

this is to support fallback trust sources for https connectivity in swupd, so that if the default trust store (configured in curl) does not work, it would try alternatives. this way, we protect the critical path for Clear Linux to be able to peform an update.

this PR is rather just an illustration to the two questions that i have:

1. i'd like to suggest an alternative way of checking connectivity. currently `check_network()` (from `src/version.c` is being used which internally fetches the version file, actually. instead, i propose that we have specialized `swupd_curl_check_network()` which, when called, also pick the appropriate trust store and if connection could not be established. additionally (this is not present yet in this PR), i think `swupd_curl_check_network()` could also be called immediately on `swupd_curl_init()`.

2. do you think we also need to make it a runtime command-line option? i noticed that most, if not all, build configuration options have their command-line analogs.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -179,6 +179,12 @@ AS_IF(
 )
 
 AC_ARG_WITH(
+  [alt-ca-paths],
+  [AS_HELP_STRING([--with-alt-ca-paths=PATHS], [List of alternative CApaths to use for https connectivity])], 
+  [AC_DEFINE_UNQUOTED([ALT_CA_PATHS], ["$withval"], [List of alternative CApaths to use for https connectivity])]
+)
+
+AC_ARG_WITH(
   [contenturl],
   [AS_HELP_STRING([--with-contenturl=URL], [Default content url])],
   [AC_DEFINE_UNQUOTED([CONTENTURL], ["$withval"], [Default content url])]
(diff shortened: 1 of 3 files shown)


matthewrsj:
> instead, i propose that we have specialized swupd_curl_check_network() which, when called, also pick the appropriate trust store and if connection could not be established.

+1, that's a good idea

> additionally (this is not present yet in this PR), i think swupd_curl_check_network() could also be called immediately on swupd_curl_init().

Hmm, would have to be careful here, since some commands do not need network and should therefore not check for network.

> do you think we also need to make it a runtime command-line option? i noticed that most, if not all, build configuration options have their command-line analogs.

This seems more like a configuration option than a runtime command-line option to me, but @phmccarty might have a different view.

busykai:
thanks @matthewrsj!

> Hmm, would have to be careful here, since some commands do not need network and should therefore not check for network.

Yes, you're right. It's better to have explicit checks on the code path where it's actually needed.

> This seems more like a configuration option than a runtime command-line option to me

OK, so it does to me, but runtime does not exclude build-time. I'll just implement it in a way so it would be easy to add a runtime analogue.

PR #297, opened 2017-09-28
Fix compiler warnings in search.c

### LINKED ISSUE #265: compiler warnings in search.c
```
src/search.c: In function 'search_main':
src/search.c:249:6: warning: argument 2 null where non-null expected [-Wnonnull]
  pos = strstr(filename, path);
  ~~~~^~~~~~~~~~~~~~~~~~~~~~~~
In file included from src/search.c:30:0:
/usr/include/string.h:329:14: note: in a call to function 'strstr' declared here
 extern char *strstr (const char *__haystack, const char *__needle)
              ^~~~~~
src/search.c:255:23: warning: argument 1 null where non-null expected [-Wnonnull]
  if (strcasestr(pos + strlen(path), search_term)) {
                       ^~~~~~~~~~~~
In file included from src/search.c:30:0:
/usr/include/string.h:384:15: note: in a call to function 'strlen' declared here
 extern size_t strlen (const char *__s)
               ^~~~~~
src/search.c:255:6: warning: argument 2 null where non-null expected [-Wnonnull]
  if (strcasestr(pos + strlen(path), search_term)) {
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from src/search.c:30:0:
/usr/include/string.h:359:14: note: in a call to function 'strcasestr' declared here
 extern char *strcasestr (const char *__haystack, const char *__needle)
              ^~~~~~~~~~
```

phmccarty:
Fixes #265 

The calls to strstr() and strcasestr() in file_search() from search.c
result in compiler warnings when -Wnonnull is specified, because GCC's
"nonnull" attribute applies to these functions.

The logic from the file_search() call sites prevent NULL pointers from
being passed to these functions, but to protect against future
refactoring introducing this bug, check that the appropriate pointers
are non-NULL before calling the string functions.

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -246,14 +246,18 @@ static bool file_search(char *filename, char *path, char *search_term)
 		return false;
 	}
 
-	pos = strstr(filename, path);
-	if (pos == NULL) {
-		return false;
+	if (path) {
+		pos = strstr(filename, path);
+		if (pos == NULL) {
+			return false;
+		}
 	}
 
-	/* match filename or substring of filename */
-	if (strcasestr(pos + strlen(path), search_term)) {
-		return true;
+	if (path && search_term) {
+		/* match filename or substring of filename */
+		if (strcasestr(pos + strlen(path), search_term)) {
+			return true;
+		}
 	}
 
 	return false;


matthewrsj:
+1

PR #299, opened 2017-09-29
Implement fallback CApath support.

busykai:
A colon-separated list of alternative CApath options can be passed to
swupd-client at the configuration time using --with-fallback-capaths
option.

In runtime, fallback CApath support is implemented as part of the
connectivity check. The implementation will try the default (built into
curl) and then will iterate through the list in the order they were
specified. It is done only once on the first call.

The code is reorganized to keep the connectivity check inside curl
wrapper:

* Deleted check_network implementation from version.c
* Removed have_network global (globals.c)
* Scoped swupd_curl_test_resume to curl.c
* Change use of check_network to swupd_curl_check_network

Also:

* Fixed an issue where SSL was only enabled if a URL was matching the
  content URL

@tmarcu, @matthewrsj please review ASAP


src/curl.c
```diff
@@ -480,7 +606,7 @@ CURLcode swupd_curl_set_basic_options(CURL *curl, const char *url)
 		}
 	}
 
-	if (strncmp(url, content_url, strlen(content_url)) == 0) {
```
busykai: This was very wrong.

src/curl.c
```diff
@@ -69,9 +75,115 @@ int swupd_curl_init(void)
...
+	c = curl_easy_duphandle(curl);
+	if (!c) {
+		return -1;
+	}
+
+	/* cycle starts with -1 which tests the default */
+	while (i < fallback_capaths_no) {
+		curl_easy_reset(c);
+
+		curl_ret = curl_easy_setopt(c, CURLOPT_URL, VERSIONURL);
```
matthewrsj: Compilation fails here because VERSIONURL is undeclared

busykai: Hm. Why travis configures a broken swupd? :) Without any endpoints configured. Sorry, will fix this...

matthewrsj:
You have a lot of test failures due to output checking. This can be fixed in this case by removing the `Attempting to download version string to memory` from most (all?) of the `lines-checked` files in `tests/functional/*/*/` directories

busykai:
@matthewrsj, thank you! i should have read your comment. now i figured it myself too. :)

src/curl.c
```diff
@@ -52,6 +52,12 @@ static int curr_version = -1;
 static int req_version = -1;
 static bool resume_download_enabled = true;
 
+/* these are used to handle alternative trust locations */
+#define CA_PATH_NO_MAX 16
+static char *capath;
+static char *fallback_capaths[CA_PATH_NO_MAX];
+static char fallback_capaths_no = 0;
```
icarus-sparry: I think just changing this to 1 will allow you to remove some special case code, see other comments.

src/curl.c
```diff
@@ -69,9 +75,108 @@ int swupd_curl_init(void)
...
+
+		curl_ret = curl_easy_setopt(c, CURLOPT_NOBODY, 1L);
+		if (curl_ret != CURLE_OK) {
+			goto cleanup;
+		}
+
+		if (i >= 0) { /* -1 stands for default */
+			curl_ret = curl_easy_setopt(c, CURLOPT_CAPATH, fallback_capaths[i]);
+			if (curl_ret != CURLE_OK) {
+				break;
```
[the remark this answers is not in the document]
busykai: the only reason setopt could fail with `CURLOPT_PATH` is when the curl backend TLS engine does not support it (never our case). i had it on continue until i figured that out... i guess continue will not harm since eventually it's going to fail anyways...

src/curl.c
```diff
@@ -129,6 +234,13 @@ double swupd_query_url_content_size(char *url)
 		return -1;
 	}
 
+	if (capath) {
```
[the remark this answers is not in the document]
busykai: that code "picks" a usable capath. this one actually uses it.

src/curl.c
```diff
@@ -69,9 +75,108 @@ int swupd_curl_init(void)
...
+		tok = strtok_r(str, ":", &ctx);
+		do {
+			if (stat(tok, &st)) {
+				continue;
+			}
+			if ((st.st_mode & S_IFMT) != S_IFDIR) {
+				continue;
+			}
+			fallback_capaths[i] = strdup(tok);
+			if (++i > CA_PATH_NO_MAX) {
```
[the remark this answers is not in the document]
busykai: yes, this is ugly. will do that.

src/curl.c
```diff
@@ -69,9 +75,108 @@ int swupd_curl_init(void)
...
+		curl_ret = curl_easy_perform(c);
+
+		switch (curl_ret) {
+			case CURLE_OK:
+				if (i >= 0) {
+					capath = fallback_capaths[i];
+				} else {
+					capath = NULL;
+				}
+				has_network = ret = 0;
```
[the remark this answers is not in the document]
busykai: this actually is a bug. :( `has_network` should be 1 here.

busykai:
@matthewrsj, I addressed the issues with the tests. finally, this "attempting to download" appears in the places where it makes sense! :)

src/curl.c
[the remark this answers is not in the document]
icarus-sparry: So at 145 we pick a usable path. What unsets the choice of the path that we need to set it again here? I see "c=curl_easy_duphandle(curl)" and think that this implies changes to c will also affect curl, so a working path will call curl_easy_init at line 71 in swupd_curl_init(), set the CA path at 146 in one of the loops, test it works at 152, then return at 176. Of course this depends on my understanding of the dup. If I am wrong, then why are we calling curl_easy_duphandle at all?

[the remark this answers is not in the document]
busykai: `c` is a comletely independent CURL handle. i could well use one `curl_easy_init()`. duplication only to make sure we actually take the right starting point if it was manipulated before (e.g. if `swupd_curl_init()` were to set any opts).

so code in `swupd_curl_check_network()` creates a dup and then destroys it. i would set CApath there (on file-scope `curl`), but the rest of the code is organized so that it constantly `curl_easy_reset()` and then re-configures before calling `curl_easy_perform()`. so it would not "stick" on global since any reset would wipe it.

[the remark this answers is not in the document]
icarus-sparry: ok, thanks. that makes sense.

src/curl.c
```diff
@@ -93,16 +101,24 @@ int swupd_curl_init(void)
...
 			}
 			fallback_capaths[i] = strdup(tok);
-			if (++i > CA_PATH_NO_MAX) {
-				fprintf(stderr, "Warning: up to %d CApaths can be specified."
-						" Ignoring the rest.\n", CA_PATH_NO_MAX);
-				break;
+			if (++i > sz) {
+				char **re_fallback_capaths, **t;
+				sz <<= 1;
+				re_fallback_capaths = (char **)malloc(sz*sizeof(char *));
```
tmarcu: Why do we care about malloc()'s return value above this, but not the two spots below?

src/curl.c
```diff
@@ -93,16 +101,24 @@ int swupd_curl_init(void)
...
+				memcpy(re_fallback_capaths, fallback_capaths, i*sizeof(char *));
```
icarus-sparry: Also this is almost exactly the behavior of realloc (malloc new array, copy values, free old one). Now it is true that realloc has some issues if you want to fail gracefully if you run out of memory, but swupd client is not one of those programs (see uses of string_or_die everywhere). I would just use the library function.

busykai:
@tmarcu, @icarus-sparry cleaned up allocation/reallocation.

src/check_update.c
```diff
@@ -139,13 +139,18 @@ static int check_update()
...
 		fprintf(stderr, "Error: Network issue, unable to check for update\n");
 		return ENOSWUPDSERVER;
 	}
 
 	read_versions(&current_version, &server_version, path_prefix);
 
+    if (server_version < 0) {
+        fprintf(stderr, "Error: server does not report any version\n");
+        return ENOSWUPDSERVER;
+    }
```
matthewrsj: Please use tabs for indentation

src/curl.c
```diff
@@ -69,9 +74,113 @@ int swupd_curl_init(void)
...
+	/* parse and initialize CA paths. */
+	{
+		char *str = strdup(FALLBACK_CAPATHS);
+		char *tok;
+		char *ctx;
+		struct stat st;
+		int i;
+		int sz = 1;
+
+		fallback_capaths = (char **)malloc(sz*sizeof(char *));
```
matthewrsj: Whitespace `sz * sizeof...`

src/curl.c
```diff
@@ -69,9 +74,113 @@ int swupd_curl_init(void)
...
+		do {
+			if (stat(tok, &st)) {
+				continue;
+			}
+			if ((st.st_mode & S_IFMT) != S_IFDIR) {
+				continue;
+			}
+			if (i == sz) {
+				sz <<= 1;
+				fallback_capaths = (char **)realloc(fallback_capaths, sz*sizeof(char *));
```
matthewrsj: Whitespace again

src/curl.c
```diff
@@ -69,9 +74,113 @@ int swupd_curl_init(void)
...
+	fallback_capaths[0] = NULL;
+	fallback_capaths_no = 1;
+#endif
+
 	return 0;
 }
 
+static void swupd_curl_test_resume(void);
+
+int swupd_curl_check_network(void) {
```
matthewrsj: The `{` should go on a newline for function definitions.

src/curl.c
```diff
@@ -69,9 +74,113 @@ int swupd_curl_init(void)
...
+	CURLcode curl_ret;
+	int ret = -1;
+	int i;
+	CURL *c;
+	static int has_network = 0;
+
+	if (has_network) return 0;
```
matthewrsj: the `return 0;` should go on a newline and surrounded by brackets even though it is a single line for consistency.

src/curl.c
```diff
@@ -69,9 +74,113 @@ int swupd_curl_init(void)
...
+			curl_ret = curl_easy_setopt(c, CURLOPT_CAPATH, fallback_capaths[i]);
+			if (curl_ret != CURLE_OK) {
+				break;
+			}
+		}
+
+		curl_ret = curl_easy_perform(c);
+
+		switch (curl_ret) {
+			case CURLE_OK:
```
matthewrsj: Bring these `case` statements back to the same indentation level as the `switch` statement.

busykai:
@matthewrsj, done!

matthewrsj:
@busykai looks good to me, please squash your commits and I'll +1. @icarus-sparry has he addressed all your feedback (looks like you still have open requests according to the github UI).

PR #308, opened 2017-10-05
Fix incorrect cycle.

busykai:
Remove the increment left-over from changing from while to for.


matthewrsj:
+1, easy to overlook

PR #309, opened 2017-10-06
Allow manpage generation to be optional

aaronovz1:
In a Yocto build environment we don't have rst2man.py and builds fail.
This allows a workaround and more flexibility in what is generated

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -270,7 +270,7 @@ release:
 	@git tag -a -m "$(PACKAGE_NAME) release $(PACKAGE_VERSION)" v$(PACKAGE_VERSION)
 	@printf "\nNew release $(PACKAGE_VERSION) tagged!\n\n"
 
-
+if ENABLE_MAN
 MANPAGES = \
 	docs/swupd.1 \
 	docs/check-update.service.4 \
@@ -290,6 +290,7 @@ man: $(dist_man_MANS)
 .rst:
 	mkdir -p "$$(dirname $@)"
 	rst2man.py "$<" > "$@.tmp" && mv -f "$@.tmp" "$@"
+endif
 
 swupd.bash: $(top_srcdir)/scripts/swupd-completion.sh $(top_srcdir)/swupd
 	bash $<
--- a/configure.ac
+++ b/configure.ac
@@ -129,6 +129,22 @@ AS_IF(
   [XATTR=no]
 )
 
+AC_ARG_ENABLE(
+  [manpages],
+  [AS_HELP_STRING([--disable-manpages], [Do not enable manpage generation (disabled by default)])]
+)
+MANPAGES="yes"
+AS_IF(
+  [test -n "$enable_manpages" -a "$enable_manpages" = "yes"],
+  [MANPAGES="$enable_manpages"]
+)
+AS_IF(
+  [test "$enable_manpages" != "no"],
+  [MANPAGES="yes"],
+  [MANPAGES="no"]
+)
+AM_CONDITIONAL([ENABLE_MAN], [test "$enable_manpages" = "yes"])
+
 TARSELINUX="yes"
 AC_ARG_ENABLE(
   [tar-selinux],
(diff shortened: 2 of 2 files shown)


aaronovz1:
In our particular Yocto setup we do not ship with manuals so there is no need to build them. The change has manuals built as default, but allows user to disable them if they don't need them. This also resolves the issue I am seeing since version 3.12.3 (commit ff1628191d8002f21b84801c124c1924a3f57b6b) which causes build issues in Yocto because rst2man.py is not found.

pohly:
Additional information: depending on the random time stamps assigned to the .rst input files and the man page output files after a git checkout, make sometimes invokes rst2man.py and sometimes it doesn't.

icarus-sparry:
In an ideal world, the man pages would not be stored in the repo, but instead generated from the restructured text source files.

pohly:
On Mon, 2017-10-09 at 18:45 +0000, icarus-sparry wrote:
> In an ideal world, the man pages would not be stored in the repo, but
> instead generated from the restructured text source files.

Then remove them and add a --enable-manpages configure option as
@aaronovz1 suggested. However, then configure should check for the
presence of rst2man.py when enabled (better fail early instead of
during compilation).
> 
>   
>   
> 
> 
> 
> 
> 
> {"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c5
> 5493e4bb","name":"GitHub"},"entity":{"external_key":"github/clearlinu
> x/swupd-client","title":"clearlinux/swupd-client","subtitle":"GitHub
> repository","main_image_url":"https://cloud.githubusercontent.com/ass
> ets/143418/17495839/a5054eac-5d88-11e6-95fc-
> 7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent
> .com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-
> b52498112777.png","action":{"name":"Open in
> GitHub","url":"https://github.com/clearlinux/swupd-
> client"}},"updates":{"snippets":[{"icon":"PERSON","message":"@icarus-
> sparry in #309: In an ideal world, the man pages would not be stored
> in the repo, but instead generated from the restructured text source
> files."}],"action":{"name":"View Pull
> Request","url":"https://github.com/clearlinux/swupd-
> client/pull/309#issuecomment-335252201"}}}

matthewrsj:
I think I would agree with @pohly here and say we shouldn't be building man pages unless passed `--enable-manpages`.

Could this PR be updated to work the other way around, with `--enable-manpages` instead of `--disable-manpages` and have the default be to not make the manpages?

aaronovz1:
Sure. I can swap it if that is final decision.

phmccarty:
@matthewrsj I think it's nicer to build/install man pages by default. If the required deps to build docs are not available, then a --disable-manpages option can skip the process entirely.

icarus-sparry:
I agree with @phmccarty, my expectation is that a unix package comes with manual pages. As I see it, we have a series of choices to make

1. Is the master source of the man pages in ReStructuredText format or not.
1. If it is, then do we want pre-generated man pages in the repo for places which don't have rst2man.py. 
1. If we do have pre-generated pages,  we probably want some additional hooks to ensure that the pages are up-to-date, maybe adding them to the "compliant" target, and not running it by default.

tmarcu:
Dealing with a couple other packages recently..it's very frustrating when the package/man pages don't ship together in a sane way. I'd support adding a --disable-manpages option, though.
@icarus-sparry is there something we could have like the build target you added to always update the swupd-cmdline options during make?

matthewrsj:
Sure, I defer to the majority, --disable-manpages works. I'll wait for @icarus-sparry to respond to @tmarcu.

icarus-sparry:
This problem shows the limitation of make depending just on timestamps, rather than the contents of the files.. There are ways around it, essentially adding another file containing checksums of the files and checking that into the repo. It can even  handle cases like where rst2man.py is not present provided that the restructured text source files are not altered.

It does require that people who do update the manual source files do do a build before the final checkin.

I will code it up.

ahkok:
I coded up the man pages and the way they're included in the build system used.

Specifically, they man pages should be included in compiled form in the `make dist` target and present in the release tarballs, which is what yocto should use and anyone who is using swupd-client in a distro.

The problem is that you're building from git.

```
$ make dist
...
$ tar tvf swupd-client-3.14.5.tar.gz |grep docs
drwxr-xr-x 1000/1000         0 2018-02-12 10:17 swupd-client-3.14.5/docs/
-rw-r--r-- 1000/1000      1404 2017-03-09 11:27 swupd-client-3.14.5/docs/swupd-update.timer.4
-rw-r--r-- 1000/1000      2014 2018-01-17 13:01 swupd-client-3.14.5/docs/swupd-add-pkg.4
-rw-r--r-- 1000/1000      1412 2017-03-09 11:27 swupd-client-3.14.5/docs/check-update.service.4
-rw-r--r-- 1000/1000      2081 2017-03-09 11:27 swupd-client-3.14.5/docs/update-triggers.target.4
-rw-r--r-- 1000/1000     13506 2018-02-12 10:16 swupd-client-3.14.5/docs/swupd.1
-rw-r--r-- 1000/1000      1575 2017-03-09 11:27 swupd-client-3.14.5/docs/swupd-update.service.4
-rw-r--r-- 1000/1000      1419 2017-03-09 11:27 swupd-client-3.14.5/docs/check-update.timer.4
```

Solution: use the release tarballs instead.

I don't support changing the way we do this. Distributions should use release tarballs or recreate the man pages that match the version of the code they distribute, and if they modify the code, they should recreate the man pages, or, they are obviously already patching the code and can patch out the man page bits themselves.

matthewrsj:
After @ahkok's comment I'm inclined to close this PR for now until we can agree on a solution or if one is even needed.

PR #310, opened 2017-10-06
test: implement validation for lines-checked; improve check_lines() error reporting

phmccarty:
The output line checking for functional tests should ensure that lines from `lines-checked` do not match any patterns in `ignore-list`.

Also, fix an issue with the update/rename test and improve check_lines() error reporting to account for `ignore-list`.

### DIFF
--- a/test/functional/swupdlib.bash
+++ b/test/functional/swupdlib.bash
@@ -106,6 +106,8 @@ check_lines() {
   if [ $ret -eq 1 ]; then
     echo -e "\nChecked lines versus actual output (note that actual output may contain ignored lines):\n"
     diff -u "$checked" "$outputfile"
+  elif [ $ret -eq 2 ]; then
+    echo -e "\n^ Error: lines-checked contains ignored lines\n"
   fi
 
   return $ret
--- a/test/functional/swupdlib.bash
+++ b/test/functional/swupdlib.bash
@@ -95,16 +95,20 @@ check_lines() {
   local checked="$DIR/lines-checked"
   local ignored="$DIR/../../ignore-list"
   local prog="$DIR/../../matcher.awk"
+  local ret
 
   echo "$outputstr" > "$outputfile"
 
   run awk -f "$prog" "$ignored" "$checked" "$outputfile"
+  ret=$status
+  echo "$output"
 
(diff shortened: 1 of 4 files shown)


phmccarty:
I restarted the Travis build to work around the issue of rst2man.py not being installed (see #311).

test/functional/swupdlib.bash
```diff
@@ -105,7 +105,18 @@ check_lines() {
 
   if [ $ret -eq 1 ]; then
     echo -e "\nChecked lines versus actual output (note that actual output may contain ignored lines):\n"
-    diff -u "$checked" "$outputfile"
+    # Omit ignored lines in the actual output to unobfuscate test error
+    # reporting.
+    diff -u "$checked" \
```
icarus-sparry: 2 comments. Normally one would use the NR == FNR idiom here (subject to trhe ignored file not being empty, or else just read the ignored file in a BEGIN clase), e.g.

awk 'NR == FNR { toignore[$0]=1 ; next} { for (i in toignore) { if ($0 ~ i) { next } } print $0 }' "$ignored" "$outputfile"

but I think the tool to use is grep, i.e.

grep -Evf "$ignored" "$output"

phmccarty: Ah, yes, thanks for the feedback. I will switch to using grep.

phmccarty:
I amended the last commit according the Icarus's feedback.

test/functional/swupdlib.bash
```diff
@@ -105,7 +105,9 @@ check_lines() {
 
   if [ $ret -eq 1 ]; then
     echo -e "\nChecked lines versus actual output (note that actual output may contain ignored lines):\n"
```
icarus-sparry: This echo should be updated as well.

phmccarty: Ugh, yes. Good catch.

matthewrsj:
^^ rst2man.py not found again.

phmccarty:
I retriggered the Travis run, and it succeeds now.

icarus-sparry:
+1

matthewrsj:
+1

PR #311, opened 2017-10-09
travis: install rst2man dependency

phmccarty:
Because rst2man.py is called when the timestamps on *.rst are newer than
their associated man pages, we need to have rst2man.py just in case we
hit this condition for the build.


icarus-sparry:
We ought to delete the man pages from the repo or else scrap the idea of using rst2man. The current middle ground of guess what timestamps git is going to give you is just wrong.

phmccarty:
@icarus-sparry That's a good point. I would be in favor of removing the static copies of the man pages, leaving them to be generated on the fly.

matthewrsj:
+1,  seeing this error in another PR reminded me this one was open.

PR #318, opened 2017-10-13
Implement ghost file handling

### LINKED ISSUE #317: Add "ghosted" file handling
When a file is "ghosted" in the manifest it should be treated as deleted for rename purposes, but swupd should not actually delete the file.

This functionality allows renames to occur on files that are garbage collected by other processes (i.e. kernels managed by clr-boot-manager) while not actually deleting those ghosted files.

matthewrsj:
Implement handling of ghosted files in manifests via the 'g' flag. These
files are treated the same as deleted files except that they are not
actually deleted during the update. This allows ghosted files to serve
as the basis for renames while not actually being deleted by swupd.

An example of where this is useful is with kernel files, which are
deleted from /usr/lib/kernel/ by clr-boot-manager but should still be
marked as deleted to utilize the rename functionality.

Adds functional tests as well.

Fixes #317


Makefile.am
```diff
@@ -185,7 +185,9 @@ BATS = \
 	test/functional/update/include-old-bundle-with-tracked-file/test.bats \
 	test/functional/update/missing-os-core/test.bats \
 	test/functional/update/newest-deleted/test.bats \
+	test/functional/update/newest-ghosted/test.bats \
 	test/functional/update/rename/test.bats \
+	test/functional/update/rename-ghosted/test.bats \
```
phmccarty: Can you move these makefile changes to the other commit which adds the tests?

test/functional/update/rename-ghosted/test.bats
```diff
@@ -0,0 +1,40 @@
...
+
+teardown() {
+  #clean_tars 10
+  #clean_tars 20
+  revert_chown_root -R "$DIR/web-dir"
+}
+
+@test "update rename ghosted file" {
+  run sudo sh -c "$SWUPD update $SWUPD_OPTS"
+
```
phmccarty: check `$status` here

matthewrsj:
@icarus-sparry that's a good point. We should not be deleting ghosted files during --fix. I also think we should not be deleting files during --picky either, since the initial use case for ghosted files is boot files, which we never want to delete.

matthewrsj:
@icarus-sparry @phmccarty ghosted files now ignored during verify --fix and verify --fix --picky. The tests I added should demonstrate that.

I would like to get this change in before next Tuesday if possible.

test/functional/update/newest-ghosted/test.bats
```diff
@@ -0,0 +1,43 @@
...
+load "../../swupdlib"
+
+os_release_file=1c1efd10467cbc1dddd8e63c3ef4d9099f36418ed5372d080a1bf0f03a49ab05
+
+setup() {
+  clean_test_dir
+  create_manifest_tar 10 MoM
+  sign_manifest_mom 10
+  create_manifest_tar 10 os-core
+  create_manifest_tar 10 test-bundle1
```
phmccarty: Do you need to include build 10 for this test? I see you're updating from build 20 to 30, and no new bundle includes are added for installed bundles in build 30, so the manifest content from build 10 would be irrelevant here. Unless I'm missing something :-)

test/functional/verify/picky-ghosted-missing/test.bats
```diff
@@ -0,0 +1,30 @@
...
+
+@test "verify ghosted file not added during --picky" {
+  run sudo sh -c "$SWUPD verify --fix --picky $SWUPD_OPTS"
+
+  check_lines "$output"
+  [ $status -eq 0 ]
+  # this should exist at the end, despite being marked as "ghosted" in the
+  # Manifest. This is an important test because internally swupd treats ghosted
+  # files as deleted for rename purposes, but we need to make sure the file
+  # doesn't actually get deleted.
```
phmccarty: This comment needs updating to describe the fact the verify will not install a missing file if it is ghosted -- and no renames are involved for this test.

test/functional/verify/picky-ghosted/test.bats
```diff
@@ -0,0 +1,31 @@
...
+
+@test "verify ghosted file skipped during --picky" {
+  run sudo sh -c "$SWUPD verify --fix --picky $SWUPD_OPTS"
+
+  check_lines "$output"
+  [ $status -eq 0 ]
+  # this should exist at the end, despite being marked as "ghosted" in the
+  # Manifest. This is an important test because internally swupd treats ghosted
+  # files as deleted for rename purposes, but we need to make sure the file
+  # doesn't actually get deleted.
```
phmccarty: Since this test is verifying correct behavior of ghosting *without* renames, I would update this comment too.

test/functional/verify/ghosted/test.bats
```diff
@@ -0,0 +1,31 @@
...
+}
+
+@test "verify ghosted file skipped" {
+  run sudo sh -c "$SWUPD verify --fix $SWUPD_OPTS"
+
+  check_lines "$output"
+  [ $status -eq 0 ]
+  # this should exist at the end, despite being marked as "ghosted" in the
+  # Manifest. This is an important test because internally swupd treats ghosted
+  # files as deleted for rename purposes, but we need to make sure the file
```
phmccarty: no renames are involved in this test

matthewrsj:
I seem to have lost my test cleanups with the various rebasing and force pushing I did (I had one or two git issues, sorry). I'll clean it up and let you know when I'm done.

test/functional/verify/ghosted/test.bats
matthewrsj: I was just trying to explain the reason for the file existence test, but I'll clarify.

matthewrsj:
Test cleanup done.

phmccarty:
+1

PR #319, opened 2017-10-16
Remove unused check_network() declaration

phmccarty:
The check_network() function definition was recently removed, but the
declaration remained. Remove the declaration as well.


matthewrsj:
+1

PR #321, opened 2017-10-19
Fix re-exec capabilities over format bumps

### LINKED ISSUE #320: Fix re-execs over format bumps
With server putting the `update` action in the new MoM **after** the format bump swupd-client was not actually re-execing itself over formats. A simpler solution than what we currently have is to just do the format change detection in the client. If the format has changed since the beginning of the update, do a re-exec.

matthewrsj:
Instead of relying on swupd-server to put the post-update action in the
correct manifest and in the correct format, do the format change
detection on the client.

Fixes #320 

### DIFF
--- a/include/swupd.h
+++ b/include/swupd.h
@@ -158,6 +158,7 @@ extern char *version_url;
 extern char *content_url;
 extern char *cert_path;
 extern long update_server_port;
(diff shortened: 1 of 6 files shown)


tmarcu:
lgtm

PR #324, opened 2017-10-20
Pretty print check_network error messages

phmccarty:
This line of code is a recent addition, and I think it is more helpful for the end user to see the pretty-printed curl error message in addition to the error number.

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -57,6 +57,12 @@ static char *capath;
 static char **fallback_capaths;
 static char fallback_capaths_no;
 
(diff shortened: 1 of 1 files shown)


matthewrsj:
Could you fix the other two curl error printfs to be consistent with this one? Right now curl errors are printed three different ways in the code base.

phmccarty:
I think I would rather add a curl error wrapper function. I'll do that, but maybe not today.

matthewrsj:
+1

PR #326, opened 2017-10-26
More a constant time check before a O(n) check

icarus-sparry:
Hopefully will drop yet another cubic operation to quadratic


matthewrsj:
+1

icarus-sparry:
Dropped the time to do an update in a test from 313 seconds to 20.5 seconds.

Issue #280, opened 2017-09-26
bsdtar: ignore UTF-8 encoding errors


pohly:
When trying with current refkit, swupd verify --install fails with:
```
	...100%zero pack downloads failed. 
Failed - Server-side error, cannot download necessary files
```

The actual error is not visible because stderr gets redirected to /dev/null (also for GNUtar - perhaps that should be changed?).

Manually unpacking the pack-os-core-from-0-to-20.tar with bsdtar reveals the actual root cause:
```
bsdtar: Linkname can't be converted from UTF-8 to current locale.
bsdtar: Linkname can't be converted from UTF-8 to current locale.
bsdtar: Linkname can't be converted from UTF-8 to current locale.
bsdtar: Linkname can't be converted from UTF-8 to current locale.
bsdtar: Error exit delayed from previous errors.
```

This is related to file name encoding, see https://github.com/libarchive/libarchive/wiki/Filenames

In my case, no locale is set and thus conversion from UTF-8 into 7-bit ASCII (or whatever the default is) fails.

I'm not sure yet what the proper solution is. Perhaps the system simply has to have a UTF-8 locale available, and it needs to be active.

matthewrsj:
Looks like the client where we did the libarchive testing is running in UTF-8 locale:
```
# localectl
   System Locale: LANG=en_US.UTF-8
       VC Keymap: n/a
      X11 Layout: n/a
```

matthewrsj:
This will be resolved by #327 once it is merged.

matthewrsj:
#327 has been merged. there is a similar issue being tracked in clearlinux/mixer-tools#92

PR #327, opened 2017-10-31
Use libarchive instead of calling out to tar

matthewrsj:
Use libarchive in order to make use of its security features and avoid
calling out to tar via a shell. The TAR_COMMAND is still used in
staging.c to complete the copy when a hardlink fails.

### DIFF
--- a/src/archives.c
+++ b/src/archives.c
@@ -1,0 +1,180 @@
+/*
+ *   Software Updater - client side
+ *
+ *      Copyright © 2017 Intel Corporation.
+ *
+ *   This program is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License as published by
+ *   the Free Software Foundation, version 2 or later of the License.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ *   Authors:
+ *         Matthew Johnson <mathew.johnson@intel.com>
+ *
+ */
+
+#include <archive.h>
+#include <archive_entry.h>
+#include <stdlib.h>
+
+#include "swupd.h"
+
+/* _archive_check_err(ar, ret)
+ *
+ * Check and print archive errors */
+static int _archive_check_err(struct archive *ar, int ret)
+{
+	int is_fatal = 0;
+
+	if (ret < ARCHIVE_WARN || ret == ARCHIVE_RETRY) {
+		/* error was worse than a warning or error was ARCHIVE_RETRY (greater
+		 * than ARCHIVE_WARN), which indicates failure with retry possible.
+		 * Regardless, indicate that the operation did not succeed.  */
+		fprintf(stderr, "Error: %s\n", archive_error_string(ar));
+		is_fatal = 1;
+	} else if (ret < ARCHIVE_OK) {
+		/* operation succeeded, warning encountered */
+		fprintf(stderr, "Warning: %s\n", archive_error_string(ar));
+		is_fatal = 0;
+	}
+
+	return is_fatal;
+}
+
+/* copy_data(ar, aw)
+ *
+ * Copy archive data from ar to aw */
+static int copy_data(struct archive *ar, struct archive *aw)
+{
+	int r;
+	const void *buffer;
+	size_t size;
+	off_t offset;
+
+	for (;;) {
+		r = archive_read_data_block(ar, &buffer, &size, &offset);
+		if (r == ARCHIVE_EOF) {
+			return ARCHIVE_OK;
+		} else if (r < ARCHIVE_OK) {
+			return r;
+		}
+
+		r = archive_write_data_block(aw, buffer, size, offset);
+		if (r < ARCHIVE_OK) {
+			return r;
+		}
+	}
+}
+
+/* extract_to(tarfile, outputdir)
+ *
+ * Extracts tar archive tarfile to outputdir. Refuses to extract any object
+ * whose final location would be altered by a symlink on disk. */
+int extract_to(const char *tarfile, const char *outputdir)
+{
+	struct archive *a, *ext;
+	struct archive_entry *entry;
+	int flags;
+	int r = 0;
+
+	/* set which attributes we want to restore */
+	flags = ARCHIVE_EXTRACT_TIME;
+	flags |= ARCHIVE_EXTRACT_PERM;
+	flags |= ARCHIVE_EXTRACT_OWNER;
+	flags |= ARCHIVE_EXTRACT_XATTR;
+
+	/* set security flags */
+	flags |= ARCHIVE_EXTRACT_SECURE_SYMLINKS;
+	flags |= ARCHIVE_EXTRACT_SECURE_NODOTDOT;
+
+	/* set up read */
+	a = archive_read_new();
+	r = archive_read_support_format_tar(a);
+	if (_archive_check_err(a, r)) {
+		goto out_read;
+	}
+
+	r = archive_read_support_filter_all(a);
+	if (_archive_check_err(a, r)) {
+		goto out_read;
+	}
+
+	/* set up write */
+	ext = archive_write_disk_new();
+	r = archive_write_disk_set_options(ext, flags);
+	if (_archive_check_err(ext, r)) {
+		goto out;
+	}
+
+	r = archive_write_disk_set_standard_lookup(ext);
+	if (_archive_check_err(ext, r)) {
+		goto out;
+	}
+
+	r = archive_read_open_filename(a, tarfile, 10240);
+	if (_archive_check_err(a, r)) {
+		/* could not open archive for read */
+		goto out;
+	}
+
+	/* read and write loop */
+	for (;;) {
+		r = archive_read_next_header(a, &entry);
+		if (r == ARCHIVE_EOF) {
+			/* reached end of file */
+			r = 0;
+			break;
+		}
+
+		if (_archive_check_err(ext, r)) {
+			goto out;
+		}
+
+		/* set output directory */
+		char *fullpath;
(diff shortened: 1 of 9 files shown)


src/download.c
```diff
@@ -293,14 +292,10 @@ int untar_full_download(void *data)
...
-		      state_dir, tarfile);
-
-	err = system(tarcommand);
-	if (WIFEXITED(err)) {
-		err = WEXITSTATUS(err);
-	}
-	free(tarcommand);
+	char *target;
+	string_or_die(&target, "%s/staged", state_dir);
+	err = extract(tarfile, target);
```
icarus-sparry: Not a fan of the name "extract", given it has two parameters, the tarfile and the directory to unpack into. Perhaps "untarto". I would expect that something called "extract" would have the name of the thing to extract as a parameter.

src/archives.c
```diff
@@ -0,0 +1,138 @@
...
+
+		r = archive_write_data_block(aw, buffer, size, offset);
+		if (r < ARCHIVE_OK) {
+			fprintf(stderr, "Error: %s\n", archive_error_string(aw));
+			return r;
+		}
+	}
+}
+
+int extract(const char *filename, const char *target)
```
icarus-sparry: At the very least this need some documentation to say what the parameters are, and that passing NULL as target is valid (but magic).

src/archives.c
```diff
@@ -0,0 +1,138 @@
...
+
+		archive_entry_set_pathname(entry, fullpath);
+		free(fullpath);
+
+		r = archive_write_header(ext, entry);
+		if (r < ARCHIVE_OK) {
+			fprintf(stderr, "%s\n", archive_error_string(ext));
+		} else if (archive_entry_size(entry) > 0) {
+			r = copy_data(a, ext);
+			if (r < ARCHIVE_OK) {
```
icarus-sparry: copy_data already prints an error in the write case. Probably should make copy_data silent and leave this message.

matthewrsj:
Having trouble getting libarchive installed in the Travis environment.

src/archives.c
matthewrsj: True, I'll add a docstring

src/archives.c
```diff
@@ -0,0 +1,138 @@
...
+		} else {
+			string_or_die(&fullpath, "%s/%s", state_dir, archive_entry_pathname(entry));
+		}
+
+		archive_entry_set_pathname(entry, fullpath);
+		free(fullpath);
+
+		r = archive_write_header(ext, entry);
+		if (r < ARCHIVE_OK) {
+			fprintf(stderr, "%s\n", archive_error_string(ext));
```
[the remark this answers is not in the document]
matthewrsj: Looks like it is missing a `if (r < ARCHIVE_WARN) return r;`

src/archives.c
icarus-sparry: Not a fan of the name "extract", given it has two parameters, the tarfile and the directory to unpack into. Perhaps "untarto". I would expect that something called "extract" would have the name of the thing to extract as a parameter.

src/archives.c
```diff
@@ -0,0 +1,138 @@
...
+		} else if (archive_entry_size(entry) > 0) {
```
icarus-sparry: Does this code, which skips copy_data for zero length mean we can't create zero length files?

matthewrsj: Quick test shows me we can create zero length files. From the man page:

> The function archive_entry_size() returns the file size, if it has been set, and 0 otherwise.  archive_entry_size() can be used to query that status.

When I created an archive with several empty files I was able to extract them using this code.

src/archives.c
```diff
@@ -0,0 +1,138 @@
...
+			fprintf(stderr, "%s\n", archive_error_string(ext));
+			return r;
+		}
+
+		if (r < ARCHIVE_WARN) {
+			return r;
+		}
+	}
+
+	archive_read_close(a);
```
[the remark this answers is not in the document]
matthewrsj: yep, I'll fix that.

src/archives.c
matthewrsj: ok, I'm thinking `extract_to(...)` then

matthewrsj:
*closed by mistake, oops

Fixes #285

src/archives.c
```diff
@@ -0,0 +1,155 @@
...
+
+/* copy_data(ar, aw)
+ *
+ * Copy archive data from ar to aw */
+static int copy_data(struct archive *ar, struct archive *aw)
+{
+	int r;
+	const void *buffer;
+	size_t size;
+	la_int64_t offset;
```
tmarcu: Just to be pedantic, from the man pages the function below is defined  archive_read_data_block(struct archive *, const void **buff, size_t *len, off_t *offset); 
I know off_t 'cause it's a standard type, so it would make sense to match them up and make it: off_t offset IMHO, then all the types are obvious and matching.

matthewrsj: Done!

tmarcu:
+1

matthewrsj:
Resolves #280 as well.

src/hash.c
```diff
@@ -319,17 +318,10 @@ int verify_bundle_hash(struct manifest *manifest, struct file *bundle)
 				free(filename);
 				break;
 			}
-			free(filename);
 
-			string_or_die(&tar, TAR_COMMAND " -C %s/%i -xf %s/%i/Manifest.%s.tar 2> /dev/null",
-				      state_dir, current->last_change, state_dir,
-				      current->last_change, current->filename);
+			ret = extract_to(filename, state_dir);
```
phmccarty: The manifest tars live in the versioned directories under statedir, not directly under statedir.

src/archives.c
```diff
@@ -0,0 +1,157 @@
...
+	struct archive *a, *ext;
+	struct archive_entry *entry;
+	int flags;
+	int r = 0;
+
+	/* set which attributes we want to restore */
+	flags = ARCHIVE_EXTRACT_TIME;
+	flags |= ARCHIVE_EXTRACT_PERM;
+	flags |= ARCHIVE_EXTRACT_ACL;
+	flags |= ARCHIVE_EXTRACT_FFLAGS;
```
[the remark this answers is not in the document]
matthewrsj: I took the flags the libarchive documentation (on their wiki) provided as an example to preserve file attributes. I'll look at the flags closer and modify.

src/archives.c
```diff
@@ -0,0 +1,157 @@
...
+
+	/* set up read */
+	a = archive_read_new();
+	archive_read_support_format_all(a);
+	archive_read_support_filter_all(a);
+
+	/* set up write */
+	ext = archive_write_disk_new();
+	archive_write_disk_set_options(ext, flags);
+	archive_write_disk_set_standard_lookup(ext);
```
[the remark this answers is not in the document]
matthewrsj: they can technically fail with ARCHIVE_FATAL, so might as well check that since I have no idea how often that can happen.

src/archives.c
```diff
@@ -0,0 +1,157 @@
...
+		fprintf(stderr, "%s\n", archive_error_string(ar));
+	}
+
+	if (ret < ARCHIVE_WARN) {
+		/* error was worse than a warning */
+		return 1;
+	}
+
+	/* just a warning */
+	return 0;
```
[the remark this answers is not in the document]
matthewrsj: Good catch, since `ARCHIVE_RETRY` indicates the operation failed, but is -10 while `ARCHIVE_WARN` is -20, so would not be caught by this.

I think we should just treat `ARCHIVE_RETRY` as an extraction failure rather than trying to figure out how many times we want to continue retrying for each separate operation.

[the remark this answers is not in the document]
phmccarty: That sounds fine to me. Implementing extraction retries based on this status code sounds overkill.

src/download.c
```diff
@@ -293,14 +292,10 @@ int untar_full_download(void *data)
...
-
-	err = system(tarcommand);
-	if (WIFEXITED(err)) {
-		err = WEXITSTATUS(err);
-	}
-	free(tarcommand);
+	char *outputdir;
+	string_or_die(&outputdir, "%s/staged", state_dir);
+	err = extract_to(tarfile, outputdir);
+	free(outputdir);
```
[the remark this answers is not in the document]
matthewrsj: Updated, other comments addressed as well.

src/archives.c
```diff
@@ -0,0 +1,173 @@
...
+	flags |= ARCHIVE_EXTRACT_OWNER;
+	flags |= ARCHIVE_EXTRACT_XATTR;
+
+	/* set security flags */
+	flags |= ARCHIVE_EXTRACT_SECURE_SYMLINKS;
+	flags |= ARCHIVE_EXTRACT_SECURE_NODOTDOT;
+
+	/* set up read */
+	a = archive_read_new();
+	r = archive_read_support_format_all(a);
```
pohly: You can save some CPU cycles by limiting the format to what will actually be present in the archive, i.e. tar.

matthewrsj: ah, you're right. I'll do that.

icarus-sparry: Ah, I thought it was deliberate that you were being flexible.

matthewrsj: the flexible part is `archive_read_support_filter_all`

src/archives.c
```diff
@@ -0,0 +1,173 @@
...
+
+		/* flush pending file attribute changes */
+		r = archive_write_finish_entry(ext);
+		if (_archive_check_err(ext, r)) {
+			goto out;
+		}
+	}
+
+out:
+	archive_write_close(ext);
```
pohly: There's no error handling here. archive_write_close() sets deferred directory attributes, so it can still fail.

matthewrsj: What should we be doing if this fails? Just set `r` and continue trying to free the write and close and free the read? That way the calling function will see the error but we will still be attempting to clean up after ourselves.

pohly:
On Fri, 2017-11-17 at 20:10 +0000, Matthew Johnson wrote
> 
> What should we be doing if this fails? Just set r and continue trying
> to free the write and close and free the read?

Yes. Basically untarring has failed and the caller needs to be alerted
about that.

src/archives.c
```diff
@@ -0,0 +1,180 @@
+/*
+ *   Software Updater - client side
+ *
+ *      Copyright © 2012-2017 Intel Corporation.
```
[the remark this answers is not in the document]
matthewrsj: Fixed

PR #329, opened 2017-11-01
Do not fail on verifytime failure

### LINKED ISSUE #234: versionstamp never created
This file might be created outside the scope of swud-client but in a Yocto/meta-swupd environment I don't see any where that creates the /usr/share/clear/versionstamp file if it doesn't exist. Shouldn't verifytime create this if the file is missing or invalid (it seems to update it if the file exists and has a timestamp).

### LINKED ISSUE #289: verifytime should exit gracefully when versionstamp does not exist
The `verifytime` program is a helper program used by swupd-client that will try to correct certain types of time skews, and it relies on the `/usr/share/clear/versionstamp` for a hint about when the current version was released (in seconds since the epoch).

However, when `/usr/share/clear/versionstamp` does not exist, for whatever reason, I think `verifytime` needs to exit gracefully because (a) it lacks required inputs, and (b) `verifytime` failing is not critical to swupd-client's operations. That is, swupd-client's success/failure should not depend on `verifytime`'s success/failure.

matthewrsj:
Fixes #234
Fixes #289

The verifytime program is a helper program used by swupd to correct
certain time skews. If verifytime fails, however, this should not be a
failure for swupd. Swupd should continue to try and verify the signature
if possible. Instead of failing print out a warning to give the user a
hint to why their signature verification may fail.

The use case this solves is when the /usr/share/clear/versionstamp file
does not exist, verifytime will fail even if the system time is in an OK
state.

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -602,12 +602,11 @@ int swupd_init(int *lock_fd)
 
 	check_root();
 
-	/* Check that our system time is reasonably valid before continuing,
-	 * or the certificate verification will fail with invalid time */
+	/* Check that our system time is reasonably valid, print a warning if not.
+	 * Certificate verification can fail with invalid time. */
 	if (timecheck) {
 		if (system("verifytime") != 0) {
-			ret = EBADTIME;
-			goto out_fds;
+			fprintf(stderr, "Warning: failed to verify system time\n");
 		}
 	}
 


tmarcu:
I'm not sure I agree with this, it has to be a failure to enforce certificate expiration. Letting it fail later on is just prolonging the exit path, assuming the certificate is correct still, which is why this check exists as the first thing in swupd_init() before anything else happens and not wasting time. If it continues and the time mismatches it will never get to verifying the signature because it checks the certificate validity range against the system time.
Maybe something can be changed in verifytime.c, or you can choose a more persistent file to read from?

phmccarty:
The argument here is that `verifytime` should be considered completely optional for swupd-client. Therefore, its exit status is irrelevant for swupd to take action on. If `verifytime` is able to fix the system time, great, report a warning and try to continue.

@tmarcu I'm not sure what you mean by "enforcing certification expiration"... Can you elaborate on that?

tmarcu:
In the general case, with swupd running on Clear, that file will always exist. However, if we are trying to support multiple OSes, then we should implement the more pedantic version of verifytime, as seen in the comment on verifytime.c. The current state was an initial fix that works for Clear, however the intended implementation is to not check any file (like versionstamp) and instead query some nameservers or such to get a real time, and use those times to get an accurate system time. That removes the dependency on any file, and keeps verifytime as the short circuit, which it was intended to be.
With the exception of bundle-list, all swupd commands need network, and if the system time is bad, it is guaranteed to fail signature verification, so we should always exit early if verifytime fails with the new (and more robust) implementation.

matthewrsj:
Closing. #340 opened to track improvements to verifytime.

PR #330, opened 2017-11-01
Fix generation of swupd.bash

icarus-sparry:
When output was changed to stderr for user output, this broke
automatic generation of the bash completion function.

Add suitable redirection to the generation script to capture stderr.

Add test to check the generated file looks reasonable.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,11 +1,11 @@
 SUFFIXES= .rst
 
-EXTRA_DIST = COPYING scripts/findstatic.pl swupd.bash
+EXTRA_DIST = COPYING scripts/findstatic.pl
 
 AM_CFLAGS = -fPIC -Iinclude/ -O2 -g -Wall -W -Wformat-security -D_FORTIFY_SOURCE=2 -fno-common -std=gnu99
 ACLOCAL_AMFLAGS = -I m4
 
-bin_PROGRAMS = swupd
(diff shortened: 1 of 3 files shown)


Makefile.am
```diff
@@ -1,11 +1,11 @@
 SUFFIXES= .rst
 
-EXTRA_DIST = COPYING scripts/findstatic.pl swupd.bash
+EXTRA_DIST = COPYING scripts/findstatic.pl
 
 AM_CFLAGS = -fPIC -Iinclude/ -O2 -g -Wall -W -Wformat-security -D_FORTIFY_SOURCE=2 -fno-common -std=gnu99
 ACLOCAL_AMFLAGS = -I m4
 
-bin_PROGRAMS = swupd
+bin_PROGRAMS = swupd swupd.bash
```
phmccarty: Was moving swupd.bash from EXTRA_DIST to bin_PROGRAMS intentional? I didn't see this mentioned in the commit description.

icarus-sparry: Yes, Otherwise swupd.bash is only made by "make dist", so "make check" fails.
I could have fixed it another way, for example having a requirement on swupd.bash for the test which checks it, but this seemed simplest.

phmccarty: Okay that makes sense now, thanks.

phmccarty:
+1

matthewrsj:
+1

PR #331, opened 2017-11-01
Add --no-scripts flag to skip post-update scripts

### LINKED ISSUE #286: add --no-scripts option to update and verify
Part of https://github.com/clearlinux/swupd-client/issues/249, originally discussed there.

Depending on whether swupd-client is invoked for the currently active filesystem or some inactive one, post-update actions vary. What's currently hard-coded into swupd-client (/usr/bin/systemctl daemon-reexec/reload, /usr/bin/systemctl restart update-triggers.target) is not applicable when updating the inactive partition. Same for run_preupdate_scripts (runs /usr/bin/clr_pre_update.sh).

Given that the current proposal is to use a combination of various swupd operations to achieve the final filesystem update, the helper script needs its own post-update hook mechanism instead of relying on systemd.

What about adding a "--no-scripts" option to "swupd update" and "swupd verify" which suppresses running these additional external helpers?

matthewrsj:
Add the --no-scripts flag to update, verify, and bundle-add commands to
allow users to skip the post update scripts for these commands. This is
useful when running in environments when systemctl is not available such
as containers.

Fixes #286


src/clr_bundle_add.c
```diff
@@ -55,6 +55,7 @@ static void print_help(const char *name)
 	fprintf(stderr, "   -S, --statedir          Specify alternate swupd state directory\n");
 	fprintf(stderr, "   -C, --certpath          Specify alternate path to swupd certificates\n");
 	fprintf(stderr, "   -t, --time              Show verbose time output for swupd operations\n");
+	fprintf(stderr, "   -N, --no-scripts        Do not install boot files to the boot partition (containers)\n");
```
phmccarty: The help string needs updating.

matthewrsj: copy pasta error

matthewrsj: done.

PR #334, opened 2017-11-09
Replace systemctl and pre-update commands

bryteise:
Instead of hardcoding systemctl commands to run post-update triggers,
add the ability to configure an executable path to be used when swupd
completes its update process. Similarly replace the use of a hardcoded
pre-update script path with a configurable executable path.

### DIFF
--- a/src/scripts.c
+++ b/src/scripts.c
@@ -30,6 +30,7 @@
 #include <sys/stat.h>
 #include <unistd.h>
 
+#include "config.h"
 #include "swupd.h"
 
 static void update_boot(void)
@@ -49,37 +50,55 @@ static void update_boot(void)
 
 static void update_triggers(bool block)
 {
-	char *restart_cmd = NULL;
-	char *block_flag = NULL;
+	char *cmd = NULL;
+	char const *block_flag = NULL;
+	char const *reexec_flag = NULL;
 	__attribute__((unused)) int ret = 0;
 
-	/* Check that systemd is working (container case)
-	 * and return if it is not after printing an error */
-	ret = system("/usr/bin/systemctl > /dev/null 2>&1");
-	if (ret != 0) {
-		fprintf(stderr, "WARNING: systemctl not operable, "
+        if (!block) {
+                block_flag = "--no-block";
+        } else {
+                block_flag = "";
+        }
+
+	if (strlen(POST_UPDATE) == 0) {
+		/* fall back to systemd if path prefix is not the rootfs
+		 * and the POST_UPDATE trigger wasn't specified */
+		if (strcmp("/", path_prefix) != 0) {
+			return;
+		}
+
+		ret = system("/usr/bin/systemctl > /dev/null 2>&1");
+		if (ret != 0) {
+			fprintf(stderr, "WARNING: systemctl not operable, "
 				"unable to run systemd update triggers\n");
-		return;
-	}
+			return;
+		}
+		/* These must block so that new update triggers are executed after */
+		if (need_systemd_reexec) {
+			ret = system("/usr/bin/systemctl daemon-reexec");
+		} else {
+			ret = system("/usr/bin/systemctl daemon-reload");
+		}
 
-	/* These must block so that new update triggers are executed after */
-	if (need_systemd_reexec) {
-		ret = system("/usr/bin/systemctl daemon-reexec");
+		string_or_die(&cmd, "/usr/bin/systemctl %s restart update-triggers.target", block_flag);
 	} else {
-		ret = system("/usr/bin/systemctl daemon-reload");
+		/* These must block so that new update triggers are executed after */
+		if (need_systemd_reexec) {
+			reexec_flag = "--reexec";
+		} else {
+			reexec_flag = "";
+		}
+
+		if (strcmp("/", path_prefix) == 0) {
+			string_or_die(&cmd, "%s %s %s", POST_UPDATE, reexec_flag, block_flag);
+		} else {
+			string_or_die(&cmd, "%s/%s %s %s %s", path_prefix, POST_UPDATE, path_prefix, reexec_flag, block_flag);
+		}
 	}
+	ret = system(cmd);
 
-	if (!block) {
-		string_or_die(&block_flag, "--no-block");
-	} else {
-		string_or_die(&block_flag, "");
-	}
-
-	string_or_die(&restart_cmd, "/usr/bin/systemctl %s restart update-triggers.target", block_flag);
-	ret = system(restart_cmd);
-
-	free(restart_cmd);
-	free(block_flag);
+	free(cmd);
 }
 
 void run_scripts(bool block)
@@ -92,7 +111,6 @@ void run_scripts(bool block)
 
 	fprintf(stderr, "Calling post-update helper scripts.\n");
 
-	/* path_prefix aware helper */
 	if (need_update_boot || need_update_bootloader) {
 		if (no_boot_update) {
 			fprintf(stderr, "WARNING: boot files update skipped due to "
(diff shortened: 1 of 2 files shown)


src/scripts.c
```diff
@@ -49,24 +50,20 @@ static void update_boot(void)
...
-		fprintf(stderr, "WARNING: systemctl not operable, "
-				"unable to run systemd update triggers\n");
+	if (strlen(POST_UPDATE) == 0) {
 		return;
 	}
 
 	/* These must block so that new update triggers are executed after */
 	if (need_systemd_reexec) {
-		ret = system("/usr/bin/systemctl daemon-reexec");
+		string_or_die(&reexec_flag, "--rexec");
```
matthewrsj: `--rexec` or `--reexec`?

src/scripts.c
```diff
@@ -75,11 +72,16 @@ static void update_triggers(bool block)
 		string_or_die(&block_flag, "");
 	}
 
-	string_or_die(&restart_cmd, "/usr/bin/systemctl %s restart update-triggers.target", block_flag);
-	ret = system(restart_cmd);
+	if (strcmp("/", path_prefix) == 0) {
+		string_or_die(&cmd, "%s %s %s", POST_UPDATE, reexec_flag, block_flag);
+	} else {
+		string_or_die(&cmd, "%s/%s %s %s %s", path_prefix, POST_UPDATE, path_prefix, reexec_flag, block_flag);
```
matthewrsj: So does this mean the executable must accept the `--rexec` flag to run `daemon-reexec` and the `--no-block` block flag to pass to `systemctl`? It seems like if this is the case it should be pretty well documented. This is placing very specific restrictions on what is supposed to be a configurable script that can be added by whoever builds swupd.

src/scripts.c
```diff
@@ -122,7 +117,17 @@ void run_preupdate_scripts(struct manifest *manifest)
 	char *script;
 	__attribute__((unused)) int ret = 0;
 
-	string_or_die(&script, "/usr/bin/clr_pre_update.sh");
+	if (strlen(PRE_UPDATE) == 0) {
+		return;
+	}
+
+	if (strcmp("/", path_prefix) == 0) {
```
matthewrsj: This took me a minute.

So if there is no `path_prefix` specified (other than root) the pre-update script is expected to run against the current root, yet if there is a `path_prefix` specified the script is expected to live under the `path_prefix` and then also needs that passed as an argument?

Can this be cleaned up so that it either always takes the argument or never does (that might be harder).

If not, the requirements for these should be well-documented somewhere.

src/scripts.c
bryteise: Well it doesn't have to accept these options but the result will be odd in the case of systemd not being reexec'd or slower. This is why the scripts probably need to be in swupd itself in order to properly handle only updating if a file in the mandb changed for instance.

The script itself can of course not take any arguments and do whatever it wants still though.

Maybe there is a way to remove the use of swupd depending on rexec or no block though.

src/scripts.c
bryteise: Well it would be hard for the script to know to run against the prefix if the prefix wasn't passed to it. We could have a heuristic that the script could do to check if it is in /usr/bin to decide if it is prefix mode or not but that seems like a bad idea.

Again the script could just decide to not handle this case at all or do something different and the OSV would need to have a good idea of how they want them to work. For now I'm fine with documenting the options required wherever you think is best. Really though these are all better off being inside of swupd proper as plugins or something.

src/scripts.c
matthewrsj: Well, we need the capability for the script to block in the event that swupd will be re-execing itself over format bumps since we don't want to start another swupd update until the post-update scripts are completed. Could we just always block on post-updates?

src/scripts.c
matthewrsj: Being inside of swupd is the right solution, you are right.

src/scripts.c
```diff
@@ -49,24 +50,20 @@ static void update_boot(void)
 
 static void update_triggers(bool block)
 {
-	char *restart_cmd = NULL;
 	char *block_flag = NULL;
+	char *cmd = NULL;
+	char *reexec_flag = NULL;
```
icarus-sparry: Can we get rid of some of this string_or_die stuff?
Here have 
     char *reexec_flag = "";
on the new line 64 replace
    string_or_die(&reexec_flag, "--reexec");
with
     reexec_flag = "--reexec";
and just don't add the new line 82, the free of reexec_flag. I would remove the '} else {' on line 65 and also line 66, but you could replace line 66 with
    reexec_flag = "";
and leve this line alone.

This is smaller and faster and just as clear.

src/scripts.c
bryteise: That's exactly the kind of requirement that makes me extra sure these should live with swupd and let each distro configure their plugins with the rest of the project.

We can do whatever we want here though, I am fine with however it looks for now.

src/scripts.c
bryteise: Ah --reexec, good catch.

src/scripts.c
bryteise: Ah that's a good cleanup, thanks.

src/scripts.c
matthewrsj: I definitely like the plugin idea since these scripts have to do certain things specific to swupd's behavior. Just having a poorly-defined requirement for the user to provide their own pre- and post-update scripts at a certain path seems difficult for anyone picking up swupd to reason out on their own.

src/scripts.c
icarus-sparry: ditto with block_flag

src/scripts.c
bryteise: Indeed, we should define how a plugin system for swupd would work and then implement it and scrap this change once that is done. I'm inclined to use this as a temporary resolution to our current problem however. Thoughts?

matthewrsj: Hm. Looking at how this is implemented though I'm beginning to see it as a step backward because of the specific requirements placed on the provided scripts. This feels more like removing functionality than solving a problem since it makes it more difficult for someone to take swupd as it exists and stand it up.

I like the decoupling *idea* but I don't like how this makes swupd more difficult to use by someone outside the Clear ecosystem. Am I missing something that makes this change necessary now instead of taking the time to implement a plugin architecture the right way?

matthewrsj: Maybe a better *now* solution would be to keep the hardcoded `systemctl` calls in place as a fallback if the configure options are not set.

icarus-sparry: We can do that by changing  lines 259 and 265 of the configure.ac file.

bryteise: It is adding requirements since there now needs to be a script in place. If a team is already using swupd and have figured out their own way to manage without using systemd they probably are already patching swupd. If they are using systemd and have setup their triggers.target accordingly they'll have to do work for this change. I'm not sure how much of an impact that will be in practice.

It may be better to have this change be moved in as a patch to the Clear Linux package as it is important there in order to properly build and update docker images and also allows populating image state directories removing boot time performance penalties. That way the upstream doesn't introduce a breaking change only to move again once a plugin system is in place.

matthewrsj: It isn't a breaking change if you leave the systemd calls in place as a fallback. Then Clear doesn't have to carry a patch internally that may have to be updated every time a new swupd release is made until the plugin feature is introduced.

src/scripts.c
```diff
@@ -49,37 +50,60 @@ static void update_boot(void)
...
-	if (!block) {
-		string_or_die(&block_flag, "--no-block");
-	} else {
-		string_or_die(&block_flag, "");
-	}
+		if (!block) {
+			block_flag = "--no-block";
+		} else {
+			block_flag = "";
+		}
```
matthewrsj: Since you do this in both the `if` and `else` clauses, how about setting the `block_flag` in one place, before the conditional?

bryteise: Sure. And fix those tabs...

bryteise:
Updated

matthewrsj:
+1

PR #349, opened 2017-12-04
heuristics: do not ignore /usr/src/kernel

### LINKED ISSUE #348: /usr/src/kernel is ignored
Clear Linux* is shipping Linux headers sources in the /usr/src/kernel directory.

Currently that directory is banned in the Heurisitcs is_state() function.

miguelinux:
Linux kernel sources are shipped in the /usr/src/kernel/<kernel_version>
directory. So, it need to be added as an exception in the is_state()
function.

Fixes #348

Debugged-by: Ornelas Aguayo, Jesus <jesus.ornelas.aguayo@intel.com>
Debugged-by: Mario Alfredo Carrillo Arevalo <mario.alfredo.c.arevalo@intel.com>
Signed-off-by: Miguel Bernal Marin <miguel.bernal.marin@linux.intel.com>

### DIFF
--- a/src/heuristics.c
+++ b/src/heuristics.c
@@ -63,6 +63,13 @@ static bool is_state(char *filename)
 		return false;
 	}
 
+	/* Compare the first part of the path, first all the entries inside
+	 * kernel directory, then only the kernel directoy */
+	if ((strncmp(filename, "/usr/src/kernel/", 16) == 0) ||
+	    ((strlen(filename) == 15) && (strncmp(filename, "/usr/src/kernel", 15) == 0))) {
+		return false;
+	}
+
 	if ((strncmp(filename, "/data", 5) == 0) ||
 	    (strncmp(filename, "/dev/", 5) == 0) ||
 	    (strncmp(filename, "/home/", 6) == 0) ||


src/heuristics.c
```diff
@@ -63,6 +63,10 @@ static bool is_state(char *filename)
 		return false;
 	}
 
+	if ((strlen(filename) == 15) && (strncmp(filename, "/usr/src/kernel", 15) == 0)) {
```
matthewrsj: This will *only* return false for the directory, but none of the files under that directory since you are only doing the strncmp for filenames that are exactly 15 characters long.

miguelinux: OK,

So, Do  I need to just compare the first part of the path?

I will send an update.

miguelinux:
@matthewrsj I've updated the patch

Could you review it?  please.

miguelinux:
Well, they are not boot files, they are source files to do some development.

As I understand, the issue is in the server not in the client.

matthewrsj:
Ah, if they are not boot files that is different. Essentially they are just normal files, so you need to do this same change in the server.

This change is still needed in client, but a very similar change is needed in server as well. It would be best to get server up-to-date first (fix inside to outside) in my opinion.

matthewrsj:
+1, I'm opening #354 to track removing heuristic matching from the client.

PR #356, opened 2018-01-03
Fix double free issue

phmccarty:
In commit e18eb69, retrieve_manifests() was refactored to support the mixer integration feature, but it introduced a double free of the "filename" pointer when following certain code paths.

One code path to reproduce the issue is when a Manifest.MoM is not present in the state directory, and swupd_curl_check_network() fails.  A free(filename) was being called immediately before swupd_curl_check_network(), and then again after jumping to the "out" label for the error condition.

Resolve the issue by resetting the filename pointer to NULL after freeing the memory to prevent a later double free. I also reset the url pointer to NULL at the second call site for similar reasoning.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -544,6 +544,7 @@ static int retrieve_manifests(int current, int version, char *component, struct 
 		goto out;
 	}
 	free(filename);
+	filename = NULL;
 
 	if (swupd_curl_check_network()) {
 		ret = -ENOSWUPDSERVER;
@@ -577,7 +578,9 @@ static int retrieve_manifests(int current, int version, char *component, struct 
 			goto untar;
 		}
 		free(filename);
+		filename = NULL;
 		free(url);
+		url = NULL;
 	}
 
 	/* Either we're not on mix or it failed, try curl-ing the file if link didn't work */


src/manifest.c
```diff
@@ -576,10 +576,11 @@ static int retrieve_manifests(int current, int version, char *component, struct
 		if (ret == 0) {
 			goto untar;
 		}
-		free(filename);
```
icarus-sparry: And add a filename=NULL here,

src/manifest.c
```diff
@@ -543,7 +543,6 @@ static int retrieve_manifests(int current, int version, char *component, struct
 		ret = 0;
 		goto out;
 	}
-	free(filename);
```
icarus-sparry: With respect, I think this is the wrong fix. Removing this line means a memory leak at about line 561.

I think the correct fix is to add a line so it says

> free(filename);
> filename=NULL;

Of course what we really want is to scrap the string_or_die function, or to add a 

> void free_string(char **s)
> {
>     if (s) {
>        free(*s);
>         *s = NULL;
> }

and use this string_free(&filename) instead of free(filename).

phmccarty: Sure, I'll amend this change to reset to NULL instead. Opened #357 regarding a new string_free() function.

PR #361, opened 2018-01-11
Fix some memory leaks

phmccarty:
See the individual commits...

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -271,11 +271,12 @@ int get_version_from_path(const char *abs_path)
 static int set_default_value_from_path(char **global, const char *path)
 {
 	int ret = -1;
-	char *ret_str;
+	char *ret_str = NULL;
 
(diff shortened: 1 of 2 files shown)


src/globals.c
```diff
@@ -271,11 +271,12 @@ int get_version_from_path(const char *abs_path)
 static int set_default_value_from_path(char **global, const char *path)
 {
 	int ret = -1;
-	char *ret_str;
+	char *ret_str = NULL;
 
 	ret = get_value_from_path(&ret_str, path, false);
-	if (ret == 0) {
+	if (ret == 0 && ret_str) {
 		string_or_die(global, "%s", ret_str);
```
icarus-sparry: I mention for completeness that one could take advantage of knowing the implementation and instead say

    if (ret == 0) {
        global = ret_str;
    }
    ret_str = NULL ;

which would be faster. However since this is not in a tight loop, not worth changing.

PR #362, opened 2018-01-12
Fix warning message in archives

cmarcelo:
Need to pass the same struct used in the operation to the
_archive_check_err helper.

This fixes issue of some Warnings appearing with "(null)" message,
since the writer did not had an error string associated.

Signed-off-by: Caio Marcelo de Oliveira Filho <caio.oliveira@intel.com>


matthewrsj:
Good catch! +1

tmarcu:
+1

PR #366, opened 2018-01-18
Add free_string function and use everywhere

### LINKED ISSUE #357: Consider adding a string_free() function
So that the code is consistent about resetting a pointer to NULL after freeing dynamic memory allocated at that location, consider adding a `string_free()` wrapper function. For an implementation idea, see the code snippet below (from #356).

```
void free_string(char **s)
{
    if (s) {
       free(*s);
        *s = NULL;
    }
}
```

matthewrsj:
Fixes #357
Add free_string() function to insure code is consistently resetting
pointers to NULL after resetting dynamic memory. Use this helper
everywhere strings are being freed.

### DIFF
--- a/src/hashdump.c
+++ b/src/hashdump.c
@@ -112,7 +112,7 @@ int hashdump_main(int argc, char **argv)
 
 	ret = set_path_prefix(NULL);
 	if (!ret) {
-		free(file->filename);
+		free_string(&file->filename);
 		free(file);
 		return EXIT_FAILURE;
 	}
@@ -143,8 +143,8 @@ int hashdump_main(int argc, char **argv)
 		}
 	}
 
-	free(fullname);
-	free(file->filename);
+	free_string(&fullname);
+	free_string(&file->filename);
 	free(file);
 	return 0;
 }
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -67,7 +67,7 @@ int rm_staging_dir_contents(const char *rel_path)
 
 	dir = opendir(abs_path);
 	if (dir == NULL) {
-		free(abs_path);
+		free_string(&abs_path);
 		return -1;
 	}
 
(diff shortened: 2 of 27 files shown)


src/hashdump.c
```diff
@@ -112,7 +112,7 @@ int hashdump_main(int argc, char **argv)
 
 	ret = set_path_prefix(NULL);
 	if (!ret) {
-		free(file->filename);
+		free_string(&file->filename);
```
icarus-sparry: This change not actually needed, but it is good for consistency

src/helpers.c
```diff
@@ -671,6 +660,14 @@ void string_or_die(char **strp, const char *fmt, ...)
 	va_end(ap);
 }
 
+void free_string(char **s)
```
icarus-sparry: The reason for checking that s is not null is just to stop cores. It is a programming error if s is null. So it might be worth declaring this as [nonnull](https://gcc.gnu.org/onlinedocs/gcc-4.7.0/gcc/Function-Attributes.html#index-g_t_0040code_007bnonnull_007d-function-attribute-2555)

src/hashdump.c
matthewrsj: There are several changes that were not actually needed, but I tried to convert all `free(some_string)` to `free_string(some_string)`, like you said, for consistency.

src/helpers.c
matthewrsj: And remove the check in free_string()?

icarus-sparry: No, they serve different purposes. The check in free_string is to avoid a runtime crash. The annotation enables a compile time warning (which we turn into a compile time error).

The warning probably will never be triggered, but defense in depth is not a bad thing.

matthewrsj: OK, but when I add nonnull the compiler complains that free_string itself is comparing a nonnull value to null.

matthewrsj: +1 for option 1, this is good to go then?

icarus-sparry: Yes

matthewrsj:
Found some additional comparisons and assignments to NULL that could be removed. This should be good to go now.

matthewrsj:
Merging this...

PR #368, opened 2018-01-25
Add info command to show OS version and URLs

### LINKED ISSUE #218: provide a comand to show update URL swupd is consuming
Currently, there is no other way for users to know the URL swupd consumes for update but by looking at `/usr/share/defaults/swupd/contenturl`.

If a user creates a mix with his/her own content, the update URL for swupd-client would differ from default clearlinux URL.

would be really helpful for those who run a mix to have a command to check the content URL swupd is consuming. e.g:

```
# swupd info contenturl
```
it could show both `contenturl` and `versionurl`.

cmarcelo:
Fixes #218.

Signed-off-by: Caio Marcelo de Oliveira Filho <caio.oliveira@intel.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -17,7 +17,8 @@ swupd_SOURCES = \
 	$(clr_bundle_rm_SOURCES) \
 	$(clr_bundle_ls_SOURCES) \
 	$(swupd_autoupdate_SOURCES) \
-	$(swupd_search_SOURCES)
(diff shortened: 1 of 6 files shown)


cmarcelo:
@matthewrsj this is not exactly what the issue asked for (info content_url), but similar because shows all the relevant information.

It helps a little bit when using mixes.

Issue #359, opened 2018-01-03
Detect when running in a container to skip CBM updates


matthewrsj:
The -b option was originally added to avoid having to detect whether or not the process was running in a container. It is trivial to check (for docker) by checking for the existence of `/.dockerenv` or for lines containing `/docker/` in `/proc/1/cgroup`. Other container types could be checked there as well.

Originally reported to https://github.com/clearlinux/clr-bundles/issues/31

tpepper:
In a good container you might not have those easy hints.  But even without the hints, the CBM run should fail more gracefully.

cmarcelo:
We could use `systemd-detect-virt -c` or see what it does and do similar.

matthewrsj:
Fixed by #374

tpepper:
Systemd's not universally available inside containers.  But I suppose this
is currently "clearlinux/swupd-client" so...ok?

bryteise:
Yea, I think updating kernels is going to always be super platform dependent. We likely eventually have a script that is called rather than cbm directly so that different OSs can make different decisions about bootloader and kernel update mechanics.

PR #374, opened 2018-02-01
Make a container check before running clr-boot-manager

bryteise:
Add a check to see if an update is being run in a container (and is
updating the rootfs) using systemd-detect-virt -c. If the update is
being run in a container skip running the clr-boot-manager update
since it will fail.

### DIFF
--- a/src/scripts.c
+++ b/src/scripts.c
@@ -33,10 +33,28 @@
 #include "config.h"
 #include "swupd.h"
 
+static bool in_container(void)
+{
+        char *detect_container_cmd = NULL;
+        bool ret = false;
+        /* systemd-detect-virt -c does container detection only *
+         * The return code is zero if the system is in a container */
+        string_or_die(&detect_container_cmd, "/usr/bin/systemd-detect-virt -c");
+        ret = !system(detect_container_cmd);
+        free_string(&detect_container_cmd);
+
+        return ret;
+}
+
 static void update_boot(void)
 {
 	char *boot_update_cmd = NULL;
 	__attribute__((unused)) int ret = 0;
+
+        /* Don't run clr-boot-manager update in a container on the rootfs */
+        if (strcmp("/", path_prefix) == 0 && in_container()) {
+                return;
+        }
 
 	if (strcmp("/", path_prefix) == 0) {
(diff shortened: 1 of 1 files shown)


matthewrsj:
Fixes #359

src/scripts.c
```diff
@@ -33,11 +33,29 @@
 #include "config.h"
 #include "swupd.h"
 
+static bool in_container(void)
```
icarus-sparry: I would just have changed the
"%s/usr/bin/clr-boot-manager update --path %s"

in update_boot to

"/usr/bin/systemd-detect-virt -c || %s/usr/bin/clr-boot-manager update --path %s"

rather than write this new function.

matthewrsj: Why? The new function makes it easier to see what is happening and why we return early.

icarus-sparry: 1. Easier to see why it is happening is subjective. If you are looking at the output of strace to try and figure out why something is not updating, then seeing `systemd-detect-virt || clr-boot-manager` is a lot easier to reason about than the typical hundreds of lines that process invocation causes.

1. Changing the string is a smaller change.

1. Eventually when it is decided that this kind of policy decision is moved outside the swupd-client to a supporting program (There is already a good argument to be made that this kind of decision is non of swupd's business) then we will only need to change the code in one place.

Random thoughts on what might be needed in the future. A container might be being used to provision a new disk, and so have access to both the EFI and root partitions. With this change we can no longer use swupd to do so, we have to run a clr-boot-manager update as another step.

 Maybe in the future something like clear containers will respond 'true' to being run in a container but still want to be able to update its kernel.

The change is fine from the point of view of code, it is a small function (Although I would have made it a lot smaller by not copying a constant string to a malloced one, using the malloced one, and then freeing it), well named etc.

bryteise: I'd expect the provision to a new disk to use a different root so this change should still allow swupd to call clr-boot-manager in that case.

I just took container detection as something that may move off to a utility at some point in time so I split it into a function. Dropping the string allocation is definitely the right move though, I had briefly considered using the path prefix'd version before I came to my senses =P.

PR #376, opened 2018-02-09
Fix invalid call to stat when FALLBACK_CAPATHS is empty

cmarcelo:
Loop wasn't taking into account the constant being set but empty. In
that case tok was set to NULL and stat was called with it. Replace
'do-while' with a 'while' to also perform the check the first time.

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -94,8 +94,7 @@ int swupd_curl_init(void)
 		fallback_capaths[0] = NULL;
 		i = 1;
 
-		tok = strtok_r(str, ":", &ctx);
-		do {
+		for (tok = strtok_r(str, ":", &ctx); tok; tok = strtok_r(NULL, ":", &ctx)) {
 			if (stat(tok, &st)) {
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


tmarcu:
+1

src/curl.c
```diff
@@ -95,7 +95,7 @@ int swupd_curl_init(void)
 		i = 1;
 
 		tok = strtok_r(str, ":", &ctx);
-		do {
+	        while (tok) {
```
matthewrsj: Please run `make compliant` to make the whitespace here consistent with the rest of the project.

cmarcelo: Done.

src/curl.c
```diff
@@ -95,7 +95,7 @@ int swupd_curl_init(void)
 		i = 1;
 
 		tok = strtok_r(str, ":", &ctx);
-		do {
+		while (tok) {
 			if (stat(tok, &st)) {
```
icarus-sparry: Don't you just hate infinite loops?

Consider what happens if stat fails!

Please can we fix this code so it is idiomatic, i.e. replace 98 with

    for ( tok = strtok_r(str, ":", &ctx); tok ; tock = strtok_r(NULL, ":", &ctx) {

and remove 97 and 111

cmarcelo: Thanks for catching this :-)  Tested with valid paths but without invalid ones.

Done.

PR #379, opened 2018-02-12
Apply all possible deltas from the pack, not only renames

cmarcelo:
Consider all deltas from a pack instead of only those who are involved
in a rename. This will allow the server to make better packs
regardless of the from/to version, with the possibility of a delta to
a file that was not necessarily marked with the rename flag.

This makes deltapeer member of file unused, so it was removed. The
peer member continues to be used for its original role: mapping the
same files (same filename) between two manifests, usually the state of
the local system and some new version to be installed.


cmarcelo:
@matthewrsj @icarus-sparry ready for a review pass.

I'm still working in: (a) adding one extra functional test   (b) check if we can simplify consolidate_files logic to ignore is_rename, so we can stop reading it. The latter likely will end up in a patch of its own.

cmarcelo:
Pushed a new version with a second commit that removes is_rename and make simplifications in consolidate_files.

@phmccarty could you take a look in these two commit messages? I've looked into past work (even before github) to try to figure out the reasoning behind some code that is being removed.

src/delta.c
```diff
@@ -38,86 +38,164 @@
...
-	base = basename(tmp2);
+	xattrs_copy(prefixed_from_file, to_staged);
 
-	string_or_die(&origin, "%s/%s/%s", path_prefix, dir, base);
-
-	ret = apply_bsdiff_delta(origin, filename, deltafile);
-	if (ret) {
-		unlink_all_staged_content(file);
-		goto out;
+	/* TODO: Remove this and rely only on hash verification only? */
```
matthewrsj: minor: you duplicate `only` in this comment

src/delta.c
```diff
@@ -38,86 +38,164 @@
...
+		goto remove_and_fail;
 	}
-	xattrs_copy(origin, filename);
 
-	if (!verify_file(file, filename)) {
-		unlink_all_staged_content(file);
-		goto out;
+	char hash[SWUPD_HASH_LEN];
+	if (!compute_hash_from_file(to_staged, &hash[0])) {
+		fprintf(stderr, "hash calculation failed\n");
```
matthewrsj: Can we improve the error message here? This doesn't indicate what step failed (delta application) or on which file it failed.

src/delta.c
```diff
@@ -38,86 +38,164 @@
...
+		goto remove_and_fail;
 	}
+	if (!hash_equal(hash, to_hash)) {
+		fprintf(stderr, "hash mismatch\n");
```
matthewrsj: Please improve error message here as well.

src/delta.c
```diff
@@ -38,86 +38,164 @@
...
+			fprintf(stderr, "Skipping invalid delta file %s/%s\n", delta_dir, delta_name);
+			continue;
+		}
+		s++;
+
+		/* Note: SWUPD_HASH_LEN accounts for the nul-terminator after the hash. */
+
+		/* Two hashes (without zero) plus one dash in the middle. */
+		if (strlen(s) != ((2 * (SWUPD_HASH_LEN - 1)) + 1)) {
+			fprintf(stderr, "Skipping invalid delta file %s/%s\n", delta_dir, delta_name);
```
matthewrsj: Remove the invalid delta file as well?

src/delta.c
```diff
@@ -38,86 +38,164 @@
...
+			continue;
+		}
+
+		if (s[SWUPD_HASH_LEN - 1] != '-') {
+			fprintf(stderr, "Skipping invalid delta file %s/%s\n", delta_dir, delta_name);
```
matthewrsj: Remove here as well

src/delta.c
```diff
@@ -38,86 +38,164 @@
...
+		char *to_staged;
+		string_or_die(&to_staged, "%s/staged/%s", state_dir, to);
+
+		struct stat stat;
+		if (lstat(to_staged, &stat) == 0) {
+			free_string(&to_staged);
+			continue;
+		}
+
+		/* TODO: Sort the list of files by hash then walk it in parallel with readdir. */
```
matthewrsj: `file_sort_hash()` exists in manifest.c if you want to do this now. You can do
```
list_sort(current_manifest->files, file_sort_hash)
```

You don't need to do this now though.

src/delta.c
```diff
@@ -38,86 +38,164 @@
...
+			ll = ll->next;
+
+			if (!file->is_deleted && !file->is_ghosted && file->is_file && hash_equal(file->hash, from)) {
+				found = file;
+				break;
+			}
+		}
+
+		if (!found) {
+			fprintf(stderr, "Skipping delta file %s/%s because the 'from' file for delta was found\n", delta_dir, delta_name);
```
matthewrsj: You mean the 'from' file for delta was **not** found?

src/delta.c
```diff
@@ -38,86 +38,164 @@
...
+
+		char *delta_file;
+		string_or_die(&delta_file, "%s/%s", delta_dir, delta_name);
+
+		if (!apply_one_delta(found->filename, to_staged, delta_file, to)) {
+			fprintf(stderr, "Failed to apply delta file %s\n", delta_file);
+		}
+
+		/* Always unlink delta files. Once applied the staged
+		 * file is there so fullfile not needed anymore. */
```
matthewrsj: You mean the delta file is not needed any more.

src/manifest.c
```diff
@@ -340,7 +281,7 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
 		}
 
 		if (c[3] == 'r') {
-			file->is_rename = 1;
+			/* rename flag is ignored */
```
matthewrsj: 👏

src/manifest.c
```diff
@@ -1075,34 +1014,8 @@ struct list *consolidate_files(struct list *files)
...
-	 *       -: Don't Care   - choose/remove either file
-	 *       X: Error State  - remove both files, LOG error
-	 *       1: choose file1 - remove file2
-	 *       2: choose file2 - remove file1
-	 *
-	 * NOTE: the code below could be rewritten to be more "efficient", but clarity
-	 *       and concreteness here are of utmost importance if we are to correctly
-	 *       maintain the installed system's state in the filesystem across updates
+	 * there are Manifest invariants to maintain.
+         * Note that "file" may be a file, directory or symlink.
```
matthewrsj: Please follow project convention of using tabs.

src/delta.c
cmarcelo: Done.

src/delta.c
cmarcelo: Done.

src/delta.c
cmarcelo: Done.

src/delta.c
cmarcelo: "Disaster prevention" flag: If we don't recognize the file, we don't remove it. Also let it around for further investigation (the deltas we know we can find in the server).

src/delta.c
cmarcelo: Ditto.

src/delta.c
cmarcelo: Done.

src/delta.c
cmarcelo: Done.

src/manifest.c
cmarcelo: `make compliant` did not complained about this one for some reason. Fixed.

src/delta.c
matthewrsj: What do you mean? It is an invalid delta file in the delta directory. afaict it won't get cleaned up and will just sit there. I have no idea why I would want to inspect that. This is the swupd state dir. If it can't do anything with the file the file should be removed. It similarly removes files with bad hashes in the staged dir.

cmarcelo:
Pushed new version. Changes:

- Per request, removing every file in the directory.
- Moved the "is this a good filename" logic to a function.
- Using a local buffer for both from and to, make it easier to understand. The impact should be negligible.

src/delta.c
```diff
@@ -38,86 +38,165 @@
...
-	free_string(&filename);
+	if (!hash_equal(hash, to_hash)) {
+		fprintf(stderr, "Couldn't use delta file %s: application resulted in wrong hash\n", delta_file);
+		goto remove_and_fail;
+	}
+
+	goto end;
+
+remove_and_fail:
+	(void)remove(to_staged);
```
matthewrsj: Why not remove the delta file here too?

matthewrsj: Ah, because it is a correctly-named delta at this point. +1.

cmarcelo:
> can you add a functional test?

Done.

matthewrsj:
Ugh, sorry about the merge commit. I tried github's "update branch" button.

cmarcelo:
Rebased and repushed.

matthewrsj:
Thanks @cmarcelo, sorry about that.

matthewrsj:
@phmccarty ping

matthewrsj:
@tmarcu @icarus-sparry looking for one more review before merging this.

matthewrsj:
Ping again. Would still like an additional review for this big change.

src/delta.c
```diff
@@ -38,86 +38,165 @@
...
+	const char *s = delta_name;
 
-	string_or_die(&deltafile, "%s/delta/%i-%i-%s-%s", state_dir,
-		      file->deltapeer->last_change, file->last_change, file->deltapeer->hash, file->hash);
-
-	ret = stat(deltafile, &sb);
-	if (ret != 0) {
-		free_string(&deltafile);
-		return;
+	s = index(s, '-');
```
tmarcu: First time I've seen index() used! From the man pages and several other locations:

> POSIX.1-2008  removes  the
>        specifications of index() and rindex(), recommending strchr(3) and str‐
>        rchr(3) instead.
> 

AFAIK this is not part of the C standard anymore, either way we should probably update it.

src/delta.c
```diff
@@ -38,86 +38,165 @@
...
-	dir = dirname(tmp);
-	base = basename(tmp2);
+	s = index(s, '-');
+	if (!s) {
+		return false;
+	}
+	s++;
 
-	string_or_die(&origin, "%s/%s/%s", path_prefix, dir, base);
+	/* Note: SWUPD_HASH_LEN accounts for the nul-terminator after the hash. */
```
tmarcu: s/nul/null

src/delta.c
```diff
@@ -38,86 +38,165 @@
...
 
-	ret = apply_bsdiff_delta(origin, filename, deltafile);
-	if (ret) {
-		unlink_all_staged_content(file);
-		goto out;
+	/* Two hashes (without zero) plus one dash in the middle. */
+	if (strlen(s) != ((2 * (SWUPD_HASH_LEN - 1)) + 1)) {
```
tmarcu: I still hate that we have to add a -1 EVERYWHERE we use this variable, if 99% of the use case is wanting it without a null terminator it should be the size we use in the #define IMO.

icarus-sparry:
NUL is the ASCII character represented by '\0'.

NULL is a value of 0, used in pointer comparisons.

src/delta.c
phmccarty: I see quite a few references to `SWUPD_HASH_LEN` is this commit without the -1 though. In this function, it's needed often though, so maybe just define/use a variable set to `SWUPD_HASH_LEN - 1`?

src/manifest.c
```diff
@@ -1113,21 +1026,39 @@ struct list *consolidate_files(struct list *files)
...
-		if (!file1->is_deleted && !file2->is_deleted && hash_equal(file1->hash, file2->hash)) {
-			/* always drop the untracked file if there is a tracked file */
+		/* If both files are present, this is the most common case. */
+		if (!file1->is_deleted && !file2->is_deleted) {
+
+			/* If the hashes don't match, we have an inconsistency in the
+			 * manifest. Exclude both from the list. */
+			if (!hash_equal(file1->hash, file2->hash)) {
+
+				/* TODO: report? */
```
phmccarty: There is a `telemetry()` function to report issues like this. Might be useful here.

src/delta.c
```diff
@@ -38,86 +38,165 @@
...
+	if (ret) {
+		fprintf(stderr, "Couldn't use delta file %s: bsdiff failed\n", delta_file);
+		goto end;
 	}
 
-	if (file->deltapeer->is_deleted) {
-		return;
+	xattrs_copy(prefixed_from_file, to_staged);
+
+	/* TODO: Remove this and rely on hash verification only? */
```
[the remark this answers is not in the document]
cmarcelo: The check mentioned in TODO already happens in the lines below, so just removing the 73-77.

src/delta.c
```diff
@@ -38,86 +38,165 @@
...
-	char *filename;
-	struct stat stat;
+	/* TODO: implement this without the indirection of creating a file... */
+	struct file f;
 
-	if (file->is_file == 0) {
-		return;
-	}
+	f.filename = filename;
+	f.use_xattrs = true;
```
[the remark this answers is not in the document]
cmarcelo: If I recall correctly, I've pulled this logic from verify_file. The goal was to perform the same hash calculation as a verify would.

src/delta.c
```diff
@@ -38,86 +38,165 @@
...
+static void apply_one_delta(char *from_file, char *to_staged, char *delta_file, char *to_hash)
+{
+	int ret;
+
+	char *prefixed_from_file;
+	string_or_die(&prefixed_from_file, "%s/%s", path_prefix, from_file);
+
+	ret = apply_bsdiff_delta(prefixed_from_file, to_staged, delta_file);
+	if (ret) {
+		fprintf(stderr, "Couldn't use delta file %s: bsdiff failed\n", delta_file);
```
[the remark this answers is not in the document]
cmarcelo: Done.

src/delta.c
cmarcelo: Changed to strchr.

src/manifest.c
cmarcelo: Done.

src/delta.c
cmarcelo: Rewrote the function with the extra variable, plus a few changes to code and comments.

src/delta.c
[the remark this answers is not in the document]
matthewrsj: I was curious about this too when I reviewed so I looked at other hash calculations in the code to make sure it was done the same. I'm ok with setting use_xattrs to be true even though we don't expect any at least for consistency. We can remove that assignment in the future after some testing if it is necessary.

Also, didn't yocto/meta-swupd use xattrs? So we need to use_xattrs for other use-cases. If none exist it shouldn't change our calculation anyways.

matthewrsj:
looks like @cmarcelo addressed comments, so you guys can do another review pass. (@phmccarty @tmarcu @icarus-sparry)

src/delta.c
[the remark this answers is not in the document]
phmccarty: There is a configure-time option (--enable-xattrs) to enable xattr support. It defaults to disabled...

test/functional/update/delta-without-rename-flag/test.bats
```diff
@@ -0,0 +1,40 @@
...
+  echo "CMD: $SWUPD update $SWUPD_OPTS"
+}
+
+teardown() {
+  clean_tars 10
+  clean_tars 20
+  revert_chown_root -R "$DIR/web-dir"
+}
+
+@test "update simple rename" {
```
phmccarty: Can you revise this test docstring to reflect the test purpose?

test/functional/update/delta-without-rename-flag/test.bats
```diff
@@ -0,0 +1,40 @@
...
+  unpack_target_from_manifest 10 MoM
+  unpack_target_from_manifest 10 os-core
+  sudo tar -C "$DIR/web-dir/20" -cf "$DIR/web-dir/20/pack-os-core-from-10.tar" delta/ staged/
+  echo "CMD: $SWUPD update $SWUPD_OPTS"
+}
+
+teardown() {
+  clean_tars 10
+  clean_tars 20
+  revert_chown_root -R "$DIR/web-dir"
```
phmccarty: You will want to add `"$DIR/target-dir"` as an additional argument to this line, since it is chowned to root in setup().

test/functional/update/delta-without-rename-flag/test.bats
```diff
@@ -0,0 +1,40 @@
...
+}
+
+@test "update simple rename" {
+  run sudo sh -c "$SWUPD update $SWUPD_OPTS"
```
phmccarty: Check exit status here to ensure the update succeeded.

test/functional/update/delta-without-rename-flag/test.bats
```diff
@@ -0,0 +1,40 @@
...
+  create_manifest_tar 10 os-core
+  create_manifest_tar 20 MoM
+  sign_manifest_mom 20
+  create_manifest_tar 20 os-core
+  set_os_release 10
+  chown_root -R "$DIR/target-dir" "$DIR/web-dir"
+  unpack_target_from_manifest 10 MoM
+  unpack_target_from_manifest 10 os-core
+  sudo tar -C "$DIR/web-dir/20" -cf "$DIR/web-dir/20/pack-os-core-from-10.tar" delta/ staged/
+  echo "CMD: $SWUPD update $SWUPD_OPTS"
```
phmccarty: This line is not needed.

src/manifest.c
```diff
@@ -1146,21 +1059,49 @@ struct list *consolidate_files(struct list *files)
...
+		if (!file1->is_deleted && !file2->is_deleted) {
+
+			/* If the hashes don't match, we have an inconsistency in the
+			 * manifest. Exclude both from the list. */
+			if (!hash_equal(file1->hash, file2->hash)) {
+				tmp = next->next;
+				list_free_item(list, NULL);
+				list_free_item(next, NULL);
+				list = tmp;
+				telemetry(TELEMETRY_INFO,
```
phmccarty: I think a `TELEMETRY_WARN` or `TELEMETRY_CRIT` severity would be more appropriate, since this condition would reflect a serious bug in manifest creation.

cmarcelo: Changed to CRIT.

test/functional/update/delta-without-rename-flag/test.bats
cmarcelo: Removed.

test/functional/update/delta-without-rename-flag/test.bats
cmarcelo: Done.

test/functional/update/delta-without-rename-flag/test.bats
cmarcelo: Done.

test/functional/update/delta-without-rename-flag/test.bats
cmarcelo: Changed to: "using delta without a rename flag in the manifest".

src/delta.c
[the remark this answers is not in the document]
cmarcelo: The configure time option controls whether the querying for xattrs is real or just give dummy answers ("no xattrs!"). See xattrs.c:41. Setting use_xattrs won't cause reading actual xattrs unless we have it enabled.

It might be worthwhile to revisit how xattrs is enabled/disabled, but for this patch doing the same thing as verify_file is the right call.

src/delta.c
```diff
@@ -38,86 +38,159 @@
...
+		if (lstat(to_staged, &stat) == 0) {
+			goto next;
+		}
+
+		/* TODO: Sort the list of files by hash then walk it in parallel with readdir. */
+
+		struct list *ll = list_head(current_manifest->files);
+
+		struct file *found = NULL;
+		while (ll) {
```
icarus-sparry: This can be simplified, 15 lines -> 12

		for (;ll;ll=ll->next) {
			struct file *file = ll->data;

			if (!file->is_deleted && !file->is_ghosted && file->is_file && hash_equal(file->hash, from)) {
		                apply_one_delta(file->filename, to_staged, delta_file, to);
				break;
			}
		}

		if (!ll) {
			fprintf(stderr, "Couldn't use delta file %s: no 'from' file to apply was found\n", delta_file);
		}

This code also misses the opportunity if there are multiple old files which match the hash, some of which are removed, to use one of the non-removed ones.

src/delta.c
```diff
@@ -38,86 +38,159 @@
...
+
+		char *delta_name = ent->d_name;
+		char *delta_file;
+		string_or_die(&delta_file, "%s/%s", delta_dir, delta_name);
+
+		char from[SWUPD_HASH_LEN] = {};
+		char to[SWUPD_HASH_LEN] = {};
+		if (!check_delta_filename(delta_name, &from[0], &to[0])) {
+			fprintf(stderr, "Invalid name for delta file: %s\n", delta_file);
+			goto next;
```
icarus-sparry: So this removes everything from delta, even if it doesn't look like a delta.

If we want to do multiple bundle processing in parallel then we need to have something like multiple delta directories, one per bundle.

src/delta.c
```diff
@@ -38,86 +38,159 @@
...
+		string_or_die(&to_staged, "%s/staged/%s", state_dir, to);
+
+		/* If 'to' file already exists, no need to apply delta. */
+		struct stat stat;
+		if (lstat(to_staged, &stat) == 0) {
+			goto next;
+		}
+
+		/* TODO: Sort the list of files by hash then walk it in parallel with readdir. */
+
```
icarus-sparry: This will reduce the number of hash compares we do a lot.

src/delta.c
```diff
@@ -38,86 +38,159 @@
...
-	if (file->deltapeer->is_deleted) {
-		return;
+static void apply_one_delta(char *from_file, char *to_staged, char *delta_file, char *to_hash)
+{
+	int ret;
+
+	char *prefixed_from_file;
+	string_or_die(&prefixed_from_file, "%s/%s", path_prefix, from_file);
+
+	ret = apply_bsdiff_delta(prefixed_from_file, to_staged, delta_file);
```
icarus-sparry: Should we check the hash of the prefixed_from_file here?
Better still should we copy the prefixed from file to a secure location and then check the copy?
Otherwise we open ourselves up to the file being changed either before or during the update, which might put bsdiff into an infinite loop.

src/delta.c
```diff
@@ -38,86 +38,159 @@
...
-	if (ret != 0) {
-		free_string(&deltafile);
-		return;
+	if (s[hash_len] != '-') {
+		return false;
 	}
 
-	tmp = strdup(file->deltapeer->filename);
-	tmp2 = strdup(file->deltapeer->filename);
+	memcpy(from, s, hash_len);
```
icarus-sparry: Probably ought to be using hash_assign here rather than memcpy.

cmarcelo: Done.

src/delta.c
cmarcelo: The original code was more conservative about deleting files, but previous reviews asked for more aggressive cleaning up.

As for multiple bundle processing, when we cross that bridge we see how it is the best fix. E.g. one can walk only through the deltas names in the bundle pack, instead of all files in the directory.

src/delta.c
cmarcelo: Changed to use the for loop, but kept the found variable. We'd have to cast it anyway, potentially eating another line.

> This code also misses the opportunity if there are multiple old files which match the hash, some of which are removed, to use one of the non-removed ones.

> This code also misses the opportunity if there are multiple old files which match the hash, some of which are removed, to use one of the non-removed ones.

In the current code there's no difference between which one we pick, since we are processing these deltas before the removal happens. What kind of opportunity are you thinking about?

src/delta.c
cmarcelo: Agreed (hence the TODO). I've opted to get the version we have in first, as it introduces a change in how we pick the deltas. Once that is in, the walk optimization should be a straightforward patch.

src/delta.c
icarus-sparry: I am thinking of something like

/bin/mail and /bin/rmail are the same file, but someone has moved /bin/mail to /bin/mail.old

src/delta.c
cmarcelo: I've kept the behavior of previous delta application code. I think some of these improvements might make sense, but I think they are independent of the main goal here, which is change how we pick what deltas to apply.

> Otherwise we open ourselves up to the file being changed either before or during the update, which might put bsdiff into an infinite loop.

I did not understand how bsdiff could be put in a loop -- but I don't really know its internals. If the a wrong file is given, then it just fail and we fallback to the fullfile.

icarus-sparry: It is only a guess that it can be put into a loop.

Many of these kinds of 'take an input stream and a set of edits to produce an output stream' can have issues, e.g. copy from the last line starting 'A' to here can be affected if you change which that last line happens to be.

src/delta.c
cmarcelo: Got it.

Mixing this and the part of the other comment: added a check of the real filename hash before consider a file "found", this covers both the real file being delete or modified before the delta application starts.

cmarcelo:
Rebased.

@tmarcu is there any other change you want suggest for this PR? (Asking because GitHub says you have outstanding requested changes :-)

cmarcelo:
Ping

src/delta.c
matthewrsj: This is an important TODO. Not going to block the merge, just noting that the TODO is important...

test/functional/update/delta-without-rename-flag/test.bats
```diff
@@ -0,0 +1,41 @@
...
+
+teardown() {
+  clean_tars 10
+  clean_tars 20
+  revert_chown_root -R "$DIR/target-dir"
+  revert_chown_root -R "$DIR/web-dir"
+}
+
+@test "using delta without a rename flag in the manifest" {
+  run sudo sh -c "$SWUPD update $SWUPD_OPTS"
```
matthewrsj: How are you sure this isn't falling back to a fullfile download in this test?

matthewrsj: I don't see anywhere in the setup where you are removing the fallback, which is how that assurance is usually met.

src/delta.c
```diff
@@ -38,86 +38,157 @@
...
+			if (!compute_hash_from_file(filename, hash) || !hash_equal(file->hash, hash)) {
+				free_string(&filename);
+				continue;
+			}
+
+			found = filename;
+		}
+
+		if (!found) {
+			fprintf(stderr, "Couldn't use delta file %s: no 'from' file to apply was found\n", delta_file);
```
matthewrsj: Let's not print these warnings. The user can't do anything about them. If anything this would be something to report to telemetry, but that again would make a bunch of useless telemetry records. I'd rather just not print this kind of thing.

matthewrsj: This indicated a `/var/lib/swupd` on my system that had a lot of extra files hanging around. Running a `swupd clean --all` made the error go away of course, but I still don't think the prints are needed.

matthewrsj:
Did a quick test to see if performance was impacted too much by this change.

Procedure:
```
# system on 21810
# 21 bundles update from 21810 to 21820 (on my system)
sudo swupd update  # system swupd update to 21820
sudo swupd verify --fix --picky -m 21810
sudo swupd clean --all
sudo ./swupd update  # locally built swupd from this PR
sudo swupd verify --fix --picky -m 21810
sudo swupd clean --all
```
I did the above 5 times and recorded update time for system swupd and local swupd. These were the results:

| system | local |
| --- | --- |
| 19.8 | 30.8 |
| 21.9 | 30.9 |
| 34.7 | 26.6 |
| 32.1 | 21.2 |
| 20.6 | 33.7 |

system avg: 25.82 s
local avg: 28.64 s
difference: 2.82 s

With the variation I saw in just doing 10 updates I believe this is insignificant. The TODO can wait until after this PR is merged.

src/delta.c
matthewrsj: From the `readdir` man page: 

> The order in which filenames are read by successive calls to readdir() depends on the 
> filesystem implementation; it is unlikely that the names will be sorted in any fashion.

So how is this a solution? Maybe `scandir` with `alphasort` (which I found after some googling).

test/functional/update/delta-without-rename-flag/test.bats
cmarcelo: Version 20 has a file (/bat) with hash `c21cc55db744a24867470b7343bdc6bde1c19e4b5a782fd82c84a86b55de6d54` that is not available as a fullfile in the test, only as a delta from a file present in version 10. If the delta fails, we would not be able to have "/bat" and the update would fail (also the output would be affected).

cmarcelo:
Thanks for doing the measurements, I'll try to replicate them locally, instrumenting only the specific part of the execution, to get rid of noise caused by network.

src/delta.c
cmarcelo: I think these are good candidates for a telemetry entry, since it means we are generating packs with files that shouldn't be there. Maybe just send one telemetry entry with the count of the deltas that we couldn't use? (I'm fine just removing it too)

> This indicated a /var/lib/swupd on my system that had a lot of extra files hanging around. 

Yes, in the first execution you would get noise since I think we were not cleaning the deltas before.

matthewrsj: Further testing shows these errors don't show up any more. I'm fine with leaving them for now.

test/functional/update/delta-without-rename-flag/test.bats
matthewrsj: Gotcha

src/delta.c
cmarcelo: Looking at a typical profile, delta application doesn't seem to be what is causing such difference, as the time it takes is lower than the difference you are seeing. I suspect there are some other noise factors at play. If we were to optimize, we should look to improve the consolidation functions first.

I'll go back to my initial position: keep this TODO (I've updated to take into account the ordering), and we later can make this improvement. I'll see if we have some low hanging fruit in the consolidation instead (independent of this PR).

```
Statistics for going from version 21810 to version 21820:

    changed bundles   : 14
    new bundles       : 0
    deleted bundles   : 0

    changed files     : 12
    new files         : 2
    deleted files     : 2

	...100%
Staging file content
Applying update
	...100%
Update was applied.
Update took 5.1 seconds

Raw elapsed time stats:
    144.57 ms: Main Update
     12.71 ms: Update Step 1: get versions
     94.40 ms: Load Manifests:
   4262.79 ms: Recurse and Consolidate Manifests
   1588.24 ms: Add Included Manifests
      0.54 ms: Pre-Update Scripts
    177.92 ms: Download Packs
     38.25 ms: Apply deltas
      5.40 ms: Create Update List
    228.30 ms: Update Loop
      0.02 ms: Run Scripts

CPU process time stats:
      3.32 ms: Main Update
      0.21 ms: Update Step 1: get versions
      1.48 ms: Load Manifests:
   2393.18 ms: Recurse and Consolidate Manifests
    477.46 ms: Add Included Manifests
      0.54 ms: Pre-Update Scripts
     13.37 ms: Download Packs
     38.06 ms: Apply deltas
      5.38 ms: Create Update List
     44.33 ms: Update Loop
      0.02 ms: Run Scripts
Update successful. System updated from version 21810 to version 21820
```

cmarcelo:
Rebased and addressed Matthew comments.

src/delta.c
matthewrsj: Thanks for your additional measurements. I wasn't actually too concerned with the time differential, as I said "With the variation I saw in just doing 10 updates I believe this is insignificant. The TODO can wait until after this PR is merged."

cmarcelo: Yes, I got that. I was just thinking that we had a bigger problem in our hands, hence the additional measurements :-)

matthewrsj:
Doing a last test on my system before merging...

PR #382, opened 2018-02-13
Forbid certain state_dir values to prevent disasters

cmarcelo:
During initialization swupd may delete the state_dir to ensure it is
in the right shape. That means that by accident, passing "-S /" to any
command is dangerous.

Since neither /, /usr or /var are really useful as state dirs, let's
prevent some damage by not allowing them.

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -353,7 +353,15 @@ bool set_state_dir(char *path)
 {
 	if (path) {
 		if (path[0] != '/') {
-			fprintf(stderr, "statepath must be a full path starting with '/', not '%c'\n", path[0]);
+			fprintf(stderr, "state dir must be a full path starting with '/', not '%c'\n", path[0]);
+			return false;
+		}
+
+		/* Prevent some disasters: since the state dir can be destroyed and
+		 * reconstructed, make sure we never set those by accident and nuke the
+		 * system. */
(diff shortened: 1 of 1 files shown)


cmarcelo:
@icarus-sparry thoughts on how to improve this? (Or if we should not)

src/globals.c
```diff
@@ -353,7 +353,15 @@ bool set_state_dir(char *path)
...
-			fprintf(stderr, "statepath must be a full path starting with '/', not '%c'\n", path[0]);
+			fprintf(stderr, "state dir must be a full path starting with '/', not '%c'\n", path[0]);
+			return false;
+		}
+
+		/* Prevent some disasters: since the state dir can be destroyed and
+		 * reconstructed, make sure we never set those by accident and nuke the
+		 * system. */
+		if (!strcmp(path, "/") || !strcmp(path, "/var") || !strcmp(path, "/usr")) {
+			fprintf(stderr, "invalid state dir: %s\n", path);
```
icarus-sparry: The idea is very good.

I think it would make sense to have a little more explicit message. If as a user I was just told "Invalid state dir" I would wonder what the rules were which make it invalid. Obviously it is a directory, so at that point one is left puzzled. Perhaps "Refusing to use %s as state dir because it will be erased first" will be a suitable heads up?

cmarcelo: Changed the message, thanks!

PR #383, opened 2018-02-13
Add 'swupd clean' command

cmarcelo:
This command removes old files from statedir. The interface is a bit
opaque so there will be room to modify the policy later.

By default it will delete staged files that are older than
DAYS_TO_KEEP_FILES. For manifests this heuristic is not good as older
manifests might still be used for the current OS version being run, so
also ensure that those do not get deleted. This prevents 'swupd
search' to redownload files.

The default behavior should be suitable to use in combination with
automatic updates, to control the size of state dir. There is also an
--all option to remove all the state files regardless of dates and
usage.

To avoid "disasters" in case some paths are set wrong, the command
explicitly delete patterns of files create by swupd.

### DIFF
--- a/src/clean.c
+++ b/src/clean.c
@@ -1,0 +1,394 @@
+/*
+ *   Software Updater - client side
+ *
+ *      Copyright © 2018 Intel Corporation.
+ *
+ *   This program is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License as published by
+ *   the Free Software Foundation, version 2 or later of the License.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#define _GNU_SOURCE
+#include <ctype.h>
+#include <errno.h>
+#include <getopt.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "swupd.h"
+
+/* Default value if --all is not passed. */
+const int DAYS_TO_KEEP_FILES = 3;
+
+static void print_help()
+{
+	fprintf(stderr,
+		"Usage:\n"
+		"	swupd clean [options]\n"
+		"\n"
+		"Remove cached data used for old updates from state directory.\n"
+		"\n"
+		"Options:\n"
+		"   --all                   Remove all the cached content not only old\n"
+		"   --dry-run               Just print files that would be removed\n"
+		"   -S, --statedir DIR      Specify alternate swupd state directory\n"
+		"   -h, --help              Display this help message\n"
+		"\n");
+}
+
+static struct {
+	int all;
+	int dry_run;
+} options;
+
+static struct {
+	int files_removed;
+} stats;
+
+static struct timespec now;
+
+static const struct option prog_opts[] = {
+	{ "help", no_argument, 0, 'h' },
+	{ "all", no_argument, &options.all, 1 },
+	{ "statedir", required_argument, 0, 'S' },
+	{ "dry-run", no_argument, &options.dry_run, 1 },
+	{ 0, 0, 0, 0 }
+};
+
+static bool parse_options(int argc, char **argv)
+{
+	int opt;
+	while ((opt = getopt_long(argc, argv, "hS:", prog_opts, NULL)) != -1) {
+		switch (opt) {
+		case '?':
+			return false;
+		case 'h':
+			print_help();
+			exit(0);
+		case 'S':
+			if (!optarg || !set_state_dir(optarg)) {
+				fprintf(stderr, "Invalid --statedir argument\n\n");
+				return false;
+			}
+			break;
+		case 0:
+			/* Handle options that don't have shortcut. */
+			break;
+		default:
+			fprintf(stderr, "Error: unrecognized option: -%c,\n\n", opt);
+			return false;
+		}
+	}
+	return true;
+}
+
+typedef bool(remove_predicate_func)(const char *dir, const struct dirent *entry);
+
+/* Remove files from path for which pred returns true.
+ * Currently it doesn't recursively remove directories.
+ *
+ * Unless --all is set, the files are also filtered by their age.
+ */
+static int remove_if(const char *path, remove_predicate_func pred)
+{
+	int ret = 0;
+	DIR *dir;
+	const int seconds = DAYS_TO_KEEP_FILES * 60 * 60 * 24;
+
+	dir = opendir(path);
+	if (!dir) {
+		return errno;
+	}
+
+	char *file = NULL;
+
+	while (true) {
+		free_string(&file);
+		ret = 0;
+
+		/* Reset errno to distinguish between a previous
+		 * failure and the end of stream. */
+		errno = 0;
+		struct dirent *entry;
+		entry = readdir(dir);
+		if (!entry) {
+			if (errno) {
+				ret = errno;
+			}
+			break;
+		}
+
+		if (!strcmp(entry->d_name, ".") || !strcmp(entry->d_name, "..")) {
+			continue;
+		}
+
+		string_or_die(&file, "%s/%s", path, entry->d_name);
+
+		struct stat stat;
+		ret = lstat(file, &stat);
+		if (ret != 0) {
+			fprintf(stderr, "couldn't access %s: %s\n", file, strerror(errno));
+			continue;
+		}
+
+		if (!options.all) {
+			if (now.tv_sec - stat.st_mtime <= seconds) {
+				continue;
+			}
+		}
+
+		if (!pred(path, entry)) {
+			continue;
+		}
+
+		if (options.dry_run) {
+			printf("%s\n", file);
+		} else {
+			if (S_ISDIR(stat.st_mode)) {
+				ret = rmdir(file);
+			} else {
+				ret = unlink(file);
+			}
+			if (ret != 0) {
+				fprintf(stderr, "couldn't remove file %s: %s\n", file, strerror(errno));
+			}
+		}
+		if (ret == 0) {
+			stats.files_removed++;
+		}
+	}
+
+	closedir(dir);
+	return ret;
+}
+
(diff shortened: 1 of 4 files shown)


cmarcelo:
This is missing Manifest*delta in statedir files. I'll add those later if the idea gets accepted.

matthewrsj:
@cmarcelo we have conflicts here.

cmarcelo:
Ping

src/clean.c
```diff
@@ -0,0 +1,397 @@
...
+	return true;
+}
+
+static bool is_manifest(const char UNUSED_PARAM *dir, const struct dirent *entry)
+{
+	static const char prefix[] = "Manifest.";
+	static const size_t prefix_len = sizeof(prefix) - 1;
+
+	const char *name = entry->d_name;
+	if (strlen(name) < prefix_len) {
```
icarus-sparry: I would have used strncmp rather than the 2 separate tests and to show that we are working on string data.

You only need the strlen(name) < prefix_len test to make sure you are not reading beyond the end of the name array,

src/clean.c
```diff
@@ -0,0 +1,397 @@
...
+	int lock_fd = 0;
+	ret = swupd_init(&lock_fd);
+	if (ret != 0) {
+		fprintf(stderr, "Failed swupd initialization, exiting now.\n");
+		return ret;
+	}
+
+	ret = clock_gettime(CLOCK_REALTIME, &now);
+	if (ret != 0) {
+		perror("error reading time");
```
icarus-sparry: A better error message would help here. If we have --all do we care if we can't read the time?

src/clean.c
```diff
@@ -0,0 +1,397 @@
...
+}
+
+typedef bool(remove_predicate_func)(const char *dir, const struct dirent *entry);
+
+/* Remove files from path for which pred returns true.
+ * Currently it doesn't recursively remove directories.
+ *
+ * Unless --all is set, the files are also filtered by their age.
+ */
+static int remove_if(const char *path, remove_predicate_func pred)
```
icarus-sparry: I very much like this functional approach.

I think you can take it further if you implement remove_predicate_func as a closure instead, probably not worth it for the simple cases here but you could create AND and OR functions to combine predicates....

src/clean.c
```diff
@@ -0,0 +1,397 @@
...
+	return !memcmp(entry->d_name, prefix, prefix_len);
+}
+
+static char *read_mom_contents(int version)
+{
+	char *mom_path = NULL;
+	char *contents = NULL;
+
+	string_or_die(&mom_path, "%s/%d/Manifest.MoM", state_dir, version);
+	FILE *f = fopen(mom_path, "r");
```
icarus-sparry: You can free_string(&mom_path) immediately after you have attempted the open.

src/clean.c
```diff
@@ -0,0 +1,397 @@
...
+	if (!f) {
+		/* This is a best effort. */
+		goto end;
```
icarus-sparry: and then use return rather than goto here

src/clean.c
```diff
@@ -0,0 +1,397 @@
...
+	}
+
+	int ret;
+	ret = fseek(f, 0, SEEK_END);
```
icarus-sparry: You could just use fstat rather than doing all this seeking.

src/clean.c
```diff
@@ -0,0 +1,397 @@
...
+	if (size < 1) {
+		goto end;
+	}
+
+	ret = fseek(f, 0, SEEK_SET);
+	if (ret != 0) {
+		goto close_then_end;
+	}
+
+	char *tmp = malloc(size + 1);
```
icarus-sparry: Not sure having tmp adds anything here rather than using contents. The compiler will probably optimize tmp out anyway.

src/clean.c
```diff
@@ -0,0 +1,397 @@
...
+	if (!tmp) {
+		goto close_then_end;
+	}
+
+	ret = fread(tmp, size, 1, f);
+	if (ret != 1) {
+		free_string(&tmp);
```
icarus-sparry: should call free, not free_string. Yes we know how free_string works, and it will do the right thing, but it is breaking the abstraction. One day we might make string_or_die use alloca and make free_string just set the pointer to NULL at which time this will turn into a memory leak.

src/clean.c
```diff
@@ -0,0 +1,397 @@
...
+		goto close_then_end;
+	}
+
+	contents = tmp;
```
icarus-sparry: You need to add the terminating NUL to the data

src/clean.c
```diff
@@ -0,0 +1,397 @@
...
+	if (!options.all) {
+		int current_version = get_current_version(path_prefix);
+		if (current_version < 0) {
+			fprintf(stderr, "Unable to determine current OS version\n");
+		} else {
+			mom_contents = read_mom_contents(current_version);
+		}
+	}
+
+	while (true) {
```
icarus-sparry: while (ret == 0) perhaps?

src/clean.c
```diff
@@ -0,0 +1,397 @@
...
+		(void)rmdir(version_dir);
+
+		free_string(&version_dir);
+		if (ret != 0) {
+			goto end;
+		}
+	}
+
+end:
+	free_string(&mom_contents);
```
icarus-sparry: free(mom_contents) rather than free_string

src/clean.c
```diff
@@ -0,0 +1,397 @@
...
+	size_t len = strlen(name);
+	if (len < (prefix_len + suffix_len)) {
+		return false;
+	}
+	return !memcmp(name, prefix, prefix_len) && !memcmp(name + len - suffix_len, suffix, suffix_len);
+}
+
+static bool is_all_digits(const char *s)
+{
+	while (*s) {
```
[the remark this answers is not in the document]
cmarcelo: You mean `for (; *s; s++)` or taking the length and using indices?

src/clean.c
```diff
@@ -0,0 +1,397 @@
...
+			continue;
+		}
+
+		char *file = NULL;
+		string_or_die(&file, "%s/%s", path, entry->d_name);
+
+		struct stat stat;
+		ret = lstat(file, &stat);
+		if (ret != 0) {
+			fprintf(stderr, "couldn't remove file %s: %s\n", file, strerror(errno));
```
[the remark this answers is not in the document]
cmarcelo: Changed to "couldn't access" file.

src/clean.c
```diff
@@ -0,0 +1,397 @@
...
+				ret = unlink(file);
+			}
+			if (ret != 0) {
+				fprintf(stderr, "couldn't remove file %s: %s\n", file, strerror(errno));
+			}
+		}
+		if (ret == 0) {
+			stats.files_removed++;
+		}
+	next:
```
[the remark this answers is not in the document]
cmarcelo: Done.

src/clean.c
```diff
@@ -0,0 +1,397 @@
...
+static int remove_if(const char *path, remove_predicate_func pred)
+{
+	int ret = 0;
+	DIR *dir;
+	const int seconds = DAYS_TO_KEEP_FILES * 60 * 60 * 24;
+
+	dir = opendir(path);
+	if (!dir) {
+		ret = errno;
+		goto end;
```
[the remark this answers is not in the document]
cmarcelo: Fixed.

src/clean.c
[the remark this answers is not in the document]
cmarcelo: Changed to `for (; *s; s++)`.

src/clean.c
cmarcelo: That is much better! Done.

src/clean.c
cmarcelo: Changed to only calculate 'now' if we need. Changed the error message.

src/clean.c
cmarcelo: Yeah, these approach could be extended if we need more complex cases.

I am assuming by closure you meant passing a data pointer to both remove_if (and then to predicate calls).

src/clean.c
cmarcelo: Good to know. Done.

src/clean.c
cmarcelo: Done.

src/clean.c
cmarcelo: Done.

src/clean.c
cmarcelo: Control flow changed a bit so tmp is gone. It was an attempt to not "fixing" contents for the return call, but had issues.

src/clean.c
cmarcelo: Fixed.

src/clean.c
cmarcelo: Good catch. Done.

src/clean.c
cmarcelo: Kept as is. It did not look very good in practice since the test doesn't cover all the breaking conditions (we have a case where the stream is over, and ret == 0).

src/clean.c
cmarcelo: Done.

cmarcelo:
> This is missing Manifest*delta in statedir files.

Removing those now. One improvement would be always remove those regardless of the date.

cmarcelo:
Thanks for the review, Icarus.

src/clean.c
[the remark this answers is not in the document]
icarus-sparry: Thank you.

src/clean.c
icarus-sparry: The usual way I implement a closure in C is to make what you currently have as a function pointer to be a struct pointer, and inside the struct have the function pointer and a void* pointer.

You can then have an AND closure which uses the void * pointer to point to a vector of closures...

src/clean.c
```diff
@@ -0,0 +1,394 @@
...
+		/* Reset errno to distinguish between a previous
+		 * failure and the end of stream. */
+		errno = 0;
+		struct dirent *entry;
+		entry = readdir(dir);
+		if (!entry) {
+			if (errno) {
+				ret = errno;
+			}
+			break;
```
icarus-sparry: Optional. Since this is the only break, you could put the closedir() and return here.

Probably not worth it.

src/clean.c
```diff
@@ -0,0 +1,394 @@
...
+		if (options.dry_run) {
+			printf("%s\n", file);
+		} else {
+			if (S_ISDIR(stat.st_mode)) {
+				ret = rmdir(file);
+			} else {
+				ret = unlink(file);
+			}
+			if (ret != 0) {
+				fprintf(stderr, "couldn't remove file %s: %s\n", file, strerror(errno));
```
icarus-sparry: Possibly remove the word "file" from the error message?

cmarcelo: Done.

PR #385, opened 2018-02-15
swupd-add-pkg: Fix upstream url not being filled in

tmarcu:
The script should be calling init-mix to fully and correctly setup the mix
directory. Without this, the upstream url was never being recorded, and the
yum-mix.conf did not get the URL substituted during chroots phase.


tmarcu:
*NOTE* This will have to be updated when the new mixer is released.

PR #387, opened 2018-02-21
Improve swupd search output

matthewrsj:
@fenrus75 wrote this patch. Opening for review.

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -29,6 +29,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <libgen.h>
 
 #include "config.h"
 #include "swupd.h"
@@ -39,6 +40,121 @@ bool init = false;
 
 static char search_type = '0';
 static char scope = '0';
+
+struct bundle_result {
+	char bundle_name[256];
+	long size;
+	double score;
+	struct list *files;
+};
+
+struct file_result {
+	char filename[PATH_MAX+1];
+	double score;
+};
+
+static struct list *results;
+
+
+static void add_bundle_file_result(char *bundlename, char *filename, double score, struct manifest *man) 
+{
+	struct bundle_result *bundle = NULL;
+	struct file_result *file;
+	struct list *ptr;
+
+	ptr = results;
+	while (ptr) {
+		struct bundle_result *item;
+
+		item = ptr->data;
+		if (strcmp(item->bundle_name, bundlename) == 0) {
+			bundle = item;
+			break;
+		}
+		ptr = ptr->next;
+	}
+	if (!bundle) {
+		bundle = calloc(sizeof(struct bundle_result), 1);
+		results = list_append_data(results, bundle);
+		strncpy(bundle->bundle_name, bundlename, 255);
+		bundle->size = man->contentsize / 1024 / 1204;
+		bundle->score = -0.1 * bundle->size;
+	}
+
+	file = calloc(sizeof(struct file_result), 1);
+	strncpy(file->filename, filename, PATH_MAX);
+	bundle->files = list_append_data(bundle->files, file);
+	file->score = score;
+	bundle->score += score;
+}
+
+static int bundle_cmp(const void *a, const void *b)
+{
+	const struct bundle_result *A, *B;
+	A = (struct bundle_result *)a;
+	B = (struct bundle_result *)b;
+	if (A->score > B->score)
+		return -1;
+	if (A->score < B->score)
+		return 1;
+	return 0;
+	
+}
+
+static int file_cmp(const void *a, const void *b)
+{
+	const struct file_result *A, *B;
+	A = (struct file_result *)a;
+	B = (struct file_result *)b;
+	if (A->score > B->score)
+		return -1;
+	if (A->score < B->score)
+		return 1;
+	return 0;
+	
+}
+static void sort_results(void)
+{
+	struct list *ptr;
+	struct bundle_result *b;
+
+	results = list_sort(results, bundle_cmp);
+
+	ptr = results;
+	while (ptr) {
+		b = ptr->data;
+		ptr = ptr->next;
+
+		b->files = list_sort(b->files, file_cmp);
+	}
+}
+
+static void print_final_results(void)
+{
+	struct bundle_result *b;
+	struct list *ptr;
+	int counter = 0;
+
+	ptr = results;
+	while (ptr && counter < 5) {
+		struct list *ptr2;
+		struct file_result *f;
+		int counter2 = 0;
+
+		b = ptr->data;
+		ptr = ptr->next;
+		counter++;
+		printf("Bundle %s   (%li Mb)\n", b->bundle_name, b->size);
+		ptr2 = b->files;
+		while (ptr2 && counter2 < 5) {
+			f = ptr2->data;
+			ptr2 = ptr2->next;
+			printf("\t%s\n", f->filename);
+			counter2++;
+		}
+		printf("\n");
+	}
+}
 
 /* Supported default search paths */
 static char *lib_paths[] = {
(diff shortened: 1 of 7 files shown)


matthewrsj:
+1 (I can't approve the PR as the PR author though)

src/search.c
```diff
@@ -40,6 +41,121 @@ bool init = false;
...
+	struct bundle_result *bundle = NULL;
+	struct file_result *file;
+	struct list *ptr;
+
+	ptr = results;
+	while (ptr) {
+		struct bundle_result *item;
+
+		item = ptr->data;
+		if (strcmp(item->bundle_name, bundlename) == 0) {
```
tmarcu: Since we know the max size at all times is 256, why not go 1 step further and
`strncmp(item->bundle_name, bundlename, sizeof(item->bundle_name)) // so we don't hardcode in magic numbers`

src/search.c
```diff
@@ -40,6 +41,121 @@ bool init = false;
...
+			bundle = item;
+			break;
+		}
+		ptr = ptr->next;
+	}
+	if (!bundle) {
+		bundle = calloc(sizeof(struct bundle_result), 1);
+		results = list_append_data(results, bundle);
+		strncpy(bundle->bundle_name, bundlename, 255);
```
tmarcu: Magic numbers.

src/search.c
```diff
@@ -40,6 +41,121 @@ bool init = false;
...
+		bundle->size = man->contentsize / 1024 / 1204;
+		bundle->score = -0.1 * bundle->size;
```
tmarcu: Some comments would be nice.

src/search.c
icarus-sparry: strncmp is usually slightly slower than strcmp, particularly if you allow strcmp to "overread".

src/search.c
icarus-sparry: Hints like why 1204 was chosen would be good as well, rather than making me suspect a typo for 1024.

src/search.c
```diff
@@ -40,6 +41,121 @@ bool init = false;
...
+}
+
+static void print_final_results(void)
+{
+	struct bundle_result *b;
+	struct list *ptr;
+	int counter = 0;
+
+	ptr = results;
+	while (ptr && counter < 5) {
```
icarus-sparry: Why not a more idiomatic for loop

    for (counter=0, ptr=results;
          ptr && counter < 5;
          ptr=ptr->next, counter++) {

src/search.c
```diff
@@ -40,6 +41,121 @@ bool init = false;
...
+		struct list *ptr2;
+		struct file_result *f;
+		int counter2 = 0;
+
+		b = ptr->data;
+		ptr = ptr->next;
+		counter++;
+		printf("Bundle %s   (%li Mb)\n", b->bundle_name, b->size);
+		ptr2 = b->files;
```
icarus-sparry: for(ptr2=ptr->data->files, counter2=0;
          ptr2 && counter2 < 5;
          ptr2=ptr2->next,counter2++) {

src/search.c
```diff
@@ -263,12 +379,58 @@ static bool file_search(char *filename, char *path, char *search_term)
...
+	double multiplier = 1.0;
+	double score = 1.0;
+
+	char bnb[PATH_MAX];
+	char dnb[PATH_MAX];
+
+	char *bn;
+	char *dn;
+
+	strcpy(bnb, file);
```
icarus-sparry: Since you are making a copy, you can just use strrchr to find the last '/' and change it into a NUL.

src/search.c
```diff
@@ -263,12 +379,58 @@ static bool file_search(char *filename, char *path, char *search_term)
...
+		score += 5;
+
+	if (strcmp(dn, "/usr/include") == 0)
+		multiplier /= 2;
+
+	if (strstr(bundle, search_term))
+		score += 0.5;
+
+	/* slightly favor shorter filenames */
+	score += (2.0 / strlen(file));
```
icarus-sparry: Division by zero if you say 

      swupd search ''

?

src/search.c
ulissesf: Btw, I believe 1204 is a typo in the MiB calculation. The score hit depending on bundle size might need a comment, yes. @fenrus75?

src/search.c
matthewrsj: That is an invalid search string actually.

matthewrsj:
OK, I'll iterate on this.

src/search.c
matthewrsj: Since it is setting an initial score of negative 1/10th the bundle size, it seems to just be an initial penalty based on the size of the bundle. I'll add a comment to that effect.

icarus-sparry:
I think that the search should look for regular expressions, rather than fixed strings.

I think that there should be an indication if the results are truncated at 5 or if that is all there is.

I think that we should have the existing ability to list all the matches, rather than just limiting it to at most 5 matches in at most 5 bundles.

I look forward to reading the documentation explaining the reasoning behind the scoring rules.

src/search.c
matthewrsj: I think the while loop is more readable than that.

src/search.c
matthewrsj: why not use `dirname`? I assume it is basically doing exactly that.

src/search.c
matthewrsj: Oh, this isn't even the search term, this is the file name. We shouldn't have zero-length file names (`/` would be the shortest).

src/search.c
```diff
@@ -263,12 +382,54 @@ static bool file_search(char *filename, char *path, char *search_term)
...
  * Report out, respecting verbosity
  */
-static void report_find(char *bundle, char *file)
+static void report_find(char *bundle, char *file, char *search_term, struct manifest *man)
 {
-	printf("'%s'  :  '%s'\n", bundle, file);
+	double score;
+
+	score = guess_score(bundle, file, search_term);
+	//	printf("'%s'  :  '%s'   (%5.1f)\n", bundle, file, score);
```
matthewrsj: This was the old way of doing swupd search. I think a flag for swupd search would be helpful to toggle `--all-results` and just print the flat list. Makes it more machine-readable.

src/search.c
matthewrsj: Also, confirmed that 1204 is a typo for 1024

src/search.c
icarus-sparry: It is making 2 copies, then calling dirname and basename. My point is it could make one copy, split it, and then use the two halves. Reduces stack space, copying, scanning.

matthewrsj:
> I think that the search should look for regular expressions, rather than fixed strings.

That would be a good improvement for the future. Let's open an issue. #388

> I think that there should be an indication if the results are truncated at 5 or if that is all there is.
> 
> I think that we should have the existing ability to list all the matches, rather than just limiting it to at most 5 matches in at most 5 bundles.

How about a flag like `--all-results` that will display all the results, maybe the way they used to be displayed so that a machine can parse it better?

> I look forward to reading the documentation explaining the reasoning behind the scoring rules.

For now these are just experimental heuristics that can be updated over time. I'll add comments with the reasoning behind them as relayed by @fenrus75

matthewrsj:
Opened #388 #389 and #390 for future improvements.

matthewrsj:
Also #391

src/search.c
matthewrsj: I see what you are saying, +1

matthewrsj:
@tmarcu @icarus-sparry I made some updates, please review again.

src/search.c
tmarcu: Is that really an issue in this code path...

icarus-sparry: It is not an issue for this code path. I was just trying to understand why you should want to change strcmp(a,b) to strncmp(a,b,MAXSIZE).

Issue #393, opened 2018-02-21
If we adopt the new swupd search, we should allow the number of results to be specified


icarus-sparry:
If we adopt #387, then we should allow the number of bundles shown to be configured, and the number of files in the bundle to be configured, rather than fixed at 5 each.

matthewrsj:
Fixed by #408

PR #396, opened 2018-02-22
search: special handling for installed bundles

### LINKED ISSUE #390: swupd search and bundle list should indicate if the listed bundle is installed
These operations should additionally report if the listed bundle is installed on the client system or not to assist users in making decisions about what they need to install.

matthewrsj:
- Show indication that bundle is installed.

  When displaying a search result for a bundle display "installed" next to bundles that have been added to the system.

- ~~Reduce score for installed bundles.~~

  ~~The assumption here is that users use the search feature when they are looking for a bundle to install a certain file. For this case already installed bundles are not what the user is looking for so reduce the score by half.~~

  ~~@icarus-sparry has already expressed concern with the second patch. Anyone else use the search functionality to figure out which installed bundles they should remove?~~

Fixes #390

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -47,6 +47,7 @@ struct bundle_result {
 	long size;
 	double score;
 	struct list *files;
+	bool is_tracked;
 };
 
 /* per-file scoring struct */
@@ -85,6 +86,8 @@ static void add_bundle_file_result(char *bundlename, char *filename, double scor
 		 * This is set to negative 1/10th of the bundle size.
 		 * NOTE this bundle->size does not include the bundle includes sizes */
 		bundle->score = -0.1 * bundle->size;
+		/* record if the bundle is tracked on the system */
+		bundle->is_tracked = is_tracked_bundle(bundlename);
 	}
 
 	file = calloc(sizeof(struct file_result), 1);
@@ -167,7 +170,10 @@ static void print_final_results(void)
 		b = ptr->data;
 		ptr = ptr->next;
 		counter++;
-		printf("Bundle %s   (%li Mb)\n", b->bundle_name, b->size);
+		printf("Bundle %s\t(%li Mb)%s\n",
+		       b->bundle_name,
+		       b->size,
+		       b->is_tracked ? "\tinstalled" : "");
 		ptr2 = b->files;
 		while (ptr2 && counter2 < 5) {
 			f = ptr2->data;


bryteise:
I rarely use search X(. I'd expect the order to be consistent however regardless of a bundle being installed or not.

matthewrsj:
@bryteise are you aware of how things are scored now? Order is managed internally to search.c now with a relevancy score. Do you think whether a bundle is installed or not should not affect that score or do you object to the entire approach?

bryteise:
The score approach is fine, I'm not convinced about the installed changing score because I like the order to be generally static; i.e if I found it once and I look in the same place by doing the same thing I'll generally see it (updating the scoring algorithm is fine though, just updating score based on system state I'm not convinced I like yet).

matthewrsj:
> just updating score based on system state I'm not convinced I like yet

Good point. @icarus-sparry, @fenrus75, peanut gallery any comments?

bryteise:
I think to the pundles case could exacerbate the problem too. If I had a lot of pundles installed and then wanted to check for something that is closely related to another thing I searched for before and installed, if I don't see the original thing I had got in my search it would confuse me.

tmarcu:
+1 for keeping consistency between searches unless new bundles are added that change the scoring. @bryteise pundles would change it the same as adding a small, more appopriate bundle though. So it should change the order IMO when pundles or new bundles are added that contain the same content.

bryteise:
@tmarcu I am completely fine with new server content changing order as well as new content should be scored correctly.

I am a little worried that pundles will be slightly annoying if the includes are limited in that we'll start to see a ton of matches for common libraries and the like with search.

matthewrsj:
> pundles will be slightly annoying if the includes are limited

A good reason to implement #393

icarus-sparry:
I agree with @bryteise about having the bundle installed shouldn't affect if something is shown or not.

When we have pundles I think this scoring approach and only showing 5 bundles is going to be pretty annoying. The good thing is that the pundle having the file will be listed first. Allowing regular expressions and allowing overrides of the 5 & 5 will help to allow someone who knows what they are doing to get useful data out when the defaults fail. Having an indication that the data is truncated will provide a clue to the expert that they need to work around the defaults.

matthewrsj:
Ok, I'm convinced that if we ever do decide to reduce score based on whether or not it is installed (client state) it needs some more thought. I'll remove it from this PR.

matthewrsj:
This is ready for review now then.

src/search.c
```diff
@@ -167,7 +170,10 @@ static void print_final_results(void)
 		b = ptr->data;
 		ptr = ptr->next;
 		counter++;
-		printf("Bundle %s   (%li Mb)\n", b->bundle_name, b->size);
+		printf("Bundle %s\t(%li Mb)\t%s\n",
+		       b->bundle_name,
+		       b->size,
+		       b->is_tracked ? "installed" : "");
```
icarus-sparry: Maybe 

		printf("Bundle %s\t(%li Mb)%s\n",
		       b->bundle_name,
		       b->size,
		       b->is_tracked ? "\tinstalled" : "");

to avoid the trailing tab?  Maybe
		printf("Bundle %s\t(%li Mb)\t%sinstalled\n",
		       b->bundle_name,
		       b->size,
		       b->is_tracked ? "" : "not-");

to keep the number of fields the same.

matthewrsj: Oooh, I like the first suggestion. Makes it obvious that this one is different. The second one you have to look at a little closer to see the `not-` in front of `installed`

PR #397, opened 2018-02-23
RFC: Report real contentsize for swupd search bundles

matthewrsj:
When reporting bundles for swupd search include their included bundles'
contentsizes as well for more accurate size reporting.

**RFC**: looking for ways to make this faster. This is considerably slower when we have to recurse all the manifests for their includes. I only do this recursion once for each bundle that we hit but the slowdown is still noticeable. Maybe an option on  `load_manifest()` to only load the header information and not the entire file list.

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -58,7 +58,7 @@ struct file_result {
 static struct list *results;
 
 /* add a bundle_result to the results list */
-static void add_bundle_file_result(char *bundlename, char *filename, double score, struct manifest *man)
+static void add_bundle_file_result(char *bundlename, char *filename, double score, struct manifest *man, struct manifest *MoM)
 {
 	struct bundle_result *bundle = NULL;
 	struct file_result *file;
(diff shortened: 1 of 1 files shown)


cmarcelo:
I meant extra installed size instead of download size.

matthewrsj:
> one suggestion is to not try to do that calculation at the time of one file being added, but later on, after you did a pass through all the bundles (i.e. do not parse manifests again, store the information you need from this first loop).

This calculation is currently only done for each new bundle that has a hit, not every time a file is added.

> One important information is the amount we will have to download, so I wonder if the output should only take into account the included bundles not installed (I haven't verified if the recurse function is already doing this for you or not).

I should have mentioned this was my plan. This PR is not complete, I just wanted to open it for comments. My plan is to say `installed size: x` for installed bundles and `size to install: x` for bundles that are not installed.

I have a couple ideas that I'll try. I'll reopen this when it is ready.

cmarcelo:
> This calculation is currently only done for each new bundle that has a hit, not every time a file is added.

I know. :)

cmarcelo:
> size to install: x for bundles that are not installed.

Nice!

matthewrsj:
> While you do the first pass, you can keep track of each bundle size & includes, and use that for your calculation.

Ok I see what you are saying, but the issue here is that we may not actually parse all the manifests if `--scope` is `o`. I suppose in that case we don't have to print the bundle size though.

PR #402, opened 2018-02-27
Remove root requirement for check-update

### LINKED ISSUE #398: Allow 'swupd' to run without root access for simple operation
Some operations such as `search`, `bundle-list` and possibly `check-update` should not need root access to be run. Today running those without root access returns an error:

```
swupd check-update 
swupd-client software update checker 3.14.6
   Copyright (C) 2012-2017 Intel Corporation

This program must be run as root..aborting.
```

matthewrsj:
Also removes a call to update an unused motd file. The motd is now
managed differently in Clear Linux and the MOTD_FILE does not even exist
where it expects it to. Since this was the only reason root was required
just remove the functionality entirely.

Fixes #398 (the best we can do for now)

### DIFF
--- a/src/check_update.c
+++ b/src/check_update.c
@@ -133,7 +133,6 @@ static int check_update()
 {
 	int current_version, server_version;
 
-	check_root();
 	if (!init_globals()) {
 		return EINIT_GLOBALS;
 	}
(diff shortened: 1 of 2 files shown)


bryteise:
I see this calling read_versions which calls get_latest_version and get_latest_version attempts to write to the statedir which makes me think this is not a happy thing (I think it should fail in swupd_download_file_start which attempts to open the statedir file).

matthewrsj:
> But that only addresses a subset of the problem. Processes running with the old ID also need to be restarted, which is harder.

Hmm, you appear to be right, but my testing shows otherwise...

matthewrsj:
Closing this as it is not a significant improvement, and priority should be #403

Issue #392, opened 2018-02-21
swupd search should indicate if its results are truncated.


icarus-sparry:
swupd search in #387 shows up to five files in up to five bundles. If there are 4 or less files then it is known that this is all there is in that bundle, but if there are five you don't know if there are five or more than five. The same applies to bundles, if it shows you 4 then you know you have everything, but if there are 5 you don't know.

matthewrsj:
Fixed by #408

Issue #391, opened 2018-02-21
swupd search should show all results by default


matthewrsj:
swupd search should ~~optionally~~ print all results in the manner it did before #387 (actually, grouped by bundle would still be nice ~~so it can be more easily machine-parse-able~~. We can also add a flag to output in CSV to be more machine-readable.

gvancuts:
What is the exact meaning you want to give to an `--all-results` flag? Is this about showing all search results (instead of limiting the output to 5 per bundle)?

As a very recent user of Clear Linux, I must admit that I have found using `swupd` *very* confusing. I did not know about this truncating of the output before I came across this feature request (while searching for something different). There is no indication of this that I have come across anywhere (nothing on the website, which contains just three `swupd` commands anyway) but more worryingly, there is nothing at all about that neither in the `sudo swupd search --help` output nor the `sudo swupd search <string>` output results.

The other aspect of `swupd search` that confuses me is that it only returns search results for bundles that are already installed, doesn't it? How can I search for things that I want to install but are not yet installed? E.g. I would like to be able to search for `kubeadm` for example by doing `sudo swupd search kubeadm` and that the output tells me which bundle(s) I can find it in. Is that covered by some of these existing feature requests or would that be a new one?

matthewrsj:
> What is the exact meaning you want to give to an --all-results flag? Is this about showing all search results (instead of limiting the output to 5 per bundle)?

Actually, I decided to make search display all results by default, with an optional `--top NUM` flag to show the top `NUM` results per bundle.

> I did not know about this truncating of the output before I came across this feature request (while searching for something different). There is no indication of this that I have come across anywhere (nothing on the website, which contains just three swupd commands anyway) but more worryingly, there is nothing at all about that neither in the sudo swupd search --help output nor the sudo swupd search <string> output results.

Yes, `swupd search` got some (very) recent rapid changes. Sorry about that. We are fixing it up right now. #408 addresses your concern here by displaying all results by default, requiring `--top` to truncate results and additionally displaying a message when the results were actually truncated.

Additionally #399 fixes an issue with the bundle size being reported.

> it only returns search results for bundles that are already installed, doesn't it?

No.

gvancuts:
>> What is the exact meaning you want to give to an --all-results flag? Is this about showing all search results (instead of limiting the output to 5 per bundle)?
> Actually, I decided to make search display all results by default, with an optional --top NUM flag to show the top NUM results per bundle.

I think this will be a big improvement!

[...]

PR #408, opened 2018-03-06
search: display all results by default

matthewrsj:
Instead of silently truncating search results to 5 files per 5 bundles
make the default search output a full list grouped by bundle. This
output will only be sorted by increasing bundle size.

Add a -t, --top=NUM argument to tell search to truncate results at NUM
files and NUM bundles. This option attempts to display the top results
based on a set of heuristics.

When truncating results due to --top add a message when the file or
bundle results are cut off.

Add the -m, --csv flag to output all search results in CSV format:

/file/result,bundle-name

This PR is based on #399. Ignore the first commit for this patchset. Only merge this after #399 has been merged.

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -36,6 +36,7 @@
 
 char *search_string;
 bool display_files = false; /* Just display all files found in Manifest set */
+bool csv_format = false;
 bool init = false;
 
 static char search_type = '0';
@@ -243,6 +244,25 @@ static void apply_size_penalty(struct list *bundle_info)
 	}
 }
 
+static void print_csv_results()
+{
+	struct bundle_result *b;
+	struct list *ptr;
+	ptr = list_head(results);
+	while (ptr) {
+		struct list *fptr;
+		struct file_result *f;
+		b = ptr->data;
+		ptr = ptr->next;
+		fptr = list_head(b->files);
+		while (fptr) {
+			f = fptr->data;
+			fptr = fptr->next;
+			printf("%s,%s\n", f->filename, b->bundle_name);
+		}
+	}
+}
+
 static void print_final_results(bool display_size)
 {
 	struct bundle_result *b;
@@ -310,6 +330,7 @@ static void print_help(const char *name)
 	fprintf(stderr, "   -b, --binary            Search paths where binaries are located for a match\n");
 	fprintf(stderr, "   -s, --scope=[query type] 'b' or 'o' for first hit per (b)undle, or one hit total across the (o)s\n");
 	fprintf(stderr, "   -t, --top=[NUM]         Only display the top NUM results for each bundle\n");
+	fprintf(stderr, "   -m, --csv               Output all results in CSV format (machine-readable)\n");
 	fprintf(stderr, "   -d, --display-files	   Output full file list, no search done\n");
 	fprintf(stderr, "   -i, --init              Download all manifests then return, no search done\n");
 	fprintf(stderr, "   -I, --ignore-time       Ignore system/certificate time when validating signature\n");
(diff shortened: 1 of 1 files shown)


matthewrsj:
Fixes #391 
Fixes #392 
Fixes #393

src/search.c
```diff
@@ -153,7 +150,83 @@ static void free_bundle_result_data(void *data)
 	free(br);
 }
 
-static void print_final_results(void)
+/* unmark all bundle_results in bl as seen */
+static void unsee_bundle_results(struct list *bl)
```
tmarcu: s/unsee/unmark?

matthewrsj: I appreciate the review, but this is the commit from #399. Can you review that commit there? This PR is based on that one.

Also, sure I suppose. I kinda like unsee though since we are tracking if we've seen the bundle. The mark is only to tell us if we've seen it. So if the function name is unmark_* what are we unmarking?

src/search.c
```diff
@@ -170,10 +243,16 @@ static void print_final_results(void)
...
-		       b->bundle_name,
-		       b->size,
-		       b->is_tracked ? "\tinstalled" : "");
+		/* do not print the size information when the scope is only one bundle ('o')
+		 * because we did not load all bundles and therefore do not have include sizes
+		 * for the result */
+		printf("Bundle %s\t%s", b->bundle_name, b->is_tracked ? "[installed]\t" : "");
+		if (scope != 'o') {
+			printf("(%li MB%s)",
+			       b->size / 1000 / 1000, /* convert from bytes->KB->MB */
```
tmarcu: why 1000 and not 1024?

phmccarty: it depends if we want to display sizes in `MB` of `MiB` units...

matthewrsj: Yes, I am reporting MB

src/search.c
```diff
@@ -226,16 +243,19 @@ static void apply_size_penalty(struct list *bundle_info)
...
-static void print_final_results(char scope)
+static void print_final_results()
 {
 	struct bundle_result *b;
 	struct list *ptr;
 	int counter = 0;
 
-	ptr = results;
+	ptr = list_head(results);
 	/* TODO: make 5 a constant, potentially configurable during build or runtime
```
icarus-sparry: Out of date TODO

src/search.c
```diff
@@ -379,6 +405,18 @@ static bool parse_options(int argc, char **argv)
 				scope = 'o';
 			}
 
+			break;
+		case 't':
+			if (!optarg) {
```
icarus-sparry: getopt_long will return an error indication if ```-t``` doesn't have an argument

src/search.c
```diff
@@ -655,9 +693,15 @@ static void do_search(struct manifest *MoM, char search_type, char *search_term)
...
 		fprintf(stderr, "Search term not found.\n");
 	}
+
+	// need recursive size regardless of num_results
 	apply_size_penalty(bundle_info);
-	sort_results();
-	print_final_results(scope);
+	if (num_results != INT_MAX) {
+		sort_results();
+	} else {
```
icarus-sparry: Personally I would like the results sorted regardless.

src/search.c
matthewrsj: Thanks, I'll remove

src/search.c
matthewrsj: If you look in the `else` clause I do `results = list_sort(results, bundle_size_cmp);` which sorts by bundle size only.

src/search.c
matthewrsj: We expect it to. `--top` requires `NUM` as an argument, or am I misunderstanding your comment?

icarus-sparry: If you say ```wibble -t``` (i.e. without an argument to tthen getopt_long will return `?` rather than `t`. If it gets to the ```case 't':``` case then optarg will point somewhere.

src/search.c
icarus-sparry: Thanks, I had the test in the ```if``` reversed in my mind.

PR #411, opened 2018-03-16
Grep pkg

tmarcu:


swupd-add-pkg
```diff
@@ -71,7 +71,7 @@ if [[ -n "$pkg" && -n "$bundle" ]]; then
 # [MAINTAINER]: Mixer User" | sudo tee "$BUNDLE_DIR/$bundle" > /dev/null
 	fi
 
-	grep "$pkg" "$BUNDLE_DIR/$bundle"
+	grep -x "$pkg" "$BUNDLE_DIR/$bundle"
```
icarus-sparry: You could have thrown in a `-F` so that any interesting characters in $pkg like `*` would not cause issues.

PR #414, opened 2018-03-18
Fix for issue 412

icarus-sparry:


src/manifest.c
```diff
@@ -239,9 +240,20 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
 		}
 		if (strncmp(line, "filecount:", 10) == 0) {
 			filecount = strtoull(c, NULL, 10);
+			if (filecount > 1000000) {
```
matthewrsj: We should think about the Manifest.full case. Right now there are currently over 427k files listed in Manifest.full for Clear release 21300. We should probably increase the limit here to account for that number increasing in the future.

src/manifest.c
```diff
@@ -239,9 +240,20 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
...
+				fprintf(stderr, "Error: preposterous (%" PRIu64 ") number of files in %s Manifest, more than a million skipping\n",
+					filecount, component);
+				goto err_close;
+			}
 		}
 		if (strncmp(line, "contentsize:", 12) == 0) {
 			contentsize = strtoull(c, NULL, 10);
+			/* if (contentsize > 2000000000) {
```
matthewrsj: Is there a reason this has been commented out? I think 2TB is a pretty reasonable number right now considering 21300 full contentsize is just over 14GB.

src/manifest.c
icarus-sparry: Sure, I actually took that number, doubled it to allow for future growth and then rounded it up to be a nice round number.

We probably shouldn't make it any bigger than 2**32/sizeof(struct file) (currently 176 on a 64/32/64 PIL build), or 12 million because of the hinted at integer overflow issues.

How about 4 million (with a comment) to allow Manifest.full to get an order of magnitude bigger than it currently is?

matthewrsj: +1 for 4 million with a comment

src/manifest.c
icarus-sparry: Yes, it makes too many tests fail!
The content-size in the MoM is pretty bogus - 13,805,671,819

matthewrsj: That... is wrong. If you look at published MoMs (https://download.clearlinux.org/update/21410/Manifest.MoM for one) contentsize is always created as 0.

matthewrsj: I'm now seeing all of our test cases define MoM contentsize to that number... those should all be 0, I wonder why they were initialized to that at all.

matthewrsj:
You have to update the test you added to report "more than 4 million" instead of "more than a million"

PR #415, opened 2018-03-22
Completion

icarus-sparry:


icarus-sparry:
The first change is actually a bug fix, if we have a lot of bundles (think pundles) the list-bundles outputs random garbage.

To reproduce, as root do

       cd /usr/share/clear/bundles
       for i in {1..1000} ; do touch XXX$i ; done
       swupd bundle-list
       # note the junk
       rm XXX[0-9]*
       swupd bundle-list
       # Note you still have junk
       cd ..
       cp -rp bundles{,.new}
       rm -rf bundles
       mv bundles{.new,}

The second patch saves a copy of the MoM in /usr/share/clear/bundles/.MoM, makes the completion code use it. It also checks in the completion code directly, rather than generating it, removes the generation script, and adds some crude filename completion to swupd hashdump, the latter is still a WIP.

src/bundle.c
```diff
@@ -928,13 +928,15 @@ int list_local_bundles()
 		return EXIT_FAILURE;
 	}
 	while ((ent = readdir(dir))) {
+		/* Should we just skip everything starting with "." ? */
```
matthewrsj: Seems reasonable to me.

src/subscriptions.c
```diff
@@ -84,7 +84,9 @@ void read_subscriptions(struct list **subs)
 	dir = opendir(path);
 	if (dir) {
 		while ((ent = readdir(dir))) {
-			if ((strcmp(ent->d_name, ".") == 0) || (strcmp(ent->d_name, "..") == 0)) {
+			if ((strcmp(ent->d_name, ".") == 0) ||
```
matthewrsj: could probably just ignore everything starting with `.` here too.

icarus-sparry: Yes. As agreed in IRC we will ignore all "dot" files, there are a few other places as well where we have this kind of logic, but didn't specifically affect adding a ".MoM" file.

matthewrsj:
lgtm after you update the bundledir reading logic to ignore all dotfiles.

PR #422, opened 2018-04-13
Update "swupd verify" documentation to cover all options

gvancuts:
Add "swupd verify" options that were missing from the documentation
(swupd.1.rst) but listed by the tool when calling "swupd verify -h".
The order was also slightly changed to be identical to the order used
by "swupd verify -h".

Signed-off-by: Geoffroy Van Cutsem <geoffroy.vancutsem@intel.com>

### DIFF
--- a/docs/swupd.1
+++ b/docs/swupd.1
@@ -380,29 +380,19 @@ control of the software update program are verified according to the
 manifest data
 .INDENT 0.0
 .IP \(bu 2
+\fI\-m, \-\-manifest\fP
+.INDENT 2.0
+.INDENT 3.5
+Verify against manifest version M.
+.UNINDENT
+.UNINDENT
+.IP \(bu 2
 \fI\-f, \-\-fix\fP
 .INDENT 2.0
 .INDENT 3.5
 Correct any issues found. This will overwrite incorrect file
 content, add missing files and do additional corrections, permissions
 etc.
-.UNINDENT
-.UNINDENT
-.IP \(bu 2
-\fI\-i, \-\-install\fP
-.INDENT 2.0
-.INDENT 3.5
-Install all files into {path} as specified by the \fI\-\-path={path}\fP
-option. Useful to generate a new system root, or verify side
-by side.
-.UNINDENT
-.UNINDENT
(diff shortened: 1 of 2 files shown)


matthewrsj:
Can you please run `make` to generate the `swupd.1` man page and check that file in as well?

matthewrsj:
Well, you seem to be missing the `bsdiff` package (https://github.com/clearlinux.bsdiff). In Clear Linux this means you just need to `swupd bundle add os-core-update-dev`. This bundle will give you all the build dependencies necessary to build swupd-client. In fedora you will have to manually install all the necessary packages.

Re: this PR. All the other subcommands only call out the flags specific to that subcommand and not all the flags that are global to all subcommands. I think I would prefer to reorganize the man page to make this more obvious rather than copying around the globals to each subcommand.

gvancuts:
> Re: this PR. All the other subcommands only call out the flags specific to that subcommand and not all the flags that are global to all subcommands. I think I would prefer to reorganize the man page to make this more obvious rather than copying around the globals to each subcommand.

Makes sense to me. I have removed all options that are global so only `-m, --manifest` and `-x, --force` are effectively added to the man page. I had a doubt as to whether I should keep `-x, --force` as there is a similar option that's global (`-f, --force`) but given the short version is different, I have kept it in (and I'm assuming they are slightly different?)

gvancuts:
> Well, you seem to be missing the bsdiff package (https://github.com/clearlinux.bsdiff). In Clear Linux this means you just need to swupd bundle add os-core-update-dev. This bundle will give you all the build dependencies necessary to build swupd-client. In fedora you will have to manually install all the necessary packages.

That's it, I had `os-core-update` and missed the fact I also needed `os-core-update-dev`. As for Fedora, the tool is installed (`bsdiff`) but there is no package providing this library, is this a specificity of the bsdiff Clearlinux fork? (I did not realise this package had been forked for Clearlinux, thanks for the pointer).

matthewrsj:
We do need the clearlinux fork for bsdiff. There are several optimizations and built-in failure cases that we need. @phmccarty is the maintainer of that project.

PR #424, opened 2018-04-14
Allow to use "update -m M"

cmarcelo:
See commit messages for details.


src/version.c
```diff
@@ -161,6 +162,17 @@ int check_versions(int *current_version,
...
+			fprintf(stderr, "Requested version for update (%d) must be greater than current version (%d)\n",
+				requested_version, *current_version);
+			return -1;
+		}
+		if (requested_version < *server_version) {
+			*server_version = requested_version;
+		}
+	}
+
 	//TODO allow policy layer to send us to intermediate version?
```
matthewrsj: This isn't my full review yet, but looks like this TODO isn't needed anymore, since this PR adds a way to get to an intermediate version.

src/main.c
```diff
@@ -1,7 +1,7 @@
 /*
  *   Software Updater - client side
  *
- *      Copyright © 2012-2016 Intel Corporation.
```
matthewrsj: Why this change? Why only to 2016 if you are updating the `from`?

src/version.c
```diff
@@ -161,6 +162,17 @@ int check_versions(int *current_version,
...
+	if (requested_version != -1) {
+		if (requested_version <= *current_version) {
+			fprintf(stderr, "Requested version for update (%d) must be greater than current version (%d)\n",
+				requested_version, *current_version);
+			return -1;
+		}
+		if (requested_version < *server_version) {
+			*server_version = requested_version;
+		}
+	}
```
matthewrsj: Just so I make sure I'm understanding, if `requested_version` is greater than `server_version` you just get the `server_version`. That's good, should it print something out that you *can't* go to the requested version?

cmarcelo: It is correct. It also works that way to not get on the way of updates that involved multiple formats (and re-executes). Because of that, unless we do more work, there's no clear place to print this -- sometimes we CAN go to the requested version, if we re-execute.

Let's say 100 -> 190 are in format 1, and 200 -> 280 are in format 2. The bump is 190-200. And we start from 150.

- `-m 140` fails because we can't rollback
- `-m 150` fails because it would be a nop
- `-m 160` goes to 160 since it is < than server version (190).
- `-m 250` goes to server_version (190), and then re-executes.
- `-m 200` goes to server_version (190), and doesn't re-execute, since going to 190, it sees itself as 200, and that is what was requested
- `-m 300` goes to server version (190), re-executes, goes to new server version (280), and that is it.

cmarcelo: We could add a note if we are not re-executing and requested_version > server_version (after the reexec check says it is not reexecing). I'm not sure how useful it is since we already explicit state in the output to what version we are really going.

src/version.c
cmarcelo: Removed.

src/main.c
cmarcelo: You are reviewing "files changed" and GitHub gets confused about the file renaming.

What is really happening is that we first moved contents of main.c to update.c, and then renamed swupd.c to main.c.

cmarcelo:
Updated manual, removed TODO.

src/version.c
matthewrsj: > we already explicit state in the output to what version we are really going.

Ok

src/main.c
matthewrsj: Ah weird. ok I see it correctly when I view the commits.

cmarcelo:
To check that I'd try to retrieve the Manifest.MoM of the version, which is essentially what the program is currently doing. It seems "wrong" because it is retrying, but I guess the reason to retry in the first place was to avoid transient errors.

I can do something else, I'm not really sure what: maybe if there is a requested version, we have a different error message?

matthewrsj:
Maybe if the return code is 404 don't try again. I can't think of a time when 404 would be a transient error.

cmarcelo:
Added an extra patch that avoids the retries in the case mentioned.

matthewrsj:
lgtm, thanks @cmarcelo

PR #426, opened 2018-04-16
Add test to ensure man pages are up to date

### LINKED ISSUE #316: Remove generated man pages from repo
As brought up in #311 by @icarus-sparry, we should not be checking man pages into the git repository:

> We ought to delete the man pages from the repo or else scrap the idea of using rst2man.  The current middle ground of guess what timestamps git is going to give you is just wrong.

We should remove them and add them to the .gitignore. This approach will require users to either have rst2man.py installed or they will have to pass something like the `--disable-manpages` flag introduced in #309.

### LINKED ISSUE #425: "make clean" deletes man pages which are part of the repository
The generated man pages are part of the repository, however "make clean" is currently deleting them, causing the work dir to have changes unrelated to changes I'm working on. My intuition would say just get rid of the `clean-local` rule, but I might be missing some nuance of autotools.

Is there a better way to solve this?

icarus-sparry:
Current idea is to have the manual pages stored in the repo in nroff
format, as well as restructured test format. Users can just use the
nroff. Add test to ensure that the pages match.

Fixes #316
Fixes #425

### DIFF
--- a/test/documentation/manpages/test.bats
+++ b/test/documentation/manpages/test.bats
@@ -1,0 +1,29 @@
+#!/usr/bin/env bats
+
+setup() {
+  # Ensure that we have the conversion program
+  type -t rst2man.py > /dev/null
+  # Change to the toplevel git directory
+  type -t git >/dev/null || skip
+  cd "$(git rev-parse --show-toplevel)"
+}
+
+@test "Check manual pages are up to date" {
+  for rst in docs/*.[1-9].rst
+  do
+    [ -f "$rst" ] || continue	# Could use shopt -s nullglob
+    echo "$rst"
+    rst2man.py "$rst" | cmp  "${rst%.rst}" -
+  done
+}
+
+@test "Check manual pages are clean" {
+  for nroff in docs/*.[1-9]
+  do
+    [ -f "$nroff" ] || continue
+    echo "$nroff"
+    git diff-index --quiet HEAD "$nroff"
+  done
+}
+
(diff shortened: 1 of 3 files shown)


cmarcelo:
The second patch does some whitespace fixes to the first. Squash that chunk into the first patch -- or just squash both patches if you prefer.

LGTM.

icarus-sparry:
Yes, I plan to squash both together if the changes are accepted.

I need to add emacs modelines to the files to stop the need to do the whitespace change as I switch editors!

test/documentation/manpages/test.bats
```diff
@@ -9,11 +9,21 @@ setup() {
...
+  for rst in ${BATS_TEST_DIRNAME}/../../../docs/*.[1-9].rst
+  do
+    echo "$rst"
+    rst2man.py "$rst" | cmp  "${rst%.rst}" -
+  done
 }
 
+@test "Check manual pages are clean" {
+  type -t git >/dev/null || skip
+  cd $(git rev-parse --show-toplevel)
```
matthewrsj: Should we use this for the first test as well? Looks cleaner than `${BATS_TEST_DIRNAME}/../../../docs.....`

test/documentation/manpages/test.bats
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bats
+
+setup() {
+  # Ensure that we have the conversion program
+  type -t rst2man.py > /dev/null
+}
+
+# teardown() {
```
matthewrsj: Commented out teardown function

Issue #322, opened 2017-10-20
configurable download retries


pohly:
I am writing tests for issue #249 that simulate network errors. Right now, the tests run very slowly due to the hard-coded retry interval of 10 seconds and MAX_TRIES = 3.

Can interval and max retries be made configurable?

How do you handle this in your own tests?

On a related note, I am seeing "swupd update" trying to download Manifest.MoM.tar four times, for an overall duration of 1:10min. It all depends on the definition of MAX_TRIES of course, but to me this looks like an off-by-one error.

If we had a command line option --max-tries=3, I would expect three GET requests at most. For --max-retries=3, one initial GET followed by three retries would also make sense.

It's trying the second manifest only once.

HTTP log:
```
2017-10-20 14:32:04,915 - oe-selftest - INFO - "GET /10/Manifest.MoM.tar HTTP/1.1" 500 -
2017-10-20 14:32:14,776 - oe-selftest - INFO - "GET /10/Manifest.MoM.tar HTTP/1.1" 500 -
2017-10-20 14:32:34,717 - oe-selftest - INFO - "GET /10/Manifest.MoM.tar HTTP/1.1" 500 -
2017-10-20 14:33:14,724 - oe-selftest - INFO - "GET /10/Manifest.MoM.tar HTTP/1.1" 500 -
2017-10-20 14:33:14,752 - oe-selftest - INFO - "GET /20/Manifest.MoM.tar HTTP/1.1" 500 -
````

Output:
```
Attempting to download version string to memory
Update started.
Preparing to update from 10 to 20
Failed to retrieve 10 MoM manifest
Retry #1 downloading from/to MoM Manifests
Failed to retrieve 10 MoM manifest
Retry #2 downloading from/to MoM Manifests
Failed to retrieve 10 MoM manifest
Retry #3 downloading from/to MoM Manifests
Failed to retrieve 10 MoM manifest
Failure retrieving manifest from server
Update took 69.8 seconds
```

This is with file:// version URL and http:// content URL. Conceptually this tests mirrors a functioning version server while the content server has some internal issue.

pohly:
Could be implemented separately or together with issue #211.

phmccarty:
I would love to see an option to disable the retry timeouts entirely. Right now, there is one functional test that intentionally hits download retries, so such an option would be useful for that purpose.

Making them configurable also sounds desirable.

matthewrsj:
[...]

Issue #221, opened 2017-06-02
provide a comand to show files into the bundle, installed and available


rlins10:
It would be nice to nice to have  "--listfile" options to show files inside a bundle that is installed and ones that are available but not installed.

(Jonh)

matthewrsj:
This idea for installed bundles would basically be a cat of a manifest that should already exist in the statedir.

This idea for available bundles is basically what swupd search is for. I don't see the use case for someone wanting to see the entire file list for a bundle without having an idea of what they are looking for.

These file lists are almost always verrrry long. It doesn't make sense to me for swupd to be outputting this in the console. Open the manifest file in your state directory in an editor or visit https://download.clearlinux.org (or whatever upstream) and view it there.

xinyunliu:
Why this feature removed? Does clearlinux provide a method to list files in a bundle?

PR #427, opened 2018-04-19
Avoid retries in local downloads for manifests and packs

cmarcelo:
If content_url is local, do not attempt certain certain download
retries. I've focused on changing only the spots that caused 'make
check' to sleep as a first step. But I agree the retry logic needs
some extra care as posted in #221.

This patch reduces the time of running 'make check' in my system from
2m44s to 24s.

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -69,6 +69,7 @@ bool verbose_time = false;
 bool have_manifest_diskspace = false; /* assume no until checked */
 char *version_url = NULL;
 char *content_url = NULL;
+bool content_url_is_local = false;
 char *cert_path = NULL;
 long update_server_port = -1;
 char *swupd_cmd = NULL;
@@ -314,7 +315,12 @@ int set_content_url(char *url)
 		return 0;
 	}
 
-	return set_url(&content_url, url, DEFAULT_CONTENT_URL_PATH);
+	int ret = set_url(&content_url, url, DEFAULT_CONTENT_URL_PATH);
+	if (ret == 0) {
+		content_url_is_local = strncmp(content_url, "file://", 7) == 0;
+	}
+
+	return ret;
 }
 
 /* Initializes the version_url global variable. If the url parameter is not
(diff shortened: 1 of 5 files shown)


src/globals.c
```diff
@@ -314,7 +315,12 @@ int set_content_url(char *url)
 		return 0;
 	}
 
-	return set_url(&content_url, url, DEFAULT_CONTENT_URL_PATH);
+	int ret = set_url(&content_url, url, DEFAULT_CONTENT_URL_PATH);
+	if (ret == 0) {
+		content_url_is_local = strncmp(content_url, "file://", 7) == 0;
+	}
```
icarus-sparry: Whilst this obviously works, wouldn't it be better to have an option to set a variable (maybe MAX_TRIES) rather than adding in magic based off the URL? I could see a hypothetical test in the future which was testing the retry facility creating the to-be-downloaded file after a certain amount of time, and I can imagine a user wanting to set a different  MAX_TRIES if their internet was flaky and they wanted to try really hard to download stuff.

matthewrsj: https://github.com/clearlinux/swupd-client/issues/322

matthewrsj: And I agree with @icarus-sparry that is a better solution. It is the solution we want to go with to fix the 'make check' problem of taking too long.

cmarcelo: > Whilst this obviously works, wouldn't it be better to have an option to set a variable (maybe MAX_TRIES) rather than adding in magic based off the URL?

I think by default we shouldn't have retries for local files, and by default we should have retries for remote files. We identify those by the URL scheme. So as I see the highlighted code will exist in some form or another.

About the hypothetical test case: you can still do it if you make the configurable number of retries override any previous existing behavior. So I see a fix #322 complementing the _behavior_ added by this patch -- not in conflict. You can also use a local webserver (in case #322 don't materialize).

If that was not enough, given the size of the change I'd strongly consider just adding it for the sake of 'make check' times.

cmarcelo: To clarify:

> If that was not enough, given the size of the change I'd strongly consider just adding it for the sake of 'make check' times.

I.e. further improvement to retry logic can easily revert the check and use a different condition instead if that turns out to be the case.

matthewrsj: Reasonable.

PR #428, opened 2018-04-24
Provide a configurable mirror URL setting

matthewrsj:
Provide a mechanism and CLI subcommand to allow users to set a configurable mirror URL on their system. See individual commit messages for more details.

### DIFF
--- a/docs/swupd.1.rst
+++ b/docs/swupd.1.rst
@@ -189,6 +189,27 @@ SUBCOMMANDS
 
     Shows the current OS version and the URLs used for updates.
 
+``mirror``
+
+    Configure a mirror URL for swupd to use instead of the defaults on the
+    system or compiled into the swupd binary.
+
+    - `-s, --set URL`
+
+        Set the content and version URLs to URL by adding configuration files to
+        <path>/etc/swupd/mirror_contenturl and
+        <path>/etc/swupd/mirror_versionurl
+
+    - `-u, --unset`
+
+        Remove the content and version URL configuration by removing
+        <path>/etc/swupd
+
+    - `-p, --path`
+
+        Optionally set the top-level directory for the swupd-managed system.
+        If not specified this will default to '/'
+
 ``search {string}``
 
     Search for matching paths in manifest data. The specified {string}
--- a/docs/swupd.1.rst
+++ b/docs/swupd.1.rst
@@ -185,6 +185,10 @@ SUBCOMMANDS
         Specify the path to use for operations. This can be used to
         point to a chroot installation of the OS or a custom mount.
 
+``info``
+
+    Shows the current OS version and the URLs used for updates.
+
 ``search {string}``
 
     Search for matching paths in manifest data. The specified {string}
@@ -334,11 +338,6 @@ SUBCOMMANDS
     - `-x, --force`
 
         Attempt to proceed even if non-critical errors found.
-
-
-``info``
-
-    Shows the current OS version and the URLs used for updates.
 
 
 EXIT STATUS
--- a/src/mirror.c
+++ b/src/mirror.c
@@ -1,0 +1,236 @@
+/*
+ *   Software Updater - client side
+ *
+ *      Copyright © 2018 Intel Corporation.
+ *
+ *   This program is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License as published by
+ *   the Free Software Foundation, version 2 or later of the License.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#define _GNU_SOURCE
+#include <assert.h>
+#include <errno.h>
+#include <getopt.h>
+#include <libgen.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "swupd.h"
+
+static void print_help(const char *name)
+{
+	fprintf(stderr, "Usage:\n");
+	fprintf(stderr, "   swupd %s [options] [mirror_url]\n\n", basename((char *)name));
+	fprintf(stderr, "Help Options:\n");
+	fprintf(stderr, "   -h, --help              Show help options\n");
+	fprintf(stderr, "   -s, --set               set mirror url\n");
+	fprintf(stderr, "   -u, --unset             unset mirror url\n");
+	fprintf(stderr, "   -p, --path=[PATH...]    Use [PATH...] as the path to set/unset the mirror url for (eg: a chroot or btrfs subvol\n");
+	fprintf(stderr, "\n");
+}
+
+static const struct option prog_opts[] = {
+	{ "help", no_argument, 0, 'h' },
+	{ "set", required_argument, 0, 's' },
+	{ "unset", no_argument, 0, 'u' },
+	{ "path", required_argument, 0, 'p' },
+	{ 0, 0, 0, 0 }
+};
+
+static int unset_mirror_url()
+{
+	char *fullpath;
+	if (path_prefix != NULL) {
+		fullpath = mk_full_filename(path_prefix, "/etc/swupd");
+	} else {
+		string_or_die(&fullpath, "/etc/swupd");
+	}
+
+	if (fullpath == NULL ||
+	    strcmp(fullpath, "/") == 0 ||
+	    strcmp(fullpath, "/etc") == 0) {
+		fprintf(stderr, "Invalid mirror configuration path\n");
+		free_string(&fullpath);
+		return 1;
+	}
+	int ret = swupd_rm(fullpath);
+	free_string(&fullpath);
+	return ret;
+}
+
+static int write_to_path(char *content, char *path)
+{
+	char *dir, *tmp = NULL;
+	string_or_die(&tmp, "%s", path);
+	dir = dirname(tmp);
+	/* attempt to  make the directory
+	 * ignore EEXIST errors */
+	int ret = mkdir(dir, S_IRWXU);
+	char *cmd;
+	string_or_die(&cmd, "mkdir -p %s", dir);
+	ret = system(cmd);
+	free_string(&cmd);
+	free_string(&tmp);
+	if (ret) {
+		return ret;
+	}
+
+	/* now try to open the file to write */
+	FILE *fp = NULL;
+	fp = fopen(path, "w");
+	if (fp == NULL) {
+		return 1;
+	}
+
+	/* and write to the file */
+	ret = fputs(content, fp);
+	fclose(fp);
+	if (ret < 0 || ret == EOF) {
+		return 1;
+	}
+
+	return 0;
+}
+
+static int set_mirror_url(char *url)
+{
+	char *content_path;
+	char *version_path;
+	int ret = 0;
+	bool need_unset = false;
+	/* concatenate path_prefix and configuration paths if necessary */
+	if (path_prefix != NULL) {
+		content_path = mk_full_filename(path_prefix, MIRROR_CONTENT_URL_PATH);
+		version_path = mk_full_filename(path_prefix, MIRROR_VERSION_URL_PATH);
+	} else {
+		string_or_die(&content_path, "%s", MIRROR_CONTENT_URL_PATH);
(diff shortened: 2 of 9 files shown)


matthewrsj:
I realize this needs man page updates.

matthewrsj:
Man page section added.

src/mirror.c
```diff
@@ -0,0 +1,233 @@
...
+	if (path_prefix != NULL) {
+		fullpath = mk_full_filename(path_prefix, "/etc/swupd");
+	} else {
+		string_or_die(&fullpath, "/etc/swupd");
+	}
+
+	assert(fullpath != NULL);
+	assert(strcmp(fullpath, "/") != 0);
+	assert(strcmp(fullpath, "/etc") != 0);
+	return swupd_rm(fullpath);
```
icarus-sparry: Leaking the memory that fullpath points to?

src/mirror.c
```diff
@@ -0,0 +1,233 @@
...
+	string_or_die(&tmp, "%s", path);
+	dir = dirname(tmp);
+	/* attempt to  make the directory
+	 * ignore EEXIST errors */
+	int ret = mkdir(dir, S_IRWXU);
+	char *cmd;
+	string_or_die(&cmd, "mkdir -p %s", dir);
+	ret = system(cmd);
+	if (ret) {
+		free_string(&cmd);
```
icarus-sparry: Could just

     ret = system(cmd);
     free_string(&cmd);
     free_string(&tmp);
     if (ret) {
          return ret;
     }

src/mirror.c
```diff
@@ -0,0 +1,233 @@
...
+	/* now try to open the file to write */
+	FILE *fp = NULL;
+	fp = fopen(path, "w");
+	if (fp == NULL) {
+		return 1;
+	}
+
+	/* and write to the file */
+	ret = fputs(content, fp);
+	if (ret < 0 || ret == EOF) {
```
icarus-sparry: If we
      ret = fputs(content, fp);
      fclose(fp);
      if (ret < 0 || ret == EOF) { ....

we only need one fclose

src/mirror.c
```diff
@@ -0,0 +1,233 @@
...
+			exit(EXIT_SUCCESS);
+		case 'p':
+			if (!set_path_prefix(optarg)) {
+				fprintf(stderr, "Invalid --path argument\n\n");
+				ret = EINVALID_OPTION;
+				goto out;
+			}
+			break;
+		case 's':
+			if (unset) {
```
icarus-sparry: I think if I was doing this I would have a function pointer to point to either a wrapper around `set_mirror_url` or `unset_mirror_url`, and make the error message more general than `"cannot set and unset at the same time"`, perhaps `"you can only have a maximum of one set ot unset"`. As it is you can `-s` a dozen times without warning.

src/mirror.c
```diff
@@ -0,0 +1,233 @@
...
+int mirror_main(int argc, char **argv)
+{
+	int ret = 0;
+	copyright_header("mirror");
+	ret = parse_options(argc, argv);
+	if (ret != 0) {
+		return ret;
+	}
+
+	if (set != NULL) {
```
icarus-sparry: Rather than this if/elseif I would just call my function pointer. The wrapper around `set_mirror_url()` would be essentially lines 206 to 211.

fenrus75:
So if we are going to support mirrors we need to do this smart, basically
1) Only for Mom and packs, but never use the mirror for fullfiles; full
files are our safety fallback.
  (We can debate the MoM since the MoM is tiny AND security sensitive.
Actually I think I prefer the MoM and signature to not come from the
mirror. so mirror is only for packs)
2) Once we get a 404, we should retry the non-mirror URL
3) If the non-mirror URL then succeeds, when should from then on no longer
use the mirror
4) if we get a hash mismatch anywhere, we must stop using the mirror (e.g.
assume compromised mirror)

Mirrors get stale at times, and we must survive even in the light of a
stale or malicious mirror

fenrus75:
I also think we should not provide a mirror for versionurl.
versionurl is like 10 bytes of content, not worth the mirror, but letting someone go stale is a security risk

matthewrsj:
@fenrus75 I understand your concerns, but this is in response to an issue seen by users working with Clear Linux in PRC. They are seeing extremely long update and installation times (like, do an install over the weekend type times). I'm willing to talk about a better approach as a next step but I think we should fix their issues first without coming up with a solution that "picks and chooses" where their content should come from.

Also, if someone decides to use a mirror, I think it is fair to say they are accepting any risks associated with using the mirror. In the specific use-case I'm addressing they know they can trust the mirror they are using. I'm of the opinion that if a user sets a mirror all content should come from the mirror and not unexpectedly fall back to the CDN.

icarus-sparry:
I think I disagree with all of @fenrus75 points. Presumably the reason for mirrors is to improve bandwidth or work around censorship issues. So let us consider someone who has 1000 machines to update behind a low bandwidth external connection.

1. Why shouldn't mirrors be used for fullfiles? The hash check still protects the content.
1. The fact that a mirror is missing one file doesn't mean that it is not still useful for other files.
1. If we get something from the non-mirror, this doesn't imply that that we want all 1000 machines all getting everything via the slow link.
1. Obviously we shouldn't use anything with a hash mismatch. But one file being wrong (perhaps the mirror was being updated and the file was not being changed atomically) does not invalidate all the other files.

fenrus75:
the point of a mirror is we fall back.
if you don't want a fallback you just do -u and use the url period....

we really should fall back.

in terms of long times, that is presumably all about packs... since those
are big. version files etc are tiny. it's the packs that are huge.

bryteise:
Hrm so I imagine that the major pain point is install (otherwise -u would be the easy fix) so packs only seems to make sense to me. We can always reassess at a later point if we need something else too.

matthewrsj:
After speaking with @fenrus75 this morning we agree this is an acceptable first step. In short-term plans we want to WARN the user if their mirror is out of date and potentially IGNORE the mirror if it becomes *too far* out of date, maybe a month or two, since this indicates a dead mirror.

Another interesting short-term plan can be to auto-detect a mirror using `gethostbyname` for `linux-ftp` and use that automatically if configured to do so.

Slightly longer-term we can provide the fallbacks to CDN, potentially a different mirror for packs-only, etc.

nesiusra:
I disagree that the point of mirroring is to fall back. The point of mirroring is nearly always to improve performance by using a cacheing pattern that protects clients from a suboptimal network experience while in-turn protecting a suboptimal network link from the clients trying to all move the same content over that link. 

This is feature that lets system owners manage their systems and computing environment. I'm fine with swupd falling-back to the default source as long as we let people opt-out of that the way we let them opt-out of privacy filtering for telemetrics, etc...   Personally I'd prefer to allow admins to opt-in to auto-fallback.

nesiusra:
Just a brainstorm... perhaps our mirror file should be a list of mirrors consulted in-order.

icarus-sparry:
Just a brainstorm... perhaps our mirror file should be a list of mirrors consulted in-order.

perfectly reasonable suggestion. More extreme brainstorming -  We could add grouping of mirrors, doing http HEAD requests in parallel and picking the fastest - building and testing this could take me to retirement :-)

matthewrsj:
@icarus-sparry ping.

src/mirror.c
```diff
@@ -0,0 +1,236 @@
...
+	}
+	int ret = swupd_rm(fullpath);
+	free_string(&fullpath);
+	return ret;
+}
+
+static int write_to_path(char *content, char *path)
+{
+	char *dir, *tmp = NULL;
+	string_or_die(&tmp, "%s", path);
```
icarus-sparry: For future reference, if all you are doing in the second argument of string_or_die is `"%s"`, and you are not going to alter the string, then you can just skip the whole allocate and clean-up dance, and just use `path` directly.

Not worth changing now.

src/mirror.c
```diff
@@ -0,0 +1,236 @@
...
+	string_or_die(&cmd, "mkdir -p %s", dir);
+	ret = system(cmd);
+	free_string(&cmd);
+	free_string(&tmp);
+	if (ret) {
+		return ret;
+	}
+
+	/* now try to open the file to write */
+	FILE *fp = NULL;
```
icarus-sparry: I worry about a symlink attack here. If you imagine a system that doesn't allow root access, but has sudo configured to allow swupd to be run, then you could create say /tmp/etc/swupd as a symbolic link to /etc/passwd, then use 'sudo swupd mirror -p /tmp 'some interesting stuff to put into /etc/passwd'`

Not sure if this is a real concern, given a typical usage pattern for swupd.

docs/swupd.1.rst
```diff
@@ -189,6 +189,26 @@ SUBCOMMANDS
...
+        <path>/etc/swupd/mirror_versionurl
+
+    - `-u, --unset`
+
+        Remove the content and version URL configuration by removing
+        <path>/etc/swupd
+
+    - `-p, --path`
+
+        Set the path prefix under which swupd will operate.
```
icarus-sparry: I would like a little more clarity on what is optional and what is required. So

      -p PATH
       set the optional top level directory for the swupd managed system. If not specified willuse the one in /

src/mirror.c
matthewrsj: considering the fact that you can run `sudo swupd verify --fix -p /some/other/mounted/system -u http://anwhere.you.want/update` you can pretty much own someone else's system no matter what if we are considering this attack pattern.

nesiusra: Well @icarus-sparry  - now that you say that - what's the purpose of --path?  What's the use case that motivates us to allow these files to live somewhere besides /etc/swupd/

nesiusra: I think it's one thing to tell swupd "look here for configs" but another to tell swupd to "write files here..." - there's no reason the sysadmin can't land the files on their own?

nesiusra: @matthewrsj  ^^

matthewrsj: The purpose of --path in this case is so an admin can set mirror URLs on other systems perhaps, or to support writing these files during an install.

icarus-sparry: The purpose is for things like the installer, where you are booted from one device but want to configure a second device to be the thing providing `"/"`.

Since @nesiusra is old enough, I can point out also things like SunOS `nd` devices.

nesiusra: The installer could land this file itself as easily as it could ask swupd to do it?

matthewrsj: The installer isn't necessarily the user of `swupd verify --install` though. I have several times installed a system to a directory for testing purposes on my own machine. I then want to keep passing --path to subsequent commands to make sure operations happen on that system. If I am somewhere the main network is slow (PRC) I'll also want the mirror configuration to be set for that installed system.

nesiusra: I think --path is fine for telling swupd where to read its configs from. It's when writing configs I get nervous for reasons Icarus pointed out.

matthewrsj: And I would argue that we don't care, considering 1) that it is an unusual setup for someone to have sudo configured to allow swupd to run yet not allow the user root access on the system and 2) if a user can run swupd they can do any arbitrary damage to their machine by passing a URL to a malicious mix they create. Therefore it is no additional risk to allow swupd to write under --path. This functionality just makes it much easier for admins to use this feature to configure other installations.

cmarcelo:
(I'm kind of late on this one, so let me know if its better to create a new thread to discuss this)

I'd suggest the configuration allow us to specify mirror(s) for specific URLs, instead of a general mirror that will apply always. So instead of having a file:

```
# /etc/swupd/mirror_contenturl
https://example.com/clear-mirror/
```

I'd prefer having a file that sort of looked like:

```
# /etc/swupd/mirrors
https://cdn.download.clearlinux.org/update/ -> https://example.com/clear-mirror/
```

As a consequence, not extra file is needed for the versionurl, as it can be specified in another line.

```
# /etc/swupd/mirrors
https://cdn.download.clearlinux.org/update/ -> https://example.com/clear-mirror/
https://download.clearlinux.org/update/ -> https://example.com/clear-mirror/
```

The main advantage would be: if I move to a mix (and back), I don't need to get the /etc/swupd/ files out of the way.

src/mirror.c
icarus-sparry: I agree that we probably don't care, hence me saying `"Not sure if this is a real concern, given a typical usage pattern for swupd"`. Perhaps we need to have something like `rbash` and `rvim` which would allow a subset of swupd features (search, bundle-add, bundle-list, bundle-remove, update, check-update) but not others (autoupdate disable,) and disallow some options, like --path.

For this particular patch, I am wondering if we should enforce something like "root must own all the components of the path, none of which should be world writable, and none of the components should be symlinks". The problem is you only need one mistake to make the system exploitable.

Perhaps a better solution would be to add a disclaimer to the manual, saying that if someone can execute swupd as root, either via something like sudo or via setting the SUID bit, then they can alter anything on the system.

PR #429, opened 2018-04-26
heuristics: do not ignore /usr/src/linux-*

bwarden:
Kernel build support files go under /usr/src/linux-$(uname -r), so
ensure that all files matching /usr/src/linux-* are not ignored.

Related to https://github.com/clearlinux/mixer-tools/pull/265


bwarden:
Looks like this is no longer needed, as I've moved the files under /usr/lib/modules

PR #435, opened 2018-05-03
Use a SSL client certificate if provided

gtkramer:
To support clients that need to authenticate with a web server
hosting protected content, a SSL certificate should be
presented for all curl requests if provided.

Signed-off-by: George Kramer <george.kramer@live.com>

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -118,6 +118,31 @@ int swupd_curl_init(void)
 	return 0;
 }
 
+/* 
+ * Cannot avoid a TOCTOU here with the current curl API.  Only using the curl
+ * API, a curl_easy_setopt does not detect if the client SSL certificate is
+ * present on the filesystem.  The only happens during curl_easy_perform.
+ * The emphasis is rather on how using a SSL client certificate is an opt-in
+ * function rather than an opt-out function.
+ */
+CURLcode swupd_curl_set_optional_client_cert(CURL* curl) {
+	CURLcode curl_ret = CURLE_OK;
+
+	if (access(SSL_CLIENT_CERT, F_OK) != -1) {
+		curl_ret = curl_easy_setopt(curl, CURLOPT_SSLCERT, SSL_CLIENT_CERT);
+		if (curl_ret != CURLE_OK) {
+			goto exit;
+		}
+		curl_ret = curl_easy_setopt(curl, CURLOPT_SSLCERTTYPE, "PEM");
+		if (curl_ret != CURLE_OK) {
+			goto exit;
+		}
+	}
+
+exit:
+	return curl_ret;
+}
+
 static void swupd_curl_test_resume(void);
 
 int swupd_curl_check_network(void)
@@ -160,6 +185,11 @@ int swupd_curl_check_network(void)
 			}
 		}
 
+		curl_ret = swupd_curl_set_optional_client_cert(c);
+		if (curl_ret != CURLE_OK) {
+			goto cleanup;
+		}
+
 		curl_ret = curl_easy_perform(c);
 
 		switch (curl_ret) {
@@ -171,6 +201,10 @@ int swupd_curl_check_network(void)
 			goto cleanup;
 		case CURLE_SSL_CACERT:
 			fprintf(stderr, "Error: unable to verify server SSL certificate\n");
+			ret = EBADCERT;
+			break;
+		case CURLE_SSL_CERTPROBLEM:
+			fprintf(stderr, "Curl: Problem with the local client SSL certificate\n");
 			ret = EBADCERT;
 			break;
 		default:
@@ -237,6 +271,11 @@ double swupd_query_url_content_size(char *url)
 		if (curl_ret != CURLE_OK) {
 			return -1;
 		}
+	}
+
+	curl_ret = swupd_curl_set_optional_client_cert(curl);
+	if (curl_ret != CURLE_OK) {
+		return -1;
 	}
 
 	curl_ret = curl_easy_perform(curl);
@@ -447,6 +486,10 @@ exit:
 			fprintf(stderr, "Curl: Bad SSL Cert file, cannot ensure secure connection\n");
 			err = -1;
 			break;
+		case CURLE_SSL_CERTPROBLEM:
+			fprintf(stderr, "Curl: Problem with the local client SSL certificate\n");
+			ret = -EBADCERT;
+			break;
 		default:
 			swupd_curl_strerror(curl_ret);
 			err = -1;
@@ -550,6 +593,11 @@ static CURLcode swupd_curl_set_security_opts(CURL *curl)
 		if (curl_ret != CURLE_OK) {
 			goto exit;
 		}
+	}
+
+	curl_ret = swupd_curl_set_optional_client_cert(curl);
+	if (curl_ret != CURLE_OK) {
+		goto exit;
 	}
 
 exit:
--- a/src/swupd.h
+++ b/src/swupd.h
@@ -42,6 +42,7 @@ extern "C" {
 #define MIX_CERT MIX_DIR "Swupd_Root.pem"
 #define MIX_BUNDLES_DIR MIX_STATE_DIR "mix-bundles/"
 #define MIXED_FILE SWUPD_DEFAULTS "mixed"
+#define SSL_CLIENT_CERT "/usr/share/clear/update-ca/Swupd_Ssl_Client.pem"
 
 #define DEFAULT_VERSION_URL_PATH "/usr/share/defaults/swupd/versionurl"
 #define MIRROR_VERSION_URL_PATH "/etc/swupd/mirror_versionurl"


src/swupd.h
```diff
@@ -42,6 +42,7 @@ extern "C" {
 #define MIX_CERT MIX_DIR "Swupd_Root.pem"
 #define MIX_BUNDLES_DIR MIX_STATE_DIR "mix-bundles/"
 #define MIXED_FILE SWUPD_DEFAULTS "mixed"
+#define SSL_CLEINT_CERT "/etc/swupd/client.crt"
```
gtkramer: I'm not sure if we want this in /etc?  I could see this being something we want as a part of the mixing process that could perhaps live in /usr/share/clear somewhere.

src/curl.c
```diff
@@ -162,6 +166,13 @@ int swupd_curl_check_network(void)
 			}
 		}
 
+		if (does_ssl_client_cert_exist()) {
```
gtkramer: Is this needed here even if the URL is not https?

src/curl.c
```diff
@@ -162,6 +166,13 @@ int swupd_curl_check_network(void)
...
+			curl_ret = curl_easy_setopt(c, CURLOPT_SSLCERT, SSL_CLEINT_CERT);
```
gtkramer: Same as previous comment.  Is this needed here even if the URL is not https?

src/curl.c
```diff
@@ -251,6 +262,13 @@ double swupd_query_url_content_size(char *url)
 		}
 	}
 
+	if (does_ssl_client_cert_exist()) {
```
gtkramer: Same as previous comment. Is this needed here even if the URL is not https?

gtkramer:
No testing has been performed yet.  Looking to start a discussion.

src/curl.c
```diff
@@ -120,6 +120,10 @@ int swupd_curl_init(void)
 	return 0;
 }
 
+bool does_ssl_client_cert_exist() {
+	return access(SSL_CLEINT_CERT, F_OK) != -1;
```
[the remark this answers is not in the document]
icarus-sparry: To expand on matthew's point, I think what you want to do is the following

rename this function to something like `set_optional_cert`,  put the

      curl_ret = curl_easy_setopt(c, CURLOPT_SSLCERT, SSL_CLEINT_CERT);

inside this function, test to see what the return value is. If it is not just something like "no such file" then this  function should return the error. If it is "no such file" then this should return CURLE_OK. Then your use of the function should be something like

              curl_ret = set_optional_cert();
              if (curl_ret != CURLE_OK) {
                       goto cleanup;
              }

You might even consider writing a helper function `set_opt_cert_and_easy_perform()`

src/swupd.h
gtkramer: I think it would make sense to have an override given we can point to a new mirror.  The new mirror could conceivably need a different SSL client cert.

matthewrsj: I'm not totally against the cert in /etc but...

> The new mirror could conceivably need a different SSL client cert.

I hope this is never the case.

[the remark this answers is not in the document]
gtkramer: Wow, I'm never coding again at midnight and then pushing for comments

matthewrsj:
@gtkramer are you still going to iterate on this?

gtkramer:
@matthewrsj, yes, I didn't have time last week, but I'm working on this today.

src/curl.c
```diff
@@ -118,6 +118,20 @@ int swupd_curl_init(void)
 	return 0;
 }
 
+CURLcode curl_set_optional_client_cert(CURL* curl, char* url) {
+	CURLcode curl_ret = CURLE_OK;
+	if (strncmp(url, "https://", 8) == 0) {
+		curl_ret = curl_easy_setopt(curl, CURLOPT_SSLCERT, SSL_CLIENT_CERT);
+		if (curl_ret == CURLE_SSL_CERTPROBLEM) {
```
mbelluzzo: Is that also the return when the file does not exist?

gtkramer: Looks like it isn't.  Fixing.

gtkramer: This is more difficult to fix than expected.  Performing a `curl_easy_setopt` returns 0 even when the client SSL certificate doesn't exist.  Only at runtime with `curl_easy_perform` does the proper error code I'm looking for come up (58).  This is still a TOCTOU issue.  After speaking with @matthewrsj, we decided to go back to the original checking implementation.  The idea is that a SSL client cert is an opt-in construct.  If it's there at the time of check, but disappears later, or is not there but then appears later, is on the user to manage correctly.

gtkramer:
```
#include <curl/curl.h>
#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>

CURLcode curl_set_optional_client_cert(CURL* curl, char* url) {
	CURLcode curl_ret = CURLE_OK;
	if (strncmp(url, "https://", 8) == 0) {
		curl_ret = curl_easy_setopt(curl, CURLOPT_SSLCERTTYPE, "PEM");
		curl_ret = curl_easy_setopt(curl, CURLOPT_SSLCERT, "/tmp/lol.pem");
		fprintf(stderr, "Curl error: (%d) %s\n", curl_ret, curl_easy_strerror(curl_ret));
		if (curl_ret == CURLE_SSL_CERTPROBLEM) {
			curl_ret = CURLE_OK;
		}
	}
	return curl_ret;
}

int main() {
	curl_global_init(CURL_GLOBAL_ALL);
	CURL* curl = curl_easy_init();
	char* url = "https://download.clearlinux.org/update/";
	curl_set_optional_client_cert(curl, url);
}
```

Save as test.c and then run `gcc test.c -lcurl -o test`.  This is the program I wrote to test this behavior.

gtkramer:
I think this looks better?  I just need to add functional tests.  To check if I hit all of the modifications required to support client SSL certificates, how could I supply one to use for running the functional tests?

src/curl.c
```diff
@@ -118,6 +118,31 @@ int swupd_curl_init(void)
...
+ * Cannot avoid a TOCTOU here with the current curl API.  Only using the curl
+ * API, a curl_easy_setopt does not detect if the client SSL certificate is
+ * present on the filesystem.  The only happens during curl_easy_perform.
+ * The emphasis is rather on how using a SSL client certificate is an opt-in
+ * function rather than an opt-out function.
+ */
+CURLcode swupd_curl_set_optional_client_cert(CURL* curl) {
+	CURLcode curl_ret = CURLE_OK;
+
+	if (access(SSL_CLIENT_CERT, F_OK) != -1) {
```
matthewrsj: Why not just check if `access(...) == 0`? It seems backwards to check if `access` returns not an error.

gtkramer:
Regarding functional testing, we could probably port this from python2 to python3:

src/curl.c
gtkramer: Makes sense.  I'll go ahead and change this.

gtkramer:
@matthewrsj any objections to changing the path for SSL_CLIENT_CERT from "/usr/share/clear/update-ca/Swupd_Ssl_Client.pem" to "/etc/swupd/client.crt"?  Then this could be something that is added as a part of a post-nonchroot script with ister, and doesn't need distributed as a part of the content of the distribution.  It would give certificate distribution authorities the flexibility to grant certificates to users, or create a generic one that allows swupd to authenticate out of the box to get protected content.

matthewrsj:
@gtkramer yes I feel much more comfortable with that suggestion than putting it under `/usr`

matthewrsj:
sorry, didn't mean to close.

tmarcu:
@gtkramer is this still something that is needed or should we close this now?

matthewrsj:
@tmarcu @gtkramer is passing this off to someone else who is working on it currently.

aaronovz1:
Is this essentially support for hostkey verification? I am asking in ticket #508 for clarification on how to use it and it looked like it was partly broken.

jwakre:
@gtkramer @matthewrsj I am writing functional tests for the client certificates using a webserver, but I don't have a good way for swupd to trust my webserver. Currently, I can authenticate my webserver by configuring swupd to add an alternative trust location (./autogen --with-fallback-capaths ...) and rebuild swupd. I would prefer another method that doesn't require a special swupd configuration and rebuild. Below are a few other ideas:

1. Add a feature to swupd to specify alternative capaths at runtime.

2. Add a feature to swupd that allows it to create insecure server connections when specified.

3. Add the webserver's certificate to the host's trust store. Can this be mocked in a test environment?

matthewrsj:
We discussed in IRC, but for the peanut gallery I have no problem with adding `--with-fallback-capaths` to testing.

gtkramer:
@jwakre has picked this up and finished the remaining integration bits.  Many thanks for the assist here!  Closing this PR and letting @jwakre open another one for the final merge.

PR #438, opened 2018-05-08
Set mirror override path perm to 0755

### LINKED ISSUE #437: /etc/swupd permissions should be 0755, not 0700
https://github.com/clearlinux/swupd-client/blob/75ef4781192ed6fffbedf942a2f9eadffdd2be4b/src/mirror.c#L77

bwarden:
When /etc/swupd is created to hold mirror URL override files, set the
permissions to 0755.

Fixes #437

### DIFF
--- a/src/mirror.c
+++ b/src/mirror.c
@@ -74,7 +74,7 @@ static int write_to_path(char *content, char *path)
 	dir = dirname(tmp);
 	/* attempt to  make the directory
 	 * ignore EEXIST errors */
-	int ret = mkdir(dir, S_IRWXU);
+	int ret = mkdir(dir, S_IRWXU | S_IRGRP | S_IXGRP | S_IROTH | S_IXOTH);
 	char *cmd;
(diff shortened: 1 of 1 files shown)


src/mirror.c
```diff
@@ -74,7 +74,7 @@ static int write_to_path(char *content, char *path)
 	dir = dirname(tmp);
 	/* attempt to  make the directory
 	 * ignore EEXIST errors */
-	int ret = mkdir(dir, S_IRWXU);
+	int ret = mkdir(dir, S_IRWXU | S_IRGRP | S_IXGRP | S_IROTH | S_IXOTH);
```
icarus-sparry: OK, I am very confused - nothing unusual there.

We have on line 77 something to make a directory, and then on lines 79 and 80 we make the same directory.

bwarden: Hmmm... tunnel vision; I didn't even notice that.

matthewrsj: That'd be my fault. Looks like I switched methods and forgot to remove the old way.

PR #440, opened 2018-05-10
swupd-add-pkg: remove sudo invocations

matthewrsj:
Instead of doing this behind the user's back expect them to do it
themselves. This also removes the inconsistencies of running some
commands with sudo -E and others without -E. If users want to provide -E
let them do it themselves.


matthewrsj:
@icarus-sparry I updated the man page.

matthewrsj:
@bktan8 this will resolve the issue you were seeing where `swupd-add-pkg` was ignoring the `-E` flag.

PR #442, opened 2018-05-14
Remove copyright header messages

matthewrsj:
swupd-client is properly open sourced and licensed. The copyright header
is unnecessary, confusing, and causes issues for scripts parsing output.

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -636,9 +636,9 @@ out_fds:
 
 /* this function prints the initial message for all utils
 */
-void copyright_header(const char *name)
+void copyright_header(void)
 {
-	fprintf(stderr, PACKAGE " %s " VERSION "\n", name);
+	fprintf(stderr, PACKAGE " " VERSION "\n");
 	fprintf(stderr, "   Copyright (C) 2012-2018 Intel Corporation\n");
 	fprintf(stderr, "\n");
 }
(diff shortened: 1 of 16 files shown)


src/helpers.c
```diff
@@ -634,15 +634,6 @@ int swupd_init(int *lock_fd)
...
 
-/* this function prints the initial message for all utils
-*/
-void copyright_header(const char *name)
-{
-	fprintf(stderr, PACKAGE " %s " VERSION "\n", name);
-	fprintf(stderr, "   Copyright (C) 2012-2018 Intel Corporation\n");
-	fprintf(stderr, "\n");
-}
-
```
phmccarty: I think we should still call this function in one spot (see other comment)...

src/main.c
```diff
@@ -106,7 +106,6 @@ static int parse_options(int argc, char **argv, int *index)
 			print_help(argv[0]);
 			exit(EXIT_SUCCESS);
 		case 'v':
-			copyright_header("swupd");
```
phmccarty: Having the copyright print out when passing `--version` is a common convention, so I think we should keep this one.

matthewrsj:
@phmccarty done, done, and done.

src/helpers.c
```diff
@@ -636,9 +636,9 @@ int swupd_init(int *lock_fd)
 
 /* this function prints the initial message for all utils
 */
-void copyright_header(const char *name)
+void copyright_header(void)
```
icarus-sparry: This is no longer a "helper", it could move into src/main.c, then made static, removed from the header...

matthewrsj: true

PR #459, opened 2018-05-23
Fix automatic mix updates for mixer integration

### LINKED ISSUE #446: mixin/mixer integration: unable to update to latest clear version
If a file in os-core is updated in a new upstream Clear version mix collisions will be detected between the locally-built os-core files and the upstream ones. Because of this the mixin build either needs to occur before the unique files check or the os-core files need to not be checked. The second option could be problematic if the custom RPMs have a dependency on an upstream RPM and the upstream RPM changes in the latest version.

matthewrsj:
When a new upstream version is detected the new local mixer build must
be done before manifest uniqueness is checked. This is due to the fact
that os-core can change in the upstream version, so it has to be
re-built locally against the upstream content in order to avoid false
positives for file colisions.

Also add additional files are files that change on the system due to the
local mix. Add these to the ignore list for file colision checks.

Fixes #446

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -1434,7 +1434,12 @@ static int is_version_data(const char *filename)
 {
 	if (strcmp(filename, "/usr/lib/os-release") == 0 ||
 	    strcmp(filename, "/usr/share/clear/version") == 0 ||
-	    strcmp(filename, "/usr/share/clear/versionstamp") == 0) {
+	    strcmp(filename, "/usr/share/clear/versionstamp") == 0 ||
+	    strcmp(filename, "/usr/share/defaults/swupd/versionurl") == 0 ||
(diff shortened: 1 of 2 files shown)


src/manifest.c
```diff
@@ -1434,7 +1434,12 @@ static int is_version_data(const char *filename)
 {
 	if (strcmp(filename, "/usr/lib/os-release") == 0 ||
 	    strcmp(filename, "/usr/share/clear/version") == 0 ||
-	    strcmp(filename, "/usr/share/clear/versionstamp") == 0) {
+	    strcmp(filename, "/usr/share/clear/versionstamp") == 0 ||
```
icarus-sparry: For the future:
Once we have more than a couple of files, it would be better to loop over an array of strings, rather than individual strcmp based compares. Whilst it is true that gcc has code to optimize compares against constant strings, I don't think this will help in this case.

PR #467, opened 2018-06-05
Update: Change log order to have a clearer output

otaviobp:
Message informing that the system was successfully updated should be
the last message. Having a message that looks like an error as the
last message in an update can be confusing for users.

### DIFF
--- a/src/update.c
+++ b/src/update.c
@@ -624,6 +624,10 @@ clean_curl:
 	}
 	swupd_deinit(lock_fd, &latest_subs);
 
+	if (nonpack > 0) {
+		printf("%i files were not in a pack\n", nonpack);
+	}
+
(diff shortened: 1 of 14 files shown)


icarus-sparry:
Needs the tests updated.

Open question is if we should enhance the test framework to allow out of order messages so changes like this don't cause churn in the tests?

matthewrsj:
> Open question is if we should enhance the test framework to allow out of order messages so changes like this don't cause churn in the tests?

Sounds like a good feature. There's a lot that could be done to improve the test framework.

+1 after tests are updated (can just re-order output for now)

otaviobp:
Sorry, I forgot about that. I'll fix it

otaviobp:
tests fixed.

PR #471, opened 2018-06-05
Curl: Fixes and api improvements

otaviobp:


bryteise:
OCD nit, Curl is capitalized in the second commit subject.

matthewrsj:
I especially like the API improvement.

src/version.c
```diff
@@ -41,10 +41,16 @@
...
+#define MAX_VERSION_CHARS 10
+#define MAX_VERSION_STR_SIZE 11
+
 	char *url = NULL;
 	char *path = NULL;
 	int ret = 0;
-	struct version_container tmp_version = { 0 };
+	struct curl_file_data tmp_version = {
+		MAX_VERSION_STR_SIZE, 0,
+		alloca(MAX_VERSION_STR_SIZE)
```
icarus-sparry: If we know the size then there is no need to use alloca, just do a normal stack allocation.

src/curl.c
```diff
@@ -331,14 +331,9 @@ int swupd_curl_get_file(const char *url, char *filename, struct file *file,
...
 
-		if (file) {
-			local = file;
-			local->fh = NULL;
-		} else {
-			local = calloc(1, sizeof(struct file));
-			if (!local) {
-				abort();
-			}
+		local = calloc(1, sizeof(struct file));
```
icarus-sparry: If we don't have an incoming struct file*, then do we need to heap allocate local at all, or can we just stack allocate it? One less potential call to abort() is always a good change in my book.

icarus-sparry: Ah, this happens in the next patch in the set. :grin:

src/version.c
otaviobp: good point. I prefer to avoid alloca too, I just haven't thought about that when I was writing this

otaviobp:
Branch updated fixing:
 - s/Curl/curl/
 - don't use alloca anymore

PR #472, opened 2018-06-06
Refactor bundle-remove to handle multiple bundles

### LINKED ISSUE #449: Refactor bundle-remove to not call remove_bundle() multiple times
The current `remove_bundle()` function was designed to only be called once, but for multiple bundle removals, it is currently called multiple times.

An improvement would be to refactor the bundle removal code to handle multiple bundles at once.

castulo:
The current remove_bundle() function was designed to only be called
once, but for multiple bundle removals, it is currently called
multiple times.

This commit refactors this function so the bundle removal code can
handle multiple bundles at once.

Fixes #449

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -399,7 +399,11 @@ out:
 /*  This function is a fresh new implementation for a bundle
  *  remove without being tied to verify loop, this means
  *  improved speed and space as well as more roubustness and
- *  flexibility. What it does is basically:
+ *  flexibility. The function removes one ore more bundles
+ *  passed in the bundles param.
+ *
+ *  For each one of the bundles to be removed what it does
+ *  is basically:
  *
  *  1) Read MoM and load all submanifests except the one to be
  *  	removed and then consolidate them.
@@ -572,7 +576,7 @@ int remove_bundles(char **bundles)
 	}
 
 	if (bad > 0) {
-		fprintf(stderr, "%i bundle(s) of %i failed to remove\n", bad, total);
+		fprintf(stderr, "%i bundle(s) of %i failed to be removed\n", bad, total);
 	} else {
 		fprintf(stderr, "%i bundle(s) were removed successfully\n", total);
 	}
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -419,7 +419,7 @@ int remove_bundles(char **bundles)
 {
 	int lock_fd;
 	int ret = 0;
-	int retCode = 0;
+	int ret_code = 0;
 	int bad = 0;
 	int total = 0;
 	int current_version = CURRENT_OS_VERSION;
@@ -573,7 +573,7 @@ int remove_bundles(char **bundles)
 			  ret);
 		if (ret) {
 			fprintf(stderr, "Error: Bundle remove failed\n");
-			retCode = ret;
+			ret_code = ret;
 		}
 	}
 
@@ -585,7 +585,7 @@ int remove_bundles(char **bundles)
 
 	swupd_deinit(lock_fd, &subs);
 
-	return retCode;
+	return ret_code;
 }
 
 /* bitmapped return
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -411,10 +411,13 @@ out:
  *  	performing a unlink(2) for each filename.
  *  6) Done.
  */
-int remove_bundle(const char *bundle_name)
+int remove_bundles(char **bundles)
 {
 	int lock_fd;
 	int ret = 0;
+	int retCode = 0;
+	int bad = 0;
+	int total = 0;
 	int current_version = CURRENT_OS_VERSION;
 	struct manifest *current_mom, *bundle_manifest = NULL;
 	struct list *subs = NULL;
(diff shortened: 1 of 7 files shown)


src/bundle.c
```diff
@@ -411,10 +411,13 @@ int show_bundle_reqd_by(const char *bundle_name, bool server)
  *  	performing a unlink(2) for each filename.
  *  6) Done.
  */
-int remove_bundle(const char *bundle_name)
+int remove_bundles(char **bundles)
```
otaviobp: Swupd in general isn't strict about having const in variables that aren't changed in the function, but as this function was already adding a const, I would keep that.

otaviobp: Also update the header documentation saying that the function now handle multiple bundles

src/bundle.c
```diff
@@ -426,126 +429,157 @@ int remove_bundle(const char *bundle_name)
...
+			  current_version,
+			  ret);
+		if (ret) {
+			fprintf(stderr, "Error: Bundle remove failed\n");
+			retCode = ret;
+		}
+	}
+
+	if (bad > 0) {
+		fprintf(stderr, "%i bundle(s) of %i failed to remove\n", bad, total);
```
otaviobp: I know you are not adding this line, but it would be nice to correct this message (in another commit) to something like: "%i bundle(s) of %i failed to be removed".
And to be honest I don't even know if my suggestion is correct.

test/functional/bundleremove/remove-multiple/test.bats
```diff
@@ -32,9 +32,19 @@ teardown() {
 }
 
 @test "bundle-remove remove multiple bundles one exist and one that dont" {
-  run sudo sh -c "$SWUPD bundle-remove $SWUPD_OPTS test-bundle3 fake-bundle"
+  run sudo sh -c "$SWUPD bundle-remove $SWUPD_OPTS fake-bundle test-bundle3"
 
-  [ ! -f "$DIR/target-dir/test-file3" ] && [ $status -eq 3 ]
+  # EBUNDLE_NOT_TRACKED == 13
```
otaviobp: +1 for adding more tests

test/functional/bundleremove/remove-multiple/test.bats
```diff
@@ -32,9 +32,19 @@ teardown() {
 }
 
 @test "bundle-remove remove multiple bundles one exist and one that dont" {
-  run sudo sh -c "$SWUPD bundle-remove $SWUPD_OPTS test-bundle3 fake-bundle"
+  run sudo sh -c "$SWUPD bundle-remove $SWUPD_OPTS fake-bundle test-bundle3"
```
otaviobp: why did you change the order of the bundles on the test?

src/bundle.c
castulo: Good point.

test/functional/bundleremove/remove-multiple/test.bats
castulo: ;)

test/functional/bundleremove/remove-multiple/test.bats
castulo: well, because I wanted to make sure the exit code of the bundle that failed to be removed was indeed being kept to be returned once all bundles had finished being removed. 
So I changed the order so the test would have a "bad remove" followed by a "good remove", but the exit code from the bad remove should not get lost.

castulo:
Thanks for your review @otaviobp

test/functional/bundleremove/remove-multiple/test.bats
otaviobp: Got it. It makes sense.

src/clr_bundle_rm.c
```diff
@@ -191,84 +191,11 @@ static bool parse_options(int argc, char **argv)
 	return false;
 }
 
-static void reload_parsed_opts(void)
-{
-	if (curopts.path_prefix) {
```
otaviobp: I think you can get rid of the curopts struct as you are not using the reload_parsed_opts() anymore. can you check that?

src/bundle.c
castulo: Turns out the compiler didn't like that, after adding the const I got a compiler warning:

src/clr_bundle_rm.c: In function ‘bundle_remove_main’:
src/clr_bundle_rm.c:200:24: warning: passing argument 1 of ‘remove_bundles’ from incompatible pointer type [-Wincompatible-pointer-types]
  return remove_bundles(bundles);
                        ^~~~~~~
In file included from src/clr_bundle_rm.c:35:
src/swupd.h:380:40: note: expected ‘const char **’ but argument is of type ‘char **’
 extern int remove_bundles(const char **bundles);
                           ~~~~~~~~~~~~~^~~~~~~

src/clr_bundle_rm.c
castulo: good catch, I'll remove that in another commit.

src/bundle.c
```diff
@@ -426,126 +433,157 @@ int remove_bundle(const char *bundle_name)
...
+			bad++;
+			goto out_free_mom;
+		}
 
-	/* deduplication needs file list sorted by filename, do so */
-	bundle_manifest->files = list_sort(bundle_manifest->files, file_sort_filename);
-	deduplicate_files_from_manifest(&bundle_manifest, current_mom);
+		/* check if bundle is required by another installed bundle */
+		struct list *reqd_by = NULL;
+		required_by(&reqd_by, *bundles, current_mom, 0);
```
icarus-sparry: How on earth did I let required_by get into the code base - I must have been asleep!

src/bundle.c
```diff
@@ -426,126 +433,157 @@ int remove_bundle(const char *bundle_name)
...
-		goto out_free_curl;
-	}
+	for (; *bundles; ++bundles, total++) {
 
-	if (!search_bundle_in_manifest(current_mom, bundle_name)) {
-		fprintf(stderr, "Bundle name is invalid, aborting removal\n");
-		ret = EBUNDLE_REMOVE;
-		goto out_free_mom;
-	}
+		fprintf(stderr, "Removing bundle: %s\n", *bundles);
```
icarus-sparry: This print ought to be after the check it is not os-core.

src/bundle.c
```diff
@@ -426,126 +433,157 @@ int remove_bundle(const char *bundle_name)
...
 
 	mix_exists = (check_mix_exists() & system_on_mix());
 
-	current_mom = load_mom(current_version, false, mix_exists);
-	if (!current_mom) {
-		fprintf(stderr, "Unable to download/verify %d Manifest.MoM\n", current_version);
-		ret = EMOM_NOTFOUND;
-		goto out_free_curl;
-	}
+	for (; *bundles; ++bundles, total++) {
```
icarus-sparry: Not a fan of 130 line for loops, hard to review on github!

I would create a new variable to avoid the need to dereference *bundles all the time.

src/bundle.c
```diff
@@ -399,7 +399,11 @@ int show_bundle_reqd_by(const char *bundle_name, bool server)
 /*  This function is a fresh new implementation for a bundle
  *  remove without being tied to verify loop, this means
  *  improved speed and space as well as more roubustness and
- *  flexibility. What it does is basically:
+ *  flexibility. The function removes one ore more bundles
```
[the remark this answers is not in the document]
castulo: Good catch, I'll update that.

src/bundle.c
castulo: Ok

src/bundle.c
castulo: Done, I'll create a new variable to store the bundle in the next commit.

src/bundle.c
matthewrsj: The compiler didn't like it because you removed `const` from `src/swupd.h`

matthewrsj: /me reads you error message closer.. nvm

src/bundle.c
```diff
@@ -411,10 +415,13 @@ int show_bundle_reqd_by(const char *bundle_name, bool server)
  *  	performing a unlink(2) for each filename.
  *  6) Done.
  */
-int remove_bundle(const char *bundle_name)
+int remove_bundles(char **bundles)
 {
 	int lock_fd;
 	int ret = 0;
+	int retCode = 0;
```
matthewrsj: camel case please

matthewrsj:
@icarus-sparry and @otaviobp any further comments?

src/bundle.c
castulo: ups, my bad! Changing it.

Issue #378, opened 2018-02-12
`/var/lib/swupd/staged` is retaining too many files


ahkok:
There is a reasonable amount of storage needed to keep updates efficiently, such that files that will be needed on the next release are kept until they are no longer needed. However, it appears that too many files are kept, resulting in wasted disk space.

I've poked at it and since I have several systems where `/var/lib/swupd/staged` is in the order of 7-9GB currently, I've got some data to discuss potential solutions.

## why doesn't systemd-tmpfiles work?

Before I start with the solutions, I'd like to discuss the problems with systemd-tmpfiles, and why it doesn't work. Most of this is due to a combination of (1) our way of timestamping python files, (2) tmpfiles' way of looking at atime/mtime and a small dose of us shipping files that are `-r--r--r-- root root` which tmpfiles entirely ignores.

The result is that systemd-tmpfiles is useless and not usable to do any reasonable amount of pruning.

We have 2 straightforward solutions, however.

## 1 - prune what we know is old

Given the manifests that are on the system, we can discover all hashes for files that are relevant for the current state of the system. Since we can compare this to the files in `/staged`, we know which files are no longer related to any known manifest on the system. This gives us a list of files that "definitely" can be deleted without ever needing to be redownloaded.

In my local system, this would recover 3708879078 bytes from a total of 9354271423 (39%)

This is not a 'cheap" operation. This solution would be costly for a shell script, requiring many seconds of processing, even up to a minute or so, to finish. If done inside `swupd` itself, it could probably be done much more efficiently.

## 2 - prune based on `mtime`

Since we mostly create files in staged and never touch them after much, we could decide to use `mtime` to distinguish between files that are considered "aged" and not used recently.

The benefits of this method is that it's really quick and can bed one in a shell script.. The downsides is that it will specifically delete lots of Python files that may be relevant and cause fullfiles to be re-downloaded.

Here's what the data looks like:

```
mtime   rm size (1)  rm size (2)
none    9354271423   3801303314
1       9294391574   3759230964
2       8704640059   3423811441
3       2880597007    795056079
4       2774649062    794839002
5       2685577506    794782008
```

As to why the gap exists between 2 and 3, this must be due to the actual update content pushing a significant rebuild out.

## Conclusion

[...]

PR #473, opened 2018-06-07
clean: use ctime instead of mtime

cmarcelo:
The ctime will represent the time the files were downloaded, instead
of the mtime generated at build time (which sometimes is the time of
build, but sometimes is not).

Fixes an issue reported by Otavio.


cmarcelo:
Related to https://github.com/clearlinux/swupd-client/issues/378#issuecomment-395487117

cmarcelo:
@otaviobp

otaviobp:
+1. Works for me

PR #474, opened 2018-06-07
Enforce braces on single line ifs in make compliant

otaviobp:
There's no way to inforce braces on single line ifs using clang-format.
Adding a grep to look for ifs without a bracket so we can at least
warn users about that.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -242,10 +242,16 @@ compliant:
 	git diff --quiet --exit-code src; ret=$$?; \
 	if [ $$ret -eq 1 ]; then \
 		echo "Code style issues found. Run 'git diff' to view issues."; \
+	fi; \
+	OUT=$$(git grep if\ \(.*\)$$ -- '*.c'); ret2=$$?; \
+	if [ $$ret2 -eq 0 ]; then \
+		echo "Missing brackets in single line if:"; \
+		echo "$$OUT"; \
 	elif [ $$ret -eq 0 ]; then \
 		echo "No code style issues found."; \
+		exit 0; \
 	fi; \
-	exit $$ret
+	exit 1
 
 release:
 	@git rev-parse v$(PACKAGE_VERSION) &> /dev/null; \
--- a/src/curl.c
+++ b/src/curl.c
@@ -128,8 +128,9 @@ int swupd_curl_check_network(void)
 	CURL *c;
 	static int has_network = 0;
 
-	if (has_network)
+	if (has_network) {
 		return 0;
+	}
 
 	if (!curl) {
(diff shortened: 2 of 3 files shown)


Makefile.am
```diff
@@ -242,10 +242,16 @@ compliant:
 	git diff --quiet --exit-code src; ret=$$?; \
 	if [ $$ret -eq 1 ]; then \
 		echo "Code style issues found. Run 'git diff' to view issues."; \
+	fi; \
+	OUT=$$(git grep if\ \(.*\)$$ -- '*.c'); ret2=$$?; \
+	if [ $$ret2 -eq 0 ]; then \
+		echo "Missing brackets in single line if:"; \
+		echo "$$OUT"; \
 	elif [ $$ret -eq 0 ]; then \
```
castulo: What if clang-format makes changes in the source files and there are no missing brackets in single if? Based on this logic both this messages would be shown at the same time, wouldn't they?
"Code style issues found. Run 'git diff' to view issues."
"No code style issues found."

Makefile.am
```diff
@@ -242,10 +242,16 @@ compliant:
...
 		echo "No code style issues found."; \
+		exit 0; \
```
castulo: What if clang-format makes changes in the source files and there are no missing brackets in single if? Based on this logic both this messages would be shown at the same time, wouldn't they?
"Code style issues found. Run 'git diff' to view issues."
"No code style issues found."

phmccarty: Yes, probably an `exit 1` should be added above. I'll add a separate comment.

Makefile.am
```diff
@@ -242,10 +242,16 @@ compliant:
 	git diff --quiet --exit-code src; ret=$$?; \
 	if [ $$ret -eq 1 ]; then \
 		echo "Code style issues found. Run 'git diff' to view issues."; \
+	fi; \
```
phmccarty: I would add an `exit 1;` above this line, to address the issue @castulo raised.

Makefile.am
otaviobp: there's no need to add an exit 1.
I'm checking if ret -eq 0 in an elif. so if we reach exit 0, ret == 0 and ret2 == 0, so no errors were found.

phmccarty: Oh, sorry, I didn't notice that a new "ret2" variable was introduced... You're correct. The logic looks okay.

castulo: You are right, now that I see it again I don't even know why I thought it wouldn't work, I must have misread some variable or something. I'll pay more attention in the future. 
So anyway, looks good ;)

Issue #478, opened 2018-06-08
Improve bundle-add code to determine if a pack download is actually necessary


matthewrsj:
Instead of immediately downloading the zero packs for all bundles that are not installed on the client system perform some heuristics on the number of files needed. If only a few files (or one or five) need to actually be downloaded and installed we could potentially just download those fullfiles and apply them immediately.

An obvious use-case for this is when adding the `git` bundle. The `git` bundle only installs `git` and its dependencies, but several other larger bundles also provide the `git` package, some of which I already have on my filesystem.

``` bash
$ git --version   # git (the package) already on filesystem
git version 2.17.1

$ time sudo swupd bundle-add git   # can bundle-add git (the bundle)
Downloading packs...

Extracting git pack for version 22690
	...100%
Installing bundle(s) files...
	...100%
Calling post-update helper scripts.
httpd.service: needs a restart (a library dependency was updated)
pacrunner.service: needs a restart (a library dependency was updated)
Bundle(s) installation done.

real	0m6.944s   # takes seven seconds
user	0m4.158s
sys	0m2.230s

$ sudo swupd bundle-remove git
Removing bundle: git
Deleting bundle files...
Total deleted files: 1   # <------ it only added one file! (the tracking file in /usr/share/clear/bundles)
Untracking bundle from system...
Success: Bundle removed
1 bundle(s) were removed successfully
```

If `bundle-add` could detect that only one fullfile was needed it could just download the single file and be on its way instead of downloading the entire pack, extracting, and *then* checking what it needs to add.

matthewrsj:
Fixed by @fenrus75 in #479

PR #479, opened 2018-06-09
Optimize bundle-add

fenrus75:
* Don't install files that are already present
* Don't parse the whole Manifest-tree of all installed manifest unless and until error recovery is required

with these changes a swupd bundle-add git becomes a 416 msec thing of which +/- 200 msec is the sync() and 180 or so msec are the triggers

Raw elapsed time stats:
      0.76 ms: Add bundles and recurse
      3.28 ms: Consolidate files from bundles
      0.03 ms: Download packs
      0.05 ms: Download missing files
    236.81 ms: Installing bundle(s) files onto filesystem

CPU process time stats:
      0.75 ms: Add bundles and recurse
      3.27 ms: Consolidate files from bundles
      0.03 ms: Download packs
      0.05 ms: Download missing files
      0.35 ms: Installing bundle(s) files onto filesystem

Raw elapsed time stats:
      0.00 ms: Prepend bundles to list
    416.82 ms: Install bundles

CPU process time stats:
      0.00 ms: Prepend bundles to list
      4.99 ms: Install bundles


src/list.c
```diff
@@ -299,3 +299,17 @@ struct list *list_deep_clone_strs(struct list *list)
 
 	return clone;
 }
+
+int list_longer_than(struct list *list, int count)
+{
+	struct list *item;
+
+	item = list_head(list);
```
mbelluzzo: I think you can use `list` itself as you did with count since you are not doing any modification in the list content for this.

src/manifest.c
```diff
@@ -1116,6 +1116,34 @@ struct list *consolidate_files(struct list *files)
 	return list;
 }
 
+/*
+ * Takes a list of files, and removes files from the list that exist on the filesystem.
+ * This is used for cases like bundle-add where we don't want to re-add files that
+ * already are on the system.
+ */
+struct list *filter_out_existing_files(struct list *files)
```
mbelluzzo: s/list/files/ should also work here.

matthewrsj:
Fixes #478 

I'll look into test failures today.

src/list.c
mbelluzzo: Common, that is a 10 line function. Compared with other stuff in this code base, this is way far from hurting the maintenance of this code.

matthewrsj:
Ah, many tests are failing due to output-matching for pack downloads. In many cases packs aren't being attempted because they provide less than 5 files.

icarus-sparry:
Another argument for improving our test tooling.

src/manifest.c
```diff
@@ -1116,6 +1116,34 @@ struct list *consolidate_files(struct list *files)
...
+
+	ret = files;
+
+	list = list_head(files);
+	while (list) {
+		next = list->next;
+
+		file = list->data;
+
+		if (verify_file_lazy(file->filename)) {
```
matthewrsj: This needs to perform the check under `path_prefix`. This is one reason so many tests are failing.
```C
if (verify_file_lazy(mk_full_filename(path_prefix, file->filename))) {
```

bryteise:
@matthewrsj One thing I'm worried about with this change is that the pack download and extraction might not being used in any tests. Do we have some tests where the 0-pack is used still?

matthewrsj:
I'll check, and if not I'll add some bundles with more files.

test/functional/bundleadd/include/test.bats
```diff
@@ -13,8 +13,8 @@ setup() {
   create_manifest_tar 10 test-bundle
   chown_root "$DIR/web-dir/10/staged/$t1_hash"
   chown_root "$DIR/web-dir/10/staged/$t2_hash"
-  #create_fullfile_tar 10 $t1_hash
-  #create_fullfile_tar 10 $t2_hash
+  create_fullfile_tar 10 $t1_hash
+  create_fullfile_tar 10 $t2_hash
```
bryteise: Mind explaining why this is being changed?

matthewrsj: I had commented it out as a debug step and committed it on accident.

PR #485, opened 2018-06-12
Be consistent in bundle-add/bundle-remove error messages

castulo:


src/bundle.c
```diff
@@ -682,7 +682,9 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
 	grabtime_start(&times, "Add bundles and recurse");
 	ret = add_subscriptions(bundles, subs, current_version, mom, false, 0);
 
-	if (ret != add_sub_NEW) {
+	/* Use a bitwise AND with the add_sub_NEW mask to determine if at least
+	 * one new bundle was subscribed */
+	if (!(ret & add_sub_NEW)) {
```
otaviobp: lgtm, but this is going to break a test. I've seen that you fixed that in another commit, but It's nice to have the broken test fixed in the same commit that breaks it.

src/bundle.c
```diff
@@ -855,10 +856,20 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
 	ret = 0;
-	fprintf(stderr, "Bundle(s) installation done.\n");
 	print_time_stats(&times);
 
 out:
+	bundles_installed = list_len(to_install_bundles);
+	bundles_failed = bundles_requested - bundles_installed;
+	/* the bundles installed could end up being more than the ones requested
+	 * in certain situations like when a bundle needs to be updated */
+	if (bundles_failed > 0) {
```
otaviobp: I don't like the error output when you try to install a bundle that is already present.

    # swupd bundle-add os-core
    Warning: Bundle "os-core" is already installed, skipping it...
    1 bundle of 1 failed to be installed

Can't we output something like: "1 bundle was already installed". What do you think about that?

src/bundle.c
```diff
@@ -855,10 +856,20 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
 	run_scripts(false);
 
 	ret = 0;
-	fprintf(stderr, "Bundle(s) installation done.\n");
 	print_time_stats(&times);
 
 out:
+	bundles_installed = list_len(to_install_bundles);
```
otaviobp: This isn't going to work. to_install_bundles contains bundles on bundles list + dependencies. So you may have a situation where bundles_installed > bundles_requested and then bundles_failed < 0. And in this case if a bundle fails to install we'll have bundles_failed <= 0 and a success message will be printed.

In a system without any desktop bundles:

    # swupd bundle-add desktop-apps-extras
    list_len(to_install_bundles) == 6
    list_len(bundles) == 1

test/functional/bundleadd/add-existing/test.bats
```diff
@@ -15,7 +15,8 @@ teardown() {
 
 @test "bundle-add an already existing bundle" {
   run sudo sh -c "$SWUPD bundle-add $SWUPD_OPTS test-bundle"
-  [ "$status" -eq 0 ]
+  echo "$status"
```
otaviobp: Probably a leftover

test/functional/bundleadd/add-directory/test.bats
```diff
@@ -22,6 +22,7 @@ teardown() {
 
 @test "bundle-add add bundle containing a directory" {
   run sudo sh -c "$SWUPD bundle-add $SWUPD_OPTS test-bundle"
+  echo "$status"
```
otaviobp: probably another leftover

src/bundle.c
castulo: Ok, I can do that. In this case, what do you think the exit status should be? 0? Or we should stick to an exit code >0?

src/bundle.c
castulo: yep, you are right. I'll solve this issue in another way so it is more reliable. I'm going to fix this commit.

test/functional/bundleadd/add-existing/test.bats
castulo: Not a leftover, I added it on purpose, when that echo is not there, and the test fails, you can only see in the test log that the test failed because it got an incorrect exit status, but you cannot see what the actual exit status is. So I think is better to have that value there all the time.

otaviobp: oh, ok then

src/bundle.c
castulo: I think the best option is to return an error code even when trying to install a bundle that is already installed. This way it will also be more consistent with the way bundle-remove works. The only way the user will get exit code = 0 is if everything he requested is installed/removed successfully.

src/bundle.c
```diff
@@ -612,7 +612,7 @@ int add_subscriptions(struct list *bundles, struct list **subs, int current_vers
 
 		file = search_bundle_in_manifest(mom, bundle);
 		if (!file) {
-			fprintf(stderr, "%s bundle name is invalid, skipping it...\n", bundle);
+			fprintf(stderr, "Warning: Bundle \"%s\" is invalid, skipping it...\n", bundle);
```
matthewrsj: We should also make how we print `Warning` messages consistent with a `print_warn` function or something. I'll open #489 to track.

src/bundle.c
```diff
@@ -880,10 +883,34 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+		if (list_search_item(bundles, to_install_manifest->component)) {
+			bundles_installed++;
+		}
+	}
+
+	/* print totals */
+	bundles_failed = bundles_requested - bundles_installed - already_installed;
+	if (bundles_failed > 0) {
+		ret = EBUNDLE_INSTALL;
+		fprintf(stderr, "%i bundle%s of %i failed to be installed\n", bundles_failed, (bundles_failed > 1 ? "s" : ""), bundles_requested - already_installed);
```
matthewrsj: I'd prefer something like `failed to install %i of %i bundle%s\n`

src/bundle.c
```diff
@@ -880,10 +883,34 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+	} else if (bundles_installed) {
+		fprintf(stderr, "%i bundle%s installed successfully\n", bundles_installed, (bundles_installed > 1 ? "s" : ""));
```
matthewrsj: Similarly, here could be `successfully installed %i bundle%s`

src/list.c
```diff
@@ -313,3 +314,18 @@ int list_longer_than(struct list *list, int count)
 	}
 	return 0;
 }
+
+bool list_search_item(struct list *list, const char *key)
```
matthewrsj: Either remove the commented-out `string_in_list` function in helpers.c  or uncomment that one and use the existing implementation instead of defining it again here.

src/bundle.c
```diff
@@ -572,15 +573,14 @@ int remove_bundles(char **bundles)
...
 			  ret);
 		if (ret) {
-			fprintf(stderr, "Error: Bundle remove failed\n");
 			ret_code = ret;
 		}
 	}
 
 	if (bad > 0) {
-		fprintf(stderr, "%i bundle(s) of %i failed to be removed\n", bad, total);
+		fprintf(stderr, "%i bundle%s of %i failed to be removed\n", bad, (bad > 1 ? "s" : ""), total);
```
matthewrsj: Since the purpose of this PR is to update our messaging, I'll mention that I don't like the passive tense here. I'd prefer something like `failed to remove %i of %i bundle%s`

src/bundle.c
```diff
@@ -572,15 +573,14 @@ int remove_bundles(char **bundles)
...
 	} else {
-		fprintf(stderr, "%i bundle(s) were removed successfully\n", total);
+		fprintf(stderr, "%i bundle%s removed successfully\n", total, (total > 1 ? "s" : ""));
```
matthewrsj: and `successfully removed %i bundle%s`

src/bundle.c
matthewrsj: I expected this comment to be listed first in the review, hence my explanation of my suggestions on the messaging.

src/bundle.c
castulo: I like that idea

src/list.c
castulo: Oh, I didn't realize there was a function for that already. I'll use the existing one then. Thanks for pointing that out.

PR #487, opened 2018-06-12
Resume download

otaviobp:
Resume download isn't working. I found this problem while working on pack downloads.
Also:
 - Improve the way resume support is detected
 - use download resume in bundle-add too.


matthewrsj:
You have a very simple merge conflict with `bundle.c`

src/curl.c
```diff
@@ -334,22 +345,22 @@ int swupd_curl_get_file_full(const char *url, char *filename,
 
 		local.staging = filename;
 
-		if (lstat(filename, &stat) == 0) {
-			if (resume_ok && resume_download_enabled) {
-				curl_ret = curl_easy_setopt(curl, CURLOPT_RESUME_FROM_LARGE, (curl_off_t)stat.st_size);
-			} else {
-				unlink(filename);
+		if (resume_ok && resume_download_enabled &&
+		    lstat(filename, &stat) == 0) {
```
matthewrsj: any reason the `lstat(...)` is on a new line?

test/functional/update/slow-server/test.bats
```diff
@@ -0,0 +1,74 @@
...
+
+load "../../swupdlib"
+
+server_pid=""
+port=""
+
+targetfile="6c27df6efcd6fc401ff1bc67c970b83eef115f6473db4fb9d57e5de317eba96e"
+
+setup() {
+  # extra debug for touchy test
```
matthewrsj: heh, reminds me when we had to debug and fix the test you copied this from. this is a situation it was good you copied the existing test.

otaviobp: oops. I'll take that out from my test

src/curl.c
otaviobp: Probably the line was too big and then I change that again. I'll make them in a line

otaviobp: Oh, I broke that because I'm used to break in 80cols, but I noticed that we don't do that on swupd. Is there a line length limit I should follow?

otaviobp:
Branch updated addressing the reported problems

src/curl.c
matthewrsj: No, the rule is just to keep it readable. Wrap when it makes it more readable to do so.

matthewrsj:
thanks @otaviobp

PR #494, opened 2018-06-15
verify: Ignore mix state when fixing to version

### LINKED ISSUE #493: Unable to verify back from mix due to /usr/share/defaults/swupd/mixed file existing
From mixer issue: 
After migrating to a local mix, /usr/share/defaults/swupd/mixed file is created.
this prevents "swupd verify --fix --picky -m X -C /usr/share/clear/update-ca/Swupd_Root.pem" from working. After removing "mixed", it works.

Example:
```
clear@clearos ~ $ sudo -E swupd verify --fix --picky -m 23000 -C /usr/share/clear/update-ca/Swupd_Root.pem
Verifying version 23000
Warning: Removing corrupt Manifest.MoM artifacts and re-downloading...
Failed to load 23000 MoM manifest
Unable to download/verify 23000 Manifest.MoM

clear@clearos ~ $ sudo rm /usr/share/defaults/swupd/mixed

clear@clearos ~ $ sudo swupd verify --fix --picky -m 23000 -C /usr/share/clear/update-ca/Swupd_Root.pem
Verifying version 23000
WARNING: the force or picky option is specified; ignoring version mismatch for verify --fix
helloclear-bundle bundle name is invalid, skipping it...
Verifying files
...100%
...0%
File /usr/bin/dumpsexp was not in a pack
...100%
```

matthewrsj:
The --fix flag is not supported for mixed systems due to special
handling of some os-core files being required, which is not currently
done. Until this is done and mixer state is handled properly ignore any
mixer related state when performing a verify --fix.

Fixes #493 

@bktan8

### DIFF
--- a/src/verify.c
+++ b/src/verify.c
@@ -685,13 +685,14 @@ int verify_main(int argc, char **argv)
 	}
 
 	/* Gather current manifests */
+	int sys_version = get_current_version(path_prefix);
 	if (!version) {
-		version = get_current_version(path_prefix);
-		if (version < 0) {
+		if (sys_version < 0) {
 			fprintf(stderr, "Error: Unable to determine current OS version\n");
 			ret = ECURRENT_VERSION;
 			goto clean_and_exit;
 		}
+		version = sys_version;
 	}
 
 	if (version == -1) {
@@ -727,9 +728,10 @@ int verify_main(int argc, char **argv)
 
 	grabtime_start(&times, "Load and recurse Manifests");
 
-	/* If upstream URL is passed, user is likely trying to get back to the official stream,
-	 * so ignore any custom mixer related state */
-	if (strncmp(content_url, DEFAULT_CONTENTURL, strlen(DEFAULT_CONTENTURL)) == 0) {
+	/* When the version we are verifying against does not match our system version
+	 * disable checks for mixer state so the user can easily switch back to their
+	 * normal update stream */
+	if (version != sys_version) {
 		official_manifest = load_mom(version, false, false);
 	} else {
 		official_manifest = load_mom(version, false, system_on_mix());


src/verify.c
```diff
@@ -727,9 +727,13 @@ int verify_main(int argc, char **argv)
...
 
-	/* If upstream URL is passed, user is likely trying to get back to the official stream,
-	 * so ignore any custom mixer related state */
-	if (strncmp(content_url, DEFAULT_CONTENTURL, strlen(DEFAULT_CONTENTURL)) == 0) {
+	/* If upstream URL is passed, user is trying to verify against the official
+	 * stream. The --fix flag is also not supported for mixed systems due to
+	 * special handling of some os-core files being required, which is not
+	 * currently done. For both these cases ignore any custom mixer related
+	 * state */
+	if (cmdline_option_fix ||
```
otaviobp: I'm not sure about that. system_on_mix() == true can't we have scenarios where "swupd verify" won't list any problem and "swupd verify --fix" will list and fix files?

otaviobp: My 2 cents on that:

I think that if a system is on a mix or not should be completely transparent to swupd. And the same for DEFAULT_CONTENTURL. This should be irrelevant. If a person have is using a mirror of the clearlinux repo, shouldn't we have the exactly same behavior as if the person is downloading strait from the upstream server?

I think we could support https:// and file:// repositories with an order of priority and just don't worry if the repository we are getting info from is a mix or the upstream one. If there's any difference in the files structure in a mix (manifest, different files, sigs, etc) we should be able to guess that from the content and not from any setting in swupd.

otaviobp: My comments on the second comment were definitely not blockers for the commit to get in. But I still think that it's possible to have different results if you are running "swupd verify" or "swupd verify --fix". I'll try to check that.

> it is not a mirror of the clearlinux repo. It is the clearlinux update stream plus local content merged in. There are essentially two sources for the update.

I know that a mix is not a mirror. But I think there's no reason not supporting mirrors for Clear Linux, so I don't like comparing the DEFAULT_CONTENTURL to take decisions in swupd. And we don't need that

> but this is not the point of the local mix content. This would mean users would have to host the entire clear update content on their machine.

No. That's why I said we could have several repositories (http and file) with order of priorities. I'll write a proposal explaining my full thoughts with more details later. It was a bit confuse to write it down in this pr.

otaviobp:
If content_url == DEFAULT_CONTENTURL the patch will have no effect because the if condition was already true before the patch. So I'm assuming that the problem occurs only when content_url != DEFAULT_CONTENTURL. And for that case:

This will fail:

    sudo -E swupd verify --picky -m 23000 -C /usr/share/clear/update-ca/Swupd_Root.pem

And this will work:

    sudo -E swupd verify --fix --picky -m 23000 -C /usr/share/clear/update-ca/Swupd_Root.pem

Can you test that?

matthewrsj:
@otaviobp I'm thinking of going with another approach (slightly) of ignoring mix state when `version != get_current_version(NULL)`, effectively disallowing version-specific verification against local mixes.

This does not, and is not intended to, fix the overall issue of messy mixer integration. We need a way to allow users to get back to upstream and I think this is the fastest way to do it.

There is a larger issue that the mixer state is messy and there are a couple of different indicators for swupd to figure out if it is on a mix or not. This definitely needs to be cleaned up and if you have ideas for a cleaner architecture I'd love to see that proposal.

otaviobp:
I found this approach a lot better then the other one. And it's working for me now.
And +1 for removing the url_comparison.

PR #496, opened 2018-06-16
delta: improve delta failure message

### LINKED ISSUE #469: More detailed error message when applying delta fails
The error message in

https://github.com/clearlinux/swupd-client/blob/master/src/delta.c#L178

is used both in the case that the "from" file is not supposed to exist (not in the manifests, so this delta shouldn't be in the delta pack) and when it should be there but the actual file has a different hash (file was modified in disk).

Based on a report in IRC:

```
<btwarden> Going from 22750 to 22780, I got a lot of these:
<btwarden> Couldn't use delta file /var/lib/swupd/delta/22690-22760-eee8dcc4e4b9a01b0e13140f12f0cbf97be9ebe3d91cc1abbed9274cb00f9722-1af976c327403cc6f300bb1bb64ea3284cbd3279d6071bbfedba430e7ac02a9a: no 'from' file to appl
```

The from hash refers to a `pyc` file (eee8dcc4e4b9a01b0e13140f12f0cbf97be9ebe3d91cc1abbed9274cb00f9722), that possibly could have been modified.

matthewrsj:
When delta application failed due to swupd being unable to calculate the
hash of the from file or a hash-mismatch against what was expected print
an error message that indicates there is a corrupted file.

Fixes #469

### DIFF
--- a/src/delta.c
+++ b/src/delta.c
@@ -155,6 +155,7 @@ void apply_deltas(struct manifest *current_manifest)
 
 		struct list *ll = list_head(current_manifest->files);
 		char *found = NULL;
+		bool bad_on_sys = false;
 
 		for (; ll && !found; ll = ll->next) {
 			struct file *file = ll->data;
@@ -169,6 +170,7 @@ void apply_deltas(struct manifest *current_manifest)
 
 			if (!compute_hash_from_file(filename, hash) || !hash_equal(file->hash, hash)) {
 				free_string(&filename);
+				bad_on_sys = true;
 				continue;
 			}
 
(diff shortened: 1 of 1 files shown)


src/delta.c
```diff
@@ -169,14 +170,20 @@ void apply_deltas(struct manifest *current_manifest)
...
 			found = filename;
 		}
 
 		if (!found) {
-			fprintf(stderr, "Couldn't use delta file %s: no 'from' file to apply was found\n", delta_file);
+			fprintf(stderr, "Couldn't use delta file %s: ", delta_file);
+			if (bad_on_sys) {
+				fprintf(stderr, "'from' file corrupted on system, consider running 'swupd verify --fix'\n");
+			} else {
+				fprintf(stderr, "no 'from' file to apply was found\n");
```
phmccarty: Hmm, does this error message now hint at a problem with mixer delta pack creation? That is, if a delta exists in STATEDIR/delta, surely the "from" file is supposed to be installed on the user's system? Because a delta pack wouldn't be downloaded unless the associated bundle was already installed to begin with...

matthewrsj: It could, but it could also indicate corruption of the user's system. I don't think users will see this specific error much after this commit is released. I think it is much more likely they will see `'from' file corrupted on system` unless they go around randomly removing files.

phmccarty: Right, I hope they will see the other error instead :-)

Issue #266, opened 2017-09-21
Do pre-bundle-add and pre-os-install disk space checks


matthewrsj:
Instead of trying to do a bundle-add or OS installation and failing due to out-of-disk errors, perform a disk-space check before attempting the operation.

This can be coupled with the bundle-list --deps functionality and the fixed contentsize in the manifests to get a required disk space metric.

icarus-sparry:
Of course this problem is not perfectly solvable, as there is a window between the disk space check and the install during which the space could be used. It gets even more complicated if you have multiple mounts. That doesn't mean to say that it is not worth attempting, just that it is not trivial.

matthewrsj:
We could at least make an educated guess that the install will fail, no? Even the contentsize calculation will only be a ceiling since there can be overlapping files.

matthewrsj:
@icarus-sparry do you have time to look at this? Even a best guess is better than what we have currently, which is just an avalanche of `not fixed` errors.

icarus-sparry:
OK, will take a stab at it.

matthewrsj:
This can potentially be made easier with the functionality being introduced in #399

phmccarty:
Also worth mentioning here is that swupd-client should report `ENOSPC` errors correctly. Out-of-disk conditions are often inferred through other errors that are less common to see (e.g. verify failing to "fix" files and reporting `not fixed` repeatedly, when it would be nicer to see `failed to fix file: out of disk space` instead).

Should we track these reporting problems in a separate issue (or issues)?

evil-brain:
IMHO reporting problems with out of disk space should be the first step toward resolving this issue.

When I started I saw this issue not just as a "check if we have enough space" problem, but more of a "swupd has problems when the FS is full" issue.

I've started by trying to fix the reporting and handling the out of space case in a sane manner.

Preemptive warning of space is good to have as well, but it makes more sense to me to start with handling the current errors/report them correctly.

With the potential for random mountpoints on the FS (along with the FS being used during updates), preemptive checking will always be a kind of 'best guess'. Unless we should check available space for each file IO operation... which seems like it'd be too much overhead.

A new issue might make sense, because there are a few places a full FS could cause problems. (ex when d/ling manifests after invoking search, d/ling bundles, decompressing, copying files...)

matthewrsj:
> I've started by trying to fix the reporting and handling the out of space case in a sane manner.

[...]

PR #501, opened 2018-06-25
Add free space check before adding bundles

evil-brain:
Prevoiusly, when running bundle-add there was no check in
place to verify free storage before attempting to add the
new bundle.

Added this check, a warning message, and a flag to ignore
the message if desired.

Some functionality from search was moved out and is re-used
for calculating bundle sizes.

related to issue #266

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -27,6 +27,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <sys/statvfs.h>
 #include <sys/types.h>
 #include <unistd.h>
 
@@ -674,10 +675,13 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
 	int already_installed = 0;
 	int bundles_installed = 0;
 	int bundles_requested = list_len(bundles);
+	long bundle_size = 0;
+	long fs_free = 0;
 	struct file *file;
 	struct list *iter;
 	struct list *to_install_bundles = NULL;
 	struct list *to_install_files = NULL;
+	struct list *ptr;
 	struct manifest *full_mom = NULL;
 	timelist times;
 
@@ -722,6 +726,49 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
 	to_install_files = consolidate_files(to_install_files);
 	to_install_files = filter_out_existing_files(to_install_files);
 	grabtime_stop(&times);
+
+	/* Step 1.5: Now that we know the bundle is valid, check if we have enough space */
+	if (!skip_diskspace_check) {
+		struct statvfs stat;
+		char *filepath = NULL;
+
+		ptr = list_head(to_install_bundles);
+		while (ptr) {
+			struct manifest *subman;
+			subman = ptr->data;
+			ptr = ptr->next;
+
+			bundle_size += subman->contentsize;
+		}
+
+		filepath = mk_full_filename(path_prefix, "/usr/");
+
+		/* Calculate free space on filepath */
+		if (statvfs(filepath, &stat) != 0) {
+			fs_free = -1;
+		} else {
+			fs_free = stat.f_bsize * stat.f_bavail;
+		}
+		free_string(&filepath);
+
+		/* Add 10% to bundle_size as a 'fudge factor' */
+		if (((bundle_size * 1.1) > fs_free && !skip_diskspace_check) || fs_free < 0) {
+			ret = EBUNDLE_INSTALL;
+
+			if (fs_free > 0) {
+				fprintf(stderr, "Error: Bundle too large by %ldM.\n", (bundle_size - fs_free) / 1000 / 1000);
+			} else {
+				fprintf(stderr, "Error: Unable to determine free space on filesystem.\n");
+			}
+
+			fprintf(stderr, "NOTE: currently, swupd only checks /usr/ "
+					"(or the passed-in path with /usr/ appended) for available space.\n");
+			fprintf(stderr, "To skip this error and install anyways, "
+					"add the --skip-diskspace-check flag to your command.\n");
+
+			goto out;
+		}
+	}
 
 	/* step 2: download neccessary packs */
 	grabtime_start(&times, "Download packs");
--- a/src/globals.c
+++ b/src/globals.c
@@ -54,6 +54,7 @@ char *mounted_dirs = NULL;
 char *bundle_to_add = NULL;
 struct timeval start_time;
 char *state_dir = NULL;
+int skip_diskspace_check = 0;
 
 /* NOTE: Today the content and version server urls are the same in
  * all cases.  It is highly likely these will eventually differ, eg:
--- a/src/swupd.h
+++ b/src/swupd.h
@@ -184,6 +184,7 @@ extern char *swupd_cmd;
 extern char *bundle_to_add;
 extern struct timeval start_time;
 extern char *state_dir;
+extern int skip_diskspace_check;
 
 extern char *version_url;
 extern char *content_url;
--- a/src/clr_bundle_add.c
+++ b/src/clr_bundle_add.c
@@ -57,6 +57,7 @@ static void print_help(const char *name)
 	fprintf(stderr, "   -t, --time              Show verbose time output for swupd operations\n");
 	fprintf(stderr, "   -N, --no-scripts        Do not run the post-update scripts and boot update tool\n");
 	fprintf(stderr, "   -b, --no-boot-update    Do not update the boot files using clr-boot-manager\n");
+	fprintf(stderr, "   --skip-diskspace-check  Do not check free disk space before adding bundle\n");
 	fprintf(stderr, "\n");
 }
 
@@ -77,6 +78,7 @@ static const struct option prog_opts[] = {
 	{ "time", no_argument, 0, 't' },
 	{ "no-scripts", no_argument, 0, 'N' },
 	{ "no-boot-update", no_argument, 0, 'b' },
+	{ "skip-diskspace-check", no_argument, &skip_diskspace_check, 1 },
 	{ 0, 0, 0, 0 }
 };
 
@@ -164,6 +166,8 @@ static bool parse_options(int argc, char **argv)
 				goto err;
 			}
 			set_cert_path(optarg);
+			break;
+		case 0:
 			break;
 		default:
 			fprintf(stderr, "error: unrecognized option\n\n");


src/bundle.c
```diff
@@ -723,6 +728,35 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+		/* do a deep copy of the includes list */
+		bundle->includes = list_deep_clone_strs(subman->includes);
+		bundle_info = list_prepend_data(bundle_info, bundle);
+		bundle_size += calculate_size(subman->component, bundle_info, 0);
+	}
+
+	fs_free = get_available_space("/usr/");
+
+	/* Add 10% to bundle_size as a 'fudge factor' */
+	if (bundle_size + (bundle_size * .10) > fs_free && !skip_diskspace_check) {
```
icarus-sparry: Multiply by 1.1 perhaps?

BTW, Very good error message.

src/bundle.c
```diff
@@ -723,6 +728,35 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
 	to_install_files = filter_out_existing_files(to_install_files);
 	grabtime_stop(&times);
 
+	/* Step 1.5: Now that we know the bundle is valid, check if we have enough space */
+	ptr = list_head(to_install_bundles);
```
icarus-sparry: The comment suggests that all this loop does is calculate the space. In that case should we skip it if the skip-diskspace-check flag is set?

If on the other hand it is doing more, in particular making a deep clone of the bundles, then the comment ought to make it clear that it is doing both.

src/globals.c
```diff
@@ -54,6 +54,7 @@ char *mounted_dirs = NULL;
 char *bundle_to_add = NULL;
 struct timeval start_time;
 char *state_dir = NULL;
+int skip_diskspace_check = false;
```
matthewrsj: `int ... false`? we have `bool` types.

src/swupd.h
```diff
@@ -398,6 +399,23 @@ typedef enum telem_prio_t {
 } telem_prio_t;
 extern void telemetry(telem_prio_t level, const char *class, const char *fmt, ...);
 
+/* fs_storage.c */
+
+/* bundle_result contains the information to print along with
```
matthewrsj: This comment could use an update since it isn't just for outputting search information now.

src/fs_storage.c
```diff
@@ -0,0 +1,77 @@
...
+			continue;
+		}
+
+		/* If installed is true the initial call to this recursive
+		 * function was for an installed bundle. In this case we want
+		 * to calculate the contentsize of all included bundles (we
+		 * know they are all installed already) so we can report the
+		 * total installed size on the system.
+		 *
+		 * Otherwise only add the contentsize of bundles not already
```
icarus-sparry: This comment worries me, What happens if a bundle grows a lot in size, e.g. editors gets neovim added as another editor?

I know that this is just "moved" code and the comment is already present, but has the calling circumstances changed too much for it still to be valid?

src/bundle.c
```diff
@@ -723,6 +728,35 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+	while (ptr) {
```
matthewrsj: You should actually skip this entire block if the user passes `--skip-diskspace-check` instead of calculating it every time regardless, *especially* since it requires recursing the manifest when we might not have the entire manifest tree on the system at this point (for example for bundles that are already installed on the system).

src/fs_storage.c
matthewrsj: Well since this is only used in the bundle-add codepath the size of bundles already on the system shouldn't change since it isn't updating them.

src/bundle.c
```diff
@@ -723,6 +728,35 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+		subman = ptr->data;
+		ptr = ptr->next;
+
+		struct bundle_result *bundle = NULL;
+		bundle = calloc(sizeof(struct bundle_result), 1);
+		/* copy relevant information over for future use */
+		strncpy(bundle->bundle_name, subman->component, BUNDLE_NAME_MAXLEN - 1);
```
otaviobp: the last parameter of the strncpy is the number of bytes to be copied including the '\0'. So you should use BUNDLE_NAME_MAXLEN and not BUNDLE_NAME_MAXLEN -1.

But anyway, there's a problem in this call. strncpy does't ensure that the dest string is null terminated, so you will have buffer overflows when the string is read.

And as the manifests are alive while you are calculating the bundle size you can have a pointer to the manifest and use the string already allocated in it. There's no need to duplicate that information.

src/bundle.c
```diff
@@ -723,6 +728,35 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+		bundle->topsize = subman->contentsize;
+		/* do a deep copy of the includes list */
+		bundle->includes = list_deep_clone_strs(subman->includes);
```
otaviobp: I have a really strong feeling that you don't need to duplicate that list to calculate the installation sizes.

src/bundle.c
```diff
@@ -723,6 +728,35 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+		bundle_info = list_prepend_data(bundle_info, bundle);
+		bundle_size += calculate_size(subman->component, bundle_info, 0);
```
otaviobp: s/0/false/

src/swupd.h
```diff
@@ -398,6 +399,23 @@ typedef enum telem_prio_t {
...
+ * the internal relevancy score used to sort the output */
+struct bundle_result {
+	char bundle_name[BUNDLE_NAME_MAXLEN];
+	long topsize;
+	long size;
+	double score;
+	struct list *files;
+	struct list *includes;
+	bool is_tracked;
+	bool seen;
```
otaviobp: bool is_tracked:1;
bool seen:1;

src/bundle.c
```diff
@@ -723,6 +728,35 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+	}
+
+	fs_free = get_available_space("/usr/");
```
otaviobp: You need to preppend the path informed by --path option

src/fs_storage.c
```diff
@@ -0,0 +1,77 @@
...
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <sys/statvfs.h>
+
+#include "swupd.h"
+
+/* recursively calculate size from a complete list of bundle_result structs */
+long calculate_size(char *bname, struct list *bundle_info, bool installed)
```
otaviobp: As far as I understood you have all information you need to calculate the bundle size in the manifest struct. I'd use this so you won't need to duplicate everything before calculating the size.

src/bundle.c
```diff
@@ -723,6 +728,35 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+		ptr = ptr->next;
+
+		struct bundle_result *bundle = NULL;
+		bundle = calloc(sizeof(struct bundle_result), 1);
+		/* copy relevant information over for future use */
+		strncpy(bundle->bundle_name, subman->component, BUNDLE_NAME_MAXLEN - 1);
+		bundle->topsize = subman->contentsize;
+		/* do a deep copy of the includes list */
+		bundle->includes = list_deep_clone_strs(subman->includes);
+		bundle_info = list_prepend_data(bundle_info, bundle);
```
otaviobp: You need to free bundle_info list

src/bundle.c
```diff
@@ -723,6 +728,35 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
 	grabtime_stop(&times);
 
+	/* Step 1.5: Now that we know the bundle is valid, check if we have enough space */
+	ptr = list_head(to_install_bundles);
+	while (ptr) {
+		subman = ptr->data;
+		ptr = ptr->next;
+
+		struct bundle_result *bundle = NULL;
+		bundle = calloc(sizeof(struct bundle_result), 1);
```
otaviobp: No free for this calloc

src/swupd.h
```diff
@@ -398,6 +399,23 @@ typedef enum telem_prio_t {
...
+};
+extern long calculate_size(char *bname, struct list *bundle_info, bool installed);
+extern long get_available_space(const char *path);
```
otaviobp: s/long/unsigned long/

src/fs_storage.c
```diff
@@ -0,0 +1,77 @@
...
+
+		/* add bundle sizes for includes as well */
+		struct list *ptr2;
+		ptr2 = list_head(bi->includes);
+		while (ptr2) {
+			char *inc = NULL;
+			inc = ptr2->data;
+			ptr2 = ptr2->next;
+			/* recursively add included sizes */
+			size += calculate_size(inc, bundle_info, installed);
```
otaviobp: There is one possible infinite recursion loop if you have an installed bundle that includes itself. I know that this won't happen in a system, but anyway this is a problem in the API of this function. You can fix this by moving the "bi->seen = true;" out of the if. I think this won't affect any other usages of this function, but it's better to check that. And It's pretty easy to get rid of this recursion too because it doesn't need to go in more than one level.

src/bundle.c
```diff
@@ -723,6 +728,35 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+		bundle_info = list_prepend_data(bundle_info, bundle);
+		bundle_size += calculate_size(subman->component, bundle_info, 0);
+	}
+
+	fs_free = get_available_space("/usr/");
+
+	/* Add 10% to bundle_size as a 'fudge factor' */
+	if (bundle_size + (bundle_size * .10) > fs_free && !skip_diskspace_check) {
+		ret = EBUNDLE_INSTALL;
+		fprintf(stderr, "Error: Bundle too large by %ldM.\n", (bundle_size - fs_free) / 1000 / 1000);
```
otaviobp: You should divide by 1024 instead of 1000 if you want the size in M

src/bundle.c
evil-brain: Agreed, will change this.

src/bundle.c
evil-brain: You're right - this step should be skipped if the flag is set.
Indeed, all this loop does is calculate the space requirements.
No sense in calculating something and not using it.

Will change this.

src/globals.c
evil-brain: bool types don't work with getopt long options :( The reference I used:  https://www.gnu.org/software/libc/manual/html_node/Getopt-Long-Options.html

That said, I should change this to  "= 0" at least because setting an int to false is incorrect.

Will change this.

src/swupd.h
evil-brain: Agreed, will update for V2.

src/bundle.c
evil-brain: @matthewrsj - Agreed, this was suggested above and I agree entirely and will change for V2.

@otaviobp - I like the for option, but it seems to me most of the code uses the while syntax. I may change this.

src/bundle.c
```diff
@@ -723,6 +728,35 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
 	to_install_files = filter_out_existing_files(to_install_files);
 	grabtime_stop(&times);
 
+	/* Step 1.5: Now that we know the bundle is valid, check if we have enough space */
+	ptr = list_head(to_install_bundles);
+	while (ptr) {
+		subman = ptr->data;
```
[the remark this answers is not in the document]
evil-brain: Agreed, subman's scope can be limited.

Will do for V2.

src/bundle.c
evil-brain: Agreed, I'll make fix the strncpy() issue and work towards removing the duplication.

src/bundle.c
evil-brain: Admittedly, this a turn-key function from search.c - I think it can be improved and will look into it for V2.

src/bundle.c
evil-brain: Agreed as well.

src/swupd.h
evil-brain: What do you mean? I should initialize the variables? (...because I probably should have...)

src/bundle.c
evil-brain: Agreed 100%, will fix.

src/bundle.c
evil-brain: Indeed I do.

Will fix.

src/bundle.c
evil-brain: Will fix.

src/bundle.c
evil-brain: Agreed, but 1000 was in use before I copy/pasted this function. Search.c uses 1000 for example.

Maybe this should be MiB instead of MB? Should we report in MiB or MB?

src/bundle.c
icarus-sparry: The point of strncpy of size X-1 after a calloc is that it will only copy up to X-1 bytes including the terminating NUL character. This leavs the final byte in the allocated buffer unchanged, and calloc will have set it to NUL, so this means that the string will be NUL terminated, regardless of size. Of course it may be trunc

src/bundle.c
matthewrsj: The intention for search was to report in MB, not MiB, this may be a different use-case though.

src/fs_storage.c
matthewrsj: Include loops are strictly prohibited in bundle generation (in various validation checks and in mixer code itself) as they cause many more issues than the issue you point out here. That said, I have no issue with fixing it here.

src/bundle.c
matthewrsj: I, for one, prefer the `while` loop because I think it's more readable but I know @icarus-sparry prefers the `for` loop. At the end of the day I'm fine with either, both are plenty readable.

icarus-sparry: I much prefer the `for` style as well, and in at least one place it would have removed a bug if it had been used. The existing style is to use while loops when traversing linked lists, and for loops for index calculations, so whilst I hate it, I have given up on this fight.

otaviobp: +1 for a LIST_FOREARCH macro

src/bundle.c
otaviobp: wow, I haven't seen that. Anyway, truncated strings will be a problem for this code because calculate_size uses a strcmp on that field. And I really think we can avoid duplicating this string by referencing the manifests.

src/swupd.h
otaviobp: that's used to use just one bit for those values. Take a look at struct file for example. Not sure though if this is going to work with bool, because unsigned int is used in other cases. can you check that?

evil-brain:
V2 uploaded.

There is an issue with the error handling (beyond just my changes) - jumping to 'out' can still print 'successfully installed bundles'. Should I file a new report? I figured it was kind of out of scope for these changes...

matthewrsj:
> Should I file a new report?

Yep, and if you know how to fix it assign it to yourself. Otherwise leave open for discussion.

src/bundle.c
```diff
@@ -723,6 +727,53 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
 	to_install_files = filter_out_existing_files(to_install_files);
 	grabtime_stop(&times);
 
+	/* Step 1.5: Now that we know the bundle is valid, check if we have enough space */
```
matthewrsj: future: the content of this `if` block should be refactored into its own function so we don't grow `install_bundles` to much. This function can be re-used when doing a `verify --install` then too.

src/bundle.c
```diff
@@ -723,6 +727,53 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+			strncpy(bundle->bundle_name, subman->component, BUNDLE_NAME_MAXLEN - 1);
+			bundle->topsize = subman->contentsize;
+			bundle->includes = subman->includes;
+			bundle_info = list_prepend_data(bundle_info, bundle);
+			bundle_size += calculate_size(subman->component, bundle_info, false);
+		}
+		list_free_list_and_data(bundle_info, free);
+
+		/* default is '/', so if '/', check 'usr', else, use the custom path. */
+		if(strcmp("/", path_prefix) == 0) {
```
matthewrsj: This block can be reduced to
```C
filepath = mk_full_filename(path_prefix, "/usr/");
fs_free = get_available_space(filepath);
```

otaviobp: You can just use  `string_or_die(&filepath, "%s/%s", path_prefix, "/usr/");` for the default path too. It's not a problem to have "//usr" as a path and the code gets easier to read.

src/bundle.c
```diff
@@ -723,6 +727,53 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+			fs_free = get_available_space("/usr/");
+		}
+		else {
+			string_or_die(&filepath, "%s/%s", path_prefix, "/usr/");
+			fs_free = get_available_space(filepath);
+		}
+
+		if (fs_free < 0){
+			fprintf(stderr, "Error: Could not determine free space in %s\n", filepath);
+			ret = EBUNDLE_INSTALL;
```
otaviobp: I prefer to print the same warning you are printing when there's no space, saying that we couldn't determine if there's enough space for the install, but users can override this using --skip-diskspace-check.

src/bundle.c
```diff
@@ -723,6 +727,53 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+			bundle_size += calculate_size(subman->component, bundle_info, false);
+		}
+		list_free_list_and_data(bundle_info, free);
+
+		/* default is '/', so if '/', check 'usr', else, use the custom path. */
+		if(strcmp("/", path_prefix) == 0) {
+			fs_free = get_available_space("/usr/");
+		}
+		else {
+			string_or_die(&filepath, "%s/%s", path_prefix, "/usr/");
```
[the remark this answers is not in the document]
matthewrsj: Ah yeah you need to have `free_string(&filepath);` before exiting

src/bundle.c
```diff
@@ -723,6 +727,53 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+			ptr = ptr->next;
+
+			struct bundle_result *bundle = NULL;
+			bundle = calloc(sizeof(struct bundle_result), 1);
+			/* point to or copy relevant information over for future use */
+			strncpy(bundle->bundle_name, subman->component, BUNDLE_NAME_MAXLEN - 1);
+			bundle->topsize = subman->contentsize;
+			bundle->includes = subman->includes;
+			bundle_info = list_prepend_data(bundle_info, bundle);
+			bundle_size += calculate_size(subman->component, bundle_info, false);
```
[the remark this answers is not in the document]
matthewrsj: nice I think you are right...

src/bundle.c
```diff
@@ -723,6 +727,53 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
...
+			fprintf(stderr, "Error: Could not determine free space in %s\n", filepath);
+			ret = EBUNDLE_INSTALL;
+			goto out;
+		}
+
+		/* Add 10% to bundle_size as a 'fudge factor' */
+		if ((bundle_size * 1.1) > fs_free && !skip_diskspace_check) {
+			ret = EBUNDLE_INSTALL;
+			fprintf(stderr, "Error: Bundle too large by %ldM.\n", (bundle_size - fs_free) / 1000 / 1000);
+			fprintf(stderr, "NOTE: currently, swupd only checks /usr/ (or a passed in path) for available space.\n");
```
matthewrsj: This is incorrect (or at least incomplete), since it *does* check /usr/ under the passed in path. This output makes it look like it is checking the root of the `--path` argument but it is actually checking `<path>/usr`

src/bundle.c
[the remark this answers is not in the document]
evil-brain: Done.

src/bundle.c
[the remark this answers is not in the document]
evil-brain: Tested this, and you're correct.

This triggers removal of quite a few changes and simplifies the commit a lot.

Thank you.

src/bundle.c
evil-brain: Done, but using mk_full_filename() as suggested above.

src/bundle.c
evil-brain: Done

src/bundle.c
evil-brain: Made output more explicit, done.

otaviobp:
+1. Lgtm now

PR #502, opened 2018-06-26
Add new test library testlib

castulo:
Work In Progress

The old way of creating new tests require a lot of manual steps to
create the test resources needed to run the tests. This commit is
part of a series of commits to improve the workflow for creation of
new tests.

This commit introduces a new test library that aims to replace the
old swupdlib.bash library which provides many funtions to create
new test resources easily.

The following functions were added in this commit:
- generate_random_content
- generate_random_name
- validate_path
- validate_item
- validate_param
- create_dir
- create_file
- create_link
- create_tar
- create_manifest
- add_to_manifest
- create_test_environment
- destroy_test_environment
- create_bundle
- set_env_variables
- sign_manifest

The tests in bundleadd_v2 would eventually replace those in bundleadd

Related to issue #258

### DIFF
--- a/test/functional/README.md
+++ b/test/functional/README.md
@@ -1,0 +1,269 @@
+# Writing tests for swupd-client
+<br/>
+
+When writing tests for swupd-client, the swupd test library (testlib.bash) should be used. This will facilitate the creation of test environments and test objects, and will also ensure that validations are performed in a consistent manner accross tests. The swupd-client uses BATS as the test framework of choice, to discover and run tests.  
+<br/>
+To use the test library you just need to source it in your shell
+```bash
+$ source testlib.bash
+```
+ or load it in your test script.  
+```bash
+load "testlib"
+```
+<br/>
+
+## Quickstart
+
+1. Source testlib.bash from your terminal, this will load all the test functions from the library into your current shell process:  
+```bash
+$ source testlib.bash
+```
+
+2. If necessary, create a directory to group the new test script with other test scripts of the same theme.  
+```bash
+$ mkdir <some_test_theme>
+$ cd <some_test_theme>
+```
+
+3. Create the new test script.  
+```bash
+$ generate_test <test_name>
+```
+
+A new test script will be generated in the current directory.  
+```bash
+#!/usr/bin/env bats
+
+load "../testlib"
+
+setup() {
+
+	create_test_environment "$TEST_NAME"
+	# create_bundle -n <bundle_name> -f <file_1>,<file_2>,<file_N> "$TEST_NAME"
+
+}
+
+teardown() {
+
+	destroy_test_environment "$TEST_NAME"
+
+}
+
+@test "<test description>" {
+
+	run sudo sh -c "$SWUPD <swupd_command> $SWUPD_OPTS <command_options>"
+	# <validations>
+
+}
+```
+
+4. The setup() function contains commands that should be executed before every test in the test script. This is usually the best place for creating test objects that are pre-requists for all tests in the script. The test library provides many [fixtures](#test-fixtures) to create all these pre-requisits easily.
+
+5. The teardown() function contains commands that should be executed after every test in the test script. It is the most common place for cleaning up test resources created for and by tests.
+
+6. Every test within the test file starts with the @test identifier followed by a description of the test. Add one or more tests to the script as needed. 
+
+7. From within each test, execute the command you want to validate by using the `run` helper.
+
+8. Validate that the program under test performed as expected by using one or more of the [assertions](#assertions) provided by the test library.
+<br/>
+
+### Test Principles
+
+* Every test needs to run in its own test environment so they don't interfere with each other.
+* Every test should clean up after itself.
+* Tests should be atomic and test only one thing per test.
+* Tests should be independent from each other, one should not need another one in order to work.
+* Test files from the same "theme" should be grouped within the theme directory. Tests from the same type can optionally be added to the same test file, for example:
+  * bundle-add  # A theme
+    * add-single-bundle.bats
+      * @test "add a bundle"
+      * @test "add a bundle that has only a directory"
+      * @test "add a bundle that has many files"
+      * @test "add a bundle that is already installed"
+      * @test "add a bundle that is invalid"
+      * etc.
+    * add-multiple-bundles.bats
+      * @test "add multiple bundles"
+      * @test "add multiple bundles, one valid, one invalid"
+      * etc.
+  * bundle-remove  # Another theme
+  * etc.
+<br/>
+
+### Running Tests
+To run a specific test script locally:  
+```bash
+$ bats <theme_directory>/<test_script>.bats
+```  
+
+To run all tests from a theme locally:  
+```bash
+$ bats <theme_directory>
+```  
+
+To run all tests locally:
+```bash
+$ cd swupd-client/test/functional
+$ bats *
+```
+
+To include tests to be run in the CI system:  
+TBD  
+<br/>
+
+## Test Fixtures
+### Test Environment
+A test environment is nothing but a directory that contains the necessary file structure that is used to emulate the existance of the following resources:
+* a target file system <test_enviroment>/target-dir
+* an alternate swupd state directory <test_enviroment>/state
+* a remote system that provides content file downloads <test_enviroment>/web-dir
+* the os-core bundle, since this bundle is required in every system it gets created by default in the content download directory, and "installed" in the target system
+
+To create a test environment called "my_env" for a test you would run the following command from within the test script.
+```bash
+# create a test environment for version 10 (default) 
+create_test_environment my_env 
+```
+Or
+```bash
+# create a test environment for version 20
+create_test_environment my_env 20
+```
+
+The test library provides the following functions for handling test environments:
+* create_test_environment
+* destroy_test_environment
+<br/>
+
+### Test Objects
+Test objects are all those elements that need to be mocked up in order to automate a test. The test library provides many functions for the creation and manipulation of these test objects. By far, the most useful test object that can be created by the test library are bundles. When creating a bundle using the test library, many other required test objects are created as a side effect, all things that are necessary for the bundle like, files, directories, manifests, packs, etc.
+
+Use the following command to crate a bundle with two files called "test-bundle" in the "my_env" test environment.
+```bash
+create_bundle -n test-bundle -f /foo/bar/test-file1,/baz/test-file2 my_env
+```
+By creating that bundle the following objects will also be created:
+* A hashed directory (to be used for the /foo, /foo/bar and /baz directories)
+* Two hashed files (test-file1 and test-file2)
+* A hashed tracking file for the bundle
+* A bundle manifest
+* A Manifest of Manifests (MoM)
+* A tar file for each file, directory and manifest
+* A zero pack for the bundle
+
+Another example:
+```bash
+create_bundle -L -n another-bundle -d /some_dir -f /baz/test-file -l /test-link my_env
+```
+This bundle, besides having characteristics similar to the previous bundle will have this new characteristics:
+* One directory without any file (/some_dir)
+* One link (test-link), this will also generate another extra file to which the symbolic link is pointing to
+* Since the -L flag was used, the bundle will not only be created in the directory for content download, but it will also be installed in the target file system, this is useful for tests that need a pre-installed bundle as pre-requisite 
+
+The following are some of the functions provided by the test library to create and handle test objects:
+* create_bundle: creates a bundle
+* create_dir: creates a hashed directory
+* create_file: creates a hashed file
+* create_link: creates a hashed file and a hashed link pointing to the file
+* create_tar: creates a tar of the specified object (manifest, full file, etc.)
+* create_manifest: creates an empty manifest (with initial headers)
+* add_to_manifest: adds the specified object (directory, file, or link) to the manifest, and creates/updates the manifest tar
+* add_dependency_to_manifest: adds another bundle as dependency in the manifest, and creates/updates the manifest tar 
+* remove_from_manifest: removes an object from the manifest, and updates the manifest tar
+* sign_manifest: signs the manifest using Swupd_Root.pem
+* update_hashes_in_mom: after modifying manifests included in a MoM, this function can be run to update the manifest hashes in the MoM
+* get_hash_from_manifest: retrieves the hash of an object within a manifest
+<br/>
+
+## Assertions
+The following assertions are included in the test library. These should be used to perform the test validations.
+
+*assert_status_is*  
+passes if the exit status matches the provided one, fails otherwise  
+Example:  
+```bash
+run <some_command>
+assert_status_is 0
+```
+
+*assert_status_is_not*  
+passes if the exit status does not match the provided one, fails otherwise  
+Example:  
+```bash
+run <some_command>
+assert_status_is_not 0
+```
+
+*assert_dir_exists*  
+passes if the provided directory exists, fails otherwise  
+Example:  
+```bash
+run <some_command>
+assert_dir_exists /some/directory
+```
+
+*assert_dir_not_exists*  
+passes if the provided directory does not exist, fails otherwise  
+Example:  
+```bash
+run <some_command>
+assert_dir_not_exists /some/directory
+```
+
+*assert_file_exists*  
+passes if the provided file exists, fails otherwise  
+Example:  
+```bash
+run <some_command>
+assert_file_exists /some/file
+```
+
+*assert_file_not_exists*  
+passes if the provided file does not exist, fails otherwise  
+Example:  
+```bash
+run <some_command>
+assert_file_not_exists /some/file
+```
+
+*assert_in_output*  
+passes if the provided text is included in the command output, fails otherwise  
+Examples:  
+```bash
+# one line strings
+run <some_command>
+assert_in_output "Successfully installed 1 bundle"
+
+# multi-line strings
+run <other_command>
+expected_output=$(cat <<-EOM
+  Some multi-ine text that needs to be present
+  in the exact order
+  some more lines, bla bla
+EOM
+)
+assert_in_output "$expected_output"
+
+# combination of both
+run <yet_another_command>
+expected_output=$(cat <<-EOM
+  Some multi-ine text that needs to be present
+  in the exact order
+  some more lines, bla bla
+EOM
+)
+assert_in_output "some literall text"
+assert_in_output "$expected_output"
+assert_in_output "more text to be checked"
+```
+
+*assert_not_in_output*  
+passes if the provided text is not included in the command output, fails otherwise  
+Examples:  
+```bash
+# one line strings
+run <some_command>
+assert_not_in_output "Error"
+```
--- a/test/functional/README.md
+++ b/test/functional/README.md
@@ -1,7 +1,11 @@
 # Writing tests for swupd-client
 <br/>
 
-When writing tests for swupd-client, the swupd test library (testlib.bash) should be used. This will facilitate the creation of test environments and test objects, and will also ensure that validations are performed in a consistent manner accross tests. The swupd-client uses BATS as the test framework of choice, to discover and run tests.  
+When writing tests for swupd-client, the swupd test library (testlib.bash) should
+be used. This will facilitate the creation of test environments and test objects,
+and will also ensure that validations are performed in a consistent manner accross
+tests. The swupd-client uses BATS as the test framework of choice, to discover and
+run tests.
 <br/>
 To use the test library you just need to source it in your shell
 ```bash
@@ -15,12 +19,14 @@ load "testlib"
 
 ## Quickstart
 
-1. Source testlib.bash from your terminal, this will load all the test functions from the library into your current shell process:  
+1. Source testlib.bash from your terminal, this will load all the test functions
+from the library into your current shell process:
 ```bash
 $ source testlib.bash
 ```
 
-2. If necessary, create a directory to group the new test script with other test scripts of the same theme.  
+2. If necessary, create a directory to group the new test script with other test
+scripts of the same theme.
 ```bash
 $ mkdir <some_test_theme>
 $ cd <some_test_theme>
@@ -37,16 +43,27 @@ A new test script will be generated in the current directory.
 
 load "../testlib"
 
-setup() {
+global_setup() {
 
-	create_test_environment "$TEST_NAME"
+	# global setup
+
+}
+
+test_setup() {
+
+	# create_test_environment "$TEST_NAME"
 	# create_bundle -n <bundle_name> -f <file_1>,<file_2>,<file_N> "$TEST_NAME"
 
 }
 
-teardown() {
+test_teardown() {
 
-	destroy_test_environment "$TEST_NAME"
+	# destroy_test_environment "$TEST_NAME"
+}
+
+global_teardown() {
+
+	# global cleanup
 
 }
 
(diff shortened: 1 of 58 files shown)


castulo:
This PR is a work in progress but I'd like to receive feedback about it. 

Also, the test "add-multiple" is failing because is is not installing the bundles correctly, there must still be bugs in the creation of the test environment, but I've been banging my head against the wall trying to resolve them with no luck, so any pointers would also be appreciated.

castulo:
I found what the problem was with the new "add-multiple" test. The test was adding a couple of bundles, and one of those bundles included this file "/etc/systemd/file2", but turns out as part of the heuristics of the bundle install the file was being filtered out in the installation process because it is in a directory that are not considered "stateless". I just modified the test to add a file in a directory not filtered.

matthewrsj:
@castulo you need to add the `bundleadd_v2` tests to the Makefile.am so they are actually run (line 111, alphabetical order).

castulo:
@matthewrsj I left the tests in bundleadd_v2 out of the Makefile on purpose for now so they are not run just yet. Those tests are not completely ready, I'm still reworking the check_lines function so all those tests need to have that validation added to them. 

Also, my idea is that we no longer need to add each test individually to the Makefile, that seems like a lot of work for something that is not necessary, we can just run something like "bats /my/dir/with/tests" and bats should be able to discover all the tests in that path by itself.

Finally, once the tests in bundeladd_v2/ are ready and reviewed to make sure they cover the same as the tests in bundleadd/, then we can remove bundleadd/ and rename bundled_v2/ as bundleadd/.

matthewrsj:
@icarus-sparry I'm starting to review this now, but I'd appreciate any comments you have on the bash bits.

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+export SWUPD="$SWUPD_DIR/swupd"
+
+
+generate_random_content() { 
+
+	local bottom_range=5
+	local top_range=100
+	local range=$((top_range - bottom_range + 1))
+	local number_of_lines=$((RANDOM%range))  
+	< /dev/urandom tr -dc 'a-zA-Z0-9-_!@#$%^&*()_+{}|:<>?=' | fold -w 100 | head -n $number_of_lines
```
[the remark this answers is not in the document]
icarus-sparry: Really? It is very straight forward. read bytes from /dev/urandom, delete any that are not listed, pack them into lines a hundred long and get the first N of them.

What is concerning is the bug in the previous line, where @castulo failed to add bottom_range into number_of_lines, so this will generate between 0 and 95 lines (inclusive) of output, rather than between 5 and 100.

test/functional/bundleadd_v2/add-rc.bats
```diff
@@ -0,0 +1,241 @@
...
+}
+
+# ------------------------------------------
+# Bad Cases (all bad bundles)
+# ------------------------------------------
+
+@test "bundle-add output: adding one bundle that is already added" {
+
+	run sudo sh -c "$SWUPD bundle-add $SWUPD_OPTS test-bundle2"
+	assert_status_is 18
```
matthewrsj: For the weird error codes such as this one let's add a comment explaining what it means, like
```bash
# 18 == EBUNDLEINSTALL
```
Just makes it easier for future developers who break tests because they change the exit status.

This should be done for the other instances of this in this file as well.

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+TEST_NAME=${TEST_FILENAME%.bats}
+export TEST_NAME
+export FUNC_DIR
+export SWUPD_DIR="$FUNC_DIR/../.."
+export SWUPD="$SWUPD_DIR/swupd"
+
+
+generate_random_content() { 
+
+	local bottom_range=5
```
icarus-sparry: Maybe have
```
            local bottom_range=${1:5}
            local top_range=${2:100}
```
to allow the amount of data required to be passed?

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+
+}
+
+generate_random_name() { 
+
+	local prefix=${1:-test-}
+	local uuid
+	
+	# generate random 8 character alphanumeric string (lowercase only)
+	uuid=$(< /dev/urandom tr -dc 'a-f0-9' | fold -w 8 | head -n 1)
```
icarus-sparry: Could use `read -n8` rather than fold+head. Of course if you do so you will need to move the echo into the subshell.

test/functional/bundleadd_v2/bad-hash-state.bats
```diff
@@ -0,0 +1,49 @@
...
+teardown() {
+
+	destroy_test_environment "$TEST_NAME"
+
+}
+
+@test "bundle-add add bundle with bad hash in state dir" {
+ 
+ 	# since one of the files needed to install the bundle is already in the state/staged
+ 	# directory, in theory this one shuld be used instead of downloading it again...
```
matthewrsj: `s/shuld/should/`

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+		echo "Please provide a valid file"
+		return 1
+	fi
+
+}
+
+validate_param() {
+
+	local param=$1
+	if [ -z "$param" ]; then
```
icarus-sparry: This doesn't allow a parameter to be set but empty. Probably OK in this context but the error message could be better.

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+		return 1
+	fi
+
+}
+
+validate_item() { 
+
+	local vfile=$1
+	if [ -z "$vfile" ] || [ ! -e "$vfile" ]; then
+		echo "Please provide a valid file"
```
icarus-sparry: Should abstract out the printing of the error message to another function. Then it will be easy to fix all of them to output to stderr, rather than stdout .

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+create_dir() { 
+	
+	local path=$1
+	local hashed_name
+	local directory
+	validate_path "$path"
+	
+	# most directories have the same hash, so we only need one directory
+	# in the files directory, if there is already one just return the path/name
+	directory=$(find "$path"/* -type d 2> /dev/null)
```
icarus-sparry: This worries me. What happens if we have two directories or more directories under $path?

test/functional/testlib.bash
```diff
@@ -0,0 +1,805 @@
...
+create_manifest() {
+
+	local path=$1
+	local name=$2
+	local version
+	validate_path "$path"
+	validate_param "$name"
+
+	version=$(basename "$path")	
+	echo -e "MANIFEST\\t3" | sudo tee "$path"/Manifest."$name" > /dev/null
```
matthewrsj: The only reason the manifest format was `3` before you touched it was because these tests were written when Clear Linux was on format 3. Now would be a good time to change this to something a little less arbitrary, like `1`

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+	sudo mv "$path"/testfile "$path"/"$hashed_name"
+	# since tar is all we use, create a tar for the new file
+	create_tar "$path"/"$hashed_name"
+	echo "$path/$hashed_name"
+
+}
+
+# Creates a symbolic link with a hashed name to the specified file in the specified path.
+# If no existing file is specified to point to, a new file will be created and pointed to
+# by the link.
```
icarus-sparry: Seems a little restrictive, this means it can not create dangling links.

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+# - BUNDLE_NAME: the name of the bundle which this manifest will be for
+create_manifest() {
+
+	local path=$1
+	local name=$2
+	local version
+	validate_path "$path"
+	validate_param "$name"
+
+	version=$(basename "$path")
```
icarus-sparry: I think this would be better structured as
```
     {
           printf 'MANIFEST\t3\n'
           printf 'version:\t%s\n' "$version"
           ...
     }  | sudo tee "$path/Manifest.$name" >/dev/null
     echo "$path/Manifest.$name"
```

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+	local name
+	local version
+	local filecount
+	local contentsize
+	local linked_file
+	validate_item "$manifest"
+	validate_item "$item"
+	validate_param "$item_path"
+	
+	item_size=$(du -b "$item" | cut -f1)
```
icarus-sparry: item_size=$(stat "%s" "$item)

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+	name=$(basename "$item")
+	version=$(basename "$(dirname "$manifest")")
+	# add to filecount
+	filecount=$(sudo cat "$manifest" | grep filecount | awk '{ print $2 }')
```
icarus-sparry: Any reason to suspect that $manifest is not readable? If not then
`filecount=$(awk '/filecount/ { print $2}' "$manifest")`

If it might not be readable then either add a `sudo` or a `sudo cat | awk` as taste dictates.

test/functional/testlib.bash
```diff
@@ -0,0 +1,805 @@
...
+	echo "HOME_URL=\"https://clearlinux.org\"" | sudo tee -a "$env_name"/target-dir/usr/lib/os-release > /dev/null
+	echo "SUPPORT_URL=\"https://clearlinux.org\"" | sudo tee -a "$env_name"/target-dir/usr/lib/os-release > /dev/null
+	echo "BUG_REPORT_URL=\"https://bugs.clearlinux.org/jira\"" | sudo tee -a "$env_name"/target-dir/usr/lib/os-release > /dev/null
+	sudo mkdir -p "$env_name"/target-dir/usr/share/clear/bundles
+
+	# state files & dirs
+	sudo mkdir -p "$env_name"/state/staged
+	sudo mkdir -p "$env_name"/state/download
+	sudo mkdir -p "$env_name"/state/delta
+	sudo mkdir -p "$env_name"/state/telemetry
```
matthewrsj: This can all be one line
```bash
sudo mkdir -p "$env_name"/state/{staged,download,delta,telemetry}
```

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+		# MoM has a contentsize of 0, so don't increase this for MoM
+		contentsize=0
+		# files, directories and links are stored already hashed, but since
+		# manifests are not stored hashed, we need to calculate the hash
+		# of the manifest before adding it to the MoM
+		name=$(sudo "$SWUPD" hashdump "$item" 2> /dev/null)
+	elif [ -L "$item" ]; then
+		item_type=L
+		# when adding a link to a bundle, we need to make sure we add
+		# its associated file too
```
icarus-sparry: We probably want to add the associated file, but I could see that we might want dangling links ....

test/functional/testlib.bash
```diff
@@ -0,0 +1,805 @@
...
+	bundle_name=${bundle_name:-$(generate_random_name test-bundle-)}
+	version=${version:-10}
+	if [ -z "$dir_list" ] && [ -z "$file_list" ] && [ -z "$link_list" ] ; then
+		# if nothing was specified to be created, at least create
+		# one directory which is the bare minimum for a bundle
+		dir_list=(/usr/bin)
+	fi
+	# all bundles should include their own tracking file, so append it to the
+	# list of files to be created in the bundle
+	file_list+=(/usr/share/clear/bundles/"$bundle_name")
```
matthewrsj: Nice, I was expecting you to miss this file :)

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+
+	local env_name=$1 
+	local version=${2:-10}
+	local mom
+	validate_param "$env_name"
+	
+	# create all the files and directories needed
+	# web-dir files & dirs
+	sudo mkdir -p "$env_name"/web-dir/version/formatstaging
+	echo "$version" | sudo tee "$env_name"/web-dir/version/formatstaging/latest > /dev/null
```
icarus-sparry: Probably worth abstracting a "write_protected" function, maybe
```
write_protected(){
        local arg
        [ "$1" = "-a" ] && { arg=-a ; shift ; }
        local file=${1?Missing output file in write_protect}
        shift
        printf "$@" | sudo tee $arg "$output" >/dev/null
}
```

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+	sudo mkdir -p "$env_name"/web-dir/"$version"/files
+	sudo mkdir -p "$env_name"/web-dir/"$version"/staged
+	mom=$(create_manifest "$env_name"/web-dir/"$version" MoM)
+	sign_manifest "$mom"
+
+	# target-dir files & dirs
+	sudo mkdir -p "$env_name"/target-dir/usr/lib
+	echo "NAME=\"Clear Linux Software for Intel Architecture\"" | sudo tee "$env_name"/target-dir/usr/lib/os-release > /dev/null
```
icarus-sparry: suggest restructure as `{ printf ; printf ... ;} | sudo tee'

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+# Destroys a test environment
+# Parameters:
+# - ENVIRONMENT_NAME: the name of the test environment to be deleted
+destroy_test_environment() { 
+
+	local env_name=$1
+	local deletable=true
+	validate_path "$env_name"
+	# since the action to be performed is very destructive, at least
+	# make sure the directory does look like a test environment
```
icarus-sparry: Nice comment and reasoning.

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+	for var in "state" "target-dir" "web-dir"; do
+		if [ ! -d "$env_name/$var" ]; then
+			deletable=false
```
icarus-sparry: I would go for print the message and return here, rather than having the local deletable variable

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+		return 1
+	fi
+
+}
+
+# Creates a bundle in the test environment. The bundle can contain files, directories or symlinks.
+create_bundle() { 
+
+	cb_usage() { 
+		echo "Usage:"
```
icarus-sparry: For something as long as this I would use cat <<EOF and put the text inline and to hell with the indentation.

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+	local env_name
+	local files_path
+	local version_path
+	local manifest
+	local local_bundle=false
+
+	set -f  # turn off globbing
+	while getopts :v:d:f:l:n:L opt; do
+		case "$opt" in
+			d)	IFS=,
```
icarus-sparry: I really don't like changing IFS globally, and even less so not either unsetting it or setting it back to a sane value after you have parsed the arguments. A helper function to read an array would be good, and in that you could declare IFS to be local.

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+		echo "Please specify a valid environment"
+		return 1
+	fi
+	version_path="$env_name"/web-dir/"$version"
+	files_path="$version_path"/files
+	target_path="$env_name"/target-dir
+
+	# 1) create the initial manifest
+	manifest=$(create_manifest "$version_path" "$bundle_name")
+	echo "Manifest -> $manifest"  # TODO(castulo): remove this msg when finish development
```
icarus-sparry: Could create a "debug" function which outputs these messages iff an environment variable is set

test/functional/testlib.bash
```diff
@@ -0,0 +1,502 @@
...
+		echo "Symlink -> $bundle_link"  # TODO(castulo): remove this msg when finish development
+		add_to_manifest "$manifest" "$bundle_link" "$val"
+		# Add the file pointed by the link to the zero pack of the bundle
+		pfile=$(readlink -f "$bundle_link")
+		sudo tar -rf "$version_path"/pack-"$bundle_name"-from-0.tar "$pfile"
+		if [ "$local_bundle" = true ]; then
+			sudo mkdir -p "$target_path$(dirname "$val")"
+			# if local_bundle is enabled copy the link to target-dir but also
+			# copy the file it points to
+			pfile_path=$(sudo cat "$manifest" | grep "$(basename "$pfile")" | awk '{ print $4 }')
```
icarus-sparry: `grep | awk` can always be replaced by `awk` (although it might be less readable). Almost always just having awk will be faster.

test/functional/bundleadd_v2/add-rc.bats
icarus-sparry: There is a good argument for defining all of these in testlib.bash,, so it could say `assert_status_is $EBUNDLEINSTALL`, which would also make the test more readable.

matthewrsj: This is a great suggestion @icarus-sparry

test/functional/README.md
```diff
@@ -0,0 +1,285 @@
...
+
+@test "<test description>" {
+
+	run sudo sh -c "$SWUPD <swupd_command> $SWUPD_OPTS <command_options>"
+	# <validations>
+
+}
+```
+
+4. The setup() function contains commands that should be executed before every test in the test script. This is usually the best place for creating test objects that are pre-requists for all tests in the script. The test library provides many [fixtures](#test-fixtures) to create all these pre-requisits easily.
```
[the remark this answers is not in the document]
castulo: ok

test/functional/README.md
```diff
@@ -0,0 +1,285 @@
...
+```bash
+$ bats <theme_directory>/<test_script>.bats
+```  
+
+To run all tests from a theme locally:  
+```bash
+$ bats <theme_directory>
+```  
+
+To run all tests locally:
```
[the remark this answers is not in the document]
castulo: I had not included that in the documentation yet because that is actually not the case yet, tests written with this library have not been added to the Makefile yet. I wanted to discuss with you if you really wanted to keep listing each test individually in the Makefile.am, I don't see the purpose in adding each one individually when we can just run something that would "discover" all the tests automatically, so people wouldn't have to worry about that with every new test. I like the idea of running `make check` as a way of running them, but I guess not the way the Makefile is including all those tests. Is there a specific reason why they are being added one by one?

test/functional/README.md
```diff
@@ -0,0 +1,285 @@
...
+create_test_environment my_env 20
+```
+
+The test library provides the following functions for handling test environments:
+* create_test_environment
+* destroy_test_environment
+<br/>
+
+### Test Objects
+Test objects are all those elements that need to be mocked up in order to automate a test. The test library provides many functions for the creation and manipulation of these test objects. By far, the most useful test object that can be created by the test library are bundles. When creating a bundle using the test library, many other required test objects are created as a side effect, all things that are necessary for the bundle like, files, directories, manifests, packs, etc.
```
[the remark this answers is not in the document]
castulo: ok

test/functional/README.md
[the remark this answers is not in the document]
matthewrsj: > we can just run something that would "discover" all the tests automatically

This would be great. Until this is done please add them to the Makefile.am so they are actually run (or just add the discovery).

test/functional/README.md
```diff
@@ -0,0 +1,285 @@
...
+
+Use the following command to crate a bundle with two files called "test-bundle" in the "my_env" test environment.
+```bash
+create_bundle -n test-bundle -f /foo/bar/test-file1,/baz/test-file2 my_env
+```
+By creating that bundle the following objects will also be created:
+* A hashed directory (to be used for the /foo, /foo/bar and /baz directories)
+* Two hashed files (test-file1 and test-file2)
```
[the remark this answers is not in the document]
castulo: the content those files will have is randomly generated so they always have different hashes. I will add that comment to the README.

test/functional/bundleadd_v2/add-rc.bats
castulo: yes, I'll include those in the test library

test/functional/README.md
```diff
@@ -0,0 +1,285 @@
+# Writing tests for swupd-client
+<br/>
+
+When writing tests for swupd-client, the swupd test library (testlib.bash) should be used. This will facilitate the creation of test environments and test objects, and will also ensure that validations are performed in a consistent manner accross tests. The swupd-client uses BATS as the test framework of choice, to discover and run tests.
```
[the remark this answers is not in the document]
castulo: done

test/functional/testlib.bash
castulo: sounds good to me

test/functional/testlib.bash
castulo: good call

test/functional/README.md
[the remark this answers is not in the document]
matthewrsj: thank you.

test/functional/testlib.bash
castulo: sure, I didn't see why this could be necessary... but it doesn't hurt anybody either and will provide flexibility, I'll add that

test/functional/testlib.bash
castulo: it should not happen, in general the content of the whole test environment is created through the test library, so we can control what goes where, and test environments are created and destroyed with every test....  and even if it happened because someone did this in one test (which would be a bad practice that could be caught when reviewing tests that are submitted) that test won't work properly, so the author of the test would only need to fix its test.

castulo: the $path that will be sent to this function will always be tests/functional/<test category>/<test theme>/<test environment>/web-dir/<version>/files, and this directory will always have only one directory

test/functional/testlib.bash
castulo: @icarus-sparry  what is a dangling link? 

This function is basically used by the create_bundle function to create a link (and its associated file) if a user chooses to add a link to its test bundle. I don't know if a dangling link is really necessary for a test bundle, I haven't even seen any test that currently uses a bundle that contains a link, I added the capability of adding links to make it more flexible

test/functional/testlib.bash
castulo: I didn't know of this method, I'll update the code. ;)

test/functional/testlib.bash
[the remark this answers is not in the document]
castulo: you are right, I forgot to add bottom_range, thanks for the comment

test/functional/bundleadd_v2/add-directory.bats
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bats
+
+load "../testlib"
+
+setup() {
```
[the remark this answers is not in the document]
castulo: That is a pretty good idea @icarus-sparry , I'll add that to the library

test/functional/testlib.bash
castulo: done

test/functional/testlib.bash
castulo: done

test/functional/testlib.bash
castulo: done

test/functional/testlib.bash
castulo: good point :)

test/functional/testlib.bash
castulo: that is a good idea

test/functional/testlib.bash
castulo: done

test/functional/testlib.bash
castulo: I modified that function, to look like this:
```bash
validate_param() {

	local param=$1
	: ${param:?"Mandatory parameter missing"}

}
```
Since the test library could also be sourced to be used from a shell instead of just from a script I wanted to find a way to terminate the program if the parameter was missing, I couldn't use exit because that would terminate the shell in case it was an interactive shell. So this way seems to do the trick.

icarus-sparry: Interesting. So if we are expecting the library to be both sourced into an interactive shell and into a shell script, this to me screams that we need to factor out the "I'm done because of an error" into a routine, and make that into something that calls exit if it is not interactive, and something like this if it is interactive.

test/functional/testlib.bash
castulo: I did something similar to the validate_param function here so the program is stopped if the file doesn't exists.

```bash
validate_item() { 

	local vfile=$1
	local param
	if [ -z "$vfile" ] || [ ! -e "$vfile" ]; then
		# since the lib could be sourced and run from an interactive shell
		# not only from a script, we cannot use exit since it would terminate
		# the shell, so we are using parameter expansion with a fake parameter
		# "param" to force an error
		: ${param:?"Please provide a valid file"}
	fi

}
```

test/functional/testlib.bash
castulo: Can you elaborate a little more please? I don't know what are dangling links or how to create/use those

icarus-sparry: A dangling symbolic link is a symbolic link that points to a file that doesn't exist.

You can create one with `ln -s /does_not_exist dangle` - after which `dandle` exists in the current directory but if you try and open it the open will fail with EEXIST as `/does_not_exist` is assumed not to exist.

test/functional/testlib.bash
castulo: what about just defining IFS local for this function?

icarus-sparry: Defining for just this junction would be better, but at about 170 lines long for this function, and it calling other functions, I still feel that the IFS change has too much scope.

test/functional/testlib.bash
castulo: done

test/functional/testlib.bash
castulo: to be honest I don't know how useful would it really be to be able to  source the library into an interactive shell, probably not much, I have been sourcing it a lot myself because I have been developing the test library, so I do it for testing purposes, but once the library is finished I don't know if it would really bring value to be able to source it. Probably everybody is going to use it from test scripts. Maybe is not worth trying to make it work fine from an interactive shell. What do you think?

castulo: I will add a function that exits depending on if the shell is interactive or not.

test/functional/testlib.bash
castulo: all messages from stdout and stderr are redirected to stderr if a test fails in bats anyway

test/functional/testlib.bash
castulo: done

matthewrsj:
You need to install the `realpath` dependency to the travis environment:
```diff
diff --git a/.travis.yml b/.travis.yml
index 0fe0e4f..7f86f0a 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ install:
         - tar -xvf v3.3.1.tar.gz
         - pushd libarchive-3.3.1 && autoreconf -fi && ./configure --prefix=/usr && make && sudo make install && popd
         - sudo apt-get install python3-docutils
+        - sudo apt-get install realpath
         - sudo ln -s /usr/share/docutils/scripts/python3/rst2man /usr/bin/rst2man.py
 
 # Ubuntu's default umask is 0002, but this break's swupd hash calculations.
```

matthewrsj:
You also need to add this line in the .gitignore
```diff
diff --git a/.gitignore b/.gitignore
index b632f70..6d2fc3c 100644
--- a/.gitignore
+++ b/.gitignore
@@ -59,4 +59,5 @@ test/functional/*/*/debug.log
 test/functional/*/*/state
 test/functional/*/*/lines-output
 test/functional/*/*/target-dir/usr/share/clear/bundles/.MoM
+test/functional/*/*.log
 verifytime
```

matthewrsj:
Other than that, +1 (travis should pass then).

test/functional/testlib.bash
```diff
@@ -0,0 +1,1015 @@
...
+
+# Exports environment variables that are dependent on the test environment
+# Parameters:
+# - ENV_NAME: the name of the test environment
+set_env_variables() {
+
+	local env_name=$1
+	local path
+	validate_path "$env_name"
+	path=$(dirname "$(realpath "env_name")")
```
icarus-sparry: BUG
Missing $ in $env_name

test/functional/testlib.bash
```diff
@@ -0,0 +1,1015 @@
...
+			sudo mkdir -p "$target_path$(dirname "$val")"
+			# if local_bundle is enabled since we cannot copy a bad link create a new one
+			# in the appropriate location in target-dir with the corrent name
+			sudo ln -s "$(generate_random_name /does_not_exist-)" "$target_path$val"
+		fi
+	done
+
+	# 6) Add the bundle to the MoM
+	add_to_manifest "$version_path"/Manifest.MoM "$manifest" "$bundle_name"
+
```
icarus-sparry: Adding `sign_manifest "$version_path"/Manifest.MoM` here gets another 25 tests to pass

matthewrsj:
@icarus-sparry you have a pending `requested changes` review. Has your review been addressed?

matthewrsj:
3... 2... 1... **SOLD** to the merge god

PR #504, opened 2018-06-27
test: Use different ports for tests to avoid colision

otaviobp:
When running all tests at the same time using make -j check,
update/slow-server and check-update/slow server were trying to start
the slow server at the same port.


otaviobp:
I closed that by accident.

matthewrsj:
So this just ups the range of ports that it tries by 1, so there is still a large overlap of ports that it tries. I guess this is fine though since it solves the issue of trying to start on the same port at the same time. The usual case is the first port will be free.

Issue #298, opened 2017-09-28
Improve pack downloads


matthewrsj:
Pack downloading is done in serial with pack extraction done after each individual pack download. Instead, extract each pack via a curl callback function so we aren't having to open new connections for each pack.

icarus-sparry:
Need to think about rate limiting, we don't want to starve the extraction by having too many partially complete downloads competing with each other - like Netscape used to do.

PR #507, opened 2018-06-29
Download

otaviobp:
Pull request with some fixes and some preparation to the fix of #298.

Please, consider a careful review of "5322aff: download: remove mutex lock in single thread code" because I don't know the reason the mutex lock was added to that file and I may have missed something.

Another point I'd like to be carefully reviewed is the "curl: Set fallback CA paths in swupd_curl_init()" commit. Is there any reason we needed to check the connection using swupd_curl_check_network() in a moment different from when we called swupd_curl_init()?

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -160,7 +160,7 @@ static size_t filesize_from_header_cb(void UNUSED_PARAM *func, size_t size, size
 	return (size_t)(size * nmemb);
 }
 
-double swupd_query_url_content_size(char *url)
+double swupd_curl_query_content_size(char *url)
 {
 	CURLcode curl_ret;
 	double content_size;
--- a/src/curl.c
+++ b/src/curl.c
@@ -288,7 +288,7 @@ int swupd_curl_get_file_full(const char *url, char *filename,
 	bool local_download = strncmp(url, "file://", 7) == 0;
 
 	if (!curl) {
-		abort();
+		return -1;
 	}
 restart_download:
 	curl_easy_reset(curl);
--- a/src/curl.c
+++ b/src/curl.c
@@ -48,10 +48,8 @@
 
 static CURL *curl = NULL;
 
-/* these are used to handle alternative trust locations */
-static char *capath;
-static char **fallback_capaths;
-static char fallback_capaths_no;
+/* alternative CA Path */
+static char *capath = NULL;
 
 /* Pretty print curl return status */
 static void swupd_curl_strerror(CURLcode curl_ret)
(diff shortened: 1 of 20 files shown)


phmccarty:
> Please, consider a careful review of "5322aff: download: remove mutex lock in single thread code" because I don't know the reason the mutex lock was added to that file and I may have missed something.

You didn't miss anything. This part of the code used to be multithreaded, but it predates the Github git history for the repo. So, removing the mutex locks is safe.

src/download.c
```diff
@@ -58,9 +58,12 @@ static struct list *failed = NULL;
  * traversing a manifest, doing hash comparisons, and (re)staging any files
  * whose hash miscompares.
  *
- * file->hash[0] acts as index into the arrays
+ * file->hash[0:1] acts as index into the arrays
  */
 #define SWUPD_CURL_HASH_BUCKETS 256
+#define HASH_VALUE(h) (h >= 'a' ? h - 'a' + 10 : h - '0')
+#define HASH_TO_KEY(hash) (HASH_VALUE(hash[0]) << 4 | HASH_VALUE(hash[1]))
```
matthewrsj: Maybe I'm being dense, but can you explain some of the magic here? Why are you shifting the first hash value left 4? Why add 10 to h - 'a' ?

src/curl.c
```diff
@@ -187,6 +148,10 @@ void swupd_curl_deinit(void)
 		curl_easy_cleanup(curl);
 	}
 	curl = NULL;
+
+	free(capath);
```
matthewrsj: Can't you just `free_string(&capath);`?

test/functional/verify/latest-missing/test.bats
```diff
@@ -5,7 +5,7 @@ load "../../swupdlib"
 @test "verify install using latest with missing version file on server" {
   run sudo sh -c "$SWUPD verify $SWUPD_OPTS --install -m latest"
 
-  check_lines "$output"
```
matthewrsj: Why aren't we checking the output any more? Are we losing the error message?

src/download.c
icarus-sparry: HASH_VALUE converts a character to its hexadecimal value (0 to 15 assuming that the input was from the set 0123456789abcdef ).
HASH_TO_KEY converts two characters to a value in the range 0 to 255.

src/curl.c
otaviobp: Yes. I should have used it.

test/functional/verify/latest-missing/test.bats
otaviobp: Good question.

This test is checking the behavior of swupd when the latest file was missing in the server. Before this patch CA Paths weren't used when downloading the latest file, so the error messages have changed. The problem is that the output will be different if the code is compiled with or without CA paths and looked enough. I haven't realized though that this is easy to fix using reg exp. I'll fix that and check the output correctly.

otaviobp:
On Tue, 2018-07-03 at 08:31 -0700, Matthew Johnson wrote:
> @matthewrsj commented on this pull request.
> curl: Aborting on invalid parameter is a bit too much: +1 on the
> change, but the commit message is confusing. This isn't an invalid
> parameter, this indicates that curl hasn't been initialized (it isn't
> in the parameter list).

My bad. I'll fix the commit message

src/download.c
otaviobp: Exactly. I'll comment this macro.

otaviobp:
Sent new version addressing the suggestions

test/functional/verify/latest-missing/lines-checked
```diff
@@ -1,2 +1,2 @@
-Unable to get latest version for install
-Error: Fix did not fully succeed
+Curl error: (37) Couldn't read a file:// file
+Failed verify initialization, exiting now.
```
phmccarty: Is this an expected change in test output?

otaviobp: yes

PR #512, opened 2018-07-03
Fix output message in bundle-add

### LINKED ISSUE #511: Incorrect message shown when bundle-add fails to download a file
There is a bug in bundle-add that causes the wrong output message to be shown when a bundle fails to be installed due to not being able to download a file (it show a success message at the bottom when it should not). 

$ sudo swupd bundle-add test-bundle

   File /test-file was not in a pack
   .
   Error for e0cf1dae844bd70fa88dd768ef52578542cd742f0aebb9cdb787671074ea48fe tarfile extraction, (check free space for swupd-client/test/functional/bundleadd_v2/tmp/state?)
   Starting download retry #1
   ..
   Error for e0cf1dae844bd70fa88dd768ef52578542cd742f0aebb9cdb787671074ea48fe tarfile extraction, (check free space for swupd-client/test/functional/bundleadd_v2/tmp/state?)
   Starting download retry #2
   ..
   Error for e0cf1dae844bd70fa88dd768ef52578542cd742f0aebb9cdb787671074ea48fe tarfile extraction, (check free space for swupd-client/test/functional/bundleadd_v2/tmp/state?)
   Starting download retry #3
   ..
   Error for e0cf1dae844bd70fa88dd768ef52578542cd742f0aebb9cdb787671074ea48fe tarfile extraction, (check free space for swupd-client/test/functional/bundleadd_v2/tmp/state?)
   ERROR: Could not download some files from bundles, aborting bundle installation.
   **Successfully installed 1 bundle**

castulo:
There is a bug in bundle-add that causes the wrong output message to
be shown when a bundle fails to be installed due to not being able to
download a file that is part of the bundle.

This commit fixes the logic so messages are shown correctly.

Fixes #511


src/bundle.c
```diff
@@ -946,7 +946,11 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
 	}
 
 	/* print totals */
-	bundles_failed = bundles_requested - bundles_installed - already_installed;
+	if (ret && bundles_installed != 0) {
+		bundles_failed = bundles_requested;
```
matthewrsj: What if you install three bundles, but only one fails. Will this print that all three bundles failed to install?

icarus-sparry: @matthewrsj thank you for writing this. A fix is clearly needed, but this change seemed too simplistic. I just couldn't put my finger on exactly why.

castulo: @matthewrsj, it depends on why it fails. So in the example you are mentioning, when trying to install 3 bundles, If one bundle fails to be installed because it was already installed, or because it is an invalid bundle (bad name), then the other two bundles will be installed and an error message like this would be shown:
*Successfully installed 1 bundle*
*1 bundle was already installed*

But if the bundle installation fails for any other reason, then yes, all three bundles will be shown as failed.
*Failed to install 3 of 3 bundles*

But based on what I see in the code I think this is fine because from this point on https://github.com/clearlinux/swupd-client/blob/master/src/bundle.c#L724, looks like if anything fails the whole process gets aborted.
- If there is no enough space on disk, it aborts the installation.
- When downloading the files, if some file fails to be downloaded, it aborts the installation.
- When installing the files in the file system
    * If the hash of one file does not match the hash in the manifest, and the bad file cannot be removed, it aborts the installation.
    * If the manifest of one of the bundles to be installed cannot be loaded, it aborts the installation.
    * If there is an error staging a file, it aborts the installation.
    * If verify_fix_path is unable to fix a path, it aborts the installation.

So looks like with almost every failure the whole bundle installation is aborted, except for the cases when the name of the bundle is bad, or the bundle is already installed. That's why I think this logic works. But please let me know if I am missing something.

Also I ran the following tests to verify this patch, all produced the correct exit code and the appropriate output message:
 ✓ bundle-add output: adding one bundle
 ✓ bundle-add output: adding multiple bundles
 ✓ bundle-add output: adding one bundle that is already added
 ✓ bundle-add output: adding one bundle that does not exist
 ✓ bundle-add output: adding multiple bundles, all invalid, both non existent
 ✓ bundle-add output: adding multiple bundles, all invalid, both already added
 ✓ bundle-add output: adding multiple bundles, all invalid, one already added, one invalid
 ✓ bundle-add output: adding multiple bundles, one valid, one already added
 ✓ bundle-add output: adding multiple bundles, one valid, one non existent
 ✓ bundle-add output: adding multiple bundles, one with a missing file and one valid

10 tests, 0 failures

matthewrsj: Ah, looks like you are right. Thanks for the code link and the writeup.

matthewrsj: Can you please add a comment explaining that reaching this point with a non-zero returncode means that no bundles were installed? That way you/we won't have to explain this to everyone else who sees this logic.

After that's done this is merge-able.

castulo: Hmm, actually I just found one scenario where this logic won't work quite well, I'll give it another thought.

castulo:
I ran the following tests in this patch (these tests were created using the new test library, they will be included as soon as the test library is approved and merged):
 ✓ bundle-add output: adding one bundle
 ✓ bundle-add output: adding multiple bundles
 ✓ bundle-add output: adding one bundle that is already added
 ✓ bundle-add output: adding one bundle that does not exist
 ✓ bundle-add output: adding multiple bundles, all invalid, both non existent
 ✓ bundle-add output: adding multiple bundles, all invalid, both already added
 ✓ bundle-add output: adding multiple bundles, all invalid, one already added, one invalid
 ✓ bundle-add output: adding multiple bundles, one valid, one already added
 ✓ bundle-add output: adding multiple bundles, one valid, one non existent
 ✓ bundle-add output: adding multiple bundles, one valid, one already added, one non existent
 ✓ bundle-add output: adding multiple bundles, one valid, one invalid, one already installed, one with a missing file

matthewrsj:
> Hmm, actually I just found one scenario where this logic won't work quite well, I'll give it another thought.

Everything good or should I still hold off?

PR #515, opened 2018-07-06
Parallel downloads

otaviobp:
Make the parallel downloader generic. Currently still using it only to download fullfiles, but it's ready to be used for any other downloads. Download resumes is still not supported and I added a warning for that.

And sorry for the large commit. I was really hard to think in a way to do this changes in smaller steps.


matthewrsj:
Please run `make compliant` against this branch.

src/hashmap.h
```diff
@@ -0,0 +1,84 @@
...
+
+/*
+ * Retrieve an element from the hashmap.
+ *
+ * Return the element or NULL if not found.
+ */
+void *hashmap_get(struct hashmap *hashmap, const void *key);
+
+/*
+ * Remove and element from the hashmap.
```
matthewrsj: `s/and/an/`

src/download.c
```diff
@@ -125,23 +115,34 @@ static int swupd_curl_hashmap_insert(struct file *file)
...
+}
+
 int start_full_download(bool pipelining)
 {
+	if (swupd_curl_hashmap) {
+		return -1;
+	}
+
 	failed = NULL;
+	swupd_curl_hashmap = hashmap_new(SWUPD_CURL_HASH_BUCKETS, file_hash_cmp, file_hash_value);
```
matthewrsj: `hashmap_new` can return `NULL` if the `calloc` fails. Are you sure all code that uses `swupd_curl_hashmap` after this point is checking for this case?

test/functional/verify/latest-missing/test.bats
```diff
@@ -5,7 +5,7 @@ load "../../swupdlib"
 @test "verify install using latest with missing version file on server" {
   run sudo sh -c "$SWUPD verify $SWUPD_OPTS --install -m latest"
 
-  check_lines "$output"
+  check_lines "$(echo "$output" | uniq)"
```
matthewrsj: Why pipe to uniq? I must be missing something.

src/download.c
```diff
@@ -29,29 +29,28 @@
...
+ * swupd_curl_parallel_download_start(), enqueue files using
+ * swupd_curl_parallel_download_enqueue() and finish the process using
+ * swupd_curl_parallel_download_end(), which will block until the previously
+ * queued downloads are completed.
+ */
 
 /*
  * The following represents a hash data structure used inside download.c
- * to track file downloads asynchonously via libcurl.  In the past a
+ * to track file downloads parallelhonously via libcurl.  In the past a
```
matthewrsj: > parallelhonously

:question:

src/download.c
```diff
@@ -74,252 +73,107 @@ static struct list *failed = NULL;
...
+ * Struct represinting a single file being downloaded.
+ */
+struct multi_curl_file {
+	struct curl_file file; /* Curl file information */
 
-	// clean up in case any prior download failed in a partial state
-	string_or_die(&tar_dotfile, "%s/download/.%s.tar", state_dir, file->hash);
-	unlink(tar_dotfile);
-	free_string(&tar_dotfile);
+	char retries; /* Number of retried performed so far */
```
matthewrsj: Why is this a `char`?

src/download.c
```diff
@@ -74,252 +73,107 @@ static struct list *failed = NULL;
...
 
 static bool file_hash_cmp(const void *a, const void *b)
 {
-	const struct file *fa = a;
-	const struct file *fb = b;
+	const struct multi_curl_file *fa = a;
+	const struct multi_curl_file *fb = b;
 
-	return hash_equal(fa->hash, fb->hash);
+	return strcmp(fa->hash, fb->hash) == 0;
```
matthewrsj: Reason for this change?

icarus-sparry: hash is supposed to be an abstract data type, so I am very much with @matthewrsj here, and would like this reverted.

src/download.c
```diff
@@ -511,115 +336,189 @@ static int poll_fewer_than(size_t xfer_queue_high, size_t xfer_queue_low)
...
+ *  - handle: Handle created with swupd_curl_parallel_download_start().
+ *  - url: The url to be downloaded.
+ *  - filename: Full path of the filename to save the download content.
+ *  - hash: Optional hex string with hash to be used as unique identifier of this
+ *    file. If NULL, filename will be used as the identifier. String MUST contain
+ *    only characters in '0123456789abcdef'.
+ *  - data: User data to be informed to success_cb().
+ *
+ * Note: This function MAY be blocked.
+ */
```
matthewrsj: Good function comment, thank you.

src/download.c
```diff
@@ -26,325 +26,154 @@
...
-		 * to retry in these situations. */
-	} else {
-		/* Only unlink when tar succeeded, so we can examine the tar file
-		 * in the failure case. */
-		unlink(tarfile);
-	}
+	h->max_xfer = max_xfer;
+	h->max_xfer_bottom = max_xfer_bottom;
+	h->success_cb = success_cb;
+	h->curl_hashmap = hashmap_new(SWUPD_CURL_HASH_BUCKETS, file_hash_cmp, file_hash_value);
```
matthewrsj: hashmap_new can return NULL if the calloc fails. Are you sure all code that uses h->curl_hashmap after this point is checking for this case?

(copied from out-of-date comment)

matthewrsj:
just a few comments. For the most part the implementation looks very good. Address or respond to my comments and I'll approve.

test/functional/verify/latest-missing/test.bats
otaviobp: Swupd will try to download the version file using each one of the ca paths informed in compile time and we are going to have one curl error for each one of them. When using uniq the test will work no matter how many ca paths were used.

src/download.c
otaviobp: Max retries is very short so we need only a byte to represent that. If we let users to set this we may change this to an int, but I can't se a reasonable reason to have max_retries > 256 (unsigned char).

src/download.c
otaviobp: Yes when I did it, but not right now. I'll fix that

src/download.c
otaviobp: I'll add one abort() after the calloc(). It looks like that's the approach we're using on swupd right now. Should we have a calloc_or_die() too? Or this is a bit too much?

src/download.c
```diff
@@ -511,115 +336,189 @@ static int poll_fewer_than(size_t xfer_queue_high, size_t xfer_queue_low)
...
+				//Remove item
+				if (l == h->failed) {
+					h->failed = l->next;
+				}
+				next = l->next;
+				list_free_item(l, NULL);
+				l = next;
+
+				fprintf(stderr, "Starting download retry #%d\n", file->retries);
+				//fprintf(stderr, "Starting download retry #%d for %s\n", file->retries, file->url);
```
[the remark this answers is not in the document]
otaviobp: Actually the commented one was supposed to be committed. I'll fix this

otaviobp:
New version with reported problems fixed and I run make compliant.

src/download.c
matthewrsj: We can add a calloc_or_die() when we are prepared to move all our malloc/callocs over to using it.

src/download.c
matthewrsj: +1

PR #517, opened 2018-07-10
Removing unnecesary code to untrack bundles

castulo:
Every bundle includes a tracking file which tells the OS the
bundle is installed, every bundle has this tracking file as
part of its manifest. When removing a bundle from a system,
all the files that are part of that bundle, and are not needed,
for any other bundle are removed, including the tracking file.

This patch removes a function call that is intended to remove
the tracking bundle from the system, since it is not necessary,
by the time the code calls that function, all files from the
manifest that can be removed are already removed, and this includes
the tracking file.

This patch also modifies a few tests to reflect the change in the
command output.


otaviobp:
This makes sense to me. +1

PR #519, opened 2018-07-13
fix file descriptor leak check.

icarus-sparry:
The readlink call was being used inefficiently (it has a lousy api).
Close extra file descriptors that are passed in to us, so we don't
complain if they are open at the end.

### DIFF
--- a/src/filedesc.c
+++ b/src/filedesc.c
@@ -30,55 +30,76 @@
 
 #include "swupd.h"
 
-void dump_file_descriptor_leaks(void)
+static void foreach_open_fd(void(pf)(int, void *), void *arg)
 {
 	DIR *dir;
 	struct dirent *entry;
+	int dir_fd;
 
 	dir = opendir("/proc/self/fd");
 	if (!dir) {
 		return;
 	}
+	/* we hold an fd open, the one from opendir above */
+	dir_fd = dirfd(dir);
 
-	while (1) {
-		char *filename;
-		char buffer[PATH_MAXLEN + 1];
-		entry = readdir(dir);
-		size_t size;
-		if (!entry) {
-			break;
-		}
-		if (strcmp(entry->d_name, ".") == 0) {
-			continue;
-		}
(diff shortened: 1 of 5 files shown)


matthewrsj:
+10 for the ignore lines removal. Now we *really* have to make sure travis passes :)

matthewrsj:
> Now we really have to make sure travis passes :)

Shoot

src/filedesc.c
```diff
@@ -81,6 +81,7 @@ static void foreach_open_fd(void(pf)(int, void *), void *arg)
 		if (n == 0 || n == 1 || n == 2 || n == dir_fd) {
 			continue;
 		}
+		fprintf(stderr, "foreach_open_fd processing %d\n", n);
```
otaviobp: Just a hint. When I need to force CI to run again I make a "git commit --amend" and don't change anything. Git will generate a different commit hash and when you push it to github it will consider a new version of the commit.

matthewrsj: There is actually a "Restart build" button in travis if you have write access to the repo (which I believe all current contributors have).

otaviobp: :man_facepalming:

icarus-sparry: Actually this was a debugging message to actually help me debug.

icarus-sparry:
@otaviobp So my current plan is to remove the dummy commit and the revert of it, reorder the 4 remaining patches to move the two which affect the travis.yml file before the two which affect the storing and reporting of the file descriptor leaks. I am happy to squash the latter two together but @matthewrsj liked them apart. Then do a force push of all 4.

matthewrsj:
@icarus-sparry +1 once you do that.

matthewrsj:
@icarus-sparry I learned recently that github sorts by author date, not by actual sequence of the commits. :confused:

icarus-sparry:
@matthewrsj Thank you. Rewrote all the author dates.

PR #523, opened 2018-07-20
Pack downloads

### LINKED ISSUE #298: Improve pack downloads

otaviobp:
Using parallel download API do download packs in parallel.

Closes #298

### DIFF
--- a/src/download.c
+++ b/src/download.c
@@ -79,10 +79,12 @@ struct swupd_curl_parallel_handle {
 	size_t mcurl_size, max_xfer, max_xfer_bottom; /* hysteresis parameters */
 	bool resume_failed;
 
-	CURLM *mcurl;			   /* Curl handle */
-	struct list *failed;		   /* List of failed downloads */
-	struct hashmap *curl_hashmap;      /* Hashmap mentioned above */
-	swupd_curl_download_cb success_cb; /* Callback to sucess function */
+	CURLM *mcurl;			  /* Curl handle */
+	struct list *failed;		  /* List of failed downloads */
+	struct hashmap *curl_hashmap;     /* Hashmap mentioned above */
+	swupd_curl_success_cb success_cb; /* Callback to success function */
+	swupd_curl_error_cb error_cb;     /* Callback to error function */
+	swupd_curl_free_cb free_cb;       /* Callback to free user data*/
 };
 
 /*
@@ -121,6 +123,10 @@ static void free_curl_file(struct swupd_curl_parallel_handle *h, struct multi_cu
 	CURL *curl = file->curl;
 	swupd_download_file_complete(CURLE_OK, &file->file);
 
+	if (h->free_cb) {
+		h->free_cb(file->data);
+	}
+
 	free_string(&file->url);
 	free_string(&file->file.path);
 	if (curl != NULL) {
@@ -138,14 +144,13 @@ static void free_curl_file(struct swupd_curl_parallel_handle *h, struct multi_cu
  * Parameters:
  *  - max_xfer: The maximum number of simultaneos downloads.
  *  - max_xfer_bottom: Minimum number of simultaneos downloads before starting
- *  - sucess_cb(): Callback called for each sucessfull download.
  *
  * Parallel download handler will retry MAX_TRIES times to download each file,
  * ading a timeout between each try.
  *
  * Note: This function is non-blocking.
  */
-void *swupd_curl_parallel_download_start(size_t max_xfer, size_t max_xfer_bottom, swupd_curl_download_cb success_cb)
+void *swupd_curl_parallel_download_start(size_t max_xfer, size_t max_xfer_bottom)
 {
 	struct swupd_curl_parallel_handle *h = calloc(1, sizeof(struct swupd_curl_parallel_handle));
 	if (!h) {
@@ -161,7 +166,6 @@ void *swupd_curl_parallel_download_start(size_t max_xfer, size_t max_xfer_bottom
 
 	h->max_xfer = max_xfer;
 	h->max_xfer_bottom = max_xfer_bottom;
-	h->success_cb = success_cb;
 	h->curl_hashmap = hashmap_new(SWUPD_CURL_HASH_BUCKETS, file_hash_cmp, file_hash_value);
 	h->timeout = RETRY_TIMEOUT;
 
@@ -169,6 +173,35 @@ void *swupd_curl_parallel_download_start(size_t max_xfer, size_t max_xfer_bottom
 error:
 	free(h);
 	return NULL;
+}
+
+/*
+ * Set parallel downloads callbacks.
+ *
+ *  - success_cb(): Called for each successful download, where data is the data
+ *                  informed on swupd_curl_parallel_download_enqueue().
+ *                  success_cb should return true if the download file was
+ *                  handled correctly. Return false to schedule a retry.
+ *  - error_cb():   Called for each failed download, where data is the data
+ *                  informed on swupd_curl_parallel_download_enqueue()
+ *                  and response is the HTTP response code. error_cb should
+ *                  return true if the error was handled by caller or false
+ *                  to schedule a retry.
+ * - free_cb():     Called when data is ready to be freed.
+ */
+void swupd_curl_parallel_download_set_callbacks(void *handle, swupd_curl_success_cb success_cb, swupd_curl_error_cb error_cb, swupd_curl_free_cb free_cb)
+{
+	struct swupd_curl_parallel_handle *h;
+
+	if (!handle) {
+		fprintf(stderr, "Invalid parallel download handle\n");
+		return;
+	}
+	h = handle;
+
+	h->success_cb = success_cb;
+	h->error_cb = error_cb;
+	h->free_cb = free_cb;
 }
 
 // Try to process at most COUNT messages from the curl multi-stack.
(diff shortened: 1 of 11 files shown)


src/packs.c
```diff
@@ -60,56 +131,47 @@ static int download_pack(int oldversion, int newversion, char *module, int is_mi
...
 	struct sub *sub = NULL;
 	int err;
 	int is_mix = 0;
 	unsigned int list_length = list_len(subs);
 	unsigned int complete = 0;
+	void *download_handle;
 
 	fprintf(stderr, "Downloading packs...\n");
+	download_handle = swupd_curl_parallel_download_start(MAX_XFER, MAX_XFER_BOTTOM);
+	download_successful(NULL);
```
matthewrsj: Why call this now? it will just return false immediately because of the `NULL` param right?

test/functional/update/boot-file/lines-checked
```diff
@@ -1,5 +1,6 @@
 Update started.
 Preparing to update from 10 to 100
+REGEXP:Error for .*pack-os-core-from-10-to-100.tar download: Response 0 - Couldn't read a file:// file
```
matthewrsj: Why are all of these necessary all of a sudden, and does your next commit make them unnecessary again?

matthewrsj:
For enhancements such as this please include some sort of analysis of the performance improvements in the PR itself.

I performed some very limited testing on my system updating from 23820 -> 23830 with the following bundles installed:
```
ansible            dev-utils-dev  kvm-host           os-core-update-dev     storage-utils-dev
bootloader         editors        libX11client       os-core-update-index*  sudo
c-basic            editors-dev    mail-utils         os-installer           sysadmin-basic
cockpit            emacs          mail-utils-dev     p11-kit                sysadmin-basic-dev
containers-basic   ethtool        mixer              package-utils          sysadmin-remote
desktop            git            network-basic      perl-basic             telemetrics
desktop-apps       go-basic       network-basic-dev  perl-basic-dev         user-basic
desktop-assets     go-basic-dev   openssh-server     python2-basic          vim
desktop-autostart  iproute2       openssl            python3-basic          web-server-basic
desktop-gnomelibs  joe            os-clr-on-clr      python-basic           which
desktop-i3         jq             os-core            redis-native           wpa_supplicant
desktop-locales    kernel-native  os-core-dev        shells                 xemacs
dev-utils          koji           os-core-update     storage-utils
```
I performed five updates each with the current client and the client from this PR

| current | new |
| --- | --- |
| 18.8s | 16.0s |
| 19.8s | 17.0s |
| 18.6s | 16.8s |
| 22.3s | 17.0s |
| 17.7s | 16.8s |
| **19.4s avg** | **16.7s avg** |

So ~14% speedup from this very limited test. Very nice. Clearly most of the time taken is in the pack extraction phase though.

test/functional/update/boot-file/lines-checked
matthewrsj: It appears it is still necessary.

matthewrsj:
This appears to work with `mixin` integration as well. Thank you very much for ensuring that @otaviobp

src/packs.c
otaviobp: My bad. That was a test I forgot to remove. I'll fix this.

test/functional/update/boot-file/lines-checked
otaviobp: These errors were occurring before, but there weren't any logs for them. And the parallel download is logging that.

matthewrsj: hmm, is there a way to ignore errors in the delta pack case? not having a delta pack is not an error since delta packs are optional.

aaronovz1:
@matthewrsj if pack extraction is a bottleneck would there be benefit in doing them in parallel as well with some kind of async task processor?

matthewrsj:
@aaronovz1 pack extraction is actually done via a callback on the download itself, so it should be happening asynchronously already. Correct @otaviobp?

otaviobp:
@aaronovz1 @matthewrsj Downloads are performed in parallel by curl. Extraction are executed in the main thread and are executed one by one. The major reason we see gain in performance is because we keep downloading files while we are extracting a pack

matthewrsj:
> The major reason we see gain in performance is because we keep downloading files while we are extracting a pack

Right, I meant asynchronously to pack downloads, not each other.

aaronovz1:
The benefit might not outweigh the complexity but it does seem that while you would be downloading multiple files at once, the extraction is still done in series and would benefit from parallel processing. Otherwise you potentially have multiple files downloaded sitting there waiting for extraction. Obviously not something for this PR/ticket but figured it was worth mentioning as a future improvement.

otaviobp:
V2:
 - not printing warning for pack downloads.
 - rebased on top of new master and fixing broken tests

matthewrsj:
With this PR test/functional/update/slow-server took about 5 minutes to complete on my system. What's the reason for that? It wasn't super obvious to me looking at the changes you made to the slow-server code.

otaviobp:
I couldn't reproduce this problem in update/slow-server test, but the add-rc test was really taking a really long time because of unnecessary retries.

 I improved the commit message explaining some changes I made in slow-server. There's a slow down because of 2 retries that are necessary to check if download resume is working. This slows down the test in 30s, so it shouldn't take 5 minutes to complete. It's taking less than 50s in my machine. All tests are running in around 6~7 minutes.

Can you retry using this new branch?

matthewrsj:
[offline convo] @otaviobp to skip the long-running test if not in travis environment.

otaviobp:
New version. Spiking slow-server test when not in travis

src/fullfile.c
```diff
@@ -99,6 +99,15 @@ static int download_loop(void *download_handle, struct list *files, int *num_dow
 	return swupd_curl_parallel_download_end(download_handle, num_downloads);
 }
 
+static bool download_error(int response, UNUSED_PARAM void *data)
+{
+	if (response == 404) {
```
matthewrsj: why not just `return (response == 404);` ?

test/functional/update/slow-server/test.bats
```diff
@@ -8,6 +8,11 @@ port=""
 targetfile="6c27df6efcd6fc401ff1bc67c970b83eef115f6473db4fb9d57e5de317eba96e"
 
 setup() {
+  # Skip if not in travis env
```
matthewrsj: Longer comment explaining why we only run the test in Travis (too slow to run for local development) with an explanation of how to run it locally (`TRAVIS=foo make check`) please.

test/functional/update/slow-server/test.bats
```diff
@@ -8,6 +8,11 @@ port=""
...
+  if [ -z "${TRAVIS}" ]; then
+    skip "Skiping test because it's too slow"
```
matthewrsj: misspelled "skipping" and please add more details in the skip message (maybe "Skipping slow test for local development, test only runs in CI")

src/fullfile.c
otaviobp: good point.

otaviobp:
new version: Fixed proposed change and improved docs.

src/download.c
```diff
@@ -156,18 +166,44 @@ void *swupd_curl_parallel_download_start(size_t max_xfer, size_t max_xfer_bottom
...
+/*
+ * Set parallel downloads callbacks.
+ *
+ *  - success_cb(): Called for each sucessfull download, where data is the data
+ *                  informed on swupd_curl_parallel_download_enqueue().
+ *                  success_cb should return true if the download file was
+ *                  handled correctly. Return false to schedule a retry.
+ *  - error_cb():   Called for each failed download, where data is the data
+ *                  informed on swupd_curl_parallel_download_enqueue()
+ *                  and response is the HTTP response code. error_cb should
```
castulo: I think you are missing something here: 

>  error_cb should ?? if the error was handled by caller or false to schedule a retry.

src/packs.c
```diff
@@ -60,56 +131,46 @@ static int download_pack(int oldversion, int newversion, char *module, int is_mi
 			return err;
 		}
 		printf("Linked %s to %s\n", url, filename);
+
+		err = finalize_pack_download(module, newversion, filename);
```
castulo: why aren't you checking if err is different than 0 here?

castulo: never mind, I see you are returning it..

test/functional/update/slow-server/test.bats
```diff
@@ -8,6 +8,12 @@ port=""
 targetfile="6c27df6efcd6fc401ff1bc67c970b83eef115f6473db4fb9d57e5de317eba96e"
 
 setup() {
+  # Skip this test if not running in Travis CI, because test takes too long for
```
castulo: the idea is that the old tests are removed when the new tests (v2) are fully integrated, so I guess it is fins to update this old test here for now, but we should be using the new library for tests instead.

test/functional/update/slow-server/server.py
```diff
@@ -15,21 +16,32 @@ def do_GET(self):
...
+
             self.wfile.write(b)
+            if partial_download:
+                break;
             time.sleep(delay)
         self.wfile.flush()
         f.close()
 
 if __name__ == '__main__':
+    first_time = True
```
evil-brain: Why a constant here and a class-level constant in SlowResponse?
My google-fu is failing me in trying to remember, and now my curiosity is getting to me :smile:

otaviobp: oops. This should be removed

test/functional/bundleadd_v2/add-rc.bats
```diff
@@ -222,19 +224,16 @@ global_teardown() {
...
+		Installing bundle(s) files...
+		.Path /bar/test-file2 is missing on the file system ... fixing
 		Error: Failed to download file /bar/test-file2 in verify_fix_path
 		Failed to install 3 of 3 bundles
 		1 bundle was already installed
 	EOM
 	)
-	assert_regex_is_output --identical "$expected_output"
-
+	assert_in_output "$expected_output"
```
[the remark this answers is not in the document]
otaviobp: true. fixed in new version

src/download.c
otaviobp: Probably I removed that by accident. 
error_cb should return true...

test/functional/update/slow-server/test.bats
otaviobp: I tried to port this test this morning because I thought I had all features I needed, but I think there's still some things to get in before porting it. If you don't mind, I prefer waiting until some updates tests are merged before porting this one. But it's on my radar.

otaviobp:
Fixed problems reported (other than the test using old lib) in new version

matthewrsj:
@evil-brain and @castulo please check that your comments have been addressed.

test/functional/update/slow-server/test.bats
castulo: perfect

PR #525, opened 2018-07-24
verify: When --install is specified, warn on disk full

### LINKED ISSUE #328: Hanging when no disk space is available
When using ister to create a virtual disk image, as swupd is performing an install of content, if the partition where content is being installed runs out of disk space, then swupd hangs indefinitely.  In this case, swupd should error out to allow ister to exit normally and log an error.  Below is a trace of the commands:

+ sudo -E ister.py -t /home/clear/clear-92be12be9967438d810c5eaa6b6778ca/image/ister.json -V https://cdn.download.clearlinux.org/update -C https://cdn.download.clearlinux.org/update -l /home/clear/clear-92be12be9967438d810c5eaa6b6778ca/log/ister.log -s /home/clear/clear-92be12be9967438d810c5eaa6b6778ca/image/Swupd_Root.pem -f 20
Reading configuration
Validating configuration
Creating virtual disk
Creating partitions
Mapping loop device
Creating file systems
Setting up mount points
Starting swupd. May take several minutes
Installing 5 bundles (and dependencies)
+ swupd verify --install --path=/tmp/ister-18720-b_v5m_u4 --manifest=18720 --contenturl=https://cdn.download.clearlinux.org/update --versionurl=https://cdn.download.clearlinux.org/update --format=20 --statedir=/var/lib/swupd -C /home/clear/clear-92be12be9967438d810c5eaa6b6778ca/image/Swupd_Root.pem

To reproduce this issue, the following ister.json file can be used (uploaded as ister.txt):
[ister.txt](https://github.com/clearlinux/swupd-client/files/1433024/ister.txt)

evil-brain:
This change adds some simple space checking on the new 'bundles' to
install when using verify.

This assumes that if you pass --install to verify you intend
to install all bundles to a 'new' folder (no other bundles installed).
This could warn erroneously if there are previously installed bundles or partially
installed bundles, because 'contentsize' could be more than the size
that's actually required to install those files.

Fixes #328

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -25,6 +25,7 @@ swupd_SOURCES = \
 	src/download.c \
 	src/extra_files.c \
 	src/filedesc.c \
+	src/fs.c \
 	src/fullfile.c \
 	src/globals.c \
 	src/hash.c \
--- a/src/verify.c
+++ b/src/verify.c
@@ -412,6 +412,47 @@ static int get_required_files(struct manifest *official_manifest, struct list *s
 	return ret;
 }
 
+/*
+ * Only called when a file has failed to be fixed during a verify or install.
+ * If a low-space warning has been printed, don't check again,
+ * but just warn the user and return.
+*/
+static void check_warn_freespace(const struct file *file)
+{
+	long fs_free;
+	char *original = NULL;
+	static bool no_freespace_flag = false;
+	struct stat st;
+
+	if (!(cmdline_option_install || cmdline_option_fix)) {
+		goto out;
+	}
+
+	/* If true, skip expensive operations for future file failures, decreasing time to completion. */
+	if (no_freespace_flag) {
+		fprintf(stderr, "\tWarning: No space left on device\n");
+		goto out;
+	}
+
+	string_or_die(&original, "%s/staged/%s", state_dir, file->hash);
+	fs_free = get_available_space(path_prefix);
+	if (fs_free < 0 || stat(original, &st) != 0) {
+		fprintf(stderr, "\tWarning: Unable to determine free space on filesystem.\n");
+		goto out;
+	}
+
+	if (fs_free < st.st_size * 1.1) {
+		fprintf(stderr, "\tWarning: File to install (%s) too large by %ldK.\n",
+			file->filename, (st.st_size - fs_free) / 1000);
+		/* set flag to skip checking space on the second failure, assume we're still out of space */
+		no_freespace_flag = true;
+	}
+
+out:
+	fprintf(stderr, "\tContinuing operation...\n");
+	free_string(&original);
+}
+
 /* for each missing but expected file, (re)add the file */
 static void add_missing_files(struct manifest *official_manifest)
 {
@@ -474,6 +515,9 @@ static void add_missing_files(struct manifest *official_manifest)
 		if ((ret != 0) || hash_needs_work(file, local.hash)) {
 			file_not_replaced_count++;
 			printf("\n\tnot fixed\n");
+
+			check_warn_freespace(file);
+
 		} else {
 			file_replaced_count++;
 			file->do_not_update = 1;
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -28,7 +28,6 @@
 #include <stdio.h>
 #include <string.h>
 #include <sys/stat.h>
-#include <sys/statvfs.h>
 #include <sys/types.h>
 #include <unistd.h>
 
@@ -743,7 +742,6 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
 	struct list *iter;
 	struct list *to_install_bundles = NULL;
 	struct list *to_install_files = NULL;
-	struct list *ptr;
 	struct manifest *full_mom = NULL;
 	timelist times;
 
@@ -793,26 +791,13 @@ static int install_bundles(struct list *bundles, struct list **subs, int current
 
 	/* Step 1.5: Now that we know the bundle is valid, check if we have enough space */
 	if (!skip_diskspace_check) {
-		struct statvfs stat;
 		char *filepath = NULL;
 
-		ptr = list_head(to_install_bundles);
-		while (ptr) {
-			struct manifest *subman;
-			subman = ptr->data;
-			ptr = ptr->next;
-
-			bundle_size += subman->contentsize;
-		}
-
+		bundle_size = get_manifest_list_contentsize(to_install_bundles);
 		filepath = mk_full_filename(path_prefix, "/usr/");
 
 		/* Calculate free space on filepath */
-		if (statvfs(filepath, &stat) != 0) {
-			fs_free = -1;
-		} else {
-			fs_free = stat.f_bsize * stat.f_bavail;
-		}
+		fs_free = get_available_space(filepath);
 		free_string(&filepath);
 
 		/* Add 10% to bundle_size as a 'fudge factor' */
--- a/src/fs.c
+++ b/src/fs.c
@@ -1,0 +1,47 @@
+/*
+ *   Software Updater - client side
+ *
+ *      Copyright (c) 2012-2018 Intel Corporation.
+ *
+ *   This program is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License as published by
+ *   the Free Software Foundation, version 2 or later of the License.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "swupd.h"
+#include <sys/statvfs.h>
+
+long get_available_space(const char *path)
+{
+	struct statvfs stat;
+
+	if (statvfs(path, &stat) != 0) {
+		return -1;
+	}
+
+	return stat.f_bsize * stat.f_bavail;
+}
+
+long get_manifest_list_contentsize(struct list *manifests)
+{
+	long total_size = 0;
+
+	struct list *ptr = NULL;
+	for (ptr = list_head(manifests); ptr; ptr = ptr->next) {
+		struct manifest *subman;
+		subman = ptr->data;
+
+		total_size += subman->contentsize;
+	}
+
+	return total_size;
+}
--- a/src/swupd.h
+++ b/src/swupd.h
@@ -395,6 +395,11 @@ typedef enum telem_prio_t {
 } telem_prio_t;
 extern void telemetry(telem_prio_t level, const char *class, const char *fmt, ...);
 
+/* fs.c */
+extern long get_available_space(const char *path);
+/* Calculate the total contentsize of a manifest list */
+extern long get_manifest_list_contentsize(struct list *manifests);
+
 extern struct file **manifest_files_to_array(struct manifest *manifest);
 extern int enforce_compliant_manifest(struct file **a, struct file **b, int searchsize, int size);
 extern void free_manifest_array(struct file **array);


evil-brain:
Also, the return on fail is EBUNDLE_INSTALL - this is "kind of" correct, but I couldn't find a better error code to use. Suggestions are welcome!

src/verify.c
```diff
@@ -805,6 +809,38 @@ int verify_main(int argc, char **argv)
 	}
 	grabtime_stop(&times);
 
+	/* When installing, check disk space availability */
+	/* This assumes there's no pre-installed/partially installed bundles in path_prefix */
+	if (cmdline_option_install) {
+		struct list *ptr = list_head(official_manifest->submanifests);
```
matthewrsj: This block of code is almost identical to the code in `src/bundle.c`. Looks like it could be its own function.

src/verify.c
```diff
@@ -805,6 +809,38 @@ int verify_main(int argc, char **argv)
...
+		/* Add 10% to total_size as a 'fudge factor' */
+		if (((total_size * 1.1) > fs_free) || fs_free < 0) {
+			ret = EBUNDLE_INSTALL;
+
+			if (fs_free >= 0) {
+				fprintf(stderr, "Error: Files to install too large by %ldM.\n", (total_size - fs_free) / 1000 / 1000);
+			} else if (fs_free < 0) {
+				fprintf(stderr, "Error: Unable to determine free space on filesystem.\n");
+			}
+			goto clean_and_exit;
```
matthewrsj: Didn't we discuss not failing for the install case? We don't want our installer frontends to fail out with no recourse for the user, especially when this is potentially a liberal check in case os-core is already installed on the target for some reason.

evil-brain: I should include some way to skip the check in case the user wants to try anyways.

In issue #328, the requested behavior was to print an error and exit.

I think the next step is to change do_staging to report out of disk errors - but without a pre-check, we will have to keep going (because we're after the "point of no return"), and print an error/run the failure path for each file in the manifests, which takes a lot of time.

As an example, using verify --install on a full filesystem to verify 8301 missing files takes swupd 93 seconds to fail, while printing ~16000 lines of errors. I think this may be why the bug submitter stated swupd hung, because I couldn't reproduce the hang, just a long processing time.

I'm not sure what to do here... I think we need to speed up the failure case, since we can't exit early. Suggestions?

Makefile.am
```diff
@@ -25,6 +25,7 @@ swupd_SOURCES = \
 	src/download.c \
 	src/extra_files.c \
 	src/filedesc.c \
+	src/fs_storage.c \
```
otaviobp: Suggestion: Rename this to fs.c so we can move later all fs related functions (mkdir_p, is_populated_dir, copy_all, etc) from helper.c to fs.c.

src/verify.c
```diff
@@ -474,6 +478,30 @@ static void add_missing_files(struct manifest *official_manifest)
...
+						}
+						fprintf(stderr, "\tContinuing with install...\n");
+					}
+				} else {
+					fprintf(stderr, "\tWarning: Unable to determine file size for staged file %s", original);
+				}
+				free_string(&original);
+			}
+			if (no_freespace_flag) {
+				fprintf(stderr, "\tWarning: No space left on device\n");
```
otaviobp: I I understood the code correctly, this message will be printed only once. You can move this print go line 489 and get rid of the no_freespace_flag value.

src/verify.c
```diff
@@ -805,6 +834,26 @@ int verify_main(int argc, char **argv)
...
+	/* When installing, check disk space availability */
+	/* Print a warning and set a flag if we lack free space */
+	if (cmdline_option_install) {
+		long fs_free = get_available_space(path_prefix);
+		long total_size = get_manifest_list_contentsize(official_manifest->submanifests);
+
+		/* Add 10% to total_size as a 'fudge factor' */
+		if (((total_size * 1.1) > fs_free) || fs_free < 0) {
+			if (fs_free >= 0) {
+				fprintf(stderr, "Warning: Files to install larger than free space by %ldM.\n", (total_size - fs_free) / 1000 / 1000);
```
otaviobp: As we aren't going to abort the installation process when we notice that we may have free space issues, can't we report the problem only on add_missing_files() if a file copy fails?

src/verify.c
```diff
@@ -474,6 +478,30 @@ static void add_missing_files(struct manifest *official_manifest)
 		if ((ret != 0) || hash_needs_work(file, local.hash)) {
 			file_not_replaced_count++;
 			printf("\n\tnot fixed\n");
+
+			if (freespace_warning_flag && cmdline_option_install) {
+				fs_free = get_available_space(path_prefix);
+				string_or_die(&original, "%s/staged/%s", state_dir, file->hash);
+				if (stat(original, &st) == 0) {
+					if (((st.st_size * 1.1) > fs_free) || fs_free < 0) {
```
otaviobp: I'd expect fs_free to be zero (or some number close to zero) when there's no disk space left in this point.

src/verify.c
```diff
@@ -474,6 +478,30 @@ static void add_missing_files(struct manifest *official_manifest)
 		if ((ret != 0) || hash_needs_work(file, local.hash)) {
```
otaviobp: I tried to trigger this error on a tmpfs filesystem and I couldn't do it. All files were created empty in the filesystem after the disk got full, so hash_needs_work returns false, even for files that weren't installed successfully.
Could you get into this if using swupd verify --install?

otaviobp: Nice. I'll test that. Maybe the behavior when ramfs is used is different.

otaviobp: It's working fine with file mounted partitions, so I think we're good about that.

matthewrsj:
/me notices you updated this PR.

I may have missed the notification, but just so you know if you force push at all it doesn't notify the watchers of the PR that it was updated. Please comment on your PRs when they need a fresh review.

src/verify.c
evil-brain: Agreed, I can't remember why I had the warning flag - I suppose I was concerned about checking space on other types of failures, but there's no real harm checking as failures here should be fairly rare.

Fixed.

src/verify.c
evil-brain: This message is printed once per file that fails (after the first file fails due to space constraints). The flag is there so we don't run a space check on each failed file - decreasing the time it takes to complete the command.

src/verify.c
evil-brain: I fixed the extra unnecessary checks I was performing here. fs_free could be -1 if we we're unable to determine the free space on disk, but I think there's no need for the OR here.

evil-brain:
Updated the PR, changed the commit message and simplified the flow - pretty concise now and ready for more reviews!

src/verify.c
```diff
@@ -474,6 +478,29 @@ static void add_missing_files(struct manifest *official_manifest)
...
 			file_not_replaced_count++;
 			printf("\n\tnot fixed\n");
+
+			if (cmdline_option_install && no_freespace_flag == false) {
+				fs_free = get_available_space(path_prefix);
+				string_or_die(&original, "%s/staged/%s", state_dir, file->hash);
```
matthewrsj: This is an interesting thing to check against, since it *could* be that someone is installing the OS to the same partition on which this file lives (installing to a directory in the same rootfs for instance). In this case it could be possible that the state directory is on the same file system and if the user got to this point it *could* still pass even if there is no more room on the device because we try to hardlink everything if we can.

This is fine though because these are just warnings, the write *did* fail, and the disk is getting full anyways.

src/verify.c
```diff
@@ -474,6 +478,29 @@ static void add_missing_files(struct manifest *official_manifest)
 		if ((ret != 0) || hash_needs_work(file, local.hash)) {
 			file_not_replaced_count++;
 			printf("\n\tnot fixed\n");
+
+			if (cmdline_option_install && no_freespace_flag == false) {
+				fs_free = get_available_space(path_prefix);
```
matthewrsj: Can you please refactor lines 483-503 into its own function please? It's getting hard to read with the deep indentation at some parts.

matthewrsj: maybe 483-499, whatever makes sense.

src/verify.c
```diff
@@ -474,6 +478,29 @@ static void add_missing_files(struct manifest *official_manifest)
...
+				string_or_die(&original, "%s/staged/%s", state_dir, file->hash);
+				if (stat(original, &st) == 0) {
```
matthewrsj:
```C
if (stat(original, &st) == 0 && st.st_size * 1.1 > fs_free) {
}
```
lets you drop a directly nested if.

src/verify.c
```diff
@@ -474,6 +478,29 @@ static void add_missing_files(struct manifest *official_manifest)
...
+					if ((st.st_size * 1.1) > fs_free) {
+						if (fs_free >= 0) {
+							fprintf(stderr, "\tWarning: File to install (%s) too large by %ldK.\n", file->filename, (st.st_size - fs_free) / 1000);
+							/* set flag to skip checking space on the second failure, assume we're still out of space */
+							no_freespace_flag = true;
+						} else if (fs_free < 0) {
```
matthewrsj: You can drop the `(fs_free < 0)` conditional here and just make it an `else` since you already check `>= 0` in the if clause.

src/verify.c
```diff
@@ -474,6 +478,29 @@ static void add_missing_files(struct manifest *official_manifest)
...
+							fprintf(stderr, "\tWarning: Unable to determine free space on filesystem.\n");
+						}
+						fprintf(stderr, "\tContinuing with install...\n");
```
matthewrsj: Move this after the below `else` so it is consistently printed after each `Unable to determine free space...` warning.

matthewrsj:
You could factor out that code block into its own function and do some real improvements to code readability. Just hacking at it real quick (no testing!) I came up with this refactor
```C
static void thing(arguments) {
	char *original;
	static bool no_freespace_flag = false;
	string_or_die(&original, "%s/staged/%s", state_dir, file->hash);
	fs_free = get_available_space(path_prefix);
	if (fs_free < 0 || stat(original, &st) != 0) {
		fprintf(stderr, "\tWarning: Unable to determine file size for staged file %s", original);
		goto out
	}

	fprintf(stderr, "\tWarning: File to install (%s) too large by %ldK.\n",
			file->filename, (st.st_size - fs_free) / 1000);
	/* set flag to skip checking space on the second failure, assume we're still out of space */
	no_freespace_flag = true;

out:
	fprintf(stderr, "\tContinuing with install...\n");
	free_string(&original);
}
```

I'm sure this could be optimized even more.

evil-brain:
V3 is up, with the suggested refactor above. Tested and ready for more review.

Thanks to all for reviewing! This is a great way to learn.

src/verify.c
```diff
@@ -412,6 +412,35 @@ static int get_required_files(struct manifest *official_manifest, struct list *s
...
+			goto out;
+		}
+		if (fs_free < st.st_size * 1.1) {
+			fprintf(stderr, "\tWarning: File to install (%s) too large by %ldK.\n",
+				file->filename, (st.st_size - fs_free) / 1000);
+			/* set flag to skip checking space on the second failure, assume we're still out of space */
+			no_freespace_flag = true;
+		}
+	} else {
+		fprintf(stderr, "\tWarning: No space left on device\n");
```
matthewrsj: So if `cmdline_option_install` is false this warning will be printed regardless of the `no_freespace_flag`?

evil-brain: True - if you run 'verify --fix', and a file fails the final hash verification, this function will print a warning for no space; even if there is enough free space.

I've verified this happens and will fix. Thanks!

evil-brain: What do you think about something like this:
`if (cmdline_option_install || cmdline_option_fix && no_freespace_flag == false)`
and then change the final printf to something like:
`fprintf(stderr, "\tContinuing with install or fix...\n");`

I think those are the only two paths that call add_missing_files().

It's also possible that stat() could fail - and the reported message isn't quite accurate (in that we could determine free space, but not how large the file we want to copy is). Should I change the output or add another if() to check stat separately?

matthewrsj: well `if (cmdline_option_install || cmdline_option_fix && no_freespace_flag == false)` will always pass if `cmdline_option_install` is true regardless of `no_freespace_flag`.

Why don't you convert this logic to a guard clause instead? Something like:
```C
static void thing(const struct file *file)
{
    if (!(cmdline_option_install || cmdline_option_fix)) {
        goto out;
    }

    if (no_freespace_flag) {
        fprintf(stderr, "Warning: ...");
        goto out;
    }

    /* the rest of your code */

out: 
    fprintf(stderr, "\tContinuing operation...\n");
    free_string(&original);
}

Comment (evil-brain) [reply src/verify.c]: This is a great idea for this case - the thought had honestly not crossed my mind... I think this is part of "knocking the rust off" :wink:

File: src/verify.c [hunk 14]
```diff
@@ -412,6 +412,41 @@ static int get_required_files(struct manifest *official_manifest, struct list *s
 	return ret;
 }
 
+static void check_warn_freespace(const struct file *file)
```
matthewrsj: This function looks much better now, thank you.

Now that it is shaping up I think it needs a few comments to describe what's going on here and what assumptions are being made by the time this function is being called. 

First, this function needs a comment that explains that it is only called when a file has failed to be fixed during a verify or install. This is the only time it makes sense to call this function because it prints warning/error messages.

Second, you have the `if (no_freespace_flag)` guard so that you don't perform the freespace check for every failure. This is awesome, but difficult to understand from reading the code if you don't make the assumption above. A comment explaining why this guard exists would be helpful.

evil-brain:
V5 is up! Added comments that hopefully make it more obvious what the new code does.

PR #533, opened 2018-07-25
Adding new assertions to the test library

castulo:
This pull request include two parts.

1) It adds 6 new assertions to the test library:
- assert_is_output
- assert_is_not_output
- assert_regex_is_output
- assert_regex_is_not_output
- assert_regex_in_output
- assert_regex_not_in_output

2) It migrates the tests from the following categories to use the
new library:
- completion
- hashdump
- mirror
- search

Finally this commit changes a few test cases so they use
"assert_is_output" instead of "assert_in_output", that way
we can have a tighter control of the output we want since
assert_is_output requires the whole output to match to a
specific text instead of doing a partial match.

### DIFF
--- a/test/functional/README.md
+++ b/test/functional/README.md
@@ -302,44 +302,97 @@ run <some_command>
 assert_file_not_exists /some/file
 ```
 
-*assert_in_output*  
-passes if the provided text is included in the command output, fails otherwise  
+*assert_is_output*  
+passes if the provided text matches the whole command output, fails otherwise  
 Examples:  
 ```bash
 # one line strings
 run <some_command>
-assert_in_output "Successfully installed 1 bundle"
+assert_is_output "Some output"
 
 # multi-line strings
+run <some_command>
+expected_output=$(cat <<-EOM
+  Some text that needs
+  to match exactly with
+  the whole command output
+EOM
+)
+assert_is_output "$expected_output"
+```
+
+*assert_is_not_output*  
+passes if the provided text does not match the whole command output, fails otherwise  
+Examples:  
+```bash
+run <some_command>
+assert_is_not_output "This should not be the command output"
+```
+
+*assert_in_output*  
+passes if the provided text is included in the command output (partial match), fails otherwise  
+Examples:  
+```bash
 run <other_command>
 expected_output=$(cat <<-EOM
-  Some multi-ine text that needs to be present
-  in the exact order
-  some more lines, bla bla
+  Some multi-line output
+  some more lines,
+  bla bla
 EOM
 )
 assert_in_output "$expected_output"
-
-# combination of both
-run <yet_another_command>
-expected_output=$(cat <<-EOM
-  Some multi-ine text that needs to be present
-  in the exact order
-  some more lines, bla bla
-EOM
-)
-assert_in_output "some literall text"
-assert_in_output "$expected_output"
-assert_in_output "more text to be checked"
 ```
 
 *assert_not_in_output*  
-passes if the provided text is not included in the command output, fails otherwise  
+passes if the provided text is not included in the command output (partial match), fails otherwise  
 Examples:  
 ```bash
-# one line strings
 run <some_command>
 assert_not_in_output "Error"
+```
+
+*assert_regex_is_output*  
+similar to assert_is_output but this assertion receives a regular expression, passes
+if the provided regular expression matches the whole command output, fails otherwise  
+Examples:  
+```bash
+# remember to skip characters that are part of the expected
+# output that match special regex characters like .*?()[] 
+run <some_command>
+expected_output=$(cat <<-EOM
+  Some expected text that can have
+  regex characters like .* in it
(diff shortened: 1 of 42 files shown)


test/functional/testlib.bash
```diff
@@ -1132,6 +1136,96 @@ assert_not_in_output() {
 
 }
 
+assert_is_output() {
+
+	local expected_output=$1
+	local sep="------------------------------------------------------------------"
+	validate_param expected_output
+
+	if [[ ! "$output" == "$expected_output" ]]; then
```
otaviobp: in swupdlib.bash we're filtering some lines from the output. I think this is a good feature to add since there're some lines that aren't very useful to check if the test is successful, like empty lines. Or lines with dots.

castulo: Yeah, I noticed that from swupdlib.bash, but I'm not sure if I agree with that approach. For example, let's say you are ignoring the dots in your output, that means your output could get full of dots and your tests would not detect that, and it would be very annoying for the end user. I think tests should be as explicit as possible.

matthewrsj: @castulo while I agree with the statement that `tests should be as explicit as possible` these tests are for the most part not designed to verify prints to stdout. The goal of the output checks is to ensure the correct steps are happening in the correct order. It is useful to ignore blank and progress lines so we don't have to change every test when we change that non-functional output.

Obviously this feature request is not a blocker for this PR.

matthewrsj:
@evil-brain can you take a look at this please?

test/functional/hashdump_v2/hashdump-file-hash.bats
```diff
@@ -0,0 +1,55 @@
...
+
+global_teardown() {
+
+	destroy_test_environment "$TEST_NAME"
+
+}
+
+@test "hashdump with prefix" {
+
+	run sudo sh -c "$SWUPD hashdump --path=$TEST_NAME/target-dir /test-hash"
```
evil-brain: Extra space in "--path=$TEST_NAME/target-dir /test-hash"" ?
-----------------------------------------------------------^

castulo: you mean between "--path=$TEST_NAME/target-dir" and "/test-hash", that is the way it is supposed to be, $TEST_NAME/target-dir are part of the --path argument, /test-hash is a different argument

test/functional/testlib.bash
castulo: @matthewrsj @otaviobp I added what you guys requested in the latest commit in this PR, whenever you have a chance please take a look at it.

castulo:
I found a bug with the use_ignore_list() function, I'll push a fix as soon as I have it

test/functional/README.md
```diff
@@ -302,46 +302,99 @@ run <some_command>
...
+if the provided regular expression matches the whole command output, fails otherwise  
+Examples:  
+```bash
+# remember to skip characters that are part of the expected
+# output that match special regex characters like .*?()[] 
+run <some_command>
+expected_output=$(cat <<-EOM
+  Some expected text that can have
+  regex characters like .* in it
+  \(skipping these pharenteses\)\.
```
phmccarty: minor issue: "parentheses" is misspelled

test/functional/README.md
```diff
@@ -357,3 +425,26 @@ Example:
...
 ```
+
+### Ignore lists
+When using assertions that compare the command output to an expected output (for example
+assert_is_output, assert_in_output, etc.), blank lines and lines containing only
+dots are removed by default before comparing outputs.  
+Besides removing empty lines and lines with dots from the output, it is also possible to
+use an *ignore list* to remove other patterns that are not important for a test when using these assertions.  
+
+There are three different ignore list files that can be used (but only one can be used at a time):
```
matthewrsj: Very nice

test/functional/README.md
```diff
@@ -302,46 +302,114 @@ run <some_command>
...
+if the provided regular expression matches the whole command output, fails otherwise  
+Examples:  
+```bash
+# remember to skip characters that are part of the expected
+# output that match special regex characters like .*?()[] 
+run <some_command>
+expected_output=$(cat <<-EOM
+  Some expected text that can have
+  regex characters like .* in it
+  \(skipping these pharentheses\)\.
```
matthewrsj: Oh no! still typoed! (p**h**arentheses)

matthewrsj:
Please comment here when you've fixed that since github doesn't notify for force pushes.

castulo:
I didn't know that @matthewrsj, I fixed the typo pointed out by @phmccarty

matthewrsj:
Still looks misspelled to me. (you have an extra 'h' after the 'P')

castulo:
Damn, I think I got it right this time.

matthewrsj:
@phmccarty how's it look to you? (requiring 2 official approvals now)

PR #542, opened 2018-08-15
test: remove manpage test

matthewrsj:
This test is unreliable and fails seemingly randomly in our CI.
Restarting our tests always makes the test pass.

Unless someone has an idea why this randomly fails I'd like to just remove it.


otaviobp:
@castulo any idea about this one?

matthewrsj:
didn't mean to close...

PR #543, opened 2018-08-15
Testlib usability improvements

castulo:
This pull request add some usability improvements to the test library.

- Print the function stack when the script fails due to missing or bad parameters.
- Print information about function usage when executing the function without any parameter in the command line.
- Allow users to skip the deletion of a test environment temporarily so they can debug what is wrong with its test, and or to verify the environment is correct as they built it.
- Adds two more assertions for validating file equality in tests.

Besides the usability improvements the PR includes a fix for some variables that were being used incorrectly (without the $ character).

### DIFF
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -1510,3 +1510,30 @@ assert_not_equal() {
 	fi
 
 }
+
+assert_files_equal() {
+
+	local val1=$1
+	local val2=$2
+	validate_item "$val1"
+	validate_item "$val2"
+
+	diff -q "$val1" "$val2"
+
+}
+
+assert_files_not_equal() {
+
+	local val1=$1
+	local val2=$2
+	validate_item "$val1"
+	validate_item "$val2"
+
+	if diff -q "$val1" "$val2" > /dev/null; then
+		echo "Files $val1 and $val2 are equal"
+		return 1
+	else
+		return 0
+	fi
+
+}
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -1238,7 +1238,7 @@ use_ignore_list() {
 assert_status_is() {
 
 	local expected_status=$1
-	validate_param expected_status
+	validate_param "$expected_status"
 
 	if [ -z "$status" ]; then
 		echo "The \$status environment variable is empty."
@@ -1263,7 +1263,7 @@ assert_status_is() {
 assert_status_is_not() {
 
 	local not_expected_status=$1
-	validate_param not_expected_status
+	validate_param "$not_expected_status"
 
 	if [ -z "$status" ]; then
 		echo "The \$status environment variable is empty."
@@ -1288,7 +1288,7 @@ assert_status_is_not() {
 assert_dir_exists() {
 
 	local vdir=$1
-	validate_param vdir
+	validate_param "$vdir"
 
 	if [ ! -d "$vdir" ]; then
 		print_assert_failure "Directory $vdir should exist, but it does not"
@@ -1300,7 +1300,7 @@ assert_dir_exists() {
 assert_dir_not_exists() {
 
 	local vdir=$1
-	validate_param vdir
+	validate_param "$vdir"
 
 	if [ -d "$vdir" ]; then
 		print_assert_failure "Directory $vdir should not exist, but it does"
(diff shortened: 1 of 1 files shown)


test/functional/testlib.bash
```diff
@@ -123,11 +123,22 @@ validate_param() {
...
-        shift
-        printf "$@" | sudo tee $arg "$file" >/dev/null
+	# If no parameters are received show usage
+	if [ $# -eq 0 ]; then
+		cat <<-EOM
+			Usage:
+			    write_to_protected_file [-a] <file> <stream>
+
+			Options:
+			    -a    If used the text will be appeneded to the file, otherwise it will be overwritten
```
otaviobp: typo: appended

test/functional/testlib.bash
```diff
@@ -1175,7 +1175,9 @@ test_setup() {
 # Default test_teardown
 test_teardown() {
 
-	destroy_test_environment "$TEST_NAME"
+	if [ "$DEBUG_TEST" != true ]; then
```
otaviobp: ++ for this commit. That's really useful.

test/functional/testlib.bash
castulo: yeah, maybe I missed something when selecting what should be in this PR, I'll check why tests are failing, thanks for reviewing

castulo:
@otaviobp, thanks for reviewing.  I found the bug I entered with this PR, it is fixed now. Let's see if Travis is happy now.

matthewrsj:
@castulo what was the issue?

castulo:
When I was creating the PR I changed this line of code within the write_to_protected_file function:
`printf "$@" | sudo tee $arg "$file" >/dev/null`
to look like this:
`printf "%s" "$@" | sudo tee $arg "$file" >/dev/null`

I did that to follow recommendation [SC2059](https://github.com/koalaman/shellcheck/wiki/SC2059), but turned out in this case it is an exception of the rule because I actually want to interpret data as a format string.

And unfortunately I missed trying to run tests locally before submitting the PR, so I didn't notice this was messing up with the manifest creation.

test/functional/testlib.bash
```diff
@@ -1510,3 +1510,30 @@ assert_not_equal() {
...
+
+assert_files_not_equal() {
+
+	local val1=$1
+	local val2=$2
+	validate_item "$val1"
+	validate_item "$val2"
+
+	if diff -q "$val1" "$val2" > /dev/null; then
+		echo "Files $val1 and $val2 are equal"
```
matthewrsj: should this library function actually print anything or should the `return 1` be enough?

test/functional/testlib.bash
matthewrsj: This should have a corresponding update to the test readme.

test/functional/testlib.bash
castulo: assert_files_equal() uses `diff -q` which will return a text like this'd a return code of 1 when files are not equal:
"Files tmp1 and tmp2 differ"

So by printing "Files $val1 and $val2 are equal" in function assert_files_not_equal() I'm attempting to being consistent in both functions.

Definitevely returning 1 is the most important thing, but I don't see why some extra info would hurt.

test/functional/testlib.bash
castulo: that is true @matthewrsj, good catch. I'll add that to the commit.

castulo:
@matthewrsj I added info about the DEBUG_TEST env variable in the README.md as requested.

test/functional/README.md
```diff
@@ -165,6 +165,23 @@ To include tests to be run in the CI system:
...
 <br/>
 
+### Debugging Tests
+
+As mentioned in the [test principles](#test-principles), automated tests should
+create their own test environment including all needed dependencies when the test
+starts, and should clean up after themselves once the test is over. This means
+the test environment will be deleted as soon as the test finishes running.
+Sometimes this is not the desired behavior, for example when writing a new test
+and is not working as expected, in this case it would be very difficult to figure
```
matthewrsj: This sentence needs re-wording. You don't specify what you are referring to when you say `for example when writing a new test and is not working as expected`. You have to specify what is not working as expected in that sentence: `for example when writing a new test and the test is not working as expected`. You can also drop the `in this case` if you leave it part of that same sentence, since the case was already described. Something like

> Sometimes this is not the desired behavior, for example when writing a new test and the test is not
> working as expected it would be very difficult to debug the problem if the test environment is wiped.

But if I wrote it I would say something like this

> Sometimes it is necessary to debug faulty tests, and therefore a persistent test environment is required.

castulo: I like it, I'll make the change

castulo:
@matthewrsj I updated the "Debugging tests" section of the README based on your suggestion.

PR #544, opened 2018-08-16
Bug fixes and improvements for the test library

castulo:
This pull request includes several bug fixes in the test library and also makes a few improvements in it.

Improvements:

- Export some more variables that are useful for some tests.
- Improves on how content size is read from a manifest.
- Allows adding dependencies to newer versions not only version 10 (default one).
- Decrease the file count and content size when removing something from a manifest.
- Updates the test template generated by the test generator with a more common output.
- Adds usage info in a couple of functions that were missing it.
- Adds the capability of creating empty test environment (with no os-core bundle).
- Removes an unnecessary directory in the creation of "default" bundles.
- Adds the capability of using existing files in the test bundles.

### DIFF
--- a/test/functional/bundleremove_v2/remove-boot-file.bats
+++ b/test/functional/bundleremove_v2/remove-boot-file.bats
@@ -13,10 +13,10 @@ test_setup() {
 
 	run sudo sh -c "$SWUPD bundle-remove $SWUPD_OPTS test-bundle"
 	assert_status_is 0
-	assert_file_not_exists "$TEST_NAME"/target-dir/usr/lib/kernel/testfile
+	assert_file_not_exists "$TARGETDIR"/usr/lib/kernel/testfile
 	expected_output=$(cat <<-EOM
 		Deleting bundle files...
-		Total deleted files: 2
+		Total deleted files: 1
 		Successfully removed 1 bundle
 	EOM
 	)
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -808,6 +808,8 @@ create_bundle() {
 		    - for every symlink created a related file will be created and added to the bundle as well (except for dangling links)
 		    - if the '-f' or '-l' options are used, and the directories where the files live don't exist,
 		      they will be automatically created and added to the bundle for each file
+		    - if instead of creating a new file you want to reuse an existing one, you can do this by adding ':' followed by the
+		      path to the file, for example '-f /usr/bin/test-1:my_environment/web-dir/10/files/\$file_hash'
 
 		Example of usage:
 
@@ -924,7 +926,14 @@ create_bundle() {
 	# 3) Create the requested file(s)
 	for val in "${file_list[@]}"; do
 		add_dirs
-		bundle_file=$(create_file "$files_path")
+		# if the user wants to use an existing file, use it, else create a new one
+		if [[ "$val" = *":"* ]]; then
+			bundle_file="${val#*:}"
+			val="${val%:*}"
+			validate_item "$bundle_file"
+		else
+			bundle_file=$(create_file "$files_path")
+		fi
 		if [ "$DEBUG" == true ]; then
 			echo "file -> $bundle_file"
 		fi
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -692,13 +692,28 @@ set_latest_version() {
 # Creates a test environment with the basic directory structure needed to
 # validate the swupd client
 # Parameters:
+# - -e: if this option is set the test environment is created empty (withouth bundle os-core)
 # - ENVIRONMENT_NAME: the name of the test environment, this should be typically the test name
 # - VERSION: the version to use for the test environment, if not specified the default is 10
 create_test_environment() { 
 
+	local empty=false
+	[ "$1" = "-e" ] && { empty=true ; shift ; }
 	local env_name=$1 
 	local version=${2:-10}
 	local mom
+	# If no parameters are received show usage
+	if [ $# -eq 0 ]; then
+		cat <<-EOM
+			Usage:
+			    create_test_environment [-e] <environment_name> [initial_version]
+
+			Options:
+			    -e    If set, the test environment is created empty, otherwise it will have
+			          bundle os-core in the web-dir and installed by default.
+			EOM
+		return
+	fi
 	validate_param "$env_name"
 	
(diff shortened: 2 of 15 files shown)


test/functional/testlib.bash
```diff
@@ -808,6 +808,8 @@ create_bundle() {
 		    - for every symlink created a related file will be created and added to the bundle as well (except for dangling links)
 		    - if the '-f' or '-l' options are used, and the directories where the files live don't exist,
 		      they will be automatically created and added to the bundle for each file
+		    - if instead of craeting a new file you want to reuse an existing one, you can do this by adding ':' followed by the
```
jwakre: craeting -> creating

castulo: thanks, will fix that

castulo: typo fixed @jwakre

test/functional/testlib.bash
```diff
@@ -1078,7 +1078,7 @@ remove_bundle() {
 	# may be used by another bundle)
 	dir_names=($(awk '/^D...\t/ { print $4 }' "$bundle_manifest"))
 	for dname in ${dir_names[@]}; do
-		sudo rmdir "$target_path$dname" 2> /dev/null
+		sudo rmdir --ignore-fail-on-non-empty "$target_path$dname" 2> /dev/null
```
otaviobp: I just leaned a new flag. haha

test/functional/testlib.bash
```diff
@@ -883,11 +883,6 @@ create_bundle() {
 	# set default values
 	bundle_name=${bundle_name:-$(generate_random_name test-bundle-)}
 	version=${version:-10}
-	if [ -z "$dir_list" ] && [ -z "$file_list" ] && [ -z "$link_list" ] && [ -z "$dangling_link_list" ] ; then
```
otaviobp: In the commit message: s/unecessary/unnecessary/

castulo: @otaviobp I fixed the typo in the commit message

test/functional/testlib.bash
```diff
@@ -1172,8 +1172,13 @@ generate_test() {
...
 		printf '@test "<test description>" {\n\n'
-		printf '\trun sudo sh -c "$SWUPD <swupd_command> $SWUPD_OPTS <command_options>"\n'
-		printf '\t# <validations>\n\n'
+		printf '\trun sudo sh -c "$SWUPD <swupd_command> $SWUPD_OPTS <command_options>"\n\n'
+		printf '\t# assert_status_is 0\n'
+		printf '\t# expected_output=$(cat <<-EOM\n'
+		printf '\t# \t<expected output>\n'
+		printf '\t# EOM\n'
+		printf '\t# )\n'
+		printf '\t# assert_is_output "$expected_output"\n\n'
```
matthewrsj: good addition.

test/functional/testlib.bash
```diff
@@ -802,7 +802,31 @@ create_bundle() {
...
 	}
-	
+
+	add_dirs_recursively() {
+
+		if [[ "$val" != "/"* ]]; then
+			val=/"$val"
+		fi
+		# if the directories the file is don't exist, add them to the bundle
+		# there are a few exceptions, the dir of the tracking file /usr/share/clear/bundles,
```
matthewrsj: I wouldn't call this an exception, since `/usr/share/clear/` is simply subtracted out by mixer due to those dirs being in `os-core`. Hence it looks like an exception because every bundle will have `/usr/share/clear/bundles/bundle` but not any of `/usr/share/clear/bundles/`. I'm not sure if you want to hardcode in the "exception" here.

test/functional/testlib.bash
```diff
@@ -802,7 +802,31 @@ create_bundle() {
 
 		EOM
 	}
-	
+
+	add_dirs_recursively() {
```
matthewrsj: not a big deal, but this doesn't actually do any recursion, it's all iterative.

test/functional/testlib.bash
castulo: you are right, my choice of words may not be the best here. What I meant is that if we don't consider that file as a "separate case" (not exception), then every test bundle we create will have at least 5 files/dirs (/usr, /usr/share, /usr/share/clear, usr/share/clear/bundles, and  /usr/share/clear/bundles/<bundle_name>), which may be a little too much for most tests. If that file is considered separately from the others, then it is possible to create a bundle with just one file (/usr/share/clear/bundles/<bundle_name>).

castulo: I will re-word the phrase

test/functional/testlib.bash
matthewrsj: but this doesn't actually remove the directory? do we not care about that?

test/functional/bundleremove_v2/remove-boot-file.bats
```diff
@@ -13,10 +13,10 @@ test_setup() {
 
 	run sudo sh -c "$SWUPD bundle-remove $SWUPD_OPTS test-bundle"
 	assert_status_is 0
-	assert_file_not_exists "$TEST_NAME"/target-dir/usr/lib/kernel/testfile
+	assert_file_not_exists "$TARGETDIR"/usr/lib/kernel/testfile
 	expected_output=$(cat <<-EOM
 		Deleting bundle files...
-		Total deleted files: 2
```
matthewrsj: wait, was this passing before as well?

test/functional/testlib.bash
castulo: that's true, this function doesn't call itself, I'll rename it

test/functional/testlib.bash
matthewrsj: +1 on keeping the same functionality, but rewording the comment.

test/functional/testlib.bash
castulo: Actually I think that is expected because it means the dir was not empty, so it may have at least one file from another test bundle that share that same directory, we don't want to delete files from another bundle.

The main use case I can think of this function is for doing partial clean ups when running many tests with the same test script, and all sharing the same test environment (using global_setup).

The whole test environment will be cleaned up after the tests in the script have finished, so there is no risk on leaving left overs.

castulo:
I made the requested changes @matthewrsj

castulo:
@evil-brain which one of the its is incorrect in that phrase? both?

PR #545, opened 2018-08-17
Re-enable manpage tests

phmccarty:
The manpage tests were recently removed due to occasional test failures that were raised. I spent some time debugging the issue and root caused the issue.

So, I've reverted the commit that the removed the tests and fixed the problem in the second commit.


otaviobp:
Great catch!!! It looks like git diff-index doesn't check for mtime in newer versions, but it is checking for that in the version we are using in travis. That's why it was harder to reproduce it in my computer.

PR #546, opened 2018-08-17
Use SSL Client Cert when provided

jwakre:
Re-opening the client certificate pull request based on @gtkramer's implementation.

In addition to enabling client certificate authentication, this pull request includes client certificate tests and new test objects to help write them.


matthewrsj:
You may need to `pip install` some things in `.travis.yml` to get the travis tests fixed.

jwakre:
I think it's using Python2 instead of Python3. I'll be more explicit.

src/curl.c
```diff
@@ -62,6 +62,33 @@ static void swupd_curl_strerror(CURLcode curl_ret)
...
+ * API, a curl_easy_setopt does not detect if the client SSL certificate is
+ * present on the filesystem.  The only happens during curl_easy_perform.
+ * The emphasis is rather on how using a SSL client certificate is an opt-in
+ * function rather than an opt-out function.
+ */
+CURLcode swupd_curl_set_optional_client_cert(CURL* curl) {
+	CURLcode curl_ret = CURLE_OK;
+	char* client_cert_path;
+
+	client_cert_path = mk_full_filename(path_prefix, SSL_CLIENT_CERT);
```
otaviobp: You need to free client_cert_path after using it.

test/functional/update_v2/update-client-certificate.bats
```diff
@@ -0,0 +1,103 @@
...
+server_pub="$TEST_DIR_PATH/server-pub.pem"
+server_key="$TEST_DIR_PATH/server-key.pem"
+client_pub="$TEST_DIR_PATH/client-pub.pem"
+client_key="$TEST_DIR_PATH/client-key.pem"
+
+global_setup() {
+
+	create_test_environment "$TEST_NAME"
+	create_test_environment "$TEST_NAME" 100
+
```
otaviobp: This a very minor thing, but avoid leaving trailing spaces and tabs in code. This is a bit annoying in some editors or when using git diff with colors.

test/functional/testlib.bash
```diff
@@ -775,6 +787,118 @@ destroy_test_environment() {
...
+				return ;;
+		esac
+	done
+	set +f  # turn globbing back on
+	
+	# get available port
+	for i in $(seq 8081 8181); do
+
+		# start server
+		python $FUNC_DIR/server.py $server_args --port $i &
```
jwakre: Travis is using Python2 here instead of Python3

test/functional/testlib.bash
```diff
@@ -775,6 +787,118 @@ destroy_test_environment() {
...
+		server_pid=$!
+
+		sleep .5
```
jwakre: When running a lot of tests in parallel, a failure can occur here. If the server.py script uses an invalid port and does not fail before 0.5 seconds, it will continue with an invalid port. I am refactoring the port selection into the python script to solve this.

src/curl.c
jwakre: yes, I missed this.

test/functional/update_v2/update-client-certificate.bats
jwakre: I'll clean this up.

test/functional/testlib.bash
```diff
@@ -4,10 +4,18 @@ FUNC_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 TEST_FILENAME=$(basename "$BATS_TEST_FILENAME")
 TEST_NAME=${TEST_FILENAME%.bats}
 THEME_DIRNAME="$BATS_TEST_DIRNAME"
+CACERT_DIR="$FUNC_DIR/test_certificates" # trusted key store path
+CLIENT_CERT_DIR="$TEST_NAME/target-dir/etc/swupd"
```
castulo: Since this environment variable depends on the value of $TEST_NAME to already exist, I think this variable should be defined within the set_env_variables https://github.com/clearlinux/swupd-client/blob/master/test/functional/testlib.bash#L148, that is its purpose.

src/curl.c
```diff
@@ -62,6 +62,33 @@ static void swupd_curl_strerror(CURLcode curl_ret)
 	fprintf(stderr, "Curl error: (%d) %s\n", curl_ret, curl_easy_strerror(curl_ret));
 }
 
+/* 
+ * Cannot avoid a TOCTOU here with the current curl API.  Only using the curl
+ * API, a curl_easy_setopt does not detect if the client SSL certificate is
+ * present on the filesystem.  The only happens during curl_easy_perform.
```
castulo: Typo here, "The only happens during curl_easy_perform."

test/functional/testlib.bash
```diff
@@ -4,10 +4,18 @@ FUNC_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
+CLIENT_CERT="$CLIENT_CERT_DIR/client.pem"
+
+PORT_FILE="$TEST_NAME/port_file.txt" # stores web server port
+SERVER_PID_FILE="$TEST_NAME/pid_file.txt" # stores web server pid
```
castulo: same here, if these variables depend on a test to already exist they should go into the set_env_variables function, that way we can have all those variables together so they can be set up at once, once the test environment exists. The main idea behind this is to be able to use the library from a test script (which would be its main use) but also have the option of using it interactively from the command line if needed.

test/functional/checkupdate_v2/chk-update-slow-server.bats
```diff
@@ -2,24 +2,27 @@
...
+	create_test_environment "$TEST_NAME"
 
-	echo "terminating web server..." >&3
-	kill "$server_pid" >&3
-	destroy_test_environment "$TEST_NAME"
+	# start slow response web server
+	start_web_server -s
 
+	# update PORT environment variable
+	set_env_variables "$TEST_NAME"
```
castulo: this function is already run from the create_test_environment function.

jwakre: I added the PORT environment variable to set_env_variables which isn't known until after the call to start_web_server. I can remove the call to set_env_variables by moving the code from test_setup to global_setup and creating a test_setup that does nothing.

test/functional/testlib.bash
jwakre: I'll move these variables and the ones listed above, thanks

test/functional/checkupdate_v2/chk-update-slow-server.bats
jwakre: Now, I think it's more clear to remove the call to set_env_variables, move the setup/teardown to the global_setup/global_teardown functions, and do nothing with test_setup/test_teardown. I plan on making the change here.

test/functional/bundleadd_v2/add-client-certificate.bats
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env bats
+
+load "../testlib"
+
+test_dir_path=$(dirname "$(realpath "$TEST_NAME")/$TEST_NAME")
```
jwakre: This command causes the current failures in Travis. I am removing this and storing the keys in /tmp.

test/functional/bundleadd_v2/add-client-certificate.bats
```diff
@@ -0,0 +1,92 @@
...
+global_teardown() {
+
+	destroy_web_server
+	destroy_trusted_cacert
+	destroy_test_environment "$TEST_NAME"
+}
+
+@test "add bundle with valid client cert" {
+
+	opts="-S $TEST_DIRNAME/state -p $TEST_DIRNAME/target-dir -F staging -u https://localhost:$PORT/$TEST_NAME/web-dir"
```
jwakre: The -following arguments are required to fix errors when using the swupd configuration specified by Travis:
-C <path to Swupd_Root.pem>
-I

jwakre: -C \<path to Swupd_Root.pem\>

test/functional/testlib.bash
```diff
@@ -60,6 +60,81 @@ generate_random_name() {
...
+
+	# only one test can use the trusted certificate store at the same time to
+	# avoid certificate contamination from other tests and race conditions.
+	until mkdir "$CACERT_DIR"; do
+		sleep 1
+		if [ "$counter" -eq "1000" ]; then
+			echo "Timeout: creating trusted certificate"
+			return 1
+		fi
+		counter=$["$counter" + 1]
```
jwakre: Travis fails here due to syntax error. The issue is resolved by changing to: counter=$[$counter + 1]

src/curl.c
```diff
@@ -62,6 +62,34 @@ static void swupd_curl_strerror(CURLcode curl_ret)
...
+ * present on the filesystem.  This only happens during curl_easy_perform.
+ * The emphasis is rather on how using an SSL client certificate is an opt-in
+ * function rather than an opt-out function.
+ */
+CURLcode swupd_curl_set_optional_client_cert(CURL* curl) {
+	CURLcode curl_ret = CURLE_OK;
+	char* client_cert_path;
+
+	client_cert_path = mk_full_filename(path_prefix, SSL_CLIENT_CERT);
+	if (access(client_cert_path, F_OK) == 0) {
```
jwakre: Earlier, I had SSL_CLIENT_CERT here instead of client_cert_path which caused the most recent Travis failures. Now, Travis test pass.

jwakre:
Since the initial pull request I have made the following changes:

1. Addressed review feedback
2. Fixed bug with web server port allocation (now this is done in Python instead of bash)
3. Added test skipping functionality for new tests when testing locally
4. Moved certificates, alternative trust store, and port/pid files to /tmp
5. Fixed bug in curl.c where SSL_CLIENT_CERT was not replaced by client_cert_path
6. Fixed bash syntax issues that caused failures on Travis

test/functional/testlib.bash
```diff
@@ -60,6 +60,81 @@ generate_random_name() {
...
+
+	# only one test can use the trusted certificate store at the same time to
+	# avoid certificate contamination from other tests and race conditions.
+	until mkdir "$CACERT_DIR"; do
+		sleep 1
+		if [ "$counter" -eq "1000" ]; then
+			echo "Timeout: creating trusted certificate"
+			return 1
+		fi
+		counter=$[$counter + 1]
```
castulo: I hadn't seen a variable being increased like that in bash before, I just learned something new.
I usually see it like this: counter=$((counter+1))

test/functional/testlib.bash
```diff
@@ -162,6 +237,22 @@ set_env_variables() {
...
 
+	export CLIENT_CERT_DIR="$TEST_NAME/target-dir/etc/swupd"
+	export CLIENT_CERT="$CLIENT_CERT_DIR/client.pem"
+	export CACERT_DIR="/tmp/swupd_test_certificates" # trusted key store path
+	export PORT_FILE="/tmp/$TEST_NAME-port_file.txt" # stores web server port
+	export SERVER_PID_FILE="/tmp/$TEST_NAME-pid_file.txt" # stores web server pid
+
+	# Add environment variables for PORT and SERVER_PID when web server used
+	if [ -f "$PORT_FILE" ]; then
+		export PORT=$(cat "$PORT_FILE")
```
castulo: declaring a variable and assigning it a value in the same step seems to be considered a bad practice in bash: https://github.com/koalaman/shellcheck/wiki/SC2155

test/functional/update_v2/update-client-certificate.bats
```diff
@@ -0,0 +1,111 @@
...
+	if [ -z "${TRAVIS}" ]; then
+		return
+	fi
+
+	create_test_environment "$TEST_NAME"
+	create_test_environment "$TEST_NAME" 100
+
+	create_bundle -n test-bundle -f /usr/bin/test-file "$TEST_NAME"
+
+	run sudo sh -c "mkdir -p $CLIENT_CERT_DIR"
```
castulo: From my understanding of bats, the run helper is intended to be used only to run the command under test, I've noticed you use it sometimes for setup steps, I think in those cases you should use a regular shell command and leave "run" only for the command to be tested.

matthewrsj:
>  One thing I like doing when writing bash scripts is to paste my code in https://www.shellcheck.net

Wow, TIL. Very cool.

test/functional/testlib.bash
```diff
@@ -1031,6 +1122,108 @@ destroy_test_environment() {
...
+			p)	server_args="$server_args --server_cert $OPTARG" ;;
+			s)	server_args="$server_args --slow_server" ;;
+			*)	cb_usage
+				return ;;
+		esac
+	done
+	set +f  # turn globbing back on
+
+	# start web server and write port/pid numbers to their respective files
+	python3 $FUNC_DIR/server.py $server_args --port_file $PORT_FILE &
```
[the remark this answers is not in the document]
jwakre: I will clean these up throughout this pull request, but I think I have one exception. The server_args variable should be split on spaces so that argparse can properly read each argument.

jwakre:
@castulo All of your feedback should be addressed now.

autogen.sh
```diff
@@ -8,7 +8,8 @@ args="\
 --sysconfdir=/etc \
 --localstatedir=/var \
 --prefix=/usr \
---enable-silent-rules"
+--enable-silent-rules \
+--with-fallback-capaths=/tmp/swupd_test_certificates"
```
otaviobp: Do we need to set the fallback-capaths as default in a dev env? Afaik all tests are skiped by default, right? I'd add a comment that you need this flag set in build time if you need to run those tests

jwakre: Yes, '--with-fallback-capaths' is not required when the tests are skipped. I will remove it from autogen.sh and make a comment.

jwakre:
@otaviobp autogen.sh has been updated to remove --with-fallback-capaths

PR #547, opened 2018-08-20
Adding two more functions to testlib

castulo:
This commit add the following two functions to testlib which will be used in upcoming PRs:

- add_to_pack
- update_manifest

Both functions aim to make those actions which are very common in tests in a consistent manner and less error prone.

### DIFF
--- a/test/functional/README.md
+++ b/test/functional/README.md
@@ -70,7 +70,13 @@ global_teardown() {
 @test "<test description>" {
 
 	run sudo sh -c "$SWUPD <swupd_command> $SWUPD_OPTS <command_options>"
-	# <validations>
+
+	# assert_status_is 0
+	# expected_output=$(cat <<-EOM
+	# 	<expected output>
+	# EOM
+	# )
+	# assert_is_output "$expected_output"
 
 }
 ```
@@ -162,7 +168,15 @@ $ make check
 ```
 
 To include tests to be run in the CI system:  
-TBD  
+In order for the CI system to run a new test the test needs to be added to the
+BATS variable in the Makefile.am file
+```bash
+BATS = \
+        test/documentation/manpages/test.bats \
+        test/functional/bundleadd_v2/add-directory.bats \
+        test/functional/bundleadd_v2/add-existing.bats \
+        ...
+```  
 <br/>
 
 ### Debugging Tests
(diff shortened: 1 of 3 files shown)


test/functional/testlib.bash
```diff
@@ -1161,6 +1160,42 @@ install_bundle() {
 
 }
 
+# Adds the specified file to the zero or delta pack for the bundle
+# Parameters:
+# - BUNDLE: the name of the bundle
+# - ITEM: the file or directory to be added into the pack
+# - FROM_VERSION: the from version for the pack, if not specified a zero pack is asumed
+add_to_pack() {
```
matthewrsj: Add this to the readme?

test/functional/testlib.bash
```diff
@@ -543,6 +543,79 @@ remove_from_manifest() {
...
+# Updates fields in an existing manifest
+# Parameters:
+# - -p: if the p (partial) flag is set the function skips updating the hashes
+#       in the MoM, this is useful if more changes are to be done in order to
+#       reduce time
+# - MANIFEST: the relative path to the manifest file
+# - KEY: the thing to be updated
+# - HASH/NAME: the file name or hash of the record to be updated (if applicable)
+# - VALUE: the value to be used for updating the record
+update_manifest() {
```
matthewrsj: Awesome addition! update readme?

castulo: This one is a good candidate for adding to the readme, I'll update it.

test/functional/testlib.bash
castulo: The way I see it, there are two types of functions in the library, those functions that are really of not use to the end user (from a test script) but are useful for other library functions (kind of like a private function), and those that are useful for users writing tests. In this case I don't see much value for a person using this function in isolation, so I'm not sure how important is to add this function to the readme.

castulo:
oh wait, let me add something else quickly, I remember @evil-brain mentioned this:

> castulo, Found another thing that should be in README.am - when using a custom file to create a bundle, the user has to also call create_tar() or their bundle isn't correct...

castulo:
ok, now is ready

test/functional/README.md
```diff
@@ -162,7 +168,15 @@ $ make check
 ```
 
 To include tests to be run in the CI system:  
-TBD  
+In orfer for the CI system to run a new test the test needs to be added to the
```
jwakre: spelling: orfer

castulo:
I'm gonna fix one typo

castulo:
ok, the typo was fixed

PR #549, opened 2018-08-21
Ability to create environment with complex bundle

castulo:
Normal test environments come with a minimal version of os-core
bundle by default (only one file and the tracking file). Users
also have the option of not having os-core at all with the -e
"empty" option.

This commit adds another option "-r" for creating a test environment
that includes a more complete version of the os-core bundle,
with a couple of useful files: os-release and format. This type
of bundle can be useful when working with some type of tests like
updates.

### DIFF
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -782,6 +782,9 @@ create_version() {
 
 	local partial=false
 	local release_files=false
+	# since this function is called with every test environment, some times multiple
+	# times, use simple parsing of arguments instead of using getopts to keep light weight
+	# with the caveat that arguments need to be provided in a specific order
 	[ "$1" = "-p" ] && { partial=true ; shift ; }
 	[ "$1" = "-r" ] && { release_files=true ; shift ; }
 	local env_name=$1
@@ -802,6 +805,9 @@ create_version() {
 			          new version in order to avoid extra processing
 			    -r    if the r flag is set (release), the version is created with hashed os-release
 			          and format files that can be used for creating updates
+
+			Note: if both options -p and -r are to be used, they must be specified in that order or
+			      one option will be ignored.
 			EOM
 		return
 	fi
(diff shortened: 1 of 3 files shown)


test/functional/testlib.bash
```diff
@@ -874,25 +874,32 @@ create_version() {
...
+#       the os-core bundle that includes release files, it is useful for some tests like update tests
 # - ENVIRONMENT_NAME: the name of the test environment, this should be typically the test name
 # - VERSION: the version to use for the test environment, if not specified the default is 10
 # - FORMAT: the format number to use initially in the environment
 create_test_environment() { 
 
 	local empty=false
+	local release_files=false
 	[ "$1" = "-e" ] && { empty=true ; shift ; }
+	[ "$1" = "-r" ] && { release_files=true ; shift ; }
```
matthewrsj: what if `-r` comes before `-e`? Will `-e` be ignored?

castulo: Yes. It is a limitation of using parameters this way, but the alternative is using a full arguments parser (getups) like create_bundle function does, which in my opinion is just too heavy for this. I'm trying to stick to simple arguments parser unless the function is too complex (lots of parameters) like create_bundle.
Also, in this case you cannot have an empty environment and one with a complex os-core, they are mutually exclusive.

castulo: I guess I could use an if ... elif clause to improve those function arguments...

matthewrsj: Or you could just document the limitation in the function comment and readme, I'd be fine with that.

castulo: I'll fix that

castulo: never mind, now that I think of it, using an if elif clause would be the same result, so the only way I can think of for not overriding those parameters would be using getopts, and I don't think is a good idea since these functions get called with every single test, it would slow down execution.

castulo: The usage info also shows that you could use one parameter or the other:
 create_test_environment [-e|-r]

castulo: @matthewrsj I updated the usage info in both functions to show the limitation when using the parameters. I didn't added it to the README because I found it to be very obvious there (not being able to create an empty environment and with a os-core bundle at the same time). If you still think is necessary to add something to the readme, just let me know what kind of message you would like to see there and I'll add it.

PR #559, opened 2018-08-29
search: Allocate the filename for each result dinamically

vcgomes:
Instead of reserving PATH_MAX bytes for the filename (of each result),
allocate it dinamically. This should use less memory overall, at the
cost of making more small allocations.

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -58,7 +58,7 @@ struct bundle_result {
 
 /* per-file scoring struct */
 struct file_result {
-	char filename[PATH_MAX + 1];
+	char *filename;
 	double score;
 };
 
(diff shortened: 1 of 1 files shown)


src/search.c
```diff
@@ -91,7 +91,7 @@ static void add_bundle_file_result(char *bundlename, char *filename, double scor
 	}
 
 	file = calloc(sizeof(struct file_result), 1);
-	strncpy(file->filename, filename, PATH_MAX);
+	file->filename = strdup(filename);
```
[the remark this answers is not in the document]
vcgomes: Sure, I can do that.

About aborting on memory allocations, I see a lot of `calloc()` calls without the same "abort on error" protection.

src/search.c
```diff
@@ -58,7 +58,7 @@ struct bundle_result {
```
jwakre: In commit message: dinamically -> dynamically

vcgomes: Ugh, fixed. Thanks.

PR #560, opened 2018-08-30
Memory

otaviobp:
@vcgomes has reported on #559 that swupd wasn't consistent on handling return value from callocs, mallocs and reallocs.
One possible solution was to create functions calloc_or_die, malloc_or_die and realloc_or_die, but I preferred to create a macro to check for NULL and abort. Any opinions about that?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -38,6 +38,7 @@ swupd_SOURCES = \
 	src/list.c \
 	src/list.h \
 	src/lock.c \
+	src/macros.h \
 	src/main.c \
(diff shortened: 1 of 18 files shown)


tmarcu:
Looks fine to me using a macro once this is rebased. There are cases you may not catch because all of those functions can still return valid pointers on weird allocations, i.e 0 sized allocs, etc, but likely won't even be the case for swupd.

matthewrsj:
needs a rebase

otaviobp:
rebased

Issue #557, opened 2018-08-29
Add support for the iterative manifest into the swupd test library


castulo:
There is a proposal for implementing iterative manifests in swupd (#498) which will affect how updates are done.

This will require adding the capability into the test library for creating these iterative manifests with the bundle updates. Having these changes ready in the test library early will allow for using a test driven development so we can make sure we don't break the clients while implementing this change.

castulo:
This issue should have been closed when PR #563 was merged, I'm closing it now.

PR #563, opened 2018-09-07
Add iterative manifest to bundle update in testlib

castulo:
This commit adds the iterative manifests (to-manifest and from-manifest)
when creating a bundle update. This manifests should contain only those
files that changed in the latest version.

Closes issue #557

### DIFF
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -405,10 +405,12 @@ create_tar() {
 # Parameters:
 # - PATH: the path where the manifest will be created
 # - BUNDLE_NAME: the name of the bundle which this manifest will be for
+# - FORMAT: the format of the manifest
 create_manifest() {
 
 	local path=$1
 	local name=$2
+	local format=${3:-1}
 	local version
 
 	# If no parameters are received show usage
@@ -424,7 +426,7 @@ create_manifest() {
 
 	version=$(basename "$path")
 	{
-		printf 'MANIFEST\t1\n'
+		printf 'MANIFEST\t%s\n' "$format"
 		printf 'version:\t%s\n' "$version"
 		printf 'previous:\t0\n'
 		printf 'filecount:\t0\n'
@@ -531,7 +533,11 @@ add_to_manifest() {
 	contentsize=$((contentsize + item_size))
 	# get the item type
 	if [ "$(basename "$manifest")" = Manifest.MoM ]; then
-		item_type=M
+		if [[ "$(basename "$item")" == Manifest.*.I.* ]]; then
+			item_type=I
+		else
+			item_type=M
+		fi
 		# MoM has a contentsize of 0, so don't increase this for MoM
 		contentsize=0
 		# files, directories and links are stored already hashed, but since
@@ -1733,6 +1739,7 @@ install_bundle() {
 # - -p: if the p (partial) flag is set the function skips updating the hashes
 #       in the MoM, and re-creating the bundle's tar, this is useful if more changes are to be done
 #       in order to reduce time
+# - -i: if the i (iterative) flag is set the iterative manifests are created with every update
 # - ENVIRONMENT_NAME: the name of the test environment
 # - BUNDLE_NAME: the name of the bundle to be updated
 # - OPTION: the kind of update to be performed { --add, --add-dir, --delete, --ghost, --rename, --rename-legacy, --update }
@@ -1742,6 +1749,8 @@ update_bundle() {
 
 	local partial=false
 	[ "$1" = "-p" ] && { partial=true ; shift ; }
+	local iterative=false
+	[ "$1" = "-i" ] && { iterative=true ; shift ; }
 	local env_name=$1
 	local bundle=$2
 	local option=$3
@@ -1759,6 +1768,7 @@ update_bundle() {
 	local fsize
 	local fhash
 	local fname
+	local filename
 	local new_fhash
 	local new_fsize
 	local new_fname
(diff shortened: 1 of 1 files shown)


test/functional/testlib.bash
```diff
@@ -1986,10 +2007,40 @@ update_bundle() {
...
 
+	if [ "$iterative" = true ]; then
+		# create the iterative "to manifest"
+		to_manifest="$bundle_manifest".I."$oldversion"
+		# get the manifest header as is
+		to_manifest_content=$(sed "/^[^\\t]*\\t[^\\t]*$/! d" "$bundle_manifest")$'\n'$'\n'
+		# get only those bundles that match the latest version
+		to_manifest_content+=$(sed "/^....\\t.*\\t$version\\t.*$/! d" "$bundle_manifest")$'\n'
+		write_to_protected_file "$to_manifest" "$to_manifest_content"
+		update_manifest "$to_manifest" filecount "$(grep -c -E "^(F|D)..." "$to_manifest")"
```
phmccarty: Are directories out of scope for iterative manifests, since deltas are not applied for them? If so, the regex should be `^F...` instead. Same comment applies for the others spots that use this regex.

castulo: that is a good question, probably directories are out of scope as you mention

matthewrsj: No, directories are out of scope for the delta from manifest, not for the iterative to manifest. The iterative to manifest has everything that changed in the latest release, including links and directories. So you need to include `L` in this regex.

test/functional/testlib.bash
```diff
@@ -1942,6 +1961,8 @@ update_bundle() {
 		sudo bsdiff "$oldversion_path"/files/"$fhash" "$version_path"/files/"$new_fhash" "$version_path"/delta/"$delta_name"
 		# create or add to the delta-pack
 		add_to_pack "$bundle" "$version_path"/delta/"$delta_name" "$oldversion"
+		# keep the modified file for the iterative "from manifest"
+		from_manifest_content="$(awk "/....\t.*\t.*\t${fname////\\/}/" "$oldversion_path"/Manifest."$bundle")"$'\n'
```
phmccarty: Not sure what the `from_manifest_content` value looks like here. Specifically, I don't understand how this form of pattern substitution work. The awk usage looks okay if you want to print contents (space separated) of the manifest body, omitting the manifest header.

castulo: The _from_manifest_content_ will have only the line from the manifest that was modified during the bundle update. The awk is looking for lines in the manifest that match this pattern:
<any four characters><a tab character><any string><a tab character><another string><a tab character><the name of the file, e.g. /foo/bar>

I guess the part of that pattern that is confusing is this ${fname////\\/}, which as you pointed out is a bash pattern substitution which is going to replace every "/" character with "\/", so basically is just escaping the \

test/functional/testlib.bash
castulo: I'll add the Links to the pattern

test/functional/testlib.bash
```diff
@@ -1986,10 +2007,40 @@ update_bundle() {
...
+		if [ -n "$from_manifest_content" ]; then
+			from_manifest="$version_path"/Manifest."$bundle".D."$oldversion"
+			if [ ! -e "$from_manifest" ]; then
+				write_to_protected_file "$from_manifest" "$(sed "/^[^\\t]*\\t[^\\t]*$/! d" "$oldversion_path"/Manifest."$bundle")"$'\n'$'\n'
+			fi
+			write_to_protected_file -a "$from_manifest" "$from_manifest_content"
+			update_manifest -p "$from_manifest" filecount "$(grep -c -E "^(F|D)..." "$from_manifest")"
+			update_manifest -p "$from_manifest" timestamp "$(date +"%s")"
+			# TODO(castulo): update the contentsize in the from_manifest
+			add_to_pack "$bundle" "$from_manifest" "$oldversion"
```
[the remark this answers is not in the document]
castulo: @jwakre don't you think it could be a good idea to keep it in $version_path also so you can debug your test easier if something goes wrong?

If you need to debug your test and the file has been deleted it will force you to unpack the tar first to be able to see the delta manifest, and since the test environment is going to be deleted anyway having an extra file there should not be a problem IMO.

test/functional/testlib.bash
```diff
@@ -2030,6 +2081,8 @@ add_to_pack() {
 		sudo tar -C "$item_path" -rf "$version_path"/pack-"$bundle"-from-"$version".tar --transform "s,^,staged/," "$(basename "$item")"
 	elif [[ "$item" = *"/delta"* ]]; then
 		sudo tar -C "$version_path" -rf "$version_path"/pack-"$bundle"-from-"$version".tar delta/"$(basename "$item")"
+	elif [[ "$item" = *"/Manifest."*".D"* ]]; then
+		sudo tar -C "$(dirname "$item")" -rf "$(dirname "$item")"/pack-"$bundle"-from-"$version".tar "$(basename "$item")"
```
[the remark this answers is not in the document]
castulo: you are right, I'll make the change, thanks

test/functional/testlib.bash
castulo: I will replace the occurrence of that pattern substitution with another variable so it only shows once, and I will add a comment about what that substitution is doing so the code is easier to read.

castulo:
@phmccarty , @jwakre , I believe I addressed your comments. Please let me know if I'm missing something, or if you have further comments.

test/functional/testlib.bash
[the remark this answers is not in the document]
jwakre: Ok, this is fine with me.

test/functional/testlib.bash
```diff
@@ -2030,6 +2084,8 @@ add_to_pack() {
 		sudo tar -C "$item_path" -rf "$version_path"/pack-"$bundle"-from-"$version".tar --transform "s,^,staged/," "$(basename "$item")"
 	elif [[ "$item" = *"/delta"* ]]; then
 		sudo tar -C "$version_path" -rf "$version_path"/pack-"$bundle"-from-"$version".tar delta/"$(basename "$item")"
+	elif [[ "$item" = *"/Manifest."*".D"* ]]; then
+		sudo tar -C "$item_path" -rf "$item_path"/pack-"$bundle"-from-"$version".tar "$(basename "$item")"
```
jwakre: There was a discussion on the [mixer delta manifest pull request ](https://github.com/clearlinux/mixer-tools/pull/414/commits/fe5f1788dcdc1fb6177d77b3cf675c65a09cc872) about where to store the manifest inside the pack. It looks like the decision was to store the delta manifest inside the TO version's directory instead of at the top level.


- delta/
    - delta files
- staged/
    - full files
- 25720/
    - Manifest.os-core.D.24710

castulo: ooh, thanks for pointing out that info @jwakre , I'll update that.

castulo:
@jwakre I added the delta manifest inside a <to_version> directory within the delta pack as you pointed out. Let me know if there is something else needed. Thanks for the review.

test/functional/testlib.bash
```diff
@@ -2030,6 +2085,8 @@ add_to_pack() {
 		sudo tar -C "$item_path" -rf "$version_path"/pack-"$bundle"-from-"$version".tar --transform "s,^,staged/," "$(basename "$item")"
 	elif [[ "$item" = *"/delta"* ]]; then
 		sudo tar -C "$version_path" -rf "$version_path"/pack-"$bundle"-from-"$version".tar delta/"$(basename "$item")"
+	elif [[ "$item" = *"/Manifest."*".D"* ]]; then
+		sudo tar -C "$version_path" -rf "$version_path"/pack-"$bundle"-from-"$version".tar "$(basename "$version_path")"/"$(basename "$item")"
```
jwakre: It looks like you accidentally swapped $(dirname "$item") with $version_path instead of item_path.

sudo tar -C "$version_path"...... should be sudo tar -C "$item_path".....

castulo: I changed that path to $version_path un purpose.

Before, the item being passed here was the delta manifest in the version directory, e.g.  <environment>/web-dir/30/Manifest.<bundle>.D.20
so $item_path was resolving to <environment>/web-dir/30 and $version_path was therefore resolving to <environment>/web-dir, so item_path was the actual version_path for that one file (I know it is confusing but it was different than the other files).

Now, after the last update, the item being passed to the function is the delta manifest in the TO directory, e.g. <environment>/web-dir/30/30/Manifest.<bundle>.D.20, this way when we add it to the pack, the Manifest goes inside the version directory like 30/Manifest.<bundle>.D.20, but because of this, now $item_path is <environment>/web-dir/30/30/ and $version_path is actually correct <environment>/web-dir/30.

I hope the explanation didn't confuse you more :P

jwakre: I see now... You had more changes than I thought and I needed more updates to my test branch. Looks good!

PR #565, opened 2018-09-11
Iterative manifest

### LINKED ISSUE #558: Enable the use of an iterative "to manifests" in swupd client
At the moment when doing a bundle update, two manifests are required: the manifest of the new version (to manifest), and the manifest of the version currently installed in the system (from manifest). The *to manifest* is necessary so swupd knows what are the things that need to be updated in the system, and the *from manifest* is necessary so swupd knows the current files to which apply deltas to.

By enabling the use of an iterative *to manifest* in swupd client, which only contains the things that need to be updated, we reduce the download time, the amount of memory needed to load that manifest and the processing time for parsing it.

This is part of the work required to implement proposal #498

castulo:
This PR enables the use of an iterative to manifest in swupd client, which only contains the things that need to be updated, therefore reducing the download time, the amount of memory needed to load that manifest and the processing time for parsing it.

Closes #558

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -55,6 +55,7 @@ char *bundle_to_add = NULL;
 struct timeval start_time;
 char *state_dir = NULL;
 int skip_diskspace_check = 0;
+bool keepcache = false;
 
 /* NOTE: Today the content and version server urls are the same in
  * all cases.  It is highly likely these will eventually differ, eg:
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -330,7 +330,7 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
 			if (ptr != NULL) {
 				/* Keep only the bundle name from the file name, the filename should be
 				 * something similar to bundle-name.I.20 */
-				file->bundlename = strndup(file->filename, ptr - file->filename - 2);
+				file->bundlename = strndup_or_die(file->filename, ptr - file->filename - 2);
 				file->from_version = atoi(ptr + 1);
 			}
 		} else {
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -701,6 +701,15 @@ struct manifest *load_manifest(int current, int version, struct file *file, stru
 retry_load:
 	ret = retrieve_manifests(current, version, file->filename, file, file->is_mix);
 	if (ret != 0) {
+		if (file->is_iterative == 1 && file->full_manifest && !retried) {
+			/* Since the iterative manifest failed to be loaded,
+			 * fall back to using the full manifest */
+			file->full_manifest->peer = file->peer;
+			file = file->full_manifest;
+			file->peer->peer = file;
+			retried = true;
+			goto retry_load;
+		}
 		fprintf(stderr, "Failed to retrieve %d %s manifest\n", version, file->filename);
 		return NULL;
 	}
@@ -954,6 +963,22 @@ void link_submanifests(struct manifest *m1, struct manifest *m2, struct list *su
 
 		if (subbed2) {
 			account_new_bundle();
+		}
+	}
+}
+
+/* Links each iterative manifest from a list with its related full manifest from another list */
+void link_iterative_manifests(struct list *iterative_manifests, struct list *full_manifests)
+{
+	struct list *list;
+	struct file *manifest_file;
+
+	list = iterative_manifests;
+	while (list) {
+		manifest_file = list->data;
+		list = list->next;
+		if (manifest_file->is_iterative == 1) {
+			manifest_file->full_manifest = search_bundle_in_manifests(full_manifests, manifest_file->bundlename);
 		}
 	}
 }
@@ -1414,7 +1439,7 @@ void filter_iterative_manifests(struct list *manifests, int version, struct list
 		file = list->data;
 		/* Keep only iterative manifests that are from the specified version and that are
 		 * already installed */
-		if (file->is_iterative == 1 && file->from_version == version  && is_tracked_bundle(file->bundlename)) {
+		if (file->is_iterative == 1 && file->from_version == version && is_tracked_bundle(file->bundlename)) {
 			*iterative_manifests = list_prepend_data(*iterative_manifests, file);
 		} else if (file->is_iterative != 1) {
 			*full_manifests = list_prepend_data(*full_manifests, file);
@@ -1429,4 +1454,7 @@ void filter_iterative_manifests(struct list *manifests, int version, struct list
 			*iterative_manifests = list_prepend_data(*iterative_manifests, file);
 		}
 	}
+
+	/* link the iterative manifest with its related full manifest */
+	link_iterative_manifests(*iterative_manifests, *full_manifests);
 }
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -460,7 +460,7 @@ static int retrieve_manifests(int current, int version, char *component, struct 
 {
 	char *url = NULL;
 	char *filename;
-	char *dir;
+	char *dir = NULL;
 	char *basedir;
 	int ret = 0;
 	struct stat sb;
@@ -517,7 +517,6 @@ static int retrieve_manifests(int current, int version, char *component, struct 
 
 untar:
 	ret = extract_to(filename, dir);
-	free_string(&dir);
 	if (ret != 0) {
 		goto out;
 	} else {
@@ -527,6 +526,7 @@ untar:
 	}
 
 out:
+	free_string(&dir);
 	free_string(&filename);
 	free_string(&url);
 	return ret;
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -907,10 +907,10 @@ void link_submanifests(struct manifest *m1, struct manifest *m2, struct list *su
 		bool subbed1, subbed2;
 		file1 = list1->data;
 		file2 = list2->data;
-		subbed1 = component_subscribed(subs1, file1->filename);
-		subbed2 = component_subscribed(subs2, file2->filename);
+		subbed1 = component_subscribed(subs1, file1->bundlename);
+		subbed2 = component_subscribed(subs2, file2->bundlename);
 
-		ret = strcmp(file1->filename, file2->filename);
+		ret = strcmp(file1->bundlename, file2->bundlename);
 		if (ret == 0) {
 			file1->peer = file2;
 			file2->peer = file1;
@@ -974,11 +974,11 @@ struct list *recurse_manifest(struct manifest *manifest, struct list *subs, cons
 		file = list->data;
 		list = list->next;
 
-		if (!server && !component && !component_subscribed(subs, file->filename)) {
+		if (!server && !component && !component_subscribed(subs, file->bundlename)) {
 			continue;
 		}
 
-		if (component && !(strcmp(component, file->filename) == 0)) {
+		if (component && !(strcmp(component, file->bundlename) == 0)) {
 			continue;
 		}
 
@@ -1277,7 +1277,25 @@ struct file *search_bundle_in_manifest(struct manifest *manifest, const char *bu
 	while (iter) {
 		file = iter->data;
 		iter = iter->next;
-		if (strcmp(file->filename, bundlename) == 0) {
+		if (strcmp(file->bundlename, bundlename) == 0) {
+			return file;
+		}
+	}
+
+	return NULL;
+}
+
+/* Search for a specific Manifest in a list of Manifests */
+struct file *search_bundle_in_manifests(struct list *manifests, const char *bundlename)
+{
+	struct list *iter = NULL;
+	struct file *file;
+
+	iter = list_head(manifests);
+	while (iter) {
+		file = iter->data;
+		iter = iter->next;
+		if (strcmp(file->bundlename, bundlename) == 0) {
 			return file;
 		}
 	}
@@ -1378,4 +1396,37 @@ int enforce_compliant_manifest(struct file **a, struct file **b, int searchsize,
 		}
 	}
 	return ret; // If collisions were found, so manifest is purely additive
+}
+
+/* The function splits a list of manifests into two lists:
+ * - full_manifests: this list will contain all the full manifests that were in the manifests list
+ * - iterative_manifests: this list will contain all the valid iterative manifests from the manifests
+ *   list, plus the full manifest of those bundles that didn't have a valid iterative manifest
+ *   A valid iterative manifest is one that can be used during an update, meaning that the iterative
+ *   manifest must match the "from version" provided, and that it belongs to an already installed bundle*/
+void filter_iterative_manifests(struct list *manifests, int version, struct list **full_manifests, struct list **iterative_manifests)
+{
+	/* Split the manifests into "iterative manifests" and "full manifests" */
+	struct list *list;
+	struct file *file;
+
+	for (list = manifests; list; list = list->next) {
+		file = list->data;
+		/* Keep only iterative manifests that are from the specified version and that are
+		 * already installed */
+		if (file->is_iterative == 1 && file->from_version == version  && is_tracked_bundle(file->bundlename)) {
+			*iterative_manifests = list_prepend_data(*iterative_manifests, file);
+		} else if (file->is_iterative != 1) {
+			*full_manifests = list_prepend_data(*full_manifests, file);
+		}
+	}
+
+	/* Add the full manifest of those bundles that do not have a valid iterative manifest
+	* into the list of iterative_manifests */
+	for (list = *full_manifests; list; list = list->next) {
+		file = list->data;
+		if (!search_bundle_in_manifests(*iterative_manifests, file->bundlename)) {
+			*iterative_manifests = list_prepend_data(*iterative_manifests, file);
+		}
+	}
 }
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -90,7 +90,7 @@ static struct manifest *alloc_manifest(int version, char *component)
 	return manifest;
 }
 
-static struct manifest *manifest_from_file(int version, char *component, bool header_only, bool latest, bool is_mix)
+static struct manifest *manifest_from_file(int version, char *component, bool header_only, bool latest, bool is_mix, bool load_iterative)
 {
 	FILE *infile;
 	char line[MANIFEST_LINE_MAXLEN], *c, *c2;
@@ -251,6 +251,11 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
 		} else if (c[0] == 'M') {
 			file->is_manifest = 1;
 		} else if (c[0] == 'I') {
+			/* only load iterative manifests when enabled*/
+			if (!load_iterative) {
+				free(file);
+				continue;
+			}
 			file->is_manifest = 1;
 			file->is_iterative = 1;
 		} else if (c[0] != '.') { /* unknown file type */
@@ -327,12 +332,16 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
 		if (file->is_iterative == 1) {
 			char *ptr;
 			ptr = strrchr(file->filename, '.');
-			if (ptr != NULL) {
-				/* Keep only the bundle name from the file name, the filename should be
-				 * something similar to bundle-name.I.20 */
-				file->bundlename = strndup_or_die(file->filename, ptr - file->filename - 2);
-				file->from_version = atoi(ptr + 1);
+			/* the filename should be something similar to bundle-name.I.20, if it is
+			 * not, drop the iterative manifest */
+			if (ptr == NULL || strncmp(ptr - 2, ".I", 2) != 0 || atoi(ptr + 1) == 0) {
+				free_string(&file->filename);
+				free(file);
+				continue;
 			}
+			/* Keep only the bundle name from the file name */
+			file->bundlename = strndup_or_die(file->filename, ptr - file->filename - 2);
+			file->from_version = atoi(ptr + 1);
 		} else {
 			file->bundlename = strdup_or_die(c);
 		}
@@ -574,7 +583,7 @@ static void remove_manifest_files(char *filename, int version, char *hash)
 
 struct manifest *load_mom(int version, bool latest, bool mix_exists)
 {
-	return load_mom_err(version, latest, mix_exists, NULL);
+	return load_mom_err(version, latest, mix_exists, NULL, false);
 }
 
 /* Loads the MoM (Manifest of Manifests) for VERSION.
@@ -589,7 +598,7 @@ struct manifest *load_mom(int version, bool latest, bool mix_exists)
  * loaded into memory, this function will return NULL. If err is passed, it is set
  * with the error code.
  */
-struct manifest *load_mom_err(int version, bool latest, bool mix_exists, int *err)
+struct manifest *load_mom_err(int version, bool latest, bool mix_exists, int *err, bool load_iterative)
 {
 	struct manifest *manifest = NULL;
 	int ret = 0;
@@ -610,7 +619,7 @@ verify_mom:
 		return NULL;
 	}
 
-	manifest = manifest_from_file(version, "MoM", false, latest, mix_exists);
+	manifest = manifest_from_file(version, "MoM", false, latest, mix_exists, load_iterative);
 
 	if (manifest == NULL) {
 		if (retried == false) {
@@ -704,9 +713,7 @@ retry_load:
 		if (file->is_iterative == 1 && file->full_manifest && !retried) {
 			/* Since the iterative manifest failed to be loaded,
 			 * fall back to using the full manifest */
-			file->full_manifest->peer = file->peer;
 			file = file->full_manifest;
-			file->peer->peer = file;
 			retried = true;
 			goto retry_load;
 		}
@@ -727,7 +734,7 @@ retry_load:
 		return NULL;
 	}
 
-	manifest = manifest_from_file(version, file->filename, header_only, false, file->is_mix);
+	manifest = manifest_from_file(version, file->filename, header_only, false, file->is_mix, false);
 
 	if (manifest == NULL) {
 		if (retried == false) {
@@ -756,7 +763,7 @@ struct manifest *load_manifest_full(int version, bool mix)
 		return NULL;
 	}
 
-	manifest = manifest_from_file(version, "full", false, false, false);
+	manifest = manifest_from_file(version, "full", false, false, false, false);
 
 	if (manifest == NULL) {
 		fprintf(stderr, "Failed to load %d Manifest.full\n", version);
@@ -1425,36 +1432,55 @@ int enforce_compliant_manifest(struct file **a, struct file **b, int searchsize,
 
 /* The function splits a list of manifests into two lists:
  * - full_manifests: this list will contain all the full manifests that were in the manifests list
- * - iterative_manifests: this list will contain all the valid iterative manifests from the manifests
+ * - consolidated_manifests: this list will contain all the valid iterative manifests from the manifests
  *   list, plus the full manifest of those bundles that didn't have a valid iterative manifest
  *   A valid iterative manifest is one that can be used during an update, meaning that the iterative
  *   manifest must match the "from version" provided, and that it belongs to an already installed bundle*/
-void filter_iterative_manifests(struct list *manifests, int version, struct list **full_manifests, struct list **iterative_manifests)
+void consolidate_manifests(struct list *manifests, int version, struct list **consolidated_manifests)
 {
-	/* Split the manifests into "iterative manifests" and "full manifests" */
+	struct list *full_manifests = NULL;
 	struct list *list;
 	struct file *file;
 
+	/* Split the manifests into "iterative manifests" and "full manifests" */
 	for (list = manifests; list; list = list->next) {
 		file = list->data;
 		/* Keep only iterative manifests that are from the specified version and that are
 		 * already installed */
-		if (file->is_iterative == 1 && file->from_version == version && is_tracked_bundle(file->bundlename)) {
-			*iterative_manifests = list_prepend_data(*iterative_manifests, file);
+		if (file->is_iterative == 1 && file->from_version <= version && is_tracked_bundle(file->bundlename)) {
+			*consolidated_manifests = list_prepend_data(*consolidated_manifests, file);
 		} else if (file->is_iterative != 1) {
-			*full_manifests = list_prepend_data(*full_manifests, file);
+			full_manifests = list_prepend_data(full_manifests, file);
 		}
 	}
 
 	/* Add the full manifest of those bundles that do not have a valid iterative manifest
-	* into the list of iterative_manifests */
-	for (list = *full_manifests; list; list = list->next) {
+	* into the list of consolidated_manifests */
+	for (list = full_manifests; list; list = list->next) {
 		file = list->data;
-		if (!search_bundle_in_manifests(*iterative_manifests, file->bundlename)) {
-			*iterative_manifests = list_prepend_data(*iterative_manifests, file);
+		if (!search_bundle_in_manifests(*consolidated_manifests, file->bundlename)) {
+			*consolidated_manifests = list_prepend_data(*consolidated_manifests, file);
 		}
 	}
 
 	/* link the iterative manifest with its related full manifest */
-	link_iterative_manifests(*iterative_manifests, *full_manifests);
+	link_iterative_manifests(*consolidated_manifests, full_manifests);
+
+	list_free_list(full_manifests);
+}
+
+/* filters a list of manifests so they only include manifests needed for an update */
+void filter_update_manifests(struct list *update_subs, struct list *manifests, struct list **update_manifests)
+{
+	struct list *list;
+	struct file *file;
+	struct sub *sub;
+
+	for (list = update_subs; list; list = list->next) {
+		sub = list->data;
+		file = search_bundle_in_manifests(manifests, sub->component);
+		if (file) {
+			*update_manifests = list_prepend_data(*update_manifests, file);
+		}
+	}
 }
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -251,9 +251,8 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
 		} else if (c[0] == 'M') {
 			file->is_manifest = 1;
 		} else if (c[0] == 'I') {
-			/* ignore this file for future iterative manifest feature */
-			free(file);
-			continue;
+			file->is_manifest = 1;
+			file->is_iterative = 1;
 		} else if (c[0] != '.') { /* unknown file type */
 			free(file);
 			goto err;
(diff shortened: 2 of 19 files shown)


src/manifest.c
```diff
@@ -327,6 +326,20 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
...
+		 * manifest so it matches the bundle name */
+		if (file->is_iterative == 1) {
+			char *ptr;
+			ptr = strrchr(file->filename, '.');
+			if (ptr != NULL) {
+				file->from_version = atoi(ptr + 1);
+			}
+			ptr = strrchr(file->bundlename, 'I');
+			if (ptr != NULL) {
+				*(ptr - 1) = '\0';
```
matthewrsj: Leave a comment here that you need to set `ptr - 1` to `\0` because of the  `.` before the `I`

src/update.c
```diff
@@ -367,6 +367,55 @@ static int main_update()
...
+	struct list *list1;
+	struct file *file1;
+
+	list1 = server_manifest->manifests;
+	while (list1) {
+		file1 = list1->data;
+		list1 = list1->next;
+		if (file1->is_iterative == 1) {
+			/* Discard iterative manifests that are not from the current version */
+			if (file1->from_version == current_version) {
```
matthewrsj: I think you are failing to track manifests that have an iterative version that does not match the current_version since the version check is nested. Wouldn't a better check be
```C
if (file1->is_iterative == 1 && file1->from_version == current_version) {
        iterative_manifests = list_prepend_data(iterative_manifests, file1);
} else {
        full_manifests = list_prepend_data(full_manifests, file1);
}
```

src/update.c
```diff
@@ -367,6 +367,55 @@ static int main_update()
...
+	}
+
+	/* Add the full manifest of those bundles that do not have a valid iterative manifest
+	* into the list of iterative_manifests */
+	struct list *list2;
+	struct file *file2;
+	bool add_manifest;
+
+	list1 = full_manifests;
+	while (list1) {
```
matthewrsj: If we don't already have a helper for finding a manifest in a list of manifests you should add one instead of embedding the logic here.

src/update.c
```diff
@@ -367,6 +367,55 @@ static int main_update()
...
+		list2 = iterative_manifests;
+		while (list2) {
+			file2 = list2->data;
+			list2 = list2->next;
+			if (strcmp(file1->bundlename, file2->bundlename) == 0) {
+				add_manifest = false;
+				break;
+			}
+		}
+		if (add_manifest) {
```
matthewrsj: Drop the `add_manifest` variable and just add it in the `if (strcmp(...) == 0) {}` block above.

src/update.c
```diff
@@ -367,6 +367,55 @@ static int main_update()
...
+			iterative_manifests = list_prepend_data(iterative_manifests, file1);
+		}
+	}
+
+	/* Replace the server manifests with the list that includes the iterative manifests */
+	server_manifest->manifests = iterative_manifests;
```
matthewrsj: At what point do the necessary full manifests get downloaded? Remember the entire point of the iterative manifests was to avoid downloading the full manifests when possible.

src/update.c
```diff
@@ -447,13 +447,23 @@ static int main_update()
 	/* The new subscription is seeded from the list of currently installed bundles
 	 * This calls add_subscriptions which recurses for new includes */
 	grabtime_start(&times, "Add Included Manifests");
+	bool reattempt_download = true;
+add_dependencies:
 	ret = add_included_manifests(server_manifest, current_version, &latest_subs);
 	grabtime_stop(&times);
 	if (ret) {
+		if (reattempt_download) {
```
matthewrsj: Ah, here's the fallback code. Ignore that comment then.

src/update.c
```diff
@@ -396,13 +447,23 @@ static int main_update()
...
+			/* Fall back to using the full manifests */
+			server_manifest->manifests = full_manifests;
+			reattempt_download = false;
+			printf("Cannot load server MoM iterative sub-manifests, trying to download full sub-manifests\n");
```
matthewrsj: This will only fail if `add_subscriptions` fails entirely right? Not when it can't get a couple iterative submanifests? I just want to make sure we aren't printing this every time we get an iterative manifest we can't use

src/update.c
```diff
@@ -37,6 +37,8 @@
 #include "signature.h"
 #include "swupd.h"
 
+#define add_sub_ERR 1
```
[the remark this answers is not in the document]
castulo: good idea

src/update.c
```diff
@@ -162,10 +164,10 @@ int add_included_manifests(struct manifest *mom, int current, struct list **subs
 
 	/* Pass the current version here, not the new, otherwise we will never
 	 * hit the Manifest delta path. */
-	if (add_subscriptions(subbed, subs, current, mom, false, 0) >= 0) {
-		ret = 0;
-	} else {
+	if (add_subscriptions(subbed, subs, current, mom, false, 0) & add_sub_ERR) {
```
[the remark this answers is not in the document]
castulo: yeah I saw that but I wasn't sure if I needed to fail on that code too, I'll add it

src/update.c
castulo: I'll update that too

src/update.c
castulo: that won't work, cause we don't want to add the manifest if the strcmp finds a match, we actually want to add it only if there was no match in the whole list, that is why if the add_manifest variable (initialized to true) if it never got set to false then means there was no match and we would add the manifest.

matthewrsj: ohhhh misread the code, sorry.

src/update.c
castulo: That is a good idea, I will add a search_bundle_in_manifests function to manifest.c (I couldn't find one)

src/update.c
castulo: This is one of the parts of the code where I'm still doing some evaluation to see what would be the best approach.

The main issue is that the _to manifests_ are downloaded as part of adding included manifests, so this is even before getting to load server sub-manifests, by the time the server sub-manifests are loaded the manifests were already downloaded.

As the code is right now, if any of the manifests fail to be downloaded while adding the included manifests, then the whole update is aborted (https://github.com/clearlinux/swupd-client/blob/master/src/update.c#L399).

Now, if this happen when using iterative manifests, we no longer want to abort the update but we want to try now with the full manifest. As the code is in this patch, if any of the manifests fail to be downloaded (iterative or full) while adding included manifests, the process of adding included manifests will be re-attempted but this time using only all full manifests (not only the full manifest of the bundle that failed). I'm seeing if there is an alternative for adding this functionality.

But yes, the message about not being able to load the iterative sub-manifests would only be shown once.

matthewrsj: So the first time it fails to download an iterative manifest it falls back entirely to full manifests? That is not ideal but maybe not a dealbreaker. The only iterative manifests that will be attempted to be downloaded are the iterative manifests from the current version, and not *all* iterative manifests, correct?

castulo: Correct.

I will continue exploring some options anyway to see if there is a way of retrying only on the failed iterative manifest without being too intrusive with the current code.

otaviobp: I agree with matthew on that. It would be better to download only the full manifest of the iterative manifest that has problems, but it isn't a dealbreaker.

src/update.c
```diff
@@ -162,10 +162,10 @@ int add_included_manifests(struct manifest *mom, int current, struct list **subs
 
 	/* Pass the current version here, not the new, otherwise we will never
 	 * hit the Manifest delta path. */
-	if (add_subscriptions(subbed, subs, current, mom, false, 0) >= 0) {
-		ret = 0;
-	} else {
+	if (add_subscriptions(subbed, subs, current, mom, false, 0) & (add_sub_ERR | add_sub_BADNAME)) {
```
otaviobp: This commit is a problem that isn't dependent from your Iterative Manifest implementation. We are going to have a problem in the current master branch too, right?
If so I would open another PR with this commit and have this merged now.

src/update.c
```diff
@@ -365,6 +367,37 @@ static int main_update()
...
+	while (list1) {
+		file1 = list1->data;
+		list1 = list1->next;
+		if (!search_bundle_in_manifests(iterative_manifests, file1->bundlename)) {
+			iterative_manifests = list_prepend_data(iterative_manifests, file1);
+		}
+	}
+
+	/* Replace the server manifests with the list that includes the iterative manifests */
+	server_manifest->manifests = iterative_manifests;
```
otaviobp: this will leak data. you need to free the list nodes from server_manifest->manifests.

src/update.c
```diff
@@ -365,6 +367,37 @@ static int main_update()
 	retries = 0;
 	timeout = 10;
 
+	/* Get a list of all the iterative manifests from the server_manifest that match
```
otaviobp: update code is way to big. You can create a function to perform this task that you are adding. I would do a

    int filter_iterative_manifests(struct list *manifests, struct list **full_manifests, struct list **iterative_manifests);

or something like this.

src/manifest.c
```diff
@@ -323,6 +322,20 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
 		c = c2;
 
 		file->filename = strdup_or_die(c);
+		file->bundlename = strdup_or_die(c);
```
otaviobp: Just an idea for you, if you like to play with strings. In this line you are allocating a string as big as filename for bundlename, but this will leave some trailing bytes, becase the bundlename length is smaller than filename length. So, instead of using strdup you could use strndup:

    ptr = strrchr(filename, '.');
    bundlename = strndup(filename, ptr - filename - 2);

And by that you won't need to look the string for the 'I' again.

src/manifest.c
```diff
@@ -323,6 +322,20 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
...
+		/* If the file is an iterative manifest, copy the from_version and rename the
+		 * manifest so it matches the bundle name */
+		if (file->is_iterative == 1) {
+			char *ptr;
+			ptr = strrchr(file->filename, '.');
+			if (ptr != NULL) {
```
otaviobp: If ptr is NULL we aren't causing any error, but from_version won't be set. I think we should raise an error for this Manifests, because we aren't going to be able to use it.

src/update.c
```diff
@@ -365,6 +367,37 @@ static int main_update()
...
+	* the "from" version, also get a list of all the full manifests in the MoM */
+	struct list *list1;
+	struct file *file1;
+
+	list1 = server_manifest->manifests;
+	while (list1) {
```
otaviobp: One this that is just a question of preference and I'm not sure if I said that before. But I find so much clearer to loop through a list using for instead of while. So we could replace this:

    list1 = server_manifest->manifests;
    while (list1) {
    list1 = list1->next;

to:

    for (list1 = server_manifest->manifests; list; list = list->next) {

With the advantage that list will be pointing to the current element of the list inside the for block.

src/update.c
castulo: yes, is unrelated... it may be a good idea to open a different PR just for this.

src/update.c
castulo: ok, sounds good

src/manifest.c
castulo: Nice!

src/manifest.c
castulo: good point.

src/update.c
castulo: I like it

castulo:
Tests still need to be added to this PR, but everything else should be now finished, so feel free to review the code. I'll add the missing tests soon.

castulo:
I just added a couple things that were needed for being able to test updates with iterative manifests:
1. A _--keepcache_ option for swupd update so the state directory is persisted after an update, which will be useful for tests to verify what manifests were actually downloaded during the update.
2. Enabled some testlib assertions (like _assert_file_exists_) to read root owned files since the state directory is not readable for non root users.

Now we just need the tests... those will be submitted shortly.

castulo:
Looks like I spoke too soon, the new tests are failing here (did not fail in my development box), I will check them. Please hold on reviews till further notice.

matthewrsj:
Make sure you are passing the same configuration options at build time as travis does when you test locally.

castulo:
I added a couple of options to travis (travis.yml) temporarily only for debugging purposes. I'll remove those once I have fixed the memory leak problem.

castulo:
@matthewrsj @otaviobp now that all tests are passing looks like the PR is finally ready to be reviewed.

castulo:
I had to push again to add the Signed-off-by to the commits.

test/functional/update_v2/update-iterative-and-full.bats
```diff
@@ -0,0 +1,60 @@
...
+}
+
+@test "update works correctly using both full and iterative manifests in the same operation" {
+
+	run sudo sh -c "$SWUPD update $SWUPD_OPTS_KEEPCACHE"
+
+	assert_status_is 0
+
+	# verify target files
+	assert_file_exists "$TARGETDIR"/foo/test-file1
```
jwakre: This verifies that the files exist at the correct locations, but it does not validate their contents for a successful update. The hashes of the files could be compared against the manifest to better validate a successful update. You could use something like the following:

hash1=$("$SWUPD" hashdump "TARGETDIR"/foo/test-file1)
hash2=$(get_hash_from_manifest "$WEBDIR"/30/test-bundle-1 /foo/test-file1)
assert_equal "$hash1" "$hash2"

castulo: Well, there are a few reasons why I didn't think on including that on the tests initially:
1) the most important one is that the actual process of applying the updates (deltas) on the files was not changed, what changed was the way the manifests were handled, so there is very small risk that the files would update incorrectly IMO. 
2) The second point is related with the first one, update and verify tests are slow tests due to all the set up they need to run, they take an average of 14 seconds each one, so I was concerned about adding extra validations that could add up more time.
3) Finally one problem I see with checking the file content like that is that we are using the system under test (swupd) to validate itself, which is not a good practice because we could mask bugs. Ideally we would have to write an independent function to validate these hashes in the testlib, or use a different tool, but I know this is probably not feasible.

jwakre: I see. This is fine with me.

test/functional/update_v2/update-iterative-and-full.bats
```diff
@@ -0,0 +1,60 @@
...
+test_setup() {
+
+	create_test_environment "$TEST_NAME"
+	create_bundle -L -n test-bundle-1 -f /foo/test-file1 "$TEST_NAME"
+	create_bundle -L -n test-bundle-2 -f /foo/test-file2 "$TEST_NAME"
+	create_version "$TEST_NAME" 20 10
+	update_bundle -i "$TEST_NAME" test-bundle-1 --update /foo/test-file1
+	create_version "$TEST_NAME" 30 20
+	update_bundle -i "$TEST_NAME" test-bundle-1 --update /foo/test-file1
+	update_bundle -i "$TEST_NAME" test-bundle-2 --update /foo/test-file2
```
jwakre: Which bundle doesn't use iterative manifests? Do you need to drop the '-i' on one of them?

src/update.c
```diff
@@ -511,6 +519,10 @@ static int main_update()
 
 clean_exit:
 	list_free_list(updates);
+	server_manifest->manifests = NULL;
+	list_free_list_and_data(original_manifests, free_file_data);
```
otaviobp: This was a good approach on freeing elements from the list, btw

src/manifest.c
```diff
@@ -695,12 +695,26 @@ struct manifest *load_mom_err(int version, bool latest, bool mix_exists, int *er
...
 retry_load:
 	ret = retrieve_manifests(current, version, file->filename, file, file->is_mix);
 	if (ret != 0) {
+		if (file->is_iterative == 1 && file->full_manifest && !retried) {
+			/* Since the iterative manifest failed to be loaded,
+			 * fall back to using the full manifest */
+			peer = file->peer;
+			file = file->full_manifest;
+			/* re-link the full manifest */
+			file->peer = peer;
```
otaviobp: This block got a bit hard to read. I think it's easier to write that this way:

    file->full_manifest->peer = file->peer;
    file = file->full_manifest;

Does it make sense?

test/functional/update_v2/update-iterative-and-full.bats
castulo: We need to use the '-i' option to tell the testlib to create an iterative manifest for the updates.
But since we are updating from version 10 to version 30, the following iterative manifests are going to exist in web-dir/30:
- Manifest.test-bundle-1.I.20 (since we are updating from 10 -> 30 this iterative manifest doesn't apply so the full manifest will be used instead)
- Manifest.test-bundle-2.I.10

otaviobp:
I performed some tests with this branch to check if it's working with real mixer data.
I installed os-clr-on-clr and desktop in a clearlinux installation and:
1) Updated from 25070 to 25090 using the current content to make sure full Manifests will be used.
 - No problem
2) Updated from 10 (25070) to 20 (25080) using a content generated using mixer with both IManifest and DManifest support:
 - I found a bug in IManifest implementation on mixer. We aren't creating Manifest tarballs. I'll fix that. I created the manifests with a script to continue the test.
 - swupd is downloading full manifests in https://github.com/clearlinux/swupd-client/blob/4ed1ce7009066d556b4e2ae1c9734f966e54fe05/src/update.c#L406
 - With a workaround on that line, the update was successful using your branch and only IManifests were downloaded. If IManifests are missing, full Manifests are downloaded.

src/manifest.c
castulo: yep, good idea, looks nicer that way, thanks for the tip

castulo:
> I performed some tests with this branch to check if it's working with real mixer data.
> I installed os-clr-on-clr and desktop in a clearlinux installation and:
> 
> 1. Updated from 25070 to 25090 using the current content to make sure full Manifests will be used.
> 
> * No problem
> 
> 1. Updated from 10 (25070) to 20 (25080) using a content generated using mixer with both IManifest and DManifest support:
> 
> * I found a bug in IManifest implementation on mixer. We aren't creating Manifest tarballs. I'll fix that. I created the manifests with a script to continue the test.
> * swupd is downloading full manifests in 
>     
>       
>         [swupd-client/src/update.c](https://github.com/clearlinux/swupd-client/blob/4ed1ce7009066d556b4e2ae1c9734f966e54fe05/src/update.c#L406)
>       
>       
>            Line 406
>         in
>         [4ed1ce7](/clearlinux/swupd-client/commit/4ed1ce7009066d556b4e2ae1c9734f966e54fe05)
>       
>       
>       
>       
>   
>           
>             
>              ret = add_included_manifests(server_manifest, current_version, &latest_subs);
> * With a workaround on that line, the update was successful using your branch and only IManifests were downloaded. If IManifests are missing, full Manifests are downloaded.

Hmm, are you sure it is downloading the full manifest of bundles that had a *valid* iterative manifest? 
server_manifest.manifests at that point only contains the iterative manifests (for those bundles that had a valid one), plus full manifests for bundles that didn't, so when looking for the bundle in the Mom it will find the iterative manifest. I don't see why it would get the full manifest (unless the iterative was not there or was invalid, for example from a "from version" that didn't match).

otaviobp:
My bad. The server side was incorrect when I tested that.
I re tested all cases and this branch is working fine for all of them.

src/manifest.c
```diff
@@ -682,12 +695,22 @@ struct manifest *load_mom_err(int version, bool latest, bool mix_exists, int *er
...
 
 retry_load:
 	ret = retrieve_manifests(current, version, file->filename, file, file->is_mix);
 	if (ret != 0) {
+		if (file->is_iterative == 1 && file->full_manifest && !retried) {
+			/* Since the iterative manifest failed to be loaded,
+			 * fall back to using the full manifest */
+			file->full_manifest->peer = file->peer;
+			file = file->full_manifest;
+			file->peer->peer = file;
```
otaviobp: Are you sure you don't need to check if file->peer isn't NULL before doing that? (I'm just asking you that because you were checking that before. It looks correct the way it is)

castulo: Well, the update workflow doesn't really need it because by the time the code gets to this part  file->peer is not going to be NULL since manifests were already linked here https://github.com/clearlinux/swupd-client/pull/565/files#diff-423d35f9c8cae7b481adc9f5a00d0f58R401. So, I was doing that just to be safe. 

But since the update workflow is the only one that uses iterative manifests I guess there is no point on adding that check, that is why I decided to remove it.

test/functional/update_v2/update-iterative-and-full.bats
jwakre: Thanks, I understand now.

castulo:
Since the PR that removes all the old tests is already in review https://github.com/clearlinux/swupd-client/pull/580, and that PR deleted the old update/ directory and renames update_v2/ to update/, I made a change in this PR so the tests for the iterative manifest go straight to the update directory instead of the update_v2 directory so we don't have to submit another patch to move them later.

src/manifest.c
```diff
@@ -1379,3 +1397,36 @@ int enforce_compliant_manifest(struct file **a, struct file **b, int searchsize,
...
+			*full_manifests = list_prepend_data(*full_manifests, file);
+		}
+	}
+
+	/* Add the full manifest of those bundles that do not have a valid iterative manifest
+	* into the list of iterative_manifests */
+	for (list = *full_manifests; list; list = list->next) {
+		file = list->data;
+		if (!search_bundle_in_manifests(*iterative_manifests, file->bundlename)) {
+			*iterative_manifests = list_prepend_data(*iterative_manifests, file);
```
jwakre: When updating, changed bundles will have an iterative manifest, but unchanged bundles will not. As a result, unchanged bundles will be added to the full_manifests list, downloaded, and parsed.

jwakre: Can we safely skip over unchanged bundles and only download updated ones with iterative manifests?

matthewrsj: Good question. If a bundle is not changed in the update we should not be attempting to download any iterative or full manifest for that bundle.

src/manifest.c
```diff
@@ -1429,4 +1454,7 @@ void filter_iterative_manifests(struct list *manifests, int version, struct list
 			*iterative_manifests = list_prepend_data(*iterative_manifests, file);
 		}
 	}
+
+	/* link the iterative manifest with its related full manifest */
+	link_iterative_manifests(*iterative_manifests, *full_manifests);
```
jwakre: I think this function should be moved before the 2nd for loop. This way, the iterative manifest list will not contain any full manifests when linking.

src/manifest.c
matthewrsj: Ok, but what if:
* manifest is given an `I` flag
* manifest is named `name.F.onebillion`

Looks like this manifest well be loaded as an iterative manifest with a `from` version of `0`. In this case we should certainly gracefully fall back to the full manifest. Although in the above example we will fallback to the full manifest (right?) because of the `0` `from` version, but what about if the manifest is named `name.F.22940`? Do we want to continue to try to use the improperly named manifest?

test/functional/update/update-iterative-download-fail.bats
```diff
@@ -0,0 +1,54 @@
...
+	assert_status_is 0
+
+	# verify target files
+	assert_file_exists "$TARGETDIR"/foo
+	assert_file_exists "$TARGETDIR"/bar
+	assert_file_exists "$TARGETDIR"/bat/baz
+
+	# since the iterative manifest was deleted the full manifest should
+	# have been downloaded
+	assert_file_exists "$STATEDIR"/20/Manifest.test-bundle
```
matthewrsj: Should you also assert that the iterative manifest is in the MoM so we know that swupd will have tried to use an iterative manifest and failed?

matthewrsj:
Just a couple questions, otherwise this looks very good.

src/manifest.c
```diff
@@ -323,6 +322,20 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
...
 		file->filename = strdup_or_die(c);
+		/* If the file is an iterative manifest, copy the from_version and obtain the
+		 * bundle name based on the file name */
+		if (file->is_iterative == 1) {
+			char *ptr;
+			ptr = strrchr(file->filename, '.');
+			if (ptr != NULL) {
+				/* Keep only the bundle name from the file name, the filename should be
+				 * something similar to bundle-name.I.20 */
+				file->bundlename = strndup(file->filename, ptr - file->filename - 2);
```
otaviobp: I forgot to comment on that. Can you create an strndup_or_die() function so we can handle strndup errors?

test/functional/update/update-iterative-download-fail.bats
matthewrsj: It's a single line of bash to make sure we don't regress and fail to add the record to the MoM. It ensures we are actually testing what we expect to be testing.

castulo: ok, I'll add that to the tests

matthewrsj:
Ok, there's something I think this PR is missing:

If a bundle is added as an include to an already-tracked bundle in the OS version the client is updating to the client needs to download the full `to` manifest for the new bundle instead of the iterative manifest, regardless of update versions. This is because on the client will need to add several files that were changed in versions older than the client OS version since it is a new bundle.

So you need to do some extra work to ensure that if a bundle is new to the system it will download a full manifest and will not attempt to download the iterative manifest.

src/manifest.c
castulo: As jwakre correctly pointed out, the full manifest for all tracked bundles that had not changed is also being downloaded and parsed. I have made some changes so this doesn't happen again. The changes will be updated in a few minutes.

src/manifest.c
castulo: sure thing

castulo:
@matthewrsj you are right in your comment above, the PR was in fact missing that. But the updated version it now covers it, and already includes a test case for that.

src/manifest.c
```diff
@@ -1379,3 +1397,36 @@ int enforce_compliant_manifest(struct file **a, struct file **b, int searchsize,
...
+	struct list *list;
+	struct file *file;
+
+	for (list = manifests; list; list = list->next) {
+		file = list->data;
+		/* Keep only iterative manifests that are from the specified version and that are
+		 * already installed */
+		if (file->is_iterative == 1 && file->from_version == version  && is_tracked_bundle(file->bundlename)) {
+			*iterative_manifests = list_prepend_data(*iterative_manifests, file);
+		} else if (file->is_iterative != 1) {
```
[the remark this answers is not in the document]
castulo: if we do it like that we may end up with some iterative manifests that do not match the criteria (like using a different from version) in the list of full_manifests.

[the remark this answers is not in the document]
castulo: The logic changed in the latest update, so probably this comment does not apply anymore. We need to have all the full manifests from the MoM in the list of full_manifests now.

src/manifest.c
castulo: If you think this is something that can actually happen in production (to have a bundle with a name like that) then we definitely need to validate the name, so we make sure we are loading the correct from_version.
As you mentioned, if we find that the name is incorrect or we couldn't determine the from_version, we can always just add a from_version of 0 which would cause the iterative manifest to be discarded later.

matthewrsj: Let's forward-proof this code so in case we add a different type of manifest in the future (like `name.F.22590`) we don't need a format bump to introduce the feature.

src/manifest.c
matthewrsj: We are going to need to update this since we plan on *not* linking manifests when delta manifests are available since delta manifests are not signed. In this case file->peer will be null and file->peer->peer will fail.

src/manifest.c
castulo: ok, let's see what I can do with this

castulo:
@otaviobp @matthewrsj @jwakre , the PR is again ready to be reviewed. I think I have addressed all your comments.

To make it easier to review, these are the steps followed by the update:

1. Loads the current MoM, it loads only the full manifests contained in the MoM.
2. Loads the server MoM, it loads all manifests (iterative and full) contained in the MoM. 
3. Consolidate the list of manifests in the server MoM to remove duplicated manifests, so there is only one manifest per bundle. If there is an iterative manifest for a bundle, and the bundle is already installed, and the iterative manifest is from the correct "from" version, it keeps that iterative manifest in the consolidated list and drops the full manifest. If no valid iterative manifest is found, it saves the full manifest in the list instead.
4. Link all the iterative manifests with its related full manifest using its _full_manifest_ member.
5. Create a new subscription list (update_subs) that contains only the subscribed bundles that need to be updated.
6. Add all the included bundles to that list of subscriptions, so the list now have all the subscribed bundles that need to be updated plus all the unsubscribed bundles that need to be installed (if any).
7. Filter the list of manifests in the current MoM using the update_subs list so we only keep the from manifests that will be used for the update.
8. Filter the list of manifests in the server MoM using the update_subs list so we only keep the to manifests that will be used for the update.
9. Download the from and to manifests based on the new lists that only contain the bundles to be updated and/or added. If an iterative manifest fails to be downloaded, the full manifest is downloaded instead.
10. The rest of the update process remains unchanged...

src/manifest.c
```diff
@@ -1425,36 +1432,55 @@ int enforce_compliant_manifest(struct file **a, struct file **b, int searchsize,
...
 	/* Add the full manifest of those bundles that do not have a valid iterative manifest
-	* into the list of iterative_manifests */
-	for (list = *full_manifests; list; list = list->next) {
+	* into the list of consolidated_manifests */
+	for (list = full_manifests; list; list = list->next) {
 		file = list->data;
-		if (!search_bundle_in_manifests(*iterative_manifests, file->bundlename)) {
-			*iterative_manifests = list_prepend_data(*iterative_manifests, file);
+		if (!search_bundle_in_manifests(*consolidated_manifests, file->bundlename)) {
+			*consolidated_manifests = list_prepend_data(*consolidated_manifests, file);
```
jwakre: I think this for loop could be consolidated into the one above.  When adding an item to the full_manifests list, we could check whether it should be added to the consolidated_manifests list too.

castulo: You need to have all the valid iterative manifests first in the list of consolidated manifests or you would not know what full manifests to add to it. After all you only need to add those full manifests of bundles that didn't have an iterative one.

jwakre: You're right, I see now.

src/update.c
```diff
@@ -208,9 +208,16 @@ static int main_update()
...
+	struct list *server_manifests = NULL;
+	struct list *consolidated_manifests = NULL;
+	struct list *current_manifest_updates = NULL;
+	struct list *server_manifest_updates = NULL;
 	struct list *updates = NULL;
 	struct list *current_subs = NULL;
-	struct list *latest_subs = NULL;
+	struct list *initial_subs = NULL;
+	struct list *update_subs = NULL;
+	struct list *nonupdate_subs = NULL;
```
jwakre: It looks like the nonupdate_subs list is populated and freed, but never used for anything. Is there or will there be value for this?

castulo: @jwakre the only reason that list exists is to make sure all the subscriptions are freed. 

Initially _current_subs_ has all the current subscribed bundles, then some of those subscriptions are copied to the _initial_subs_ , and _update_subs_ lists, then included bundles may be added to the _update_subs_ list. So some subscriptions end up being shared in many lists and others only appear in one list. When cleaning's up, we know all the subscriptions are in two lists without being duplicated: __update_subs_ and _non_update_subs_ . So if we want to free all the subscriptions we can just free them through these two lists.

src/manifest.c
```diff
@@ -1379,3 +1397,36 @@ int enforce_compliant_manifest(struct file **a, struct file **b, int searchsize,
...
+{
+	/* Split the manifests into "iterative manifests" and "full manifests" */
+	struct list *list;
+	struct file *file;
+
+	for (list = manifests; list; list = list->next) {
+		file = list->data;
+		/* Keep only iterative manifests that are from the specified version and that are
+		 * already installed */
+		if (file->is_iterative == 1 && file->from_version == version  && is_tracked_bundle(file->bundlename)) {
```
otaviobp: I have the feeling we can use file->from_version <= version

If we are updating from 20 to 30 and we have Manifest.bundle.I.10 it means there was no change in bundle between 10 and 30. So we can use it. @matthewrsj is there any corner case where this isn't valid?

But if we are updating from 10 to 30 and there is a Manifest.bundle.I.20 we can't use it, because we will miss files updated from 10 to 20.

src/globals.c
```diff
@@ -55,6 +55,7 @@ char *bundle_to_add = NULL;
 struct timeval start_time;
 char *state_dir = NULL;
 int skip_diskspace_check = 0;
+bool keepcache = false;
```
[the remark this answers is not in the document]
castulo: done

src/manifest.c
```diff
@@ -460,7 +460,7 @@ static int retrieve_manifests(int current, int version, char *component, struct
 {
 	char *url = NULL;
 	char *filename;
-	char *dir;
+	char *dir = NULL;
```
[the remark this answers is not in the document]
castulo: done

src/globals.c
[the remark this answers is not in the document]
castulo: https://github.com/clearlinux/swupd-client/pull/594

src/manifest.c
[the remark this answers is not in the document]
castulo: https://github.com/clearlinux/swupd-client/pull/593

otaviobp:
Github hid my comment, so I'm pasting it here

    > + * - full_manifests: this list will contain all the full manifests that were in the manifests list
    + * - iterative_manifests: this list will contain all the valid iterative manifests from the manifests
    + *   list, plus the full manifest of those bundles that didn't have a valid iterative manifest
    + *   A valid iterative manifest is one that can be used during an update, meaning that the iterative
    + *   manifest must match the "from version" provided, and that it belongs to an already installed bundle*/
    +void filter_iterative_manifests(struct list *manifests, int version, struct list **full_manifests, struct list **iterative_manifests)
    +{
    +	/* Split the manifests into "iterative manifests" and "full manifests" */
    +	struct list *list;
    +	struct file *file;
    +
    +	for (list = manifests; list; list = list->next) {
    +		file = list->data;
    +		/* Keep only iterative manifests that are from the specified version and that are
    +		 * already installed */
    +		if (file->is_iterative == 1 && file->from_version == version  && is_tracked_bundle(file->bundlename)) {
I have the feeling we can use file->from_version <= version

If we are updating from 20 to 30 and we have Manifest.bundle.I.10 it means there was no change in bundle between 10 and 30. So we can use it. @matthewrsj is there any corner case where this isn't valid?

But if we are updating from 10 to 30 and there is a Manifest.bundle.I.20 we can't use it, because we will miss files updated from 10 to 20.

otaviobp:
This makes sense. ahhahahaha.
But in this case if this bundle is included by another bundle won't the full Manifest be downloaded because of the add_included_manifests()?

otaviobp:
> @otaviobp, but if we are updating from 20 to 30 we don't really need Manifest.bundle.I.10 because that was updated in a previous version. That should have been already updated when going from v10 to v20.

but in this case we are downloading the full manifest. We don't need any information about this manifest? consolidate_manifests() is adding that to the consolidated_manifests

jwakre:
> > @otaviobp, but if we are updating from 20 to 30 we don't really need Manifest.bundle.I.10 because that was updated in a previous version. That should have been already updated when going from v10 to v20.
> 
> but in this case we are downloading the full manifest. We don't need any information about this manifest? consolidate_manifests() is adding that to the consolidated_manifests

If I understand things correctly, not every manifest in the consolidated_manifests list is downloaded. The 'filter_update_manifests' function creates 'server_manifest_updates' which is a subset of consolidated_manifests that contains manifests that will be downloaded.

castulo:
@otaviobp I tried the following scenario:
1. Create bundleA in version 10 and install it.
2. Create a version 20 (with no updates)
3. Create a version 30 (with no updates)
4. Create a version 40 with an update for bundleA
5. set the current version of the target system to 20
6. run an update

Result: the full manifest for bundleA was used
Expected result: the iterative manifest should be used

I think you are probably right that this is caused because of this if within the consolidate_manifests function:
```
if (file->is_iterative == 1 && file->from_version == version && is_tracked_bundle(file->bundlename)) {
	*consolidated_manifests = list_prepend_data(*consolidated_manifests, file);
} else if (file->is_iterative != 1) {
	full_manifests = list_prepend_data(full_manifests, file);
}
```

castulo:
I fixed the issue you mentioned @otaviobp and added one more test to make sure we don't regress.

PR #566, opened 2018-09-12
Return error if adding subscriptions fails

castulo:
When adding included manifests during an update the add_subscriptions
function is used, this function returns a value of either add_sub_ERR=1,
or add_sub_BADNAME=4 when it fails to add a subscription, and the code
is currently considering any value > 0 to be a successful value.

This commit fixes the issue by making sure the add_subscriptions
function doesn't return a code of 1 or 4 which would mean it failed
adding a subscription.


castulo:
I just updated the commit message.

PR #567, opened 2018-09-13
Group generic libraries in the same dir and add propose some changes to list implementation

otaviobp:
As we are working on a Thread Pool for swupd, we thought about having all generic libraries (list, hashmap and ThreadPool) isolated in a single directory to distinguish them from swupd code.

When I was moving this code I decided to take some time to suggest some improvements in the current List API.


otaviobp:
It looks like a better idea. I'll take a look at  them

matthewrsj:
+1 for the first commit though.

otaviobp:
The first commit of this message will be included in the thread pool pull request and I'll move the discussion of the glibc list to some time in the future. So I'm closing this pull request

PR #574, opened 2018-09-17
clean: remove manifest hash hints from cache

### LINKED ISSUE #484: Unlink manifest cache files at the end of an operation
When a manifest cache file (\<VER\>/Manifest.\<BUNDLE\>.\<HASH\>) exists in the state directory for a manifest, a hash check for that manifest ends prematurely without computing its hash because the cache file and manifest file are hardlinked and therefore share the same inode.

To protect against the possibility of swupd-client corrupting either a bundle manifest or an associated cache file, I think the cache files should be unlinked after a swupd operation (e.g. `swupd update`) is complete. This way, the optimization provided by the cache files (i.e. not duplicating manifest hash checks for deeply nested bundle hierarchies) is still present for the duration of the operation, and by unlinking the cache files afterward, we avoid the possibility of future corruption, as the manifest hashes will be recomputed and cache files recreated at the beginning of the next operation.

matthewrsj:
Fixes #484

The manifest files with their hashes appended to the filename in the
cache are used internally to prevent swupd from re-calculating manifest
hashes every time they are opened. These should only persist for the
duration of the update and should not persist afterwards to prevent
root-priveleged owners from modifying internally-used data. This patch
causes swupd to remove them by default without having to pass --all. The
files will be automatically removed upon update.

### DIFF
--- a/src/clean.c
+++ b/src/clean.c
@@ -189,11 +189,55 @@ static bool is_all_digits(const char *s)
 	return true;
 }
 
+static bool is_all_xdigits(const char *s)
+{
+	for (; *s; s++) {
+		if (!isxdigit(*s)) {
+			return false;
+		}
+	}
+	return true;
+}
+
 static bool is_manifest(const char UNUSED_PARAM *dir, const struct dirent *entry)
 {
 	static const char prefix[] = "Manifest.";
 	static const size_t prefix_len = sizeof(prefix) - 1;
 	return !strncmp(entry->d_name, prefix, prefix_len);
+}
+
+static bool is_hashed_manifest(const char UNUSED_PARAM *dir, const struct dirent *entry)
+{
+	/* check that this has the manifest prefix */
+	if (!is_manifest(dir, entry)) {
+		return false;
+	}
+
+	int i;
+	int start = sizeof("Manifest.");
+	const char *ename = entry->d_name + start;
+
+	/* check for the correct number of '.' characters (i.e.
+	 * Manifest.bundlename.hashvalue). Note, this function returns false for
+	 * iterative manifests (Manifest.bundlename.I.version) and delta manifests
+	 * (Manifest.bundlename.D.version) because of the number of '.' characters
+	 * in the file name. Expect 1 '.' because we are starting after the prefix
+	 * 'Manifest.'
+	 */
+	for (i = 0; ename[i]; ename[i] == '.' ? i++ : *ename++) {
+		/* this for loop is counting the occurrances of '.' by incrementing a
+		 * counter (i) when the character is encountered. The 'i' variable is
+		 * used as an index into the string as well. The pointer to the string
+		 * is advanced when a '.' is not encountered, so the index into the
+		 * original string is num_dots + num_not_dots.
+		 */
+	}
+	if (i != 1) {
+		return false;
+	}
+
+	/* grab hash suffix by splitting on last '.' then iterating past it (+ 1) */
+	return is_all_xdigits(strrchr(entry->d_name + start, '.') + 1);
 }
 
 static bool is_manifest_delta(const char UNUSED_PARAM *dir, const struct dirent *entry)
@@ -288,16 +332,19 @@ static int clean_staged_manifests(const char *path, bool dry_run, bool all)
 			continue;
 		}
 
+		char *version_dir;
+		string_or_die(&version_dir, "%s/%s", state_dir, name);
+
 		/* This is not precise: it may keep Manifest files that we don't use, and
 		 * also will keep the previous version. If that extra precision is
 		 * required we should parse the manifest. */
 		if (mom_contents && strstr(mom_contents, name)) {
-			continue;
+			/* Remove all hash-hint manifest files. */
+			ret = remove_if(version_dir, dry_run, is_hashed_manifest);
+		} else {
+			/* Remove all manifest files, including hash-hints */
+			ret = remove_if(version_dir, dry_run, is_manifest);
 		}
-
-		char *version_dir;
-		string_or_die(&version_dir, "%s/%s", state_dir, name);
-		ret = remove_if(version_dir, dry_run, is_manifest);
 
 		/* Remove empty dirs if possible. */
 		(void)rmdir(version_dir);


src/clean.c
```diff
@@ -288,15 +309,28 @@ static int clean_staged_manifests(const char *path, bool dry_run, bool all)
 			continue;
 		}
 
+		char *version_dir;
+		string_or_die(&version_dir, "%s/%s", state_dir, name);
+
+		/* Remove all hash-hint manifest files.
+		 * Don't perform mom_contents check because we always want to remove
+		 * these */
+		ret = remove_if(version_dir, dry_run, is_hashed_manifest);
```
otaviobp: When --all, --dry-run is going to print hashed manifests twice and we will check for the prefix "Manifest." twice for all files. I prefer to run remove_if only once, using is_manifest when name isn't in mom_contents and is_hashed_manifest otherwise.

src/clean.c
```diff
@@ -189,6 +189,27 @@ static bool is_all_digits(const char *s)
 	return true;
 }
 
+static bool is_all_xdigits(const char *s)
+{
+	for (; *s; s++) {
+		if (!isxdigit(toupper(*s))) {
```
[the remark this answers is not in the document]
matthewrsj: Ah, ok. The manpage only mentions the uppers so I made an assumption.

src/clean.c
matthewrsj: ah, good catch.

src/clean.c
```diff
@@ -189,6 +189,27 @@ static bool is_all_digits(const char *s)
...
+			return false;
+		}
+	}
+	return true;
+}
+
+static bool is_hashed_manifest(const char UNUSED_PARAM *dir, const struct dirent *entry)
+{
+	static const char prefix[] = "Manifest.";
+	static const size_t prefix_len = sizeof(prefix) - 1;
```
phmccarty: Any reason for using static storage here? `prefix_len` will be recomputed every time from `prefix`, so I'm not seeing why it's needed.

otaviobp: If you set it as static it will compute it only once. (I have the felling the compiler will optimize that anyway, but setting this as static makes sense to me)

matthewrsj:
@otaviobp @phmccarty addressed comments.

src/clean.c
```diff
@@ -189,13 +189,39 @@ static bool is_all_digits(const char *s)
...
 static bool is_manifest(const char UNUSED_PARAM *dir, const struct dirent *entry)
 {
 	static const char prefix[] = "Manifest.";
 	static const size_t prefix_len = sizeof(prefix) - 1;
 	return !strncmp(entry->d_name, prefix, prefix_len);
 }
 
+static bool is_manifest_type(const char UNUSED_PARAM *dir, const struct dirent *entry)
+{
+	return is_manifest(dir, entry) || is_hashed_manifest(dir, entry);
```
otaviobp: Isn't all hashed_manifests also manifests? I think clean --all is already cleaning hashed manifests

matthewrsj: Ah, yep, fixed and pushed.

src/clean.c
```diff
@@ -189,6 +189,27 @@ static bool is_all_digits(const char *s)
...
+	return true;
+}
+
+static bool is_hashed_manifest(const char UNUSED_PARAM *dir, const struct dirent *entry)
+{
+	static const char prefix[] = "Manifest.";
+	static const size_t prefix_len = sizeof(prefix) - 1;
+	if (strncmp(entry->d_name, prefix, prefix_len)) {
+		return false;
+	}
```
jwakre: Can you call the is_manifest function instead of using the above code?

matthewrsj: nice. yes.

matthewrsj:
Ha, I just noticed we have a bundle called `bc` which has an associated manifest called `Manifest.bc` which is identified as a hashed manifest...

matthewrsj:
we would have the same problem with a `Manifest.deadbeef`, `Manifest.deadcafe`, `Manifest.feedface` etc.

phmccarty:
Perhaps you could check for the expected number of hex digits instead?

matthewrsj:
That's an idea, I was going to count the number of `.`s but maybe that's better...

matthewrsj:
Actually going with my approach because 1) dots are not allowed in bundle names and 2) someone could have a very long bundle name that is all valid hex xD (deadbeefcafeface....)

matthewrsj:
ok @otaviobp @jwakre @phmccarty the `is_hashed_manifest` function changed a little bit to account for hex-named bundles.

src/clean.c
```diff
@@ -189,13 +189,51 @@ static bool is_all_digits(const char *s)
...
+	int i;
+	char *ename, *tmp = NULL;
+	tmp = ename = strdup_or_die(entry->d_name);
+
+	/* check for the correct number of '.' characters
+	 * (i.e. Manifest.bundlename.hashvalue). Note, this will not clean up
+	 * iterative manifests (Manifest.bundlename.I) because I is not a valid
+	 * xdigit.
+	 */
+	for (i = 0; ename[i]; ename[i] == '.' ? i++ : *ename++) {
```
otaviobp: you could use entry->d_name here so you won't need to duplicate memory. *ename++ looks wrong too.
I would use strchr for that and then you can get the pointer to inform is_all_xdigits so you won't need to look for '.' again in line 234

src/clean.c
```diff
@@ -189,13 +189,51 @@ static bool is_all_digits(const char *s)
...
+		return false;
+	}
+
+	int i;
+	char *ename, *tmp = NULL;
+	tmp = ename = strdup_or_die(entry->d_name);
+
+	/* check for the correct number of '.' characters
+	 * (i.e. Manifest.bundlename.hashvalue). Note, this will not clean up
+	 * iterative manifests (Manifest.bundlename.I) because I is not a valid
```
otaviobp: I would check the hash size too. This isn't a problem for D manifests because of the number of '.'s, but we got close.

src/clean.c
matthewrsj: why does *ename++ look wrong to you? I am marching through the string by updating the pointer to it. I copy the memory because I can't (and wouldn't want to) modify the `const char *entry->d_name`

src/clean.c
matthewrsj: As you said the delta manifests will fail this check because of the mismatch of '.'s so I don't see the point of adding another check there. Also, my iterative manifest comment is incorrect since the format for that is `Manifest.bundlename.I.version` so it will also fail this test. I'll update that comment.

src/clean.c
otaviobp: It is wrong because you are not moving the pointer. *ename++ is changing the value of the memory pointed by ename, so you are updating the string and not moving the pointer forward. that's why you need to move i forward too. If you use a string "012345" to this loop the string will be "612345" after that.

If you are doing a read only operation you never need to duplicate the string. Just move the pointer: 

    char *tmp;
    for (tmp = entry->d_name; tmp; tmp++)

But you don't need to do that. As you are already looking for the first '.' in the string you just need to look for the second one. You can do this:

    #define MANIFEST_PREFIX "Manifest."
    
    suffix = strchr(entry->d_name + sizeof(MANIFEST_PREFIX), '.'); // Look for next '.' after Manifest.
    return suffix && is_all_xdigits(suffix + 1); // if '.' was found check if all digits after '.' are xdigits

otaviobp: ops, my bad. the comment about  *ename++ is incorrect. It's not updating the content. but anyway, I don't get why moving i and ename foward.

otaviobp: Ok, now I got that. It's to count the number of occurrences at the same time you're moving forward. haha. If you want to keep this you can, but add some comment about that.

matthewrsj:
@otaviobp added comments.

otaviobp:
You still don't need to strdup the string as you aren't changing the content of entry->d_name.

And a nice-to-have improvement is to start from "entry->d_name + sizeof("Manifest.")" so you don't need to look for the first '.'. Also, you don't need to count all dots in the string. because if it has at least 2 dots you can call is_all_xdigits() and it will return false for Iterative and Delta Manifests.
By that you would read each byte of the string only once if it's a manifest hash and even less than that if it's a Iterative or Delta manifest.

matthewrsj:
> Also, you don't need to count all dots in the string. because if it has at least 2 dots you can call is_all_xdigits() and it will return false for Iterative and Delta Manifests.

Actually for iterative and delta manifests `is_all_xdigits` will return true because the trailing version are all xdigits.

matthewrsj:
^ and I've confirmed that

matthewrsj:
Still counting the number of '.' characters, but starting at sizeof("Manifest.") and not duplicating the string.

otaviobp:
> Actually for iterative and delta manifests is_all_xdigits will return true because the trailing version are all xdigits.

If you use strchr it won't. "Manifest.bundle.D.1234" will call is_all_xdigits("D.1234") that will return false.

But I'm ok with your implementation too.

PR #580, opened 2018-09-20
Testlib update verify

castulo:
This PR includes the following items:
- Adds a wrapper function to the test library to print to console from within the tests
- Replaces all the old *update* tests that used the swupd.bash library with an equivalent test that now uses the testlib.bash library.
- Replaces all the old *verify* tests that used the swupd.bash library with an equivalent test that now uses the testlib.bash library.
- All the old tests from other themes (bundle-add, bundle-remove, etc.) were removed since they already had an equivalent newer test.
- Directories for the test themes were renamed from <theme>_v2/ to <theme>/
- Since there are no longer tests that use the old test infrastructure it was removed so all future tests are written using the new test library.


castulo:
Ups, I forgot about the certificate. I'll fix this right away.

test/functional/testlib.bash
```diff
@@ -147,6 +147,16 @@ print_stack() {
```
jwakre: Spelling: frmework -> framework (in commit message)

test/functional/generate-cert.prereq
```diff
@@ -1,11 +1,20 @@
...
+test_teardown() {
+	:
+}
 
 @test "Generate certificate for signing Manifest.MoM" {
-  run sudo sh -c "rm -rf $CERT $BATS_TEST_DIRNAME/private.pem"
-  generate_cert
+
+  run sudo sh -c "rm -rf "$FUNC_DIR"/private.pem "$FUNC_DIR"/Swupd_Root.pem"
+  generate_certificate "$FUNC_DIR"/private.pem "$FUNC_DIR"/Swupd_Root.pem
```
jwakre: Functionally, there are some differences between 'generate_cert' and 'generate_certificate'.

1. Subjects are different: Does this matter for the MoM certificate? I had to set CN=localhost to get a webserver to work with the generated certificate.

2. The 'generate_certificate' function doesn't consume the certattributes.cnf file. Maybe this is necessary for the MoM certificate? If we don't consume this file with 'generate_certificate', then it is no longer used.

test/functional/testlib.bash
castulo: damn typos! haha, thanks

test/functional/generate-cert.prereq
castulo: you are right, I missed the part where the generate_cert function was using the certattributes.cnf file. To be honest I have no idea if this is needed or not, I don't know much about certificates, but I'll add as a optional parameter to the generate_certificate function. Good catch @jwakre !

matthewrsj:
@castulo the test that you don't think is testing anything (https://github.com/clearlinux/swupd-client/tree/master/test/functional/bundleadd/fix-missing-file) is intended to test that a bundle-add works when there is no zero pack to download. In that case it falls back to downloading the full files for the bundle. It seems this test is working.

The test you thought it might be a duplicate of (https://github.com/clearlinux/swupd-client/tree/master/test/functional/bundleadd/verify-fix-path) is validating something else. That test is checking that it will fall back to downloading full files if the zero pack is *incomplete* and missing one of the files needed in the bundle-add. This test makes sure that file makes it onto the target regardless of pack completeness.

tl;dr
test 1 - test that bundle-add works with no zero pack
test 2 - test that bundle-add works with incomplete zero pack

matthewrsj:
In general looks good to me as well, can you just confirm that you are covering those two cases I mention above?

matthewrsj:
Actually, both of these tests are not going to attempt to download the zero packs at all due to 

https://github.com/clearlinux/swupd-client/blob/master/src/bundle.c#L825

which only downloads the zero pack if there are greater than 10 files needed to download. So those tests need an update to add at least 11 files so that packs are initially attempted.

castulo:
> @castulo the test that you don't think is testing anything (https://github.com/clearlinux/swupd-client/tree/master/test/functional/bundleadd/fix-missing-file) is intended to test that a bundle-add works when there is no zero pack to download. In that case it falls back to downloading the full files for the bundle. It seems this test is working.
> 
Well, if the test is attempting to test that, then it is correct. However, I'm still not sure that is the purpose of the test. Consider this: 
1. There are many other tests (in the old tests) that don't have zero packs, so this would be a duplicate.
2. The description of the test says this: "bundle-add verify_fix_path support", which suggests that the test is validating the verify_fix_path function (https://github.com/clearlinux/swupd-client/blob/9edc02c15b50c3ed8cb35df797abf12491e3151c/src/helpers.c#L702) is being used some how, but I don't even see this function being called in this test. Maybe it is intended for what you say it is but someone copied the test description from other test and forgot to update it.

> The test you thought it might be a duplicate of (https://github.com/clearlinux/swupd-client/tree/master/test/functional/bundleadd/verify-fix-path) is validating something else. That test is checking that it will fall back to downloading full files if the zero pack is _incomplete_ and missing one of the files needed in the bundle-add. This test makes sure that file makes it onto the target regardless of pack completeness.
> 
I think you may be mistaken about this test, I believe this test is intended to validate that bundle-add uses the verify_fix_path function here https://github.com/clearlinux/swupd-client/blob/c8ca0d3bdf37e0fa33f42f657bade8736343f33b/src/staging.c#L100 to create the missing directories so the file in the manifest can be added. This test looks fine.

castulo:
Hey @matthewrsj I just added a few tests to replace the old https://github.com/castulo/swupd-client/tree/master/test/functional/bundleadd/fix-missing-file test, I think the new ones make more sense for what the original test was validating. I had to cherry-pick this [commit](
https://github.com/clearlinux/swupd-client/pull/580/commits/638f1e0a8f21cfa9e9711bfdbe2fdaf4000803af) from PR #565 because I needed that functionality for these tests. So whatever PR merges second it will need a rebase.

matthewrsj:
@castulo did you add the case to validate that bundle-add will fallback to full-files when there are more than 10 files to download but the zero pack doesn't exist? Or was that case already covered? I don't see it in your latest commit.

castulo:
@matthewrsj I added one test that adds a bundle that has more than 10 files, but one of the files is missing in the zero pack, so the full-file for tat one file is used. There is no case where the whole zero pack is missing. I can add one like that if you think it'd be good to have it.

matthewrsj:
Please add it. May not be a different code path right now but it could be in the future.

test/functional/testlib.bash
```diff
@@ -68,20 +68,31 @@ generate_certificate() {
...
+		openssl req -x509 -sha256 -nodes -days 365 -newkey rsa:2048 \
+			-keyout "$key_path" -out "$cert_path" \
+			-subj "/C=US/ST=Oregon/L=Portland/O=Company Name/OU=Org/CN=localhost"
+	else
+		validate_item "$config"
+		openssl req -x509 -sha256 -nodes -days 365 -newkey rsa:2048 \
+			-keyout "$key_path" -out "$cert_path" \
+			-subj "/C=US/ST=Oregon/L=Portland/O=Company Name/OU=Org/CN=localhost" \
+			-config "$config"
+	fi
```
jwakre: To avoid duplication, can we move the identical sections of the openssl command outside of the condition statement? I think you can store -

config_cmd=""

if [ -n "$config"]; then
           config_cmd="-config $config"
fi

openssl ......... "$config_cmd"

castulo: I tried that @jwakre  but it didn't work correctly, so I went with adding the full command.

Issue #211, opened 2017-05-08
Retry code rework


tmarcu:
The retry code in swupd is not an ideal implementation, making it difficult to use throughout the code base. Some parts retry differently based on needing to support retries for network errors, or retry based on IO errors. Refactoring this code into a consistent, callable implementation would benefit many codepaths and potentially reduce the complexity and obscure bugs.

PR #581, opened 2018-09-20
Delta Manifests

### LINKED ISSUE #561: Enable "From Manifests" in swupd client
During an update, swupd downloads the current (from) version's complete manifests for all bundles on the system. These manifests are used to provide the "from" file path so that swupd knows which files to apply  deltas to. Only the lines in the manifests which contain deltas are useful for an update, so there is time wasted performing unnecessary downloads and manifest parsing. To improve performance, a manifest that includes information for changed files will be included inside the delta packs. Swupd must be able to consume manifests when they are found in delta packs and use the current "from" manifests implementation when delta packs don't include a manifest.

This is part of the work required to implement proposal #498

jwakre:
This PR enables the consumption of delta manifests in the swupd client. Delta manifests are manifests that may be found inside delta packs and contain a list of files (from the current version) that have been changed in the TO version.

This PR also cherry picks some changes from #565 which are used by the added tests. Once these cherry picked changes are merged into master, they will be removed from this PR.

Closes #561


src/update.c
```diff
@@ -365,7 +365,45 @@ static int main_update()
...
+	grabtime_start(&times, "Add Included Manifests");
+	ret = add_included_manifests(server_manifest, current_version, &latest_subs);
+	grabtime_stop(&times);
+	if (ret) {
+		ret = EMANIFEST_LOAD;
+		goto clean_exit;
+	}
+	set_subscription_versions(server_manifest, current_manifest, &latest_subs);
+
+	grabtime_start(&times, "Download Packs");
```
jwakre: The "Recurse and Consolidate Manifests" time recording block starts before loading the server MoM, so now it includes pack downloading in its time statistics. I think that the "Recurse and Consolidate Manifests" time block should be moved so that it starts before loading the subscribed manifests instead of before loading the server MoM. I will open a separate issue for this.

Makefile.am
```diff
@@ -233,6 +233,10 @@ BATS = \
 	test/functional/update/slow-server/test.bats \
 	test/functional/update/statedir-bad-hash/test.bats \
 	test/functional/update_v2/update-client-certificate.bats \
+	test/functional/update_v2/update-delta-manifest-invalid.bats \
+	test/functional/update_v2/update-delta-manifest-no-iterative.bats \
+	test/functional/update_v2/update-delta-manifest-and-full.bats \
+	test/functional/update_v2/update-delta-manifest.bats \
```
castulo: since there is already a PR that removes the update_v2 directory I'd suggest to add this directly to the update directory.

test/functional/update_v2/update-delta-manifest.bats
```diff
@@ -0,0 +1,58 @@
...
+		Finishing download of update content...
+		Staging file content
+		Applying update
+		Update was applied.
+		Calling post-update helper scripts.
+		Update successful. System updated from version 10 to version 30
+	EOM
+	)
+	assert_in_output "$expected_output"
+}
```
castulo: What about the case when a bundle has updates during a couple of versions?
For example:

create_test_environment "$TEST_NAME" 10
create_bundle -L -n bundle1 -v 10 -f /file1 "$TEST_NAME"
create_version "$TEST_NAME" 20 10 staging
update_bundle -i "$TEST_NAME" bundle1 --update /file1
create_version "$TEST_NAME" 30 20 staging
update_bundle -i "$TEST_NAME" bundle1 --update /file1

So if the current version is 10 and we do an update, it should update all the way to version 30, iterative and delta manifests are going to exist for each version, but since the jump is larger than one version the delta manifests should not be used, correct? (I know the iterative manifest should not be used)
Maybe a test for this scenario is worth adding.

Makefile.am
jwakre: Good idea, I will update this.

test/functional/update_v2/update-delta-manifest.bats
jwakre: My understanding is that the server would look like the following:

* 20
    * pack-bundle1-from-0.tar
    * pack-bundle1-from-10.tar
* 30
    * pack-bundle1-from-0.tar
    * pack-bundle1-from-10.tar
    * pack-bundle1-from-20.tar

When updating FROM 10 TO 30, a valid delta manifest would be extracted from /30/pack-bundle1-from-10.tar. This delta manifest should include entries for the files that changed during the update FROM 10 TO 30.

src/manifest.c
```diff
@@ -679,38 +684,54 @@ struct manifest *load_mom_err(int version, bool latest, bool mix_exists, int *er
...
 
 retry_load:
-	ret = retrieve_manifests(current, version, file->filename, file, file->is_mix);
-	if (ret != 0) {
-		fprintf(stderr, "Failed to retrieve %d %s manifest\n", version, file->filename);
-		return NULL;
-	}
+	if (delta_manifest) {
+		if (file->peer != NULL) {
+			delta_manifest_version = file->peer->last_change;
```
jwakre: I am going to condense this to one condition statement:

if (delta_manifest && file->peer != NULL), use delta manifest
else, don't use delta manifest

I can also check for the case when the manifest is unchanged and skip attempting to load a delta manifest.

src/manifest.c
```diff
@@ -1010,7 +1031,7 @@ struct list *recurse_manifest(struct manifest *manifest, struct list *subs, cons
 			version1 = version2;
 		}
 
-		sub = load_manifest(version1, version2, file, manifest, false);
+		sub = load_manifest(version1, version2, file, manifest, false, delta_manifest);
```
jwakre: This attempts to load bundles that have not changed. The load_manifest function will fail to find an unchanged delta manifest and fallback to downloading the entire bundle manifest. I will add a case to skip bundles that have not changed when looking for delta_manifests.

jwakre:
I added a change so that when loading FROM bundles, unchanged bundles will be skipped. Earlier, unchanged bundles would attempt to download a delta manifest and fallback to download a full manifest.

test/functional/update_v2/update-delta-manifest.bats
jwakre: @otaviobp Is my assumption wrong here? Are delta manifests only valid for the N-1 version?

If delta manifests only work for the N-1 verison, I will need to not look for old delta manifests.

matthewrsj: Yes @jwakre right now we are only publishing delta manifests for bundle->previous (not necessarily N-1 or N-10)

matthewrsj: Wait, no that's wrong. Delta manifests are valid for the FROM version to the TO version. I was getting my iterative and delta manifests mixed up again (I've been doing that a lot). If you download a pack from FROM (to your TO) it will have all the records of files used in deltas.

matthewrsj: tl;dr @jwakre is right

jwakre: Ok, thanks for the clarification. I think a test case for the scenario when a bundle has been updated more than once since an update, so I will add that.

src/update.c
jwakre: Below is the timing issue:

#584

jwakre:
Based on offline feedback from @matthewrsj to replace functions that take a boolean delta_manifest argument with wrapper functions, I intend to do the following:

1. Instead of passing the delta_manifest argument to the recurse_manifest and load_manifest functions, I will add a use_delta_manifest variable to the manifest struct which will only be set to true for the FROM manifest.

2. I will create a wrapper function for the manifest_from_file function to abstract loading a delta manifest vs loading a normal manifest.

jwakre:
> Based on offline feedback from @matthewrsj to replace functions that take a boolean delta_manifest argument with wrapper functions, I intend to do the following:
> 
> 1. Instead of passing the delta_manifest argument to the recurse_manifest and load_manifest functions, I will add a use_delta_manifest variable to the manifest struct which will only be set to true for the FROM manifest.

The change described above has been pushed. Adding the 'use_delta_manifest' variable removes most of the API changes.

src/clean.c
```diff
@@ -203,6 +203,16 @@ static bool is_manifest_delta(const char UNUSED_PARAM *dir, const struct dirent
 	return !strncmp(entry->d_name, prefix, prefix_len);
 }
 
+static bool is_delta_manifest(const char UNUSED_PARAM *dir, const struct dirent *entry)
+{
+	static const char delta_str[] = ".D.";
+
+	if (is_manifest(dir, entry) && strstr(entry->d_name, delta_str) != NULL) {
```
matthewrsj: this is a `bool` return function. Just do
```c
return is_manifest(dir, entry) && strstr(entry->d_name, delta_str) != NULL;
```

src/clean.c
```diff
@@ -288,14 +299,22 @@ static int clean_staged_manifests(const char *path, bool dry_run, bool all)
 			continue;
 		}
 
+		/* Delta manifests are only valid when updating FROM the current version
+		 * TO the new version and should not be cached. */
+		string_or_die(&version_dir, "%s/%s", state_dir, name);
```
matthewrsj: `version_dir` gets reassigned to the same thing again before `remove_if(version_dir, dry_run, is_manifest)` below. Remove the reassignment down there.

matthewrsj: This also changes a bit in #574 which should be merged soon, so this will need an update.

src/manifest.c
```diff
@@ -110,7 +110,12 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
 	} else {
 		basedir = MIX_STATE_DIR;
 	}
-	string_or_die(&filename, "%s/%i/Manifest.%s", basedir, version, component);
+
+	if (delta_manifest_version > 0) {
+		string_or_die(&filename, "%s/%i/Manifest.%s.D.%i", basedir, delta_manifest_version, component, version);
```
matthewrsj: Leave a comment here that delta manifests are extracted to the `delta_manifest_version` version directory (which is the `last_change` version) but have the `version` in the name.

src/manifest.c
```diff
@@ -683,34 +689,48 @@ struct manifest *load_manifest(int current, int version, struct file *file, stru
...
-			retried = true;
-			goto retry_load;
+		if (!header_only) {
+			ret = verify_bundle_hash(mom, file);
+		}
+
+		if (ret != 0) {
+			if (retried == false) {
+				remove_manifest_files(file->filename, version, file->hash);
+				retried = true;
```
matthewrsj: /me really wants #211 to happen...

src/manifest.c
```diff
@@ -683,34 +689,48 @@ struct manifest *load_manifest(int current, int version, struct file *file, stru
...
 		if (retried == false) {
-			remove_manifest_files(file->filename, version, file->hash);
-			retried = true;
+			if (delta_manifest) {
+				delta_manifest = false;
+				delta_manifest_version = 0;
+			} else {
+				remove_manifest_files(file->filename, version, file->hash);
+				retried = true;
+			}
```
matthewrsj: ok, so if manifest failed to load and retried is false, we set delta_manifest to false, delta_manifest_version to 0 (essentially turning off delta_manifest options for this manifest) and retry without setting retried to true. This means we will try to download the full manifest and still allow it to retry one more time.

This flow is good, but please add a lengthy comment explaining the flow. It took me a while to figure that out just by reading the code here.

src/manifest.c
```diff
@@ -801,6 +823,17 @@ struct list *create_update_list(struct manifest *server)
 			}
 		}
 
+		/* Delta manifests don't include unchanged files in the FROM file list,
+		 * so files in the TO file list without a peer are not new when the file
+		 * exists on the system. */
+		filename = mk_full_filename(path_prefix, file->filename);
+		ret = verify_file_lazy(filename);
```
matthewrsj: we talked about just using the TO file list as "new or changed" files and combining that metric into one report. Is that feasible to do?

src/manifest.c
```diff
@@ -866,13 +900,30 @@ void link_manifests(struct manifest *m1, struct manifest *m2)
 			continue;
 		} /* else ret > 0  m1/file1 is after m2/file2 */
 		list2 = list2->next;
-		account_new_file();
+
+		/* Delta manifests don't include unchanged files in the FROM file list,
+		 * so files in the TO file list without a peer are only new when they
+		 * don't exist on the system. */
+		filename = mk_full_filename(path_prefix, file2->filename);
+		if (!verify_file_lazy(filename)) {
```
matthewrsj: Same question as above.

src/manifest.c
```diff
@@ -866,13 +900,30 @@ void link_manifests(struct manifest *m1, struct manifest *m2)
...
 	// Capture new files if they are at the tail end of the file list
 	while (list2) {
+		file2 = list2->data;
 		list2 = list2->next;
-		account_new_file();
+
+		/* Delta manifests don't include unchanged files in the FROM file list,
+		 * so files in the TO file list without a peer are only new when they
+		 * don't exist on the system. */
+		filename = mk_full_filename(path_prefix, file2->filename);
```
matthewrsj: Same question as above, but anyways at this point you are repeating code a lot. Make this its own function if you end up keeping it.

src/manifest.c
```diff
@@ -801,6 +803,17 @@ struct list *create_update_list(struct manifest *server)
 			}
 		}
 
+		/* Delta manifests don't include unchanged files in the FROM file list,
+		 * so files in the TO file list without a peer are not new when the file
+		 * exists on the system. */
+		filename = mk_full_filename(path_prefix, file->filename);
```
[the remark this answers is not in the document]
otaviobp: it looks like this comment doesn't make sense anymore after Matthews comment

src/manifest.c
jwakre: We can already count changed files because delta manifests will have entries for them. The problem here is that unchanged files will not be tracked in the FROM file list when using delta manifests. As a result, the old method could not differentiate unchanged files from new ones.

I don't think combining unchanged and new files would be useful.

matthewrsj:
> I have a feeling that a great refactor in create_update_list() is needed to avoid downloading the current manifest from server. We could use current version of the system to decide if we need to update a file or not instead of using the version of that file in the current manifest.

Very good point @otaviobp. That block of code was put in to optimize rename detection iirc, but we don't do things the same way any more. Regardless, if we refactor that code it will need careful review.

src/manifest.c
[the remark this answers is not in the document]
otaviobp: My suggestion isn't going to work. If an old bundle is included in a new version of a new bundle, the version of the file will be older than current version and it will be installed anyway.

I was trying to understand the best approach to that and I wrote a patch, but I still have 2 tests failing. https://github.com/otaviobp/swupd-client/commit/fcbf00934f13fde580470cbaad7ce7e23a93c380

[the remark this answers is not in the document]
jwakre: I think we can solve this case by doing the following:

1. Add a variable to the file struct that is set to true when a file is included

2. Initialize variable for loaded files to false

3. Mark files pulled in by included bundles true (in add_subscriptions).

4. Check for included files with an older version than the current.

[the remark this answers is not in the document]
jwakre: Since we can't use delta manifests to decide information about file updates, I no longer see a need for the link_manifests function. I think we could consolidate the file statistic counting to create_update_list and remove the link_manifests function.

matthewrsj:
Going to comment here about a few cases that need to be handled when not checking against the previous manifest and instead using the version of the installed OS.
1. file in the to manifest is a newer version than the OS, **add to update list**.
2. file in the to manifest is an older version than the OS, but the manifest was added as a new include and was not tracked on the system, and the file itself is only provided by the newly included bundle, if the file is *missing* or *invalid* on the file system, **add to update list**.
3. file in the to manifest is an older version than the OS and the manifest was already tracked on the file system, **do not add to update list** regardless of if that file is *also* provided by a newly included bundle.

matthewrsj:
A note on minversion bumps:
```
13:59 < mrsj> the only time a file will get an updated version number with no hash or content change or file name change is when you have a minversion update
14:00 < mrsj> which raises an issue...
14:00 < mrsj> our current plan will cause ALL OS files to be downloaded for a minversion update
14:00 < mrsj> because a minversion marks all files with a new version
14:00 < mrsj> :/
14:01 < mrsj> so our plan would put those under Case1, where the file is greater than the current OS version
14:01 < mrsj> so for every tracked file you will be adding those to the manifest
14:01 < mrsj> CURRENTLY...
14:01 < mrsj> in manifest.c,
14:01 < mrsj> we can detect minversion updates because we know the file version changed by the hash and flags stayed the same
14:01 < mrsj> can't do that when we aren't using the old manifests any more
14:02 < mrsj> we need another solution
14:02 < mrsj> one solution:
14:02 < mrsj> add a manifest header field to the MoM
14:02 < mrsj> minversion:\t<minversion>
14:02 < mrsj> and if <minversion> is greater than current OS version we do the fallback to verifying all files with versions greater than our own OS version
14:03 < mrsj> make sense jwakre, otaviobp?
14:03 < mrsj> this is unfortunate
14:03 < mrsj> but we should be able to add a minversion field to the MoM in mixer without requiring a format bump for the client
14:03 < mrsj> because swupd ignores header fields it doesn't know about
14:04 < jwakre> What is a minversion?
14:05 < mrsj> a minversion is when all content is "updated" on the server-side
14:05 < mrsj> by mixer
14:05 < otaviobp> mrsj: this sounds correct to me
14:05 < mrsj> so all file versions are updated to that "minversion" build regardless of whether they actually physically changed or not
14:06 < jwakre> oh, so a minimum version for all files on the system
14:06 < mrsj> right now in manifest.c (search for "minversion") the client falls back to verifying all files in the update against the system
14:06 < otaviobp> gtkramer said they usually do it once a mounth, so it's releavant to be optimized for that
14:06 < mrsj> to prevent having to download the ENTIRE OS
14:06 < otaviobp> I thought that this was a rare thing, but it's not
14:06 < mrsj> no it is often
14:06 < mrsj> because it allows devops to free up space on their builder
14:06 < mrsj> :/
14:06 < mrsj> apparently disk space is not cheap
14:06 < otaviobp> disk isn't that cheap, apparently 
14:07 < mrsj> despite what everyone tells us ;P
14:09 < mrsj> that was a lot of writing... I should have put it in the PR...
14:09  * mrsj copies
```

jwakre:
PR Updates:

1. Rebased on to master_new_manifests branch which includes iterative manifests and the refactored update list.

2. @otaviobp's feedback was addressed by the refactored update list PR which has already been committed.

3. @matthewrsj 's feedback:
* Added comments for delta manifest flow
* Removed unnecessary version_dir assignment in clean.c
* Removed condition statement in is_delta_manifest function in clean.c
* "TO" file list feedback was addressed by refactored update list PR.

src/update.c
```diff
@@ -447,38 +469,20 @@ static int main_update()
 	server_manifest->files = files_from_bundles(server_manifest->submanifests);
 	server_manifest->files = consolidate_files(server_manifest->files);
 
-	set_subscription_versions(server_manifest, current_manifest, &update_subs);
```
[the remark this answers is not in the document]
castulo: cool

test/functional/testlib.bash
```diff
@@ -2071,6 +2071,10 @@ update_bundle() {
 		contentsize=$((contentsize + (new_fsize - fsize)))
 		# update the zero-pack with the new file
 		add_to_pack "$bundle" "$version_path"/files/"$new_fhash"
+
+		# TODO: Delta packs should be generated for an update from the current version
+		#       to the latest version. If a bundle is updated more than once in a single
+		#       update, an incorrect delta pack will be generated. Issue 615
```
castulo: good catch

src/update.c
```diff
@@ -430,6 +436,7 @@ static int main_update()
 	 * First load up the old (current) manifests. Statedir could have been cleared
 	 * or corrupt, so don't assume things are already there. Updating subscribed
 	 * manifests is done as part of recurse_manifest */
+	link_submanifests(current_manifest, server_manifest, initial_subs, initial_subs, false);
```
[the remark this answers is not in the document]
jwakre: To find where a delta manifest is extracted to inside the state dir, I need to know its corresponding peer bundle's last change. The link_submanifests function must be called earlier so that the peer is set before I look for the delta manifest.

The peer->last_change is used to set the 'delta_manifest_state_dir' variable in load_manifests.

test/functional/update/update-delta-manifest-and-full.bats
```diff
@@ -0,0 +1,60 @@
...
+		Starting download of remaining update content. This may take a while...
+		Finishing download of update content...
+		Staging file content
+		Applying update
+		Update was applied.
+		Calling post-update helper scripts.
+		Update successful. System updated from version 10 to version 20
+	EOM
+	)
+	assert_in_output "$expected_output"
```
castulo: if possible is better to use _assert_is_output_ than _assert_in_output_

src/update.c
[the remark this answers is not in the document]
jwakre: I could move this above the comment though...

test/functional/update/update-delta-manifest-invalid.bats
```diff
@@ -0,0 +1,51 @@
...
+test_setup() {
+
+	create_test_environment "$TEST_NAME" 10
+	create_bundle -L -n "bundle1" -v 10 -f /file1 "$TEST_NAME"
+
+	create_version "$TEST_NAME" 20 10 staging
+	update_bundle -i "$TEST_NAME" bundle1 --update /file1
+
+	# Overwrite delta manifest with invalid file
+	sudo sh -c "echo 1 > $WEBDIR/20/20/Manifest.bundle1.D.10"
```
castulo: it is ok like tit is now, but just FYI there is a function in the test library to do that:
```
Usage:
    write_to_protected_file [-a] <file> <stream>

Options:
    -a    If used the text will be appended to the file, otherwise it will be overwritten
```

test/functional/update/update-delta-manifest-invalid.bats
```diff
@@ -0,0 +1,51 @@
...
+		Starting download of remaining update content. This may take a while...
+		Finishing download of update content...
+		Staging file content
+		Applying update
+		Update was applied.
+		Calling post-update helper scripts.
+		Update successful. System updated from version 10 to version 20
+	EOM
+	)
+	assert_in_output "$expected_output"
```
castulo: same comment as above

jwakre:
Thanks for the feedback @castulo. I made the following changes to address your feedback:

* Moved link_submanifests above the recurse_manifest comment
* Changed assert_in_output -> assert_is_output for added tests
* Replaced the echo file write with write_to_protected_file

otaviobp:
Just wait for @matthewrsj to check if he has something else to add and then we can merge this.

PR #585, opened 2018-09-27
Add minversion support to testlib

jwakre:
Soon, a minversion header field will be included in the MoM. This change
updates the test library to add a minversion header field to MoMs and adds
functions to create a minversion update. Also, the skip-verified-fullfiles
test was replaced by the update-minversion test. The update-minversion test
verifies that unchanged files with a version bump caused by a minversion
update are skipped during an update.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -170,6 +170,7 @@ BATS = \
 	test/functional/update/update-include.bats \
 	test/functional/update/update-include-old-bundle.bats \
 	test/functional/update/update-include-old-bundle-with-tracked-file.bats \
+	test/functional/update/update-minversion.bats \
 	test/functional/update/update-missing-os-core.bats \
 	test/functional/update/update-newest-deleted.bats \
 	test/functional/update/update-newest-ghosted.bats \
@@ -178,7 +179,6 @@ BATS = \
 	test/functional/update/update-re-update-bad-os-release.bats \
 	test/functional/update/update-re-update-required.bats \
 	test/functional/update/update-skip-scripts.bats \
-	test/functional/update/update-skip-verified-fullfiles.bats \
 	test/functional/update/update-slow-server.bats \
 	test/functional/update/update-statedir-bad-hash.bats \
 	test/functional/update/update-status.bats \
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -423,6 +423,72 @@ create_tar() {
 
 }
 
+# Inside the version directory that matches the minversion, update the MoM, bundle manifests, and
+# fullfiles for the minversion update.
+# Parameters:
+# - minversion: new minversion.
+update_minversion() {
+	local minversion=$1
+	local minversion_dir="$WEBDIR"/"$minversion"
+	local line
+	local bundle
+	local version
+
+	# Read the MoM in the minversion's directory to determine bundles that need updates
+	while read line; do
+		# Header fields will not have an integer in 3rd column, so they will be skipped
+		version=$(echo "$line" | awk '{print $3;}')
+		if [[ ! "$version" =~ ^[0-9]+$ ]]; then
+			continue
+		fi
+
+		# Copy bundles older than the minversion to the minversion directory
+		bundle=$(echo "$line" | awk '{print $4;}')
+		if [ "$version" -lt "$minversion" ]; then
+			sudo cp "$WEBDIR"/"$version"/Manifest."$bundle" "$minversion_dir"
+		fi
+
+		# Copy old fullfiles from bundle manifests and update their versions to
+		# reflect the new minversion
+		update_bundle_minversion "$bundle" "$minversion"
+
+	done < "$minversion_dir"/Manifest.MoM
+
+	# Update manifest hashes/versions in MoM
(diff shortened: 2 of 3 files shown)


test/functional/testlib.bash
```diff
@@ -411,6 +477,7 @@ create_manifest() {
 	local path=$1
 	local name=$2
 	local format=${3:-1}
+	local minversion="0"
```
matthewrsj: Is this meant to default to '0' but set to $3 if it is provided?

test/functional/update/update-minversion.bats
```diff
@@ -0,0 +1,38 @@
...
+	create_version "$TEST_NAME" 20 10 staging
+	update_minversion 20
+}
+
+@test "skip fullfile download for unchanged file in minversion update " {
+
+	run sudo sh -c "$SWUPD update $SWUPD_OPTS"
+	assert_status_is 0
+
+	expected_output=$(cat <<-EOM
```
matthewrsj: This test is verifying that the file wasn't downloaded just by checking the output of swupd. Can you also *remove* the fullfile from the webdir before performing the update so the update will fail if it actually tries to download the file.

What sometimes happens when swupd output changes is that the developer comes and blanket fixes all the failing output tests without verifying if it should be updated or not. By removing the fallback we can make sure the test will always fail if it does the wrong thing.

test/functional/testlib.bash
jwakre: I intended to default the manifest minversion to 0 at creation. I thought the format variable was for the format header field.

test/functional/update/update-minversion.bats
jwakre: Ok, I'll clean the files directory.

test/functional/testlib.bash
matthewrsj: Sorry, I didn't mean to reference format variable, I just meant to ask if you intended this to be set via a user-passed argument.

jwakre: I didn't support variable minversions at creation because it would add complexity to support functionality that I don't think I need for delta manifests tests.

I haven't fully scoped the requirements, but I would need to make sure that all added bundles adhere to the minversion. Currently, my implementation performs the minversion update all at once.

If you want testlib minversion support to be more robust, I could look closer.

matthewrsj: Nope this is fine for now.

test/functional/update/update-minversion.bats
jwakre: done in new revision

Makefile.am
```diff
@@ -204,6 +204,7 @@ BATS = \
 	test/functional/search_v2/search-content-check-positive.bats \
 	test/functional/search_v2/search-client-certificate.bats \
 	test/functional/update/apply-full-file-delta/test.bats \
+	test/functional/update/update-minversion.bats \
```
castulo: this should be in alphabetical order

jwakre: I'll fix that

test/functional/testlib.bash
```diff
@@ -741,6 +814,18 @@ update_manifest() {
 	version | previous | filecount | timestamp | contentsize)
 		sudo sed -i "s/$key.*/$key:\\t$var/" "$manifest"
 		;;
+	minversion)
+		# Update minversion field which should only exist in MoM
+		sudo sed -i "s/$key.*/$key:\\t$var/" "$manifest"
```
otaviobp: use:
sudo sed -i "s/^$key:.*/$key:\\t$var/" "$manifest"

Take a look at #586

test/functional/update/update-minversion.bats
```diff
@@ -0,0 +1,40 @@
...
+		Update started.
+		Preparing to update from 10 to 20
+		Downloading packs...
+		Statistics for going from version 10 to version 20:
+		    changed bundles   : 2
+		    new bundles       : 0
+		    deleted bundles   : 0
+		    changed files     : 4
```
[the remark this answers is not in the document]
jwakre: 2 Changed bundles: 
* bundle1
* os-core

4 changed files: Every file in the bunde1 and os-core manifests was updated:
* /core
* /usr/share/clear/bundles/os-core
* /file1
* /usr/share/clear/bundles/bundle1

With the current implementation changed bundles are counted when the TO version > FROM version. These 'changed bundles' are only added to the update list when their hashes are not the same.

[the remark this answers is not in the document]
jwakre: Do we want to change this? I am about to refactor the implementations for update statistics and the update list to enable delta manifests.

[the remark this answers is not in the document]
castulo: Pff you are right, I forgot we add the os-core bundle by default to the test environment unless specified otherwise.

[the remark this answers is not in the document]
castulo: > Do we want to change this? I am about to refactor the implementations for update statistics and the update list to enable delta manifests.

That's more of a question for @matthewrsj

test/functional/testlib.bash
jwakre: Nice! I'll fix that

PR #587, opened 2018-09-28
Use release os-core manifest for update boot tests

jwakre:
When creating the test environment, some files and directories, like
/usr/lib, are created without corresponding entries in the os-core
manifest. This is a problem for update tests that add boot files, like
/usr/lib/kernel/file. Without an entry for /usr/lib in the FROM os-core
manifest, an update will consider /usr and /usr/lib new files even when
they already exist on the system. The '-r' argument can be passed to
create_test_environment to create a more complete os-core manifest.

### DIFF
--- a/test/functional/update/update-boot-file.bats
+++ b/test/functional/update/update-boot-file.bats
@@ -4,7 +4,7 @@ load "../testlib"
 
 test_setup() {
 
-	create_test_environment "$TEST_NAME"
+	create_test_environment -r "$TEST_NAME"
 	create_version -p "$TEST_NAME" 100 10
 	update_bundle "$TEST_NAME" os-core --add /usr/lib/kernel/testfile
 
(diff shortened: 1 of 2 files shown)


castulo:
@jwakre I'm curious why you say this is required for these tests?
I originally didn't add that to these tests because update tests are slow, and using the _-r_ option makes them a even slower, and since I considered that not being important for what these tests were validating I left that out.

jwakre:
> @jwakre I'm curious why you say this is required for these tests?
> I originally didn't add that to these tests because update tests are slow, and using the _-r_ option makes them a even slower, and since I considered that not being important for what these tests were validating I left that out. Adding the -r will create a test environment that more closely resembles a real environment, but it comes with the price of a slower test.

I am rebasing my changes that refactor the update list onto the updated test library and correcting failures. This test was failing earlier and I thought it was because of the shortcut it takes in the manifest. I tried to replicate the failure, but it seems to be working fine now without it. I'm fine dropping this PR.

jwakre:
I am rebasing my changes that refactor the update list onto the updated test library and correcting failures. This test was failing earlier and I thought it was because of the shortcut it takes in the manifest. I tried to replicate the failure, but it seems to be working fine now without it. I'm fine dropping this PR.

PR #593, opened 2018-10-02
Fixing memory leak in retrieve_manifests

castulo:
This commit fixes a memory leak with the dir variable in the retrieve_manifests function.


castulo:
As requested by @otaviobp this commit was part of PR #565, but since it is really not related to that PR it was included in its own PR.

PR #594, opened 2018-10-02
Adding the --keepcache option when doing updates

castulo:
After running an update, the state directory is cleaned up.
Sometimes it is useful to be able to peek at the state directory
after an update, this is specially true when testing updates,
this way tests can verify what manifests were downloaded during
the update.

This commit adds the --keepcache option to swupd update so it can
be used when writing tests for updates.

### DIFF
--- a/src/update.c
+++ b/src/update.c
@@ -545,7 +545,9 @@ clean_curl:
 	}
 	/* clean up our cached content from the update. It is likely much more than
 	 * we need and the clean helps us prevent cache bloat. */
-	clean_statedir(false, false);
+	if (!keepcache) {
+		clean_statedir(false, false);
+	}
 	free_subscriptions(&latest_subs);
 	swupd_deinit();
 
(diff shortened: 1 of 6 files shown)


src/update.c
```diff
@@ -622,6 +625,7 @@ static void print_help(const char *name)
 	fprintf(stderr, "   -n, --nosigcheck        Do not attempt to enforce certificate or signature checking\n");
 	fprintf(stderr, "   -I, --ignore-time       Ignore system/certificate time when validating signature\n");
 	fprintf(stderr, "   -S, --statedir          Specify alternate swupd state directory\n");
+	fprintf(stderr, "   -k, --keepcache         Do not delete the swupd state directory after updating the system\n");
```
otaviobp: We don't really delete the state directory. Can you say "delete the swupd state directory content ..." or "Do not clean the swupd state directory cache..."

castulo: sure

castulo:
so @otaviobp, regarding adding the flag to swiped.1.rst, I thought of this when I initially worked on this commit, the reason why I decided no to add it to the manual is because looking at what is already in the manual regarding updates, there are actually a lot of the flags missing there, from the 20 options that are shown when using _swupd update --help__, only 5 are shown in the manual. So I figured that the manual only contains the most important ones, and --keepcache is not that important.

castulo:
I didn't realize there was an _OPTIONS_ section that listed the common flags, I'll add --keepcache to the update list

castulo:
@otaviobp I updated  docs/swupd.1.rst with the --keepcache flag.

PR #596, opened 2018-10-02
Run extraction in a thread

evil-brain:


src/download.c
```diff
@@ -109,6 +112,27 @@ struct multi_curl_file {
...
 
+/*
+ * Used for wrapping the success callback and tracking return values from threads
+ * and a pointer to the associated return value's file data, to re-schedule on failure
+ */
+struct multithread_data {
+	struct multi_curl_file *file;   /* Curl file information */
+	swupd_curl_success_cb callback; /* Holds original success callback to be wrapped */
+
+	bool retval; /*return value from callback*/
```
otaviobp: Can't you move retval to struct multi_curl_file and then you won't need to have this struct?
And instead of mt_handles you can iterate curl_hashmap using HASHMAP_FOREACH to check all files.
Then you can get rid of the success wrapper too

src/lib/tp.c
```diff
@@ -0,0 +1,152 @@
...
+
+	tp->num_threads = num_threads;
+	if (pipe2(tp->pipe_fds, O_DIRECT | O_CLOEXEC) < 0) {
+		goto error;
+	}
+
+	// Create threads
+	for (i = 0; i < num_threads; i++) {
+		if (pthread_create(&tp->threads[i], NULL, thread_run, &tp->pipe_fds[0]) != 0) {
+			tp->num_threads = i;
```
jwakre: Is there a reason that tp->num_threads is set earlier in the function and inside the loop? Can we remove it from the loop? Also, the value i would be from 0 to (num_threads - 1). Is that correct? So if num_threads=5, the final value set in tp->num_threads would be 4.

otaviobp: The num_threads is updated in the loop if pthread_create fails, so we can release the correct number of threads in tp_complete() call in error_threads goto. At this point we have 'i' threads that were successfully created.

jwakre: Ahhh, i see nvm

src/download.c
```diff
@@ -166,6 +180,12 @@ void *swupd_curl_parallel_download_start(size_t max_xfer)
 		goto error;
 	}
 
+	/*
+	 * Libarchive is not thread safe when using the archive_disk_write api.
+	 * So for now, use only 1 thread to handle untar/extraction
+	 */
+	h->thpool = tp_start(1);
```
jwakre: Would it make more sense to use a macro for the number of threads?

src/lib/tp.c
```diff
@@ -0,0 +1,152 @@
...
+	void *data;
+};
+
+static void *thread_run(void *data)
+{
+	int *fd;
+	ssize_t r;
+	struct task task;
+
+	fd = data;
```
tmarcu: You could check that `*fd` is actually a valid integer > 0 before trying to read anything here, and exit early if it's not.

src/lib/tp.c
tmarcu: You could also just do `tp->num_threads++` under the `if` statement here to just count successful cases; both work just throwing out more options :+1:

src/lib/tp.h
```diff
@@ -0,0 +1,47 @@
...
+/*
+ * Wait for all scheduled tasks to be completed, finishes all threads and
+ * release all memory used by the thread pool.
+ */
+void tp_complete(struct tp *tp);
+
+/*
+ * Get number of CPU cores.
+ */
+int tp_get_num_cores();
```
tmarcu: Where is this func defined?

otaviobp:
I'm closing this PR for now. We can open that back after the code rework

PR #599, opened 2018-10-03
Add more compile options to info and --version

matthewrsj:
When invoking swupd info or swupd --version print more information about
compile-time configuration. This feature was requested after a bug was
introduced to an unreleased Clear Linux build when swupd was built with
the wrong format ID. This allows QA to determine exactly which
compile-time options were set during runtime without inspecting the
binary.

### DIFF
--- a/src/swupd-build-opts.h
+++ b/src/swupd-build-opts.h
@@ -46,7 +46,67 @@
 #define OPT_BSDTAR "-BSDTAR"
 #endif
 
+#ifdef SWUPD_WITH_XATTRS
+#define OPT_XATTRS "+XATTRS"
+#else
+#define OPT_XATTRS "-XATTRS"
+#endif
+
+#ifdef SWUPD_TAR_SELINUX
+#define OPT_SELINUX "+TAR_SELINUX"
+#else
+#define OPT_SELINUX "-TAR_SELINUX"
+#endif
+
+#ifdef OS_IS_STATELESS
+#define OPT_STATELESS "+STATELESS"
+#else
+#define OPT_STATELESS "-STATELESS"
+#endif
+
+#ifdef CONTENTURL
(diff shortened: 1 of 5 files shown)


src/helpers.c
```diff
@@ -1098,3 +1099,9 @@ int untar_full_download(void *data)
 	}
 	return err;
 }
+
+void print_compile_opts(void)
+{
+	printf("Compile-time options: %s\n", BUILD_OPTS);
+	printf("Comile-time configuration:\n%s\n", BUILD_CONFIGURE);
```
gvancuts: s/comile/compile

otaviobp:
> Eh, I like having a swupd info command that gives info. It probably wouldn't make sense to require users not on a mirror to call the swupd mirror command to see their system. The idea is that the information printed from these commands can diverge at some point. swupd mirror might just print the URLs while swupd info might additionally print the format.

true

matthewrsj:
ok updated

src/swupd-build-opts.h
```diff
@@ -46,7 +46,73 @@
...
+#ifdef FORMATID
+#define OPT_FORMAT FORMATID
+#else
+#define OPT_FORMAT "no default set"
+#endif
+
+#ifdef SYSTEMD_UNITDIR
+#define OPT_SYSTEMD SYSTEMD_UNITDIR
+#else
+#define OPT_SYSTEMD "no default set"
```
jwakre: Should this be included in BUILD_CONFIGURE? I don't see this used anywhere.

matthewrsj: You are right. Removed from this commit and I added a commit to remove it from configure.ac and Makefile.am since it looks like the option isn't used anywhere anymore.

PR #602, opened 2018-10-04
Verify server version is read in check_versions

### LINKED ISSUE #589: Check for errors when comparing versions on main_update() in update.c
Reported swupd output:

Version on server (-404) is not newer than system version (60)

In: https://github.com/clearlinux/swupd-client/blob/c9ee84c70c7c6dc386fb9f9e25bba32f951ba47d/src/update.c#L302

castulo:
check_versions is a function that performs some validations using
the target system current version, the server version, and the
requested_version (if any), but if there was an error while retrieving
the server version no action would be taken.

This commit adds a verification to make sure the server version was
able to be identified.

Closes #589


otaviobp:
check_update is validating the versions too. I think it's a good idea if you can merge those 2 functions in 1. I'm trying to avoid code duplication. The only difference is that check_update doesn't have a requested_version.

castulo:
> check_update is validating the versions too. I think it's a good idea if you can merge those 2 functions in 1. I'm trying to avoid code duplication. The only difference is that check_update doesn't have a requested_version.

I decided not to merge both functions, I think they make more sense separate as they are, but I moved the validations of the server and current versions to the read_versions function, this should take care of the duplicated code.

PR #604, opened 2018-10-04
Make parallel downloads flag global, used for fullfile downloads too.

ahkok:
Adjust getopts, since this is relatively new. -D stays the same
and now applies to fullfiles too.


evil-brain:
Seems to help when using fullfiles, and does what it says it does. Still testing a bit, but tomorrow this will get a +1 after conflicts are fixed.

otaviobp:
merged

PR #605, opened 2018-10-05
Avoid print time stats if the queue is empty.

### LINKED ISSUE #600: Swupd verify segfault with -t into blank directory
Swupd verify segfaults in this specific corner case:
* `/home/blovin/swupd_verify_testing/testing1` is an empty directory
* Found on swupd 3.17.15

```
sudo -E swupd verify -t --fix --picky --path=/home/blovin/swupd_verify_testing/testing1 --contenturl=https://cdn.download.clearlinux.org/update --versionurl=https://cdn.download.clearlinux.org/update --format=25 --statedir=/home/blovin/swupd_verify_testing/testing1/var/lib/swupd

Raw elapsed time stats:
Segmentation fault
```

This is a corner case, and probably related to the memory leaks from using -t #575 (though that's with bundle-add).

ericho:
If the -t parameter is set for a given empty directory, then
the time queue will be empty causing that the `TAILQ_FOREACH_REVERSE`
macro segfaults.

The proposed solution is to check if the head is NULL at the beginning
of the `print_time_stats` function.

Fixes #600

Signed-off-by: Erich Cordoba <erich.cm@yandex.com>

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -131,7 +131,7 @@ void grabtime_stop(timelist *head)
 
 void print_time_stats(timelist *head)
 {
-	if (verbose_time == false) {
+	if (verbose_time == false || TAILQ_FIRST(head) == NULL) {
 		return;
 	}
 


otaviobp:
Nice fix :)
+1 for the commit, but it didn't close #600 yet. In verify_main If sys_version is incorrect the timelist won't be initialized before calling print_stats(), so we may still have invalid memory access.

PR #606, opened 2018-10-08
time: initialize time list before using it

### LINKED ISSUE #600: Swupd verify segfault with -t into blank directory

otaviobp:
Fixes #600

@ericho Can you take a look at this commit, as it's related to #605


ericho:
LGTM!

PR #611, opened 2018-10-09
Hashdump - minor fixes

otaviobp:


src/hashdump.c
```diff
@@ -102,36 +97,33 @@ int hashdump_main(int argc, char **argv)
...
 	} else {
-		printf("%s\n", file->hash);
-		if (file->is_dir) {
-			if (is_directory_mounted(fullname)) {
-				fprintf(stderr, "!! dumped hash might not match a manifest hash because a mount is active\n");
-			}
+		printf("%s\n", file.hash);
+		if (file.is_dir && is_directory_mounted(fullname)) {
+			fprintf(stderr, "!! dumped hash might not match a manifest"
+					"hash because a mount is active\n");
```
jwakre: I think this will print: "manifesthash" instead of "manifest hash"

otaviobp: thanks. fixed.

src/hashdump.c
```diff
@@ -83,35 +83,30 @@ int hashdump_main(int argc, char **argv)
...
-	ret = set_path_prefix(NULL);
-	if (!ret) {
-		free_string(&file->filename);
+	if (!set_path_prefix(NULL)) {
 		free(file);
-		return EXIT_FAILURE;
+		return EINIT_GLOBALS;
 	}
 
+	file.filename = strdup_or_die(argv[optind]);
```
jwakre: I think this line should be moved to the commit that changes file to static allocation.

otaviobp: fixed.

PR #617, opened 2018-10-11
Improve error reporting when there is a failure during an update

### LINKED ISSUE #323: abort operation when network fails and return network-related  error code
When the network goes down, I'd expect swupd to retry the same request a few times and then give up entirely (ideally configurably, see issue #322). The overall return code then should indicate a network error (see issue #296).

What I see happening instead when simulating network issues starting with the pack-os-core-from-10.tar is that it tries to download the pack only once (why no retry?) and then it falls back to downloading individual files. In addition, the exit code is the generic value 1, i.e. the caller cannot determine that the problem was because of a network issue.

HTTP log:
```
2017-10-20 16:29:22,892 - oe-selftest - INFO - "GET /10/Manifest.MoM.tar HTTP/1.1" 200 -
2017-10-20 16:29:22,908 - oe-selftest - INFO - "GET /20/Manifest.MoM.tar HTTP/1.1" 200 -
2017-10-20 16:29:22,922 - oe-selftest - INFO - "GET /10/Manifest.os-core.tar HTTP/1.1" 200 -
2017-10-20 16:29:23,114 - oe-selftest - INFO - "GET /20/Manifest-os-core-delta-from-10 HTTP/1.1" 200 -
2017-10-20 16:29:23,151 - oe-selftest - INFO - code 500, message test server is intentionally down
2017-10-20 16:29:23,151 - oe-selftest - INFO - "GET /20/pack-os-core-from-10.tar HTTP/1.1" 500 -
2017-10-20 16:29:23,165 - oe-selftest - INFO - code 500, message test server is intentionally down
2017-10-20 16:29:23,166 - oe-selftest - INFO - "GET /20/files/fbfc8883ed443f3dfbc5c97c5f533b226f453c37f1bb922b22800c5db5a8f844.tar HTTP/1.1" 500 -
2017-10-20 16:29:23,178 - oe-selftest - INFO - code 500, message test server is intentionally down
2017-10-20 16:29:23,178 - oe-selftest - INFO - "GET /20/files/44d4a9eb71ab2e552fe189f638b04c39ea47a63108a5c7f84d7a40aaa28c3b38.tar HTTP/1.1" 500 -
2017-10-20 16:29:23,190 - oe-selftest - INFO - code 500, message test server is intentionally down
2017-10-20 16:29:23,190 - oe-selftest - INFO - "GET /20/files/e4e5e0ae2c60df90f913e58bbdcfdaf5bd6f4762702c092a19074ebf6d410b08.tar HTTP/1.1" 500 -
2017-10-20 16:29:23,202 - oe-selftest - INFO - code 500, message test server is intentionally down
2017-10-20 16:29:23,202 - oe-selftest - INFO - "GET /20/files/c702727015ed7fb561646f0a5f6280b95e7c2058edacce1ee7a60b586f1bc250.tar HTTP/1.1" 500 -
2017-10-20 16:29:23,213 - oe-selftest - INFO - code 500, message test server is intentionally down
2017-10-20 16:29:23,214 - oe-selftest - INFO - "GET /20/files/f6feec20bdf0ff89dff3688a5d9c662276f33f69691d9d91ec404d251ccc877b.tar HTTP/1.1" 500 -
...
```

Output:
```
swupd-client software update 3.12.0
   Copyright (C) 2012-2017 Intel Corporation

Update started.
Preparing to update from 10 to 20
Running script 'Pre-update'
Downloading packs...
	...100%

Statistics for going from version 10 to version 20:

    changed bundles   : 1
    new bundles       : 0
    deleted bundles   : 0

    changed files     : 11
    new files         : 2
    deleted files     : 1

File /etc/os-release was not in a pack
	...92%
Error for fbfc8883ed443f3dfbc5c97c5f533b226f453c37f1bb922b22800c5db5a8f844 download: Received 500 response
Error for 44d4a9eb71ab2e552fe189f638b04c39ea47a63108a5c7f84d7a40aaa28c3b38 download: Received 500 response
Error for e4e5e0ae2c60df90f913e58bbdcfdaf5bd6f4762702c092a19074ebf6d410b08 download: Received 500 response
Error for c702727015ed7fb561646f0a5f6280b95e7c2058edacce1ee7a60b586f1bc250 download: Received 500 response
Error for f6feec20bdf0ff89dff3688a5d9c662276f33f69691d9d91ec404d251ccc877b download: Received 500 response
Error for fdd9cdad7808d054db30f814df78b79b987f5a940f881de68a1bf708b58fea68 download: Received 500 response
Error for 10bd4b83febc4cf5dac47dc48d315b861743d2a54ca5d2f433ccee1842d73aa3 download: Received 500 response
Error for 9dd3a9120a7b346f512426efe61f7d1ff6ac01aa430676320b1f794c64de1650 download: Received 500 response
Error for eaf9a9fcf2c7d73d256229a992511bddd8788718b2e5ef75e41fa6c4c107835f download: Received 500 response
Error for 28cb7c6cad23920f344157f48ae353caca34e9a2f5cff934a4fdaaf7509f2062 download: Received 500 response
Error for 699eba4e65f383e8f34f79004652a25ea5346eaa69dc48019d5b46ab38056afa download: Received 500 response
Starting download retry #1
	...100%
Error for 699eba4e65f383e8f34f79004652a25ea5346eaa69dc48019d5b46ab38056afa download: Received 500 response
Error for 28cb7c6cad23920f344157f48ae353caca34e9a2f5cff934a4fdaaf7509f2062 download: Received 500 response
Error for eaf9a9fcf2c7d73d256229a992511bddd8788718b2e5ef75e41fa6c4c107835f download: Received 500 response
Error for 9dd3a9120a7b346f512426efe61f7d1ff6ac01aa430676320b1f794c64de1650 download: Received 500 response
Error for 10bd4b83febc4cf5dac47dc48d315b861743d2a54ca5d2f433ccee1842d73aa3 download: Received 500 response
Error for fdd9cdad7808d054db30f814df78b79b987f5a940f881de68a1bf708b58fea68 download: Received 500 response
Error for f6feec20bdf0ff89dff3688a5d9c662276f33f69691d9d91ec404d251ccc877b download: Received 500 response
Error for c702727015ed7fb561646f0a5f6280b95e7c2058edacce1ee7a60b586f1bc250 download: Received 500 response
Error for e4e5e0ae2c60df90f913e58bbdcfdaf5bd6f4762702c092a19074ebf6d410b08 download: Received 500 response
Error for 44d4a9eb71ab2e552fe189f638b04c39ea47a63108a5c7f84d7a40aaa28c3b38 download: Received 500 response
Error for fbfc8883ed443f3dfbc5c97c5f533b226f453c37f1bb922b22800c5db5a8f844 download: Received 500 response
Starting download retry #2
	...100%
Error for fbfc8883ed443f3dfbc5c97c5f533b226f453c37f1bb922b22800c5db5a8f844 download: Received 500 response
Error for 44d4a9eb71ab2e552fe189f638b04c39ea47a63108a5c7f84d7a40aaa28c3b38 download: Received 500 response
Error for e4e5e0ae2c60df90f913e58bbdcfdaf5bd6f4762702c092a19074ebf6d410b08 download: Received 500 response
Error for c702727015ed7fb561646f0a5f6280b95e7c2058edacce1ee7a60b586f1bc250 download: Received 500 response
Error for f6feec20bdf0ff89dff3688a5d9c662276f33f69691d9d91ec404d251ccc877b download: Received 500 response
Error for fdd9cdad7808d054db30f814df78b79b987f5a940f881de68a1bf708b58fea68 download: Received 500 response
Error for 10bd4b83febc4cf5dac47dc48d315b861743d2a54ca5d2f433ccee1842d73aa3 download: Received 500 response
Error for 9dd3a9120a7b346f512426efe61f7d1ff6ac01aa430676320b1f794c64de1650 download: Received 500 response
Error for eaf9a9fcf2c7d73d256229a992511bddd8788718b2e5ef75e41fa6c4c107835f download: Received 500 response
Error for 28cb7c6cad23920f344157f48ae353caca34e9a2f5cff934a4fdaaf7509f2062 download: Received 500 response
Error for 699eba4e65f383e8f34f79004652a25ea5346eaa69dc48019d5b46ab38056afa download: Received 500 response
Starting download retry #3
	...100%
Error for 699eba4e65f383e8f34f79004652a25ea5346eaa69dc48019d5b46ab38056afa download: Received 500 response
Error for 28cb7c6cad23920f344157f48ae353caca34e9a2f5cff934a4fdaaf7509f2062 download: Received 500 response
Error for eaf9a9fcf2c7d73d256229a992511bddd8788718b2e5ef75e41fa6c4c107835f download: Received 500 response
Error for 9dd3a9120a7b346f512426efe61f7d1ff6ac01aa430676320b1f794c64de1650 download: Received 500 response
Error for 10bd4b83febc4cf5dac47dc48d315b861743d2a54ca5d2f433ccee1842d73aa3 download: Received 500 response
Error for fdd9cdad7808d054db30f814df78b79b987f5a940f881de68a1bf708b58fea68 download: Received 500 response
Error for f6feec20bdf0ff89dff3688a5d9c662276f33f69691d9d91ec404d251ccc877b download: Received 500 response
Error for c702727015ed7fb561646f0a5f6280b95e7c2058edacce1ee7a60b586f1bc250 download: Received 500 response
Error for e4e5e0ae2c60df90f913e58bbdcfdaf5bd6f4762702c092a19074ebf6d410b08 download: Received 500 response
Error for 44d4a9eb71ab2e552fe189f638b04c39ea47a63108a5c7f84d7a40aaa28c3b38 download: Received 500 response
Error for fbfc8883ed443f3dfbc5c97c5f533b226f453c37f1bb922b22800c5db5a8f844 download: Received 500 response
ERROR: Could not download all files, aborting update
Update took 70.9 seconds
44 files were not in a pack
swupd-update-partition: swupd: return code 1 = unspecific failure
```

castulo:
This PR includes commits to improve the way errors in updates are being handled so the final status of an update is more accurately communicated to the user. The PR includes the following items:
- It merges the load_mom and load_mom_err functions into since the only difference between them was that load_mom was a wrapper that didn't return an error code in case the function failed.
- Shows an actual message of "Update filed" when there is an error during the update process.
- If one or more files fail to be downloaded, abort the update.

Closes #323


src/update.c
```diff
@@ -315,25 +315,28 @@ static int main_update()
...
 			increment_retries(&retries, &timeout);
-			fprintf(stderr, "Retry #%d downloading from/to MoM Manifests\n", retries);
+			fprintf(stderr, "Retry #%d downloading current MoM manifest\n", retries);
 			goto load_current_mom;
 		}
-		fprintf(stderr, "Failure retrieving manifest from server\n");
-		ret = EMOM_NOTFOUND;
+		if (manifest_err == -ENET404 || manifest_err == -ENOENT) {
+			fprintf(stderr, "The current MoM manifest was not found\n");
+		}
```
jwakre: I think this would be easier to read:

if (manifest_err == -ENET404 || manifest_err == -ENOENT) {
....
} else if (retries < MAX_TRIES) {
.....
}

castulo: ok, I'll make that change @jwakre

castulo:
@jwakre I made the requested change

src/update.c
```diff
@@ -315,25 +315,27 @@ static int main_update()
...
+			fprintf(stderr, "The current MoM manifest was not found\n");
+		} else if (retries < MAX_TRIES) {
 			increment_retries(&retries, &timeout);
-			fprintf(stderr, "Retry #%d downloading from/to MoM Manifests\n", retries);
+			fprintf(stderr, "Retry #%d downloading current MoM manifest\n", retries);
 			goto load_current_mom;
 		}
-		fprintf(stderr, "Failure retrieving manifest from server\n");
-		ret = EMOM_NOTFOUND;
+		ret = EMOM_LOAD;
```
jwakre: I think that since we already have manifest_err, we should return that instead of guessing EMOM_LOAD.

src/search.c
```diff
@@ -805,10 +805,10 @@ static int download_manifests(struct manifest **MoM, struct list **subs)
 		return ECURRENT_VERSION;
 	}
 
-	*MoM = load_mom(current_version, false, false);
+	*MoM = load_mom(current_version, false, false, NULL);
```
jwakre: My understanding of #323 is that part of the issue is error code propagation. Is there a reason that we still want to not get the exact error codes from load_mom? We could return the error code from load_mom instead of using EMOM_LOAD as a generic error.

src/download.c
```diff
@@ -581,6 +582,11 @@ int swupd_curl_parallel_download_end(void *handle, int *num_downloads)
 		}
 	}
 
+	// If there are download failures at this point return an error
+	if (h->failed != NULL) {
+		ret = -EFULLDOWNLOAD;
```
otaviobp: As you are changing this, can you rename this constant? This name doesn't make sense anymore. And can you document the new return code in the function header?

src/update.c
```diff
@@ -479,8 +467,9 @@ static int main_update()
 		(void)update_device_latest_version(server_version);
 		printf("Update was applied.\n");
 	} else if (ret < 0) {
-		// Ensure a positive exit status for the main program.
+		/* Ensure a positive exit status for the main program. */
```
otaviobp: any reason to change the comment style? afaik we're using both in swupd

src/search.c
castulo: The other day I had a talk with @otaviobp about this, we talked about what return codes we should be returning back to the end user. He mentioned that although we can use many different codes within the functions of the swupd code, the codes we are supposed to return to the end user are only those listed in https://github.com/clearlinux/swupd-client/blob/master/src/swupd-error.h.
This make sense to me, because at the moment internally we also use error codes from curl or from errno.h or errno-base.h, so if we propagate those errors to the end user they may not make any sense to the user anyway. 

I understand that part of #323 is error code propagation, and that part has been addressed by adding the exact error code to the load_mom function.  But in the end we need to translate that error into one of the swupd-errors before returning to the user, so we cannot really always propagate all the codes up to the user.

I think the most important part of what #323 complains about is the fact that the error returned was misleading and the presented output didn't provide much help to determine what caused the error. I think that has been fixed now with the changes in the PR, even if the return code received by the end user is a generic EMOM_LOAD error (which is more accureate than EMOM_NOTFOUND) the messages that are printed as part of the output will now be useful to determine the problem.

src/update.c
castulo: For some reason I thought we were only using this style /* */, but now that you mentioned this I just started looking the other style of comments everywhere too. My mind was playing tricks on me. I'll return that comment the way it was.

src/update.c
castulo: see my answer to your comment below and let me know what you think

src/search.c
jwakre: I understand and this looks good to me now. Thanks for the clarification.

otaviobp: I would add just one thing to @castulo comment. All error codes returned by a Linux application should be a positive integer number from 0 to 255, where 0 is success. And all exit codes should be in swupd-error.h and documented on doc/swupd.1.

src/download.c
castulo: ok, I'll rename it as EFILEDOWNLOAD, I think it will be more useful that way

castulo:
@otaviobp I made the change you requested

castulo:
I forgot to change the comment of EFILEDOWNLOAD in the test library, I updated that.

PR #619, opened 2018-10-11
Adding a menu of functions for testlib

castulo:
This commit adds a menu of the functions that are available for
users when writing tests or just using the library to create
resources to test their code.

### DIFF
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -39,7 +39,7 @@ export EBADCERT=24  # unable to verify server SSL certificate
 # global constant
 export zero_hash="0000000000000000000000000000000000000000000000000000000000000000"
 
-generate_random_content() { 
+generate_random_content() { # swupd_function
 
 	local bottom_range=${1:-5}
 	local top_range=${2:-100}
@@ -49,7 +49,7 @@ generate_random_content() {
 
 }
 
-generate_random_name() { 
+generate_random_name() { # swupd_function
 
 	local prefix=${1:-test-}
 	local uuid
@@ -64,7 +64,7 @@ generate_random_name() {
 # Parameters:
 # - key_path: path to private key
 # - cert_path: path to public key
-generate_certificate() {
+generate_certificate() { # swupd_function
 
 	local key_path=$1
 	local cert_path=$2
@@ -99,7 +99,7 @@ generate_certificate() {
 # Creates the trusted certificate store and adds one certificate
 # Parameters:
 # - cert_path: path of certificate to add to trust store
-create_trusted_cacert() {
+create_trusted_cacert() { # swupd_function
 
 	local cert_path=$1
 	local counter=0
@@ -136,7 +136,7 @@ create_trusted_cacert() {
 }
 
 # Deletes the trusted certificate store
-destroy_trusted_cacert() {
+destroy_trusted_cacert() { # swupd_function
 
 	# only the test that created CACERT_DIR can erase it which stops
 	# tests from erasing CACERT_DIR when they fail before creating it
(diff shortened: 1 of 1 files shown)


icarus-sparry:
I don't think the implementation is a good idea. POSIX only allows the 'foo()' format for declaring a function. 'function foo' is an extension which means different things in ksh, bash, dash etc. In dash it is invalid, in bash it is the same and in ksh it introduces different scoping rules for local variables.

I think a better implementation would use a "magic comment" to mark things that should be ignored.

 I wonder if the sorts are a good idea. The fact that you are splitting function names which contain "assert" apart suggests that a pure sorted list is not exactly what is needed.

Perhaps you could instead look for lines matching say '^## [0-9]+ ', sort them, and then remove the first two 'words' would allow a more general arrangement of the output. It would also allow you to remove the special case code for the headings, you could have '## 0 heading for section 1' and '## 2 heading for section 3' lines.

otaviobp:
> I don't think the implementation is a good idea. POSIX only allows the 'foo()' format for declaring a function. 'function foo' is an extension which means different things in ksh, bash, dash etc. In dash it is invalid, in bash it is the same and in ksh it introduces different scoping rules for local variables.

We are using bats for testing and only bash is supported, so I don't see a reason on being concerned about that.

> I think a better implementation would use a "magic comment" to mark things that should be ignored.

Anyway, I think it's a better idea to use a magic comment. But instead of a comment for ignore I would add a comment for listing it. You could have

     foo() { #swupd_function

and 

    assert_bla() { #swupd_assert

and grep for that comment.

> I wonder if the sorts are a good idea. The fact that you are splitting function names which contain "assert" apart suggests that a pure sorted list is not exactly what is needed.

I'm ok on just a regular sort for now as we don't have a really big number of functions. But I agree that it's nice to have that organized by sections

castulo:
Thanks @icarus-sparry and @otaviobp that is really good feedback

test/functional/testlib.bash
```diff
@@ -2693,3 +2705,40 @@ assert_files_not_equal() {
...
+# Parameters:
+# - prefix: a prefix that could be used to filter functions
+testlib() {
+
+	local prefix=$1
+	local funcs
+	local assertions
+
+	funcs=$(grep "$prefix.*() { [#] swupd_function" "$FUNC_DIR"/testlib.bash | cut -f1 -d "(")
+	assertions=$(grep "$prefix.*() { [#] assertion" "$FUNC_DIR"/testlib.bash | cut -f1 -d "(")
```
jwakre: For these regex to match the function prefixes, they should start at the beginning of the line.

test/functional/testlib.bash
```diff
@@ -2693,3 +2705,40 @@ assert_files_not_equal() {
...
+	echo ""
+	if [ -z "$prefix" ]; then
```
otaviobp: I prefer to leave the header even if prefix is set.

test/functional/testlib.bash
otaviobp: You are right. But I prefer the way it is now.
So:

     $ testlib not

will print all functions with not in the name, like:

    assert_status_is_not
    assert_dir_not_exists
    assert_file_not_exists
    assert_not_in_output
    assert_is_not_output
    assert_regex_not_in_output
    assert_regex_is_not_output
    assert_not_equal
    assert_files_not_equal

jwakre: In that case, I think the prefix argument name and description should be changed to remove confusion.

castulo: that is a good idea @jwakre, I'll update that

test/functional/testlib.bash
castulo: I will fix that @otaviobp

castulo:
@otaviobp, @jwakre, I have addressed your issues in this new PR update.

PR #623, opened 2018-10-16
test: Test if make install is installing all necessary files

otaviobp:
Check if all files expected to be installed in a full installation are
really installed

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -221,9 +221,11 @@ dist_check_SCRIPTS += test/functional/generate-cert.prereq
 
 parallel_tests = $(BATS:.bats=.log)
 $(parallel_tests) : test/functional/generate-cert.log
+
+install-check:
+	test/installation/test.bats
+
 endif
-
-
 
 if COVERAGE
 AM_CFLAGS += --coverage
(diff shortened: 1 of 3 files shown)


Makefile.am
```diff
@@ -221,14 +221,16 @@ dist_check_SCRIPTS += test/functional/generate-cert.prereq
...
+install-check:
+	test/installation/test.bats
 
+endif
 
 if COVERAGE
 AM_CFLAGS += --coverage
 
-coverage: coverage-clean
+coverage: coverage-clean ]
```
phmccarty: This looks like an accidental change

otaviobp: I wonder how this is not breaking any test. I'll remove that. it was an accident. thanks

test/installation/test.bats
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bats
```
castulo: Remember that if you don't specify a [setup](https://github.com/clearlinux/swupd-client/blob/master/test/functional/testlib.bash#L2335) function a test environment is going to be created by default.
The same applies for the [teardown](https://github.com/clearlinux/swupd-client/blob/master/test/functional/testlib.bash#L2342) function.

If you want to save that extra time, just add an empty setup and teardown methods to your test:

```
test_setup() {

	# do nothing
	return

}

test_teardown() {

	# do nothing
	return

}
```

PR #625, opened 2018-10-17
Remove 404 from the list of possible swupd return codes

otaviobp:
Linux programs return code should be a number between 0 and 255, so 404 is an
invalid return code. Removing it from docs and from swupd-error.h file.

As it wasn't being used, this won't be a problem.


otaviobp:
Related to commend I did on #617

PR #626, opened 2018-10-17
Adding the author and testID info to test template

castulo:
This commit adds a few improvements to the test template that aim to
make easier to track tests.

As we are trying to organize the swupd tests better, it is useful to
have an easy way to identify who the original author of the test was
in case there are questions regarding the purpose of the test.

It is also useful sometimes to add a long description to a complex
test, so people reading the test understand what the test is attempting
to verify, and why is useful or different.

Lastly, it is a good practice to add a unique ID to every test case,
this makes it easier to track it, and to refer to it.

### DIFF
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -2245,6 +2245,8 @@ generate_test() {
 
 	{
 		printf '#!/usr/bin/env bats\n\n'
+		printf '# Author: <author name>\n'
+		printf '# Email: <author email>\n\n'
 		printf 'load "../testlib"\n\n'
 		printf 'global_setup() {\n\n'
 		printf '\t# global setup\n\n'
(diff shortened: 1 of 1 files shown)


test/functional/testlib.bash
```diff
@@ -2259,15 +2261,16 @@ generate_test() {
 		printf 'global_teardown() {\n\n'
 		printf '\t# global cleanup\n\n'
 		printf '}\n\n'
-		printf '@test "<test description>" {\n\n'
+		printf '@test "<test ID>: <test description>" {\n\n'
```
jwakre: I am concerned that it would be annoying to figure out the next available test ID value when adding a new test. Could this function populate the test ID with an open number? Maybe a grep command could be used to read existing IDs and help pick an available one.

otaviobp: For now we can leave like this and plan to have a script to create IDs later. Let's see how this work first. If it's getting complicated we can write a better solution

jwakre: Ok

PR #627, opened 2018-10-18
Fix a bug in some verify tests

castulo:
Some time ago the "-u" option of the create_test_environment and
create_version functions was renamed to "-r". Four of the verify
tests were not updated, so they were crashing during the setup,
while trying to create the test environment, so the test would not
be run since the setup failed but the teardown would still run and
the teardown would attempt to delete a non existant test environment
which would cause the destroy_test_environment to exit the script
causing an inconclusive test result.

This commit fixes the issue by doing the following actions:
- Fixes the tests by using the correct option "-r".
- Modifies the destroy_test_environment function so it does not exit
if a test environment doesn't exist, but instead it just returns a non
zero code and warns the user so the teardown works as expected.

### DIFF
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -1366,7 +1366,15 @@ destroy_test_environment() { # swupd_function
 			EOM
 		return
 	fi
-	validate_path "$env_name"
+	validate_param "$env_name"
+
+	# if the test environment doesn't exist warn the user but don't terminate script
+	# execution, this function could be called from a test teardown and terminating
+	# at that point will cause incorrect test reporting
+	if [ ! -d "$env_name" ]; then
+		echo "Warning: test environment \"$env_name\" doesn't exist, nothing to destroy."
+		return 1
+	fi
 
 	# since the action to be performed is very destructive, at least
 	# make sure the directory does look like a test environment
--- a/test/functional/verify/verify-format-mismatch-override.bats
+++ b/test/functional/verify/verify-format-mismatch-override.bats
@@ -4,9 +4,9 @@ load "../testlib"
 
 test_setup() {
 
-	create_test_environment -u "$TEST_NAME" 10 1
+	create_test_environment -r "$TEST_NAME" 10 1
 	bump_format "$TEST_NAME"
-	create_version -u "$TEST_NAME" 40 30 2
+	create_version -r "$TEST_NAME" 40 30 2
 	update_bundle "$TEST_NAME" os-core --add-dir /usr/bin
 
 }
--- a/test/functional/verify/verify-format-mismatch.bats
+++ b/test/functional/verify/verify-format-mismatch.bats
@@ -4,9 +4,9 @@ load "../testlib"
 
 test_setup() {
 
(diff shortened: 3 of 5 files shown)


jwakre:
Spelling: s/accouonted/accounted

Otherwise +1

otaviobp:
>  so they were crashing during the setup, so the failure was not accouonted for

We need a fix for that. If the setup is failing the test should fail. This is a big problem. if it's not an easy fix, open an issue for that.

otaviobp:
btw, s/crete_version/create_version

castulo:
> > so they were crashing during the setup, so the failure was not accouonted for
> 
> We need a fix for that. If the setup is failing the test should fail. This is a big problem. if it's not an easy fix, open an issue for that.

So after further debugging this, I found out that the problem is actually not because there was an error in the setup method.... but the real problem was with the teardown method.

Looks like this is the current behavior in BATS; if something causes the test execution to exit during setup or during the test body, the test shows as failed correctly, but if exit occurs during teardown then the test shows no result. 

For example, this case:

```
#!/usr/bin/env bats

setup() {
	exit 1
}

teardown() {
	echo "teardown"
}


@test "some test" {
	exit 1
}
```
Running the test above will yield the correct result:
 **✗ some test
1 test, 1 failure**

Now, consider this other one:

```
#!/usr/bin/env bats

setup() {
	exit 1
}

teardown() {
	exit 1
}

@test "some test" {
	exit 1
}
```

The test above will yield to an incorrect inconclusive result:
   **some test                                                                                                                                         
1 test, 0 failures**

It doesn't pass but it doesn't fail either.

The misbehaving verify tests where using a wrong option when creating the test environment, which would cause the create_test_environment function to fail, and issue an `exit 1`, which should be fine,  but then the tests were running their teardown function which was attempting to destroy the test environment using the destroy_test_environment function using a non existing environment as parameter, this was causing this function to trigger an `exit 1` which would cause the inconclusive test result.

I will update this PR to address that and will update the commit message.

test/functional/testlib.bash
```diff
@@ -1366,7 +1366,15 @@ destroy_test_environment() { # swupd_function
...
 	fi
-	validate_path "$env_name"
+	validate_param "$env_name"
+
+	# if the test environment doesn't exist warn the user but don't terminate script
+	# execution, this function could be called from a test teardown and terminating
+	# at that point will cause incorrect test reporting
+	if [ ! -d "$env_name" ]; then
+		echo "Warning: test environment \"$env_name\" doesn't exist, nothing to destroy."
+		return 1
```
otaviobp: I'm not as familiar with bats as you, but I think bats will set the test as fail if you return 1 from any function called by a test, so this is different from what is commented.
It looks like this isn't a problem for teardown, as you said. So just change the comment (if this is really the case)

castulo: I'm not sure I understand your comment @otaviobp. Returning 1 to mark the test as failed if something in the teardown fails is ok, but exiting the script from the teardown is not since it will live the test with no result.

otaviobp: ok. got that :)

PR #628, opened 2018-10-18
test: Check if swupd is able to perform updates in last releases

otaviobp:
Tests added:
 - Check if bundle-list --all is working (because this affect the next tests)
 - Check if we can install an old version of clear and update interactively
   until last release
 - Check if we can install and old version and update to the most recent one

As this test takes a really long time to run, I'm not adding it to the
tests list to be executed to make check, but this can be used to validate
swupd before a release.

### DIFF
--- a/test/real_content/swupd_update.bats
+++ b/test/real_content/swupd_update.bats
@@ -1,0 +1,235 @@
+#!/usr/bin/env bats
+
+load "../functional/testlib"
+
+# Test swupd with real official content published on clearlinux.org
+# This test assumes all functional tests to be passing in order to work,
+# because we aren't going to test all swupd functionalities. The idea is
+# to test major swupd commands and standard use cases
+
+#TODO: Add test for crossing a format barrier
+
+# Optional parameters
+# $URL: Clearlinux content and version urls
+# $MAX_VERSIONS: The maximum number of versions we are going to be looking to
+#                test updates. Should be # at least 2 (current + update).
+#                Default is 10.
+# $BUNDLE_LIST: List of bundles to be installed on the system besides os-core
+#               and os-core-update. If empty, install all bundles.
+
+export SWUPD_OPTS # Swupd options to use in tests
+export SWUPD_OPTS_SHORT # Swupd options to use in tests (for bundle-list)
+export FORMAT # Format of the last available version
+export VERSION # Array of last available versions. First item is the oldest
+export ROOT_DIR # Root directory of the installation
+
+# get_format: Echos the format of the version in $1
+# TODO: This is not the best way to handle that. we should have in swupd
+# codebase the information about the supported format of this swupd.
+get_format() {
+	v="$1"
+	curl -f "${URL}/${v}/format" 2>/dev/null
+	return 0
+}
+
+# get_format_from_versions: Echos the first format available in a version from
+# a list of versions
+get_format_from_versions() {
+	formats=("$@")
+	i="${#formats[@]}"
+	i=$((i-1))
+	while [ "$i" -ge 0 ]; do
+		format=$(get_format "${formats[$i]}")
+		if [ -z "$format" ]; then # if version isn't published, discard
+			continue
+		else
+			echo "$format"
+			break
+		fi
+		i=$((i-1))
+	done
+}
+
+test_setup() {
+	# Fill defaults in optional parameters
+	if [ -z "$URL" ]; then
+		URL=https://cdn.download.clearlinux.org/update
+	fi
+	if [ -z "$MAX_VERSIONS" ]; then
+		MAX_VERSIONS=10
+	fi
+
+	# Using github clr-bundles project to get official clear releases
+	# version list is sorted from newer to older
+	# TODO: Discover available versions to work with custom mixers
+	read -r -d '\n' -a version_list < <(git ls-remote --tags https://github.com/clearlinux/clr-bundles.git | grep -v {} | grep -v latest | sed "s/.*refs\\/tags\\///" | sort -gr | head -n "$MAX_VERSIONS" | sort -g) || echo
+	if [ "${#version_list[@]}" -eq 0 ]; then
+		echo "Impossible to get versions list"
+		return 1
+	fi
+
+	# Get the last format
+	FORMAT=$(get_format_from_versions "${version_list[@]}")
+	export FORMAT
+
+	# Fill a list of last available versions (up to ${MAX_VERSIONS})
+	i=0
+	for v in "${version_list[@]}"; do
+		format=$(get_format "$v")
+		# if version isn't published or is from a different version, discard
+		if [ -z "$format" ] || [ "$FORMAT" -ne "$format" ]; then
+			continue
+		fi
+
+		VERSION[$i]=$v
+		i=$((i+1))
+	done
+
+	# Not enough versions to run the test
+	if [ -z "${VERSION[0]}" ] || [ -z "${VERSION[1]}" ]; then
+		# TODO: This is going to break on first release in a new format
+		echo "We need at least 2 versions in format $FORMAT to continue with this test"
+		return 1
+	fi
+
+	sudo rm -rf "$TEST_NAME"
+	sudo mkdir -p "$TEST_NAME"
+	ROOT_DIR="${PWD}/${TEST_NAME}"
+
+	# TODO: use -S ${ROOT_DIR}/var/lib/swupd/ when #665 gets fixed
+	# TODO: Add -t
+	SWUPD_OPTS_SHORT="-u ${URL} -p ${ROOT_DIR} -S ${ROOT_DIR}/swupd/"
+	SWUPD_OPTS="$SWUPD_OPTS_SHORT --no-scripts"
+}
+
+test_teardown() {
+	sudo rm -rf "$TEST_NAME"
+}
+
+verify_system() {
+	# TODO: This is a bug in verify --picky #632
+	sudo rm -f "${ROOT_DIR}/usr/share/clear/bundles/.MoM"
+
+	run sudo sh -c "$SWUPD verify --picky $SWUPD_OPTS 2>/dev/null"
+	if [ -n "$output" ]; then
+		echo "Verify found extra files in the system:"
+		echo "$output"
+		return 1
+	fi
+
+	run sudo sh -c "$SWUPD verify $SWUPD_OPTS"
+	assert_not_in_output "did not match"
+
+	return
+}
+
+# check_version: Check if version in /usr/lib/os-release is equal $1.
+# Fails otherwise.
+check_version() {
+	version="$1"
+	cur_version=$(grep VERSION_ID "${ROOT_DIR}/usr/lib/os-release" | cut -d = -f 2)
+	if [ "$version" -ne "$cur_version" ]; then
+		echo "Version $version is different from expected $cur_version"
+		return 1
+	fi
+}
+
+install_bundles() {
+	if [ -z "$BUNDLE_LIST" ]; then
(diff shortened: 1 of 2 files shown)


test/real_content/test_swupd.bats
```diff
@@ -0,0 +1,211 @@
...
+		return 1
+
+	fi
+
+	# Fill a list of last available versions (up to ${MAX_VERSIONS})
+	i=0
+	for v in $version_list; do
+		format=$(get_format $v)
+		# if version isn't published or is from a different version, discard
+		if [ -z "$format" ] || [ "$FORMAT" -ne "${format}" ]; then
```
phmccarty: `"$format"` and `"${format}"` are equivalent in this case. It looks like the `"$format"` matches the style of the rest of the file for non-array variable references.

test/real_content/test_swupd.bats
```diff
@@ -0,0 +1,211 @@
...
+			continue
+		fi
+
+		VERSION[$i]=$v
+		i=$(($i+1))
```
phmccarty: Other test files omit the leading `$` for variable references in arithmetic expressions. For consistency, you can write this as `i=$((i+1))`.

The same comment applies to several other places in this source file, so I won't comment on those lines.

test/real_content/test_swupd.bats
```diff
@@ -0,0 +1,211 @@
...
+	#assert_not_in_output "did not match"
+
+	return
+}
+
+# check_version: Check if version in /usr/lib/os-release is equal $1.
+# Fails otherwise.
+check_version() {
+	version="$1"
+	cur_version=$(grep VERSION_ID ${ROOT_DIR}/usr/lib/os-release|cut -d = -f 2)
```
phmccarty: optional: I like using AWK instead:

```shell
cur_version=$(awk -F= '$1 == "VERSION_ID" { print $2 }' ${ROOT_DIR}/usr/lib/os-release)
```

test/real_content/test_swupd.bats
```diff
@@ -0,0 +1,211 @@
...
+
+	echo "Install one package"
+	run sudo sh -c "$SWUPD bundle-add $SWUPD_OPTS os-core-update -F $FORMAT"
+	assert_status_is 0
+	verify_system
+
+	echo "Install all packages"
+	all_packages=$(sudo $SWUPD bundle-list --all $SWUPD_OPTS)
+	all_packages=$(echo $all_packages) #take \n out
+	run sudo sh -c "$SWUPD bundle-add $SWUPD_OPTS $all_packages"
```
jwakre: Usually after a 'run' command, the following line calls an assert (assert_status_is 0).

test/real_content/test_swupd.bats
```diff
@@ -0,0 +1,211 @@
...
+	verify_system
+
+	i=1
+	version=""
+	while [ ! -z ${VERSION[$i]} ]; do
+		version="${VERSION[$i]}"
+		i=$(($i+1))
+	done
```
jwakre: To get the last item in the array, you could do one of the following instead of using a loop:

${VERSION[-1]} # requires bash > v4.1
${VERSION[${#VERSION[@]} -1]}

https://unix.stackexchange.com/questions/198787/is-there-a-way-of-reading-the-last-element-of-an-array-with-bash

test/real_content/test_swupd.bats
```diff
@@ -0,0 +1,211 @@
...
+		if [ -z "$format" ]; then # if version isn't published, discard
+			continue
+		else
+			echo $format
+			break
+		fi
+	done
+}
+
+test_setup() {
```
jwakre: I think some of the contents of test_setup only need to be run once and could be moved to the global_setup function.

test/real_content/test_swupd.bats
```diff
@@ -0,0 +1,211 @@
...
+
+	sudo rm -rf "$TEST_NAME"
+	sudo mkdir -p "$TEST_NAME"
+	ROOT_DIR="${PWD}/${TEST_NAME}"
+
+	SWUPD_OPTS="-u ${URL} -p ${ROOT_DIR} -S ${ROOT_DIR}/var/lib/swupd/"
+}
+
+test_teardown() {
+	if [ "$DEBUG_TEST" != true ]; then
```
jwakre: I don't think this is very useful because it will only keep the state of the last test to run (update from first to last). For instance, if one of the 'incremental updates' tests fail, its test directory will be deleted by the 'update from first to last' test setup.

test/real_content/test_swupd.bats
```diff
@@ -0,0 +1,211 @@
...
+get_format() {
+	v="$1"
+	curl -f ${URL}/${v}/format 2>/dev/null
+	return 0
+}
+
+# get_format_from_versions: Echos the first format available in a version from
+# a list of versions
+get_format_from_versions() {
+	for v in "$@"; do
```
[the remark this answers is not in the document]
otaviobp: I really didn't know that. Thanks

test/real_content/test_swupd.bats
otaviobp: I agree, but I couldn't find an easy way to do that.
Is there a way to export an env variable from one test to the other?

test/real_content/test_swupd.bats
otaviobp: I followed the same approach of the teardown in the testlib. So we have a problem in all tests with more than one test when DEBUG_TEST is used.

test/real_content/test_swupd.bats
otaviobp: For this case I see no reason to use awk, but I know shell-script devs usually hate grep|cut

test/real_content/test_swupd.bats
otaviobp: hahaha. I should have googled that.

test/real_content/test_swupd.bats
jwakre: I see what you mean. The env variables don't persist between tests... Currently, I think the only ways to export variables between the tests are to call the set_env_variables function in setup() or to give them global scope. Maybe we could add a function to the testlib in the future to make this easier

jwakre: Now, I think this is fine as is. It would be useful to add a feature to testlib to solve this issue in the future.

test/real_content/test_swupd.bats
phmccarty: I'm okay with this change, I just prefer the other way. One improvement would be to grep for `^VERSION_ID=` to avoid matching some other key (or value).

otaviobp:
Fixed proposed changes, added --no-scripts to verify and update commands, added a new variable $BUNDLE_LIST to make it possible to run a faster test and added a new test to check if fullfile downloads are working.

To run it in the fast mode just do:

    $ BUNDLE_LIST=os-core bats -t test/real_content/test_swupd.bats

test/real_content/test_swupd.bats
```diff
@@ -0,0 +1,234 @@
...
+# get_format: Echos the format of the version in $1
+# TODO: This is not the best way to handle that. we should have in swupd
+# codebase the information about the supported format of this swupd.
+get_format() {
+	v="$1"
+	curl -f ${URL}/${v}/format 2>/dev/null
+	return 0
+}
+
+# get_format_from_versions: Echos the first format available in a version from
```
castulo: one of the things I see you are often overlooking is using variables without enclosing them in double quotes.
> Quoting variables prevents word splitting and glob expansion, and prevents the script from breaking when input contains spaces, line feeds, glob characters and such. - https://github.com/koalaman/shellcheck/wiki/SC2086

>

test/real_content/test_swupd.bats
```diff
@@ -0,0 +1,234 @@
...
+# a list of versions
+get_format_from_versions() {
+	for v; do
```
castulo: where are you defining variable "v" here?

phmccarty: `for v; do ... done` is equivalent to `for v in "$@"; do ... done`. @otaviobp changed this in response to my review from a couple days ago.

test/real_content/test_swupd.bats
```diff
@@ -0,0 +1,234 @@
...
+	# Fill a list of last available versions (up to ${MAX_VERSIONS})
+	i=0
+	for v in $version_list; do
+		format=$(get_format $v)
+		# if version isn't published or is from a different version, discard
+		if [ -z "$format" ] || [ "$FORMAT" -ne "$format" ]; then
+			continue
+		fi
+
+		VERSION[$i]=$v
```
castulo: in line 22 you defined a local variable called VERSIONS:
`local VERSIONS # Array of last available versions. First item is the oldest`

but here you are using VERSION and I don't see VERSIONS anywhere else. did you mean to use VERSIONS here?

test/real_content/test_swupd.bats
```diff
@@ -0,0 +1,234 @@
...
+
+# Optional parameters
+# $URL: Clearlinux content and version urls
+# $MAX_VERSIONS: The maximum number of versions we are going to be looking to
+#                test updates. Should be # at least 2 (current + update).
+#                Default is 10.
+# $BUNDLE_LIST: List of bundles to be installed on the system besides os-core
+#               and os-core-update. If empty, install all bundles.
+
+local SWUPD_OPTS # Swupd options to use in tests
```
castulo: you are using local to define some variables here, but local is only valid within functions

otaviobp:
> Actually one thing I used to do in the last project I worked on, is that we had shellcheck running in our CI server on every bash script and it would reject the patch if it the scripts didn't comply with shellcheck. It would save reviewers time if we do something like that.

I forgot about that again. Can you set that up for travis? Unless it takes a really long time it would be nice to have this run on every script we change.

castulo:
@otaviobp there is still one comment you have not addressed though:

> in line 22 you defined a local variable called VERSIONS:
> `local VERSIONS # Array of last available versions. First item is the oldest`
> 
> but here you are using VERSION and I don't see VERSIONS anywhere else. did you mean to use VERSIONS here?

So didn't you mean to call the variable at line 23 as VERSION instead of VERSIONS?

otaviobp:
> So didn't you mean to call the variable at line 23 as VERSION instead of VERSIONS?

yes. fixed.

castulo:
@otaviobp did you say you ran shellcheck in your script and fixed all the issues reported by it? I just did and I see 24 issues reported by shellcheck. Starting with the variables you are declaring as local at the beginning of the script. _local_ only make sense inside a function.

otaviobp:
Ouch. I renamed the file to a name that was in .gitignore. So I sent the outdated version. Thank you for noticing.
I have just pushed a new version with a change to .gitignore.
I'm also always using shellchecker binary with the seds in Makefile because shellcheck can't recognize load and @test.

test/real_content/swupd_update.bats
```diff
@@ -0,0 +1,235 @@
...
+# $MAX_VERSIONS: The maximum number of versions we are going to be looking to
+#                test updates. Should be # at least 2 (current + update).
+#                Default is 10.
+# $BUNDLE_LIST: List of bundles to be installed on the system besides os-core
+#               and os-core-update. If empty, install all bundles.
+
+export SWUPD_OPTS # Swupd options to use in tests
+export SWUPD_OPTS_SHORT # Swupd options to use in tests (for bundle-list)
+export FORMAT # Format of the last available version
+export VERSIONS # Array of last available versions. First item is the oldest
```
castulo: lol, now VERSIONS is back instead of VERSION 😂

otaviobp: hahahaha

otaviobp:
well, I think we're fine now

PR #631, opened 2018-10-18
Safely convert strings to numbers

### LINKED ISSUE #601: Type mismatch for variables loaded from manifest
In the manifest_from_file function in manifest.c, the strtoull function is used to load several variables from a manifest. The strtoull function returns an unsigned long long int, but its return value is frequently stored in a variable with a different type. This is a problem because large return values from strtoull will overflow other types.

jwakre:
Various functions, like strtol, were used to convert strings to numbers.
Frequently, the results of these conversion functions were stored in
variables with types that did not match the function return value. As a
result, there was a potential risk that an overflow would occur when
type casting the string conversion function result to store it in a
mismatched type.

The most common case is to convert a string to an int, so this commit
creates the strtoi_err function which converts a string to an int and
provides error handling. Also, the error checking for other string to
number conversions was improved by adding error handling with errno.

Fixes #601

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -595,6 +595,37 @@ out_fds:
 	return ret;
 }
 
+/* The strtol function is commonly used to convert a string to a number and
+ * the result is frequently stored in an int type, but type casting a long to
+ * an int can cause overflows. This function returns negative error codes based
+ * on the errno table. The -ERANGE code is returned when the string is out of
+ * range for strtol and the -EOVERFLOW code is returned when the long to int
+ * type conversion overflows.
+*/
+int strtoi_err(const char *str, char **endptr, int *value)
+{
+	long num;
+	int err;
+
+	errno = 0;
+	num = strtol(str, endptr, 10);
+	err = -errno;
+
+	/* When the return value of strtol overflows the int type, don't overflow
+	 * and return an overflow error code. */
+	if (num > INT_MAX) {
+		num = INT_MAX;
+		err = -EOVERFLOW;
+	} else if (num < INT_MIN) {
+		num = INT_MIN;
+		err = -EOVERFLOW;
+	}
+
+	*value = (int)num;
+
+	return err;
+}
+
 /* Return a duplicated copy of the string using strdup().
  * Abort if there's no memory to allocate the new string.
 */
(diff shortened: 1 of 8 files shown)


jwakre:
I am unsure on the best way to create tests for this change because different systems will have different type ranges. On my system I verified that type overflows was a problem (updating to version 2^32 + 100 identical to updating to version 100). With this change, type overflows were caught on my system.

src/helpers.c
```diff
@@ -595,6 +595,32 @@ int swupd_init(void)
...
+ * int can cause overflows. This function sets 'err' to a nonzero value when
+ * either the strtol return value or the long to int type conversion overflows.
+*/
+int strtoi_err(const char *str, char **endptr, int base, int *err)
+{
+	long num;
+
+	errno = 0;
+	num = strtol(str, endptr, base);
+	*err = errno;
```
otaviobp: Errno is always positive. So you should use -errno here

src/helpers.c
```diff
@@ -595,6 +595,32 @@ int swupd_init(void)
...
+
+	/* When the return value of strtol overflows the int type, don't overflow
+	 * and return an error. */
+	if (num > INT_MAX) {
+		num = INT_MAX;
+		*err = -1;
```
otaviobp: You can use the same codes of errno table here. -EOVERFLOW is more appropriate than -1 (EPERM)

otaviobp: And say in the doc above that you are running negative numbers correspondent to errno table.

src/helpers.c
```diff
@@ -595,6 +595,32 @@ int swupd_init(void)
 	return ret;
 }
 
+/* The strtol function is commonly used to convert a string to a number and
+ * the result is frequently stored in an int type. Type casting a long to an
+ * int can cause overflows. This function sets 'err' to a nonzero value when
+ * either the strtol return value or the long to int type conversion overflows.
+*/
+int strtoi_err(const char *str, char **endptr, int base, int *err)
```
otaviobp: We are always using base 10 and I don't see a reason to use other bases in any other case we are handling, so you can take the base out of the arguments list.
I prefer to always return the error and use the parameter for the value you wan't to retrieve. This is particularly good for this case because if you use incorrect int sizes the compiler will warn you.

So if you have

    int strtoi_err(const char *str, char **endptr, int *value);

and use

    char value;
    err = strtoi_err("1", NULL, &value);

The compiler will warn you.

otaviobp:
nice changes

Issue #610, opened 2018-10-09
Swupd is inconsistent in return codes


otaviobp:
Swupd doesn't return error codes consistently. We need to review all return codes and return the appropriate code.

For example:
 - 404 is an invalid code and it's never returned
 - swupd verify, update and others may return the undocumented code 1 (EXIT_FAILURE).

Related to #323

castulo:
So these are the tasks I've been thinking need to be done in order to have consistent return codes in swupd. Since it is a lot of work I think its better to just open an issue for each one and close this issue so we can track progress and make reviews easier.

TO DOs:
- Rename the file from swupd-error.h to swupd-codes.h so we can include “successful” codes if needed.
- Rename existing error codes following these rules:
    - All codes should start with SWUPD_<something> so we never get confused what code comes from swupd
    - Always separate codes that contain multiple words using an underscore SWUPD_MULTI_WORD_ERROR instead of using SWUPDMULTIWORDERROR
    - Code 0 is reserved for SWUPD_SUCCESS (or SWUPD_OK)
    - Code 1 is reserved to be used by commands that return a boolean state (e.g. _swupd autoupdate_)
- When a function returns a boolean state, explicitly return true or false instead of returning 0 or 1
- Some functions mask error codes in a “catch all” error code, for example bundle-add which only uses EBUNDLE_INSTALL or bundle-remove which only uses EBUNDLE_REMOVE. Remove the catch all return codes from these functions so we can have a more meaningful error.
- swupd should always exit with a code defined in swupd-codes.h
- Internally, swupd can use errors that comply with these rules:
    - it can use errors defined in any included file as long as these don’t get propagated to the end user. For example errors from: 
        - /usr/include/asm-generic/errno-base.h
        - /usr/include/asm-generic/errno.h
        - /usr/include/curl/curl.h
    - The errors used internally should preferably be negative values. For example prefer using -ENOENT instead of ENOENT.
- Update the documentation about swupd codes on swupd.1 and swupd.1.rst
- (Consider creating a document with a more detailed explanation of each code, like this -> https://curl.haxx.se/libcurl/c/libcurl-errors.html)

castulo:
> > Rename the file from swupd-error.h to swupd-codes.h so we can include “successful” codes if needed.
> 
> swupd_codes.h (maybe swupd_exit_codes.h)
> 
Yep, that sounds good to me
> > Internally, swupd can use errors that comply with these rules:
> 
[...]

PR #637, opened 2018-10-19
tests: add verify --install tests

matthewrsj:
This adds several tests for os installation functionality.
* install - regular installation of just os-core
* install-multiple - install os-core and a test-bundle
* install-bad - attempt to install an invalid bundle name
* install-no-zero-packs - install os-core and test-bundle with no
  zero-packs available. Fullfiles still available as a fallback.
* install-no-fullfile-fallback - attempt to install with no zero packs
  and no fullfiles available for fallback.


test/functional/install/install-bad.bats
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bats
+
+load "../testlib"
+
+test_setup() {
+
+	create_test_environment "$TEST_NAME" 10
+	create_bundle -n test-bundle -f /file_1,/file_2 "$TEST_NAME"
+	sudo mkdir -p "$TARGETDIR"/usr/share/clear/bundles
```
castulo: this line doesn't hurt, but just FYI, this directory is already created with the test environment:
https://github.com/clearlinux/swupd-client/blob/master/test/functional/testlib.bash#L1326

test/functional/install/install-multiple.bats
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bats
+
+load "../testlib"
+
+test_setup() {
+
+	create_test_environment "$TEST_NAME" 10
+	create_bundle -n test-bundle -f /file_1,/file_2 "$TEST_NAME"
+	sudo mkdir -p "$TARGETDIR"/usr/share/clear/bundles
+	sudo touch "$TARGETDIR"/usr/share/clear/bundles/test-bundle
```
castulo: same here, every bundle that is created with the test library includes the tracking file by default:
https://github.com/clearlinux/swupd-client/blob/master/test/functional/testlib.bash#L1585

test/functional/install/install-multiple.bats
```diff
@@ -0,0 +1,28 @@
...
+
+
+}
+
+@test "install multiple bundles" {
```
castulo: I'm confused about the description of this test, it mentions installing multiple bundles but yet I see only one bundle being explicit here. The "multiple" bundles is because of the os-core bundle that comes by default?

castulo: Never mind, I see now that it must be it. Maybe adding `assert_file_exists "$TARGETDIR"/core` to the list of verified files will make it easier to see that

Makefile.am
```diff
@@ -156,6 +156,11 @@ BATS = \
 	test/functional/checkupdate/chk-update-client-certificate.bats \
 	test/functional/completion/completion-basic.bats \
 	test/functional/hashdump/hashdump-file-hash.bats \
+	test/functional/install/install.bats \
+	test/functional/install/install-bad.bats \
+	test/functional/install/install-multiple.bats \
+	test/functional/install/install-no-fullfile-fallbacks.bats \
+	test/functional/install/install-no-zero-packs.bats \
```
castulo: don't you think that it would be better to add these tests to the verify directory instead of install? I see we already have a couple of tests related to install in the verify directory:
https://github.com/clearlinux/swupd-client/blob/master/Makefile.am#L211

test/functional/install/install-no-fullfile-fallbacks.bats
```diff
@@ -0,0 +1,29 @@
...
+	sudo rm -rf "$WEBDIR"/10/files/*
+
+}
+
+@test "install bundle with no zero packs and no fullfile fallbacks" {
+
+	run sudo sh -c "$SWUPD verify $SWUPD_OPTS --install -m 10"
+
+	# no way to get content, everything should fail
+	assert_status_is 1
```
jwakre: At first I was confused why this status was compared against '1' instead of a defined error, but I see that there is no error defined for exit status 1 in swupd-error.h. Is there a reason that an entry for exit status 1 does not exist there?

test/functional/install/install-multiple.bats
matthewrsj: Correct.

test/functional/install/install-no-fullfile-fallbacks.bats
otaviobp: verify.c is returning EXIT_FAILURE(1) and it shouldn't. We have a ticket for that #610.
Not a problem in this patch though.

PR #641, opened 2018-10-19
Adding missing tests for the -t option in the bundle-add command.

reaganlo:
Adding missing tests for the -t option in the bundle-add command.

### DIFF
--- a/test/functional/bundleadd/add-install-time.bats
+++ b/test/functional/bundleadd/add-install-time.bats
@@ -1,0 +1,51 @@
+#!/usr/bin/env bats
+
+load "../testlib"
+
+test_setup() {
+
+        create_test_environment "$TEST_NAME"
+        create_bundle -n bundle1 -f /file1,/file2 "$TEST_NAME"
+
+}
+
+@test "ADD030: Add a bundle showing detailed installation time" {
+
+	run sudo sh -c "$SWUPD bundle-add $SWUPD_OPTS bundle1 -t"
+
+	assert_status_is 0
+
+	expected_output=$(cat <<-EOM
+		Starting download of remaining update content. This may take a while...
+		Finishing download of update content...
+		Installing bundle\(s\) files...
+		Calling post-update helper scripts.
+		Successfully installed 1 bundle
+		Raw elapsed time stats:
+		.* ms: Total execution time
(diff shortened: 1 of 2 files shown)


test/functional/bundleadd/add-option-t.bats
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bats
+
+load "../testlib"
+
+test_setup() {
+
+	create_test_environment "$TEST_NAME"
+        create_bundle -n bundle1 -f /file1,/file2 "$TEST_NAME"
```
castulo: there is a weird indent here

castulo:
One more thing @reaganlo , in order for that test to run in travis, you need to add that test (in alphabetic order) to https://github.com/clearlinux/swupd-client/blob/master/Makefile.am#L114

test/functional/bundleadd/add-install-time.bats
```diff
@@ -0,0 +1,53 @@
...
+load "../testlib"
+
+test_setup() {
+
+	create_test_environment "$TEST_NAME"
+	create_bundle -n bundle1 -f /file1,/file2 "$TEST_NAME"
+
+}
+
+test_teardown() {
```
castulo: It is ok to leave it if you want but FYI, If you don't specify a test_teardown, the default test_teardown will be used which has _destroy_test_environment "$TEST_NAME"_  so you could remove the test_teardown in this case.

test/functional/bundleadd/add-install-time.bats
```diff
@@ -0,0 +1,53 @@
...
+		Raw elapsed time stats:
+			.* ms: Prepend bundles to list
+			.* ms: Install bundles
+		CPU process time stats:
+			.* ms: Prepend bundles to list
+			.* ms: Install bundles
+	EOM
+	)
+
+	assert_regex_in_output "$expected_output"
```
castulo: I would use _assert_regex_is_output_ instead of _assert_regex_in_output_, the difference between _is_ and _in_ is that _is_ expects the command output to match what you specified in your test, and nothing more, while using _in_ you are only verifying a partial output. So it is better to use _assert_regex_is_output_ when possible.

PR #644, opened 2018-10-22
bundle-add to return 0 if bundle already installed

### LINKED ISSUE #510: bundle-add returns non-zero exit code if no work needs done
If dev-utils is already installed, the following output is shown:
```
Warning: Bundle "dev-utils" is already installed, skipping it...
1 bundle was already installed
```

Returning an error code of 18.  This should return an error code of 0, much like check-update, to indicate when no work needs done.

castulo:
Previously, the only way to get a return code of 0 (success) during
a bundle-add operation was if everything that was attempted to be
added had been successful. That means that bundle-add command was
returning EBUNDLE_INSTALL whenever it would try to add a bundle that
was already installed.

This commit changes the return code of such operation since it was
determined that it is a more appropriate behavior to consider an
attempt to add a bundle that is already added as successful, even when
no bundle was added during the operation.

Closes #510

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -766,8 +766,7 @@ static int install_bundles(struct list *bundles, struct list **subs, struct mani
 	/* Use a bitwise AND with the add_sub_NEW mask to determine if at least
 	 * one new bundle was subscribed */
 	if (!(ret & add_sub_NEW)) {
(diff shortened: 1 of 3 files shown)


castulo:
I had missed updating one test, Travis should be happy now.

PR #649, opened 2018-10-23
test: consolidate test web server implementations

jwakre:
There were two web server implementations in the functional tests and the
test library's slow server implementation was not functional. These issues
were addressed by the following changes:

- Consolidate test web server implementations into test library
- Add partial download support to start_web_server test library function
- Fix test library slow server functionality.


test/functional/checkupdate/chk-update-slow-server.bats
```diff
@@ -28,15 +29,11 @@ global_teardown() {
...
-	slow_opts="-p $TEST_NAME/target-dir -F staging -u http://localhost:$PORT/"
-
-	# test
-	run sudo sh -c "$SWUPD check-update $slow_opts"
+	run sudo sh -c "$SWUPD check-update $SWUPD_OPTS_HTTP_NO_CERT"
 	assert_status_is 0
 	expected_output=$(cat <<-EOM
 		Current OS version: 10
-		There is a new OS version available: 99990
+		There is a new OS version available: 20
```
phmccarty: The reason the original slow server used version `99990` was to have the test web server write one byte at a time, with a small delay between writes to make it seem that the server was responding, but slowly.

I'm not sure the revised slow server does this anymore (see other comment below).

test/functional/server.py
```diff
@@ -2,29 +2,67 @@
...
+            self.send_response(200)
         self.send_header("Content-type", "text/html")
         self.end_headers()
-        response = "99990"
-        delay = 0.00001 # seconds
-        for c in response:
-            self.wfile.write(str.encode(c))
 
+        while True:
+            b = f.read(100)
```
phmccarty: For the slow server, processing 100 bytes at a time is much different than 1 byte at a time. Perhaps:

```python
if slow_server:
    b = f.read(1)
else:
    b = f.read(100)
```

otaviobp: This was written for the retry code and I used 100 because it was good enough to make suer versions and small manifests were downloaded all at one, but for sure testing with 1 will be better. Just make sure we're not forcing unnecessary retries. Not sure if 1 by 1 isn't too slow.

phmccarty: Good point, I was primarily thinking about the check-update slow server test, which only asks for the `latest` file. Maybe the slow server could read 1 byte at a time for `latest`, and 100 at a time for everything else?

jwakre: So would passing the delay as an argument would work?

phmccarty: More specifically, if the server writes 100 bytes for `latest`, that means that only one libcurl callback is needed to read the response, and the bug I linked above would not be triggered. By writing 1 byte at a time, likely more than one libcurl callback is needed. Also, the longer the version number in `latest`, the better; maybe the test should use a `latest` contains more than 5 bytes...

jwakre: I recorded the test execution times of update-slow-server when reading 1 byte and 100 bytes:

read(1): 134 seconds
read(100): 63 seconds

The test did not perform any unnecessary retries and it is normally skipped, so I will change this to f.read(1).

otaviobp: nice. I'm good with that

jwakre:
New version has the following changes to address feedback:

* Server reads 1 byte at a time.
* Changed latest version in check version test to 99990.

Additionally, the previous implementation had a problem with the logic to match the file name from self.path with partial_download_file. The old implementation used the 'endswith' function which would match incorrectly. Now, the 'split' function is used to correctly match the last path item from self.path with partial_download_file.

PR #653, opened 2018-10-24
Use verify --fix --picky with older versions

### LINKED ISSUE #608: "swupd verify --fix --picky -m <ver>" fails if a bundle doesn't exist in <ver>
I was at Clear ver 25180.  This was when the zstd bundle was introduced into my system.  I tried to "swupd verify --fix --picky -m 24180" and it failed because zstd didn't exist back then.  swupd gave a warning that the bundle is invalid and "skipping it...", errored out, and stopped.

The workaround was to manually delete the offending bundle(s) in /usr/share/clear/bundles/ and retry "swupd verify --fix --picky...".  This is what I think swupd should do automatically.

castulo:
When you do a "verify --fix --picky -m older_version", and one
or more of the currently installed bundles didn't exist back then,
a non useful error message was being used, and the verify operation
would fail.

This commit makes the following changes.

1) With this commit, if the mentioned situation happens, the verify
operation will still be stopped, but a useful message will now be
displayed to inform the user of the situation, and he/she will be
presented with an alternative to proceed using the --force option.

2) If the user uses the --force option, the user will still be warned
but the verify operation will continue and the offending bundles will
be removed from the system.

3) Fixes a bug when using verify --picky. Only files that were orphaned
were being counted as files being removed from the system, files removed
because of the picky options were not being counted. This commit fixes
that.

Closes #608

### DIFF
--- a/src/extra_files.c
+++ b/src/extra_files.c
@@ -90,9 +90,11 @@ static int qsort_helper(const void *A, const void *B)
 	return strcmp(((struct filerecord *)A)->filename, ((struct filerecord *)B)->filename);
 }
 
-static void handle(const char *filename, bool is_dir, bool fix)
+/* return true if the function deletes the specified filename, otherwise false */
+static bool handle(const char *filename, bool is_dir, bool fix)
 {
 	char *temp;
+	bool ret = true;
 
 	if (fix) {
 		string_or_die(&temp, "%s%s", path_prefix, filename);
@@ -100,19 +102,24 @@ static void handle(const char *filename, bool is_dir, bool fix)
 			is_dir ? "REMOVING DIR %s/\n" : "REMOVING %s\n",
 			filename);
 		if (remove(temp)) {
-			fprintf(stderr, "Removing %s failed: %s", temp, strerror(errno));
+			fprintf(stderr, "Removing %s failed: %s\n", temp, strerror(errno));
+			ret = false;
 		}
 		free_string(&temp);
 	} else {
 		printf("%s%s\n", filename, is_dir ? "/" : "");
+		ret = false;
 	}
+
+	return ret;
 }
 
 /* expect the start to end in /usr and be the absolute path to the root */
-int walk_tree(struct manifest *manifest, const char *start, bool fix, const regex_t *whitelist)
+int walk_tree(struct manifest *manifest, const char *start, bool fix, const regex_t *whitelist, struct file_counts *counts)
 {
 	/* Walk the tree, */
 	int rc;
+	int ret;
 	path_prefix_len = strlen(path_prefix);
 	path_whitelist = whitelist;
 	rc = nftw(start, &record_filename, 0, FTW_ACTIONRETVAL | FTW_PHYS | FTW_MOUNT);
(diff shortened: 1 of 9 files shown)


src/extra_files.c
```diff
@@ -90,26 +90,33 @@ static int qsort_helper(const void *A, const void *B)
...
 	}
+
+	/* return true if the function deleted a file or directory, false otherwise */
+	return ret;
 }
 
 /* expect the start to end in /usr and be the absolute path to the root */
-int walk_tree(struct manifest *manifest, const char *start, bool fix, const regex_t *whitelist)
+int walk_tree(struct manifest *manifest, const char *start, bool fix, const regex_t *whitelist,
+	      int *extraneous_files, int *files_deleted, int *files_not_deleted)
```
icarus-sparry: Rather than adding 3 int* parameters, consider adding a pointer to a struct that contains 3 integers.

castulo: Thanks for reviewing @icarus-sparry , that was a good suggestion. I have pushed an update that includes a struct for keeping those file counts. Please let me know if you have any concern with the new implementation.

src/verify.c
```diff
@@ -818,6 +818,27 @@ int verify_main(int argc, char **argv)
 	}
 
 	ret = add_included_manifests(official_manifest, &subs);
+	/* if we are verifying against an older version, and one or more bundles were
+	 * not found, continue only if --force was used since those bundles will be
+	 * removed */
+	if (version < sys_version && ret == -add_sub_BADNAME) {
```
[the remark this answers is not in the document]
castulo: hmm, yeah you are probably right, I guess that could happen if for example a bundle was removed in a newer version

castulo:
@otaviobp I made the changes you requested and I also took the opportunity to regroup the changes into more appropriate commits according to what they were doing so they are more organized.

otaviobp:
Fixed proposed changes, added --no-scripts to verify and update commands, added a new variable $BUNDLE_LIST to make it possible to run a faster test and added a new test to check if fullfile downloads are working.

To run it in the fast mode just do:

    $ BUNDLE_LIST=os-core bats -t test/real_content/test_swupd.bats

otaviobp:
oops, commented in the wrong PR

src/verify.c
```diff
@@ -818,6 +818,27 @@ int verify_main(int argc, char **argv)
...
+		if (force) {
+			if (cmdline_option_fix && cmdline_option_picky) {
```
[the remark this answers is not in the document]
otaviobp: That's right. Even bundle-list will list the bundle.

src/extra_files.c
```diff
@@ -90,29 +90,36 @@ static int qsort_helper(const void *A, const void *B)
...
+			fprintf(stderr, "Removing %s failed: %s\n", temp, strerror(errno));
+			ret = false;
 		}
 		free_string(&temp);
 	} else {
 		printf("%s%s\n", filename, is_dir ? "/" : "");
+		ret = false;
 	}
+
+	/* return true if the function deleted a file or directory, false otherwise */
```
icarus-sparry: I would put this comment just before the 'static bool handle' definition line, and then question if it was correct. In the case that you are attempting to remove a directory but fail you may well delete some files but still return false.

Perhaps a better comment would be "return true if the function deletes the specified filename, otherwise false".

Failure examples are hard, because you are running as root, but could happen. A race condition where a new file is added to a directory whilst `remove` is busy removing a subtree might cause the eventual removal of the directory to fail as it is not empty. A read-only file system might be mounted on a sub directory.

Thank you for the struct file_counts change - I think it is good.

castulo: I really appreciate your comments @icarus-sparry I/m learning so much. Thanks for reviewing. I'll update the comment as you suggested.

PR #654, opened 2018-10-24
makefile: Group variables and targets

otaviobp:
When mixing variables and targets make was attributing some variables as
targets. Because of that make install was running some tests defined in TEST
variable. Reordering all variable definitions to make sure all are defined
before first target definition.

Fix travis error in #653

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -93,13 +93,6 @@ systemdunit_DATA = \
 	data/swupd-update.timer \
 	data/verifytime.service
 
-distclean-local:
-	rm -rf aclocal.m4 ar-lib autom4te.cache config.guess config.h.in config.h.in~ config.sub configure depcomp install-sh ltmain.sh m4 Makefile.in missing compile
-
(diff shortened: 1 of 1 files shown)


Makefile.am
```diff
@@ -278,22 +285,15 @@ release:
...
-dist_man_MANS = \
-	$(MANPAGES)
-
 man: $(dist_man_MANS)
 
 .rst:
 	mkdir -p "$$(dirname $@)"
 	rst2man.py "$<" > "$@.tmp" && mv -f "$@.tmp" "$@"
 
-test/functional/completion/completion-basic.bats: swupd.bash
```
phmccarty: Did you intend to remove this line?

otaviobp: yes. I should have added a comment about that in the commit message.
swupd.bash was autogenerated a long time ago, but it isn't anymore, so we don't need this line

phmccarty: Okay, thanks.

Issue #486, opened 2018-06-12
Correctly report out-of-disk errors


phmccarty:
As mentioned in #266, swupd-client does not report out-of-disk errors very well (e.g. when `ENOSPC` is returned from a library call). Those errors should at least be reported correctly, especially if there is no recovery path to free up disk space.

PR #656, opened 2018-10-25
archives: Add output for out of space errors

evil-brain:
Libarchive's archive_error_string() method returns a textual
error message suitable for display.
However, this method returns only "Write Failed" if libarchive
runs out of hard disk space.

Check specifically for ENOSPC and print a correct
warning message to the user.

### DIFF
--- a/src/archives.c
+++ b/src/archives.c
@@ -23,6 +23,7 @@
 #include <archive.h>
 #include <archive_entry.h>
 #include <stdlib.h>
+#include <sys/errno.h>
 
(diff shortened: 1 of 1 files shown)


evil-brain:
I think this may fix https://github.com/clearlinux/swupd-client/issues/486 - unless we can think of any more side cases... libcurl reports a little weird (Error for $FILE download: Response 200 - Failed writing received data to disk/application), but I think that is a good enough pointer to check disk space...

PR #657, opened 2018-10-25
Improvements for DEBUG_TEST in test library

### LINKED ISSUE #651: Test leftovers when using DEBUG_TEST created interference with next executions
If I run:

    $ DEBUG_TEST=true make check

the tests directories are not going to be removed, as expected. But if I run the tests again I'll see some errors because of the left overs. We need to remove tmp directories before running a new test or we need to warn the user about that.

castulo:
When running tests, there is the possibility of preserving your
test environment for debugging purposes.

We were preserving the environment by skipping the execution of
the destroy_test_environment function when the DEBUG_TEST env
variable was set. This had one limitation, this was being done in
the default teardown function, so if the user would define his own
teardown then the DEBUG_TEST had no effect, the same if the user
used global_teardown instead. Another limitation the current
implementation had was that when an environment had been preserved
if the user would forget to clean it up manually before re-running
the test it would cause unexpected results in the test.

This commit removes both limitations by looking for the DEBUG_TEST
variable directly within the destroy_test_environment function so
the env will be preserved regardless of where the function is called
from and removes the environment at the beginning of the test if it
exists.

Closes #651

### DIFF
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -1,4 +1,4 @@
-#!bin/bash
+#!/usr/bin/bash
 
 FUNC_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 TEST_FILENAME=$(basename "$BATS_TEST_FILENAME")
@@ -175,7 +175,7 @@ print() { # swupd_function
 	fi
 
 	local msg=$1
-	echo "$msg" >&3
+	echo -e "$msg" >&3
 
 }
 
@@ -1364,6 +1364,9 @@ destroy_test_environment() { # swupd_function
 		cat <<-EOM
 			Usage:
 			    destroy_test_environment <environment_name>
+
+			Note: if you want your test environment to be preserved for debugging
+			      purposes, set DEBUG_TEST=true before running your test.
 			EOM
 		return
 	fi
@@ -1375,6 +1378,12 @@ destroy_test_environment() { # swupd_function
 	if [ ! -d "$env_name" ]; then
 		echo "Warning: test environment \"$env_name\" doesn't exist, nothing to destroy."
 		return 1
+	fi
+
+	if [ "$DEBUG_TEST" = true ]; then
+		local msg="Warning: DEBUG_TEST is set to true, the test environment will be preserved"
+		print "$msg\n" 2>/dev/null || echo "$msg"
+		return
 	fi
 
 	# since the action to be performed is very destructive, at least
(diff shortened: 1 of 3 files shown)


test/functional/testlib.bash
```diff
@@ -2315,6 +2324,12 @@ generate_test() { # swupd_function
 
 setup() {
 
+	# if a test environment already exists for the current test, remove it
+	# regardless of the value of DEBUG_TEST, this is important to avoid
+	# corrupt data
+	if [ -d "$TEST_NAME" ]; then
+		DEBUG_TEST=false destroy_test_environment "$TEST_NAME"
```
otaviobp: +1 for this

test/functional/testlib.bash
```diff
@@ -1377,6 +1380,12 @@ destroy_test_environment() { # swupd_function
 		return 1
 	fi
 
+	if [ "$DEBUG_TEST" = true ]; then
```
otaviobp: We still have one problem. If the bats file has 2 tests the second will remove the env from the first. 
Can't we create different envs for each test in the bats file?
like ${TEST_NAME}_1 or something similar?

castulo: yep, sounds like we could create the environment with something like this:
`"$TEST_NAME"_"$BATS_TEST_NUMBER"`

castulo:
some tests, are failing, I'll take a look at this

castulo:
There was an issue when cleaning up an old environment in the scenario where global_setup was being used, since the rest of the tests share one environment. To be able to fix this I had to implement the use of multiple environment, which is what you were referring to @otaviobp . It should work now.

test/functional/testlib.bash
```diff
@@ -2313,13 +2322,47 @@ generate_test() { # swupd_function
...
 setup() {
 
-	# the first time setup runs, run the global_setup
+	# a global environment will always use number 1, so check for that
+	# environment first
+	if [ -d "$TEST_NAME"_1 ]; then
+		# if a global environment is found we want to continue using that
+		# as our test environment for the rest of the tests, if not continue
+		# numbering the envs according to the test number
+		if [ -e "$TEST_NAME"_1/.global_env ]; then
```
jwakre: If I understand this correctly, only 1 test environment will be saved when using a global test setup, even when multiple tests exist. In this case, would it make sense to copy the state of "$TEST_NAME" to "$TEST_NAME"_"$BATS_TEST_NUMBER" before the teardown step for each test? This would save the state of each test before tearing it down and starting the next one

castulo: I guess we can do that @jwakre, I'm not sure how useful can that be but it should not hurt either. I'll include that.

PR #660, opened 2018-10-26
Test: file cleanup

### LINKED ISSUE #650: Tests get frozen if /tmp/swupd_test_certificates exists in the system
If you run:

    $ TRAVIS=true make check

And hit a ctrl-c during the execution of any certification tests the directory /tmp/swupd_test_certificates will be left behind and the user needs to remove the directory in order to continue. We should warn the user about that in a way that's easier to realize what's going on. With current implementation tests get frozen for a really long time and there's no info on what's going on.

Suggestion: Can't we use the swupd-client root folder instead of /tmp to store this? So it's easier to clean everything before running a new batch of tests.

jwakre:
Some tests (tests using web servers) wrote files to /tmp which were hard to clean up. This change moves the files/directories that were written to /tmp to more obvious locations. The trusted certificate store is now written to the top level directory of the repo and the other files are written to their corresponding test environment directories. Also this change adds a trap to clean up tests when they are interrupted.

Fixes #650

### DIFF
--- a/test/functional/bundleremove/remove-client-certificate.bats
+++ b/test/functional/bundleremove/remove-client-certificate.bats
@@ -2,15 +2,15 @@
 
 load "../testlib"
 
-server_pub="/tmp/$TEST_NAME-server-pub.pem"
-server_key="/tmp/$TEST_NAME-server-key.pem"
-client_pub="/tmp/$TEST_NAME-client-pub.pem"
-client_key="/tmp/$TEST_NAME-client-key.pem"
+server_pub="$PWD/$TEST_NAME"_1/server-pub.pem
+server_key="$PWD/$TEST_NAME"_1/server-key.pem
+client_pub="$PWD/$TEST_NAME"_1/client-pub.pem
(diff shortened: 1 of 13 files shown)


test/functional/bundleremove/remove-client-certificate.bats
```diff
@@ -60,13 +60,6 @@ global_teardown() {
 		return
 	fi
 
-	destroy_web_server
```
castulo: isn't destroying the web server still required though?

jwakre: Yes. I moved destroy_web_server into destroy_test_environment. This makes it easier to kill the web server with 'trap'.

jwakre:
I need to resolve an issue after rebasing. Attempting to put the certificate keys in $TEST_NAME, but it is changed to $TEST_NAME_<#> and is not updated in the certificate key path.

jwakre:
@otaviobp Your feedback should be addressed. Additionally, I updated the certificates paths to store them in directory named "$TEST_NAME"_1 instead of "$TEST_NAME"

Issue #662, opened 2018-10-29
Fix reported warnings from shellcheck in testlib


otaviobp:
Shellcheck is reporting some warnings on testlib and we should fix the ones that are relevant. If there are checks that doesn't apply, comment here.

PR #663, opened 2018-10-29
Shellcheck

otaviobp:


castulo:
So from what I see here, you are not including testilb.bash yet, right? until we fix all the current errors?

otaviobp:
> So from what I see here, you are not including testilb.bash yet, right? until we fix all the current errors?

yes. Issue: #662

PR #665, opened 2018-10-30
Adding the list of bundle-add tests

castulo:
Since every test should include a unique ID, it is necessary to add
the list of tests to a plain text document so it is easier to know
what is the following available ID when creating tests.

This commit also adds the ID to a bundle-add test that missing it.


otaviobp:
Nice :)

But I prefer to add a function to testlib that print that as output instead of committing the test file. And you could print just one output instead of one per file, like: 

    ==========
    Bundle-add
    ==========
    ADD001: Adding one bundle
    ADD002: Adding multiple bundles
    ADD003: Try adding one bundle that is already added
    
    =============
    Bundle-remove
    =============
    ....

castulo:
@otaviobp I made some changes based on your feedback. I updated the PR description to reflect this too.

castulo:
@jwakre this PR adds the functionality that you had suggested in a previous PR.

PR #670, opened 2018-11-02
Fixing a bug in the test list in Makefile

castulo:
The list is missing a "\" at the end of one test which is causing
a group of tests to be ignored by the makefile, therefore they are
not being run.


castulo:
Looks like one of the tests book by this commit 
https://github.com/clearlinux/swupd-client/commit/0f160138557c6902f1889a6bbb07df7f1e6810ff#diff-a7e08b5cc0d732f4df545b0edc5ae058 
I submitted a fixed patch

PR #672, opened 2018-11-02
Adding a missing line break in the verify command

castulo:
There is a missing line break that shows up When running
the "swupd verify" command and there is a hash mismatch.

The output looks like this:
```
$ sudo swupd verify
Verifying version 25590
Verifying files
   ...1%Hash mismatch for file: /usr/bin/tmux
   ...100%
Inspected 292278 files
  1 files did not match
Verify successful
```

This commit fixes the issue by adding a line break at the 
beginning of the output.

This PR depends on PR #670  as soon as that one merges I will rebase this one.


castulo:
Rebased

PR #667, opened 2018-10-31
Add consistency between verify and verify --fix

castulo:
Running "swupd verify" should be a dry-run of running
"swupd verify --fix", meaning that "swupd verify" should find
the same issues to be resolved than "swupd verify --fix" but it
should just inform about them while --fix should inform + fix them.
In occasions this does not happen, for example when there are
tracked files that need to be removed from the system.

Closes #607
Closes #536 

Signed-off-by: Castulo Martinez <castulo.martinez@intel.com>

PR #673, opened 2018-11-06
Adding ID for verify tests and add missing tests for verify --picky

### LINKED ISSUE #315: Add functional tests for verify --fix --picky
There are no tests at the moment for verify's --picky option. We should add tests to cover the special cases it handles.

castulo:
Adding tests for verify --picky

This commit adds some missing tests for "verify --picky", the rest
of the tests were added recently as part of other commits.

This commit also add the test ID to every verify test, and remove
duplicates.

This PR is built on top of #667, once that PR merges I will rebase this one.

Closes #315


Makefile.am
```diff
@@ -188,14 +188,11 @@ BATS = \
 	test/functional/update/update-verify-fix-path-hash-mismatch.bats \
 	test/functional/update/update-verify-fix-path-missing-dir.bats \
 	test/functional/update/update-verify-fullfile-hash.bats \
-	test/functional/verify/verify-add-missing-directory.bats \
```
otaviobp: The only thing I couldn't track on this pull request is where are the tests that replaces the tests in verify-add-missing-directory and verify-check-missing-directory

castulo: yes, I removed those tests because I added the tests in verify-missing-file: https://github.com/clearlinux/swupd-client/pull/673/files#diff-6d1105117307cb164dc89f95a6ee4110, and those tests include a missing file and a missing directory. In those tests file _/foo/bar_ is missing, so directory _/foo_ has to be added before file _bar_ can be added.

otaviobp: nice.

Issue #294, opened 2017-09-28
Be consistent about where output goes.


matthewrsj:
Progress output should go to stderr so as not to pollute stdout for processes that read swupd output.

EDIT: we should be consistent about what output we send to stderr or stdout throughout the entire project, not just progress output.

phmccarty:
A followup question here is "what should swupd output entail?". Most of its current output *is* progress output, so answering this question at the same time will also be useful.

matthewrsj:
That's a good point, I was mostly thinking about the cases like "swupd bundle-list" where the list should be the only output to go to stdout. This would probably be different for update etc. where the progress and a success message are most of the output.

icarus-sparry:
In general things should go to stdout if it makes sense for the output to be consumed, either by piping it to another program or using it in a command substitution (i.e. version=$(sudo swupd tell_me_the_version shells) as a hypothetical example). Things like progress meters are of no interest, ditto copyright messages (can we get rid of them by default and add a new option to show it?). It would also be good to have a quiet/verbose setting to turn off the noise.

So something like 'verify' can output to stdout the totals and the filenamess which are bad (preferably with some text which makes it easy to identify that this is a bad name so it doesn't get confused with a total if the bad name is "total: 12345").

matthewrsj:
I would like to rework the swupd CLI implementation and try to standardize swupd output across commands as much as possible. This would help with standardizing progress output. Perhaps this work would be best to wait until we have a plan-of-action for that.

matthewrsj:
@asifhg that is expected. We print the copyright header for all swupd operations.

phmccarty:
Also note that all of that output from `swupd check-update` goes to stderr, not stdout. So, if you are running the command in a script, you could do:

```
sudo swupd check-update 2> /dev/null
status=$?
case $status in
  0) sudo swupd update ;; # perform update
  1) break ;; # ignore, since no update available
  2) echo "check-update failed" ;; # handle fatal error
esac
```

otaviobp:
I disagree that this issue is low priority.

My first impression using swupd-client was that it was hard to understand what was going on because of the quality of the output messages. I've also heard similar complaints about it. And I'm very annoyed that there's no easy way to get errors messages and output messages redirected to different files.

Suggestions:
 - Review all output messages
 - send all errors to stderr and all other output (including progress, but we need to change the dot system in that case) to stdout
[...]

PR #677, opened 2018-11-07
Continue with update when a bundle was removed upstream

### LINKED ISSUE #624: Error updating if bundles went missing from mix.
Scenario:
- mix with bundle X
- remove X from mix
- swupd on client with bundle X installed
- swupd fails to update

```
# swupd update
Update started.
Preparing to update from 140 to 150
Warning: Bundle "xemacs" is invalid, skipping it...
Warning: Bundle "openssl" is invalid, skipping it...
Update took 0.2 seconds
# echo $?
14
```

This should probably continue, or print a really serious error message. Especially auto-updates should succeed at any cost...

castulo:
If a bundle happens to be removed from the content server (or mix) it
means the bundle won't be in the MoM anymore, so the bundle in the
system will look like an invalid bundle. When this happens the update
is currently stopping.

This PR fixes this so If this happens, the user is informed, and
the update continues.

Closes #624

### DIFF
--- a/src/update.c
+++ b/src/update.c
@@ -395,8 +395,14 @@ load_current_submanifests:
 	ret = add_included_manifests(server_manifest, &latest_subs);
 	timelist_timer_stop(global_times);
 	if (ret) {
-		ret = EMANIFEST_LOAD;
-		goto clean_exit;
+		if (ret == -add_sub_BADNAME) {
+			/* this means a bundle(s) was removed in a future version */
+			printf("WARNING: One or more installed bundles are no longer available at version %d.\n",
+			       server_version);
+		} else {
+			ret = EMANIFEST_LOAD;
+			goto clean_exit;
+		}
 	}
 
 load_server_submanifests:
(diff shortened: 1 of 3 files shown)


src/update.c
```diff
@@ -395,8 +395,14 @@ static int main_update()
 	ret = add_included_manifests(server_manifest, &latest_subs);
 	timelist_timer_stop(global_times);
 	if (ret) {
-		ret = EMANIFEST_LOAD;
-		goto clean_exit;
+		if (ret == -add_sub_BADNAME) {
```
evil-brain: I know these definitions are already merged, but it feels odd to have constant names mix case.

src/update.c
```diff
@@ -395,8 +395,14 @@ static int main_update()
...
+			/* this means a bundle(s) was removed in a future version */
+			printf("WARNING: One or more installed bundles are no longer available at version %d.\n",
```
evil-brain: Seems swupd prints most warnings to stderr. Not asking to change this in this commit, because there are other examples in the code where warnings are printed to stdout.

src/update.c
castulo: true

src/update.c
castulo: yes, there is an open issue that talks about that #294

PR #680, opened 2018-11-08
Output hints to check free space on download failures

### LINKED ISSUE #486: Correctly report out-of-disk errors

evil-brain:
Currently some of the output from swupd can be cryptic when
files fail to download. Add a hint to the user to check
their free space if Curl returns write errors.

Closes #486

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -448,6 +448,7 @@ exit:
 			break;
 		case CURLE_WRITE_ERROR:
 			fprintf(stderr, "Curl: Error downloading to local file - %s\n", filename);
+			fprintf(stderr, "Check free space for %s ?", state_dir);
 			err = -EIO;
 			break;
 		case CURLE_OPERATION_TIMEDOUT:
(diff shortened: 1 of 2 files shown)


src/curl.c
```diff
@@ -448,6 +448,7 @@ static int swupd_curl_get_file_full(const char *url, char *filename,
 			break;
 		case CURLE_WRITE_ERROR:
 			fprintf(stderr, "Curl: Error downloading to local file - %s\n", filename);
+			fprintf(stderr, "Check free space for %s ?", state_dir);
```
castulo: isn't this line missing the \n?

castulo: also, is it ok to have a space between the path and the "?" ?

evil-brain: Agreed on the \n, and there is one example already merged that doesn't use a space. I'll update to that standard.

otaviobp:
merged.

Issue #283, opened 2017-09-27
Add support for openssl 1.1


phmccarty:
Some distros have deprecated use of openssl 1.0.x, so it would be nice to add support for openssl 1.1.x to swupd-client.

tpepper:
Given the isolated usage of openssl, do you think it's viable to have this be a build time option?  Could enable testing and transition phase.

phmccarty:
@tpepper Yeah, I would be in favor of wrapping the transition code in the build time option. Then we can flip the switch to the new code whenever the time is right.

phmccarty:
@otaviobp There is currently some discussion about moving Clear Linux to openssl 1.1.x in the near future, so this issue should be on the radar.

ahkok:
This is now confirmed broken:
```
src/signature.c: In function 'terminate_signature':
src/signature.c:153:2: warning: 'ERR_remove_thread_state' is deprecated [-Wdeprecated-declarations]
  ERR_remove_thread_state(NULL);
  ^~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/openssl/e_os2.h:13,
                 from /usr/include/openssl/bio.h:13,
                 from src/signature.c:35:
/usr/include/openssl/err.h:259:1: note: declared here
 DEPRECATEDIN_1_1_0(void ERR_remove_thread_state(void *))
 ^~~~~~~~~~~~~~~~~~
src/signature.c: In function 'validate_certificate':
src/signature.c:399:19: error: dereferencing pointer to incomplete type 'X509_STORE_CTX' {aka 'struct x509_store_ctx_st'}
  return verify_ctx->error;
                   ^~
src/signature.c:400:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
```

ahkok:
Switching from enhancement to bug ;)

PR #682, opened 2018-11-09
[RFC] - Update signature verification for OpenSSL 1.1

tmarcu:
The ERR_remove_thread_state() function is deprecated as the thread
handling and many memory management operatoins have been re-written to
be internally handled by the API. The error code should be retrieved
by the proper function now from X509 store contexts.

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -150,7 +150,6 @@ void terminate_signature(void)
 	if (cert) {
 		cert = NULL;
 	}
-	ERR_remove_thread_state(NULL);
 	CRYPTO_cleanup_all_ex_data();
 }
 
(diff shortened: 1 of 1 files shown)


ahkok:
Neat! did you also verify this against OpenSSL-1.0?

castulo:
awesome, will this close issue #283?

otaviobp:
Really nice!!!

I tested that with libcryto 1.0 and it works fine, so we don't need to concern about compatibility, bumps, etc. But I had some problems with libcrypto 1.1 in last clear (26170).

I got the following warning

    /usr/bin/ld: warning: libcrypto.so.1.0.0, needed by /usr/lib64/gcc/x86_64-generic-linux/8.2.1/../../../../lib64/libcurl.so, may conflict with libcrypto.so.1.1

and a segmentation fault when running swupd. Last curl available is linked agains libcrypto 1.0 and not 1.1, so we need to wait for curl with 1.1 to fix that. Probably we are going to have that soon. Same problem with libarchive. So I would wait for them to merge.

ahkok:
> Last curl available is linked agains libcrypto 1.0 and not 1.1, so we 
> need to wait for curl with 1.1 to fix that.


I pushed a rebuild of curl against 1.1 earlier this afternoon to koji, 
so this will be very soon.

ahkok:
`libarchive` also getting updated now.

otaviobp:
I'm having this problem when running swupd verify with this patch:

    139928164279488:error:21075075:PKCS7 routines:PKCS7_verify:certificate verify error:crypto/pkcs7/pk7_smime.c:285:Verify error:unsupported certificate purpose

PR #686, opened 2018-11-09
Real content

otaviobp:
I've been running those tests for currently release and they are quite stable right now. Sending for review

### DIFF
--- a/test/real_content/real_content_lib.bash
+++ b/test/real_content/real_content_lib.bash
@@ -57,7 +57,7 @@ install_bundles() {
 		BUNDLE_LIST=$(echo "$output" | tr '\n' ' ')
 	fi
 
-	echo "Install bundles: $BUNDLE_LIST"
+	print "Install bundles: $BUNDLE_LIST"
 
 	run sudo sh -c "$SWUPD bundle-add $SWUPD_OPTS $BUNDLE_LIST"
 	assert_status_is 0
@@ -70,7 +70,7 @@ check_version() {
 	version="$1"
 	cur_version=$(grep VERSION_ID "${ROOT_DIR}/usr/lib/os-release" | cut -d = -f 2)
 	if [ "$version" -ne "$cur_version" ]; then
-		echo "Version $version is different from expected $cur_version"
+		print "Version $version is different from expected $cur_version"
 		return 1
 	fi
 }
@@ -79,8 +79,8 @@ verify_system() {
 
 	run sudo sh -c "$SWUPD verify --picky $SWUPD_OPTS 2>/dev/null"
 	if [ -n "$output" ]; then
-		echo "Verify found extra files in the system:"
-		echo "$output"
+		print "Verify found extra files in the system:"
+		print "$output"
 		return 1
 	fi
 
@@ -104,7 +104,7 @@ test_setup() {
 	# TODO: Discover available versions to work with custom mixers
 	read -r -d '\n' -a version_list < <(git ls-remote --tags https://github.com/clearlinux/clr-bundles.git | grep -v {} | grep -v latest | sed "s/.*refs\\/tags\\///" | sort -gr | head -n "$MAX_VERSIONS" | sort -g) || echo
 	if [ "${#version_list[@]}" -eq 0 ]; then
-		echo "Impossible to get versions list"
+		print "Impossible to get versions list"
 		return 1
 	fi
 
@@ -128,7 +128,7 @@ test_setup() {
 	# Not enough versions to run the test
 	if [ -z "${VERSION[0]}" ] || [ -z "${VERSION[1]}" ]; then
 		# TODO: This is going to break on first release in a new format
-		echo "We need at least 2 versions in format $FORMAT to continue with this test"
+		print "We need at least 2 versions in format $FORMAT to continue with this test"
 		return 1
 	fi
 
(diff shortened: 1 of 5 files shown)


test/real_content/swupd_update.bats
```diff
@@ -1,156 +1,13 @@
...
-	echo "Install bundles: $BUNDLE_LIST"
-
-	run sudo sh -c "$SWUPD bundle-add $SWUPD_OPTS --skip-diskspace-check $BUNDLE_LIST"
-	assert_status_is 0
-	verify_system
-}
-
-@test "test bundle-list --all" {
+# Test if bundle-list --all is working because all other tests depend on that
+@test "RC001: test bundle-list --all" {
```
jwakre: This file is called swupd_update, but this is a bundle-list test. Does it make sense to move this one?

test/real_content/swupd_update_fullfiles.bats
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bats
+
+load "real_content_lib"
+
+# Test massive fullfile downloads
+@test "RC004: Update from first to last with --fix" {
+	# shellcheck disable=SC2153
+	echo "Install minimal system with oldest version (${VERSION[0]})"
```
castulo: all these messages that you are printing with "echo" won't show on the test's output unless the test fails.
I would suggest to use the _print_ function provided by the test library so you can see those messages as part of the output when using _bats -t_ to run a test.

test/real_content/swupd_update.bats
otaviobp: The only reason I'm testing a bundle-list here is that if bundle-list is failing we can't trust the results of the tests. And I'd like to run that before next test.

test/real_content/swupd_update_fullfiles.bats
otaviobp: Thanks. I'll fix that

otaviobp:
Fixed proposed changes and squashed last commit with the one spliting files because it makes more sense to have them together

test/real_content/swupd_verify_install.bats
```diff
@@ -0,0 +1,25 @@
...
+	if [ -z "$BUNDLE_LIST" ]; then
+		run sudo sh -c "$SWUPD bundle-list --all $SWUPD_OPTS_SHORT -F $FORMAT"
+		BUNDLE_LIST=$(echo "$output" | tr '\n' ' ')
+	fi
+	for b in $BUNDLE_LIST; do
+		sudo touch "$ROOT_DIR/usr/share/clear/bundles/$b"
+	done
+
+	# shellcheck disable=SC2153
+	echo "Install complete system with newest version (${VERSION[$i]})"
```
castulo: you forgot one echo here

test/real_content/swupd_update.bats
castulo: It is not a good practice to have tests that depend from other tests, but sometimes it can be necessary. So if you think the result of tests RC002-04 depend on the good execution of RC001 maybe you can create a file like .RC001 at the end of the first test, and run the following tests only if .RC001 exists. Then you can remove .RC001 in the global_teardown (and in the global_setup in case it was left over for some reason).

test/real_content/swupd_verify_install.bats
otaviobp: Doh

test/real_content/real_content_lib.bash
```diff
@@ -0,0 +1,150 @@
+#!/bin/bash
+
+source "test/functional/testlib.bash"
```
jwakre: Why did this change from: load "../functional/testlib"?

With source, I can only run the tests from the top level directory

test/real_content/swupd_update.bats
otaviobp: Ok, maybe I wasn't clear in my comments about that test. Actually this test isn't really necessary and it looks out of scope for real content tests. We could remove it, but there's a reason I added that.

Almost all real content tests count on bundle-list --all to get a list of bundles to be tested. If by any change this command is broken and it's returning, for example, just one bundle. In this case we would run the tests only on that bundle and even if all tests are successful we can't assure that swupd is working fine for other cases.

So we don't need to run RC001 before any other test, so I didn't add it as a dependency (like we do with generate-cert). No sure if blocking other tests is the best way to go on this case.

Do you think we really need to have a strategy like you suggested.

test/real_content/real_content_lib.bash
otaviobp: I just got that. load is a command in bats and this is a bash script, so we need to use source. The path thing is very inconvenient, you are right about that. Is there any way to fix it? I really don't know.

jwakre: Yaaa, I see that now.... We could get the path to the script:

**SCRIPT_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"**

test/real_content/swupd_update.bats
otaviobp: I realized we don't need to have this test here. We can test that every time because it's not time consuming, so we can get better results. I'll remove RC001 and improve the other tests.

otaviobp:
I have reorganized the commits and fixed the addressed issues. I removed the test "list bundles" and added a check where this is used to make sure it's working fine.

test/real_content/real_content_lib.bash
```diff
@@ -57,7 +57,7 @@ install_bundles() {
 		BUNDLE_LIST=$(echo "$output" | tr '\n' ' ')
 	fi
 
-	echo "Install bundles: $BUNDLE_LIST"
+	print "Install bundles: $BUNDLE_LIST"
```
jwakre: @castulo @otaviobp I ran passing/failing tests and I do not see output from the 'print' command. Does bats actually show the output of 'print' or am I missing something here?

castulo: bats will show that output on passing test when using the _-t_ option (bats -t your_test.bats), or it will show it in failing tests without using -t

jwakre: ahhh I see

PR #685, opened 2018-11-09
Fix a bug in bundle-add --time

castulo:
This PR fixes a bug in bundle-add --time that was causing the command to show wrong values (and duplicated)

Closes #675

PR #692, opened 2018-11-14
Time: Add a fancy decoration to timelist output

otaviobp:
As suggested by @castulo on #685


otaviobp:
I'm closing this one because @castulo has a better solution

PR #693, opened 2018-11-14
verifytime: modify set_time() to take single argument

gvancuts:
Modify the set_time() function used in verifytime.c to take a single
argument instead of two. The second argument that was previously used
can easily be derived from the first and it avoids situations where
these two are not coherent (which happens to the case in the current
version of verifytime.c)

Signed-off-by: Geoffroy Van Cutsem <geoffroy.vancutsem@intel.com>


otaviobp:
nice. this is really a bug. thanks

PR #694, opened 2018-11-14
Prettifying the --time output

castulo:
Changing the way time is displayed when using the --time option so it
shows as a tree view so it is easier to see what items are sub-tasks
of other tasks.

### DIFF
--- a/src/timelist.c
+++ b/src/timelist.c
@@ -24,6 +24,7 @@
 #include <time.h>
 
 #include "macros.h"
+#include "swupd.h"
 #include "timelist.h"
 
 struct time {
@@ -31,7 +32,7 @@ struct time {
 	struct timespec procstop;
 	struct timespec rawstart;
 	struct timespec rawstop;
-	const char *name;
+	char *name;
 	bool complete;
 	TAILQ_ENTRY(time)
 	times;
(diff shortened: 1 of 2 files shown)


src/timelist.c
```diff
@@ -90,8 +96,32 @@ void timelist_timer_start(timelist *head, const char *name)
...
+		number_of_spaces = (count - 1) * 4;
+	}
+	char spaces[number_of_spaces];
+	for (int i = 0; i < number_of_spaces; i++) {
+		spaces[i] = ' ';
+	}
+	spaces[number_of_spaces] = '\0';
+
+	/* build the decorated name that will be inserted */
+	string_or_die(&decorated_name, "%s%s%s", spaces, count ? "|-- " : "", name);
```
otaviobp: Just a hack using "%*s". If you do this:

    string_or_die(&decorated_name, "%*s%s%s", number_of_spaces, "", count ? "|-- " : "", name);

otaviobp: And you need to free memory allocated by string_or_die on timelist_free.

castulo: > Just a hack using "%*s". If you do this:
> 
> ```
> string_or_die(&decorated_name, "%*s%s%s", number_of_spaces, "", count ? "|-- " : "", name);
> ```

Nice, I tried doing this before but it didn't work, but know I see what I was doing wrong. 😅 Thanks for the tip

castulo:
@otaviobp I made the changes you requested, thanks for reviewing 💪 

I would like to hold on to merging this PR until this other PR is merged though #641, I don't like submitting code without adding a test for the new code. I am not adding a test to this PR because it would duplicate the one in #641. Once #641 I will update that test with the changes in this PR and then we can merge it.

castulo:
I rebased and updated the _bundle-add --time_ test to include this format.

castulo:
Just removing a couple of shellcheck warnings

PR #696, opened 2018-11-15
Continue check-update when time is not correct

### LINKED ISSUE #678: swupd operations will fail when the platform time is not correct
swupd operations will fail when the platform time is not right. While it is right to fail (certificate validation fails) in such scenario, it is hard to track what caused the failure. Having platforms that do not have the time right is a regular occurrence when working in embedded development where some boards do not have an on-board battery (and typically have the date/time set to 1970).

It would be a great help to have swupd check the platform time and issue a warning when the date looks off (e.g. a date that is in the 70's or a date that is in the past compared to the swupd client build time).

castulo:
When the system time is way off, the certificate validation will
fail. For this reason, many swupd commands attempt to fix the
system time if wrong. check-update was not doing this so the
command would fail in this situation.

This commit adds a verification for this so check-update attempts
to fix the system time and continue instead of failing.

Closes #678 

### DIFF
--- a/src/check_update.c
+++ b/src/check_update.c
@@ -29,6 +29,7 @@
 #include <unistd.h>
 
 #include "swupd.h"
+#include "verifytime.h"
 
 static void print_help(const char *name)
 {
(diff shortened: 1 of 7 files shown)


castulo:
@otaviobp  I added the -I and -C options to check-update

src/check_update.c
```diff
@@ -42,8 +42,10 @@ static void print_help(const char *name)
 	fprintf(stderr, "   -F, --format=[staging,1,2,etc.]  the format suffix for version file downloads\n");
 	fprintf(stderr, "   -x, --force             Attempt to proceed even if non-critical errors found\n");
 	fprintf(stderr, "   -n, --nosigcheck        Do not attempt to enforce certificate or signature checking\n");
+	fprintf(stderr, "   -I, --ignore-time       Ignore system/certificate time when validating signature\n");
 	fprintf(stderr, "   -p, --path=[PATH...]    Use [PATH...] as the path to verify (eg: a chroot or btrfs subvol\n");
 	fprintf(stderr, "   -S, --statedir          Specify alternate swupd state directory\n");
+	fprintf(stderr, "   -C, --certpath          Specify alternate path to swupd certificates\n");
```
otaviobp: nice catch. check-update wasn't working with alternative certificates.

castulo:
@otaviobp I just modified the PR to use the verify_time() function as requested.

PR #705, opened 2018-11-16
verifytime: Fix return value of verify_time function

otaviobp:
There was some confusion about shell return code and bools.
Shell returns 0 for true and bools are 0 for false.

### DIFF
--- a/src/verifytime_main.c
+++ b/src/verifytime_main.c
@@ -20,5 +20,9 @@
 
 int main()
 {
-	return verify_time() ? 1 : 0;
+	if (verify_time()) {
+		return 0; // Success
+	}
+
+	return 1; // Error
 }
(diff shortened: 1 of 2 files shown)


otaviobp:
@castulo Can you test if this is really fixing the problem?

castulo:
> @castulo Can you test if this is really fixing the problem?

Looks like it does! 👍

src/verifytime_main.c
```diff
@@ -20,5 +20,5 @@
 
 int main()
 {
-	return verify_time() ? 1 : 0;
+	return verify_time() ? 0 : 1;
```
castulo: I think it is a good idea to add comments in both files explaining why you are returning false instead of true since it is confusing, so we don't forget why is like that in the future

otaviobp: ok. that's true

PR #707, opened 2018-11-16
locale: Set locale to system locale

### LINKED ISSUE #445: Warning: Linkname can't be converted from UTF-8 to current locale.
When extracting symlinks with non-ASCII (UTF-8) characters in the name, swupd (actually libarchive) complains about converting even if the current locale is UTF-8.

We first investigated whether Mixer was generating something wrong, but it seems to be generating valid tar files. That investigation is in https://github.com/clearlinux/mixer-tools/issues/92. The next stop is looking a libarchive, as the error message seems misleading.

For debugging libarchive, a starting point:

```
<thiago> if anyone wnats to debug this, put a breakpoint in archive_entry.c line 1301 with condition r != 0
<thiago> the debug fuse doesn't work for me (container)
```

otaviobp:
Standard locale on C applications is C and we use in clearlinux UTF-8
for file names. Libarchive checks the program locale to convert
filenames so we need to set this.

According to setlocale documentation, setlocale(LC_ALL, "") sets the locale
to system default (in the case of ClearLinux, UTF-8).

More information on the libarchive problem:
https://github.com/libarchive/libarchive/wiki/Filenames and
https://github.com/libarchive/libarchive/issues/587

Fixes #445

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -19,6 +19,7 @@
 
 #define _GNU_SOURCE // for basename()
 #include <getopt.h>
+#include <locale.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -157,6 +158,12 @@ int main(int argc, char **argv)
 	int index;
 	int ret;
 
+	/* Set locale to system locale
+	 * Change from the standard (C) to system locale, so libarchive can
+	 * handle filename conversions.
+	 * More information on the libarchive problem:
+	 * https://github.com/libarchive/libarchive/wiki/Filenames */
+	setlocale(LC_ALL, "");
 	if (parse_options(argc, argv, &index) < 0) {
 		return EINVALID_OPTION;
 	}


src/main.c
```diff
@@ -157,6 +158,12 @@ int main(int argc, char **argv)
 	int index;
 	int ret;
 
+	/* Set locale to system locale
+	 * Change from the standard (C) to system locale, so libarchive can
+	 * handle filename conversions.
+	 * More information on the libarchive problem:
+	 * https://github.com/libarchive/libarchive/wiki/Filenames */
+	setlocale(LC_ALL, "");
```
jameshilliard: I would recommend using `LC_CTYPE` here instead of `LC_ALL` as libarchive should only care about the `LC_CTYPE` and setting `LC_ALL` could cause unexpected side effects, I've found on some embedded linux systems that `LC_ALL` doesn't work while `LC_CTYPE` does.

PR #713, opened 2018-11-20
Removing option --force where is not needed

### LINKED ISSUE #636: Option --force in bundle-add doesn't do anything
bundle-add --help show the --force option but it doesn't really do anything for bundle-add, we should probably remove this option or change its description, there are probably no problems that are not critical when installing a bundle.

Usage:
   swupd bundle-add [options] [bundle1 bundle2 (...)]

Help Options:
   -h, --help              Show help options
   -u, --url=[URL]         RFC-3986 encoded url for version string and content file downloads
   -c, --contenturl=[URL]  RFC-3986 encoded url for content file downloads
   -v, --versionurl=[URL]  RFC-3986 encoded url for version string download
   -P, --port=[port #]        Port number to connect to at the url for version string and content file downloads
   -p, --path=[PATH...]    Use [PATH...] as the path to verify (eg: a chroot or btrfs subvol
   -F, --format=[staging,1,2,etc.]  the format suffix for version file downloads
   **-x, --force             Attempt to proceed even if non-critical errors found**
   -n, --nosigcheck        Do not attempt to enforce certificate or signature checking
   -I, --ignore-time       Ignore system/certificate time when validating signature
   -S, --statedir          Specify alternate swupd state directory
   -C, --certpath          Specify alternate path to swupd certificates
   -t, --time              Show verbose time output for swupd operations
   -N, --no-scripts        Do not run the post-update scripts and boot update tool
   -b, --no-boot-update    Do not update the boot files using clr-boot-manager
   -D, --max-parallel-downloads=[n] Set the maximum number of parallel downloads
   --skip-diskspace-check  Do not check free disk space before adding bundle

castulo:
The --force option is included in bundle-add, bundle-remove,
check-update, update and verify, but it is really only used in
verify.

This commit removes the option from the commands that are not using
it so it is not misleading to users. If at some point we need to
implement somthing with --force it can be added back at that point.

Closes #636

### DIFF
--- a/src/check_update.c
+++ b/src/check_update.c
@@ -41,7 +41,6 @@ static void print_help(const char *name)
 	fprintf(stderr, "   -v, --versionurl=[URL]  RFC-3986 encoded url for version string download\n");
 	fprintf(stderr, "   -P, --port=[port #]     Port number to connect to at the url for version string and content file downloads\n");
 	fprintf(stderr, "   -F, --format=[staging,1,2,etc.]  the format suffix for version file downloads\n");
-	fprintf(stderr, "   -x, --force             Attempt to proceed even if non-critical errors found\n");
 	fprintf(stderr, "   -n, --nosigcheck        Do not attempt to enforce certificate or signature checking\n");
 	fprintf(stderr, "   -I, --ignore-time       Ignore system/certificate time when validating signature\n");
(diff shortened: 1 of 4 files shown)


bryteise:
+1 for the idea here, I'd like to see --force never come back though and instead be --force=thing1,thing2 just as a request.

Issue #640, opened 2018-10-19
Path with a single quote fails bundle-add


evil-brain:
Example:
`sudo -E swupd bundle-add -t --path="/testing'" --contenturl=https://cdn.download.clearlinux.org/update --versionurl=https://cdn.download.clearlinux.org/update --format=25 --statedir=/home/blovin/swupd_verify_testing/testing/var/lib/swupd vim`

results in this failure:
```
Downloading packs...
        ...100%
Starting download of remaining update content. This may take a while...
        ...100%
Finishing download of update content...
Installing bundle(s) files...
        ...40%sh: -c: line 0: unexpected EOF while looking for matching `''
sh: -c: line 1: syntax error: unexpected end of file
Path /usr/share/vim/vim81/doc is missing on the file system ... fixing
sh: -c: line 0: unexpected EOF while looking for matching `''
sh: -c: line 1: syntax error: unexpected end of file
Error: Path /usr/share/vim/vim81/doc failed to stage in verify_fix_path
sh: -c: line 0: unexpected EOF while looking for matching `''
sh: -c: line 1: syntax error: unexpected end of file
Failed to install 1 of 1 bundles
```

otaviobp:
One question, do you really have one path called /testing' ?

Related to #573

phmccarty:
Related issue: #671

castulo:
🍾

PR #714, opened 2018-11-20
Global

### LINKED ISSUE #571: Make it easier to add a global option
Swupd documentation states that there are some options that are applicable to most subcommands. We could have a better way to handle parameters so we wouldn't need to duplicate them in all subcommands that use them.
Currently even help documentation on what that parameter is doing is duplicated.

Commit 15b987c is one example on how it's hard to maintain global options.

### LINKED ISSUE #573: Sanitize all user variables read through command line parameters
Some sanitization is already done, but we could improve that to prevent errors when invalid parameters are used.

otaviobp:
There is a list of parameters in swupd that are really configuration changes that applies to almost every command (and they should apply to any command that uses any feature that this setting is configuring).

But those parameters were being set specifically for each swupd command, making it a hard task to keep documentation updated and make sure that all parameters are really added to all commands that may benefit from using them.

When I was reviewing parameters sanitization I decided to take some time to close #571 to make our lives easier.

So this pull request consolidate parameters in global.c and fixes some sanitization problems I found.

Note that we have a problem related to weird path names (#640) but this isn't a sanitization problem because that path is valid. 

Fixes: #571

### DIFF
--- a/src/clean.c
+++ b/src/clean.c
@@ -26,20 +26,19 @@
 
 #include "swupd.h"
 
-static void print_help()
+static void print_help(void)
 {
 	fprintf(stderr,
 		"Usage:\n"
-		"	swupd clean [options]\n"
+		"   swupd clean [OPTION...]\n\n"
 		"\n"
 		"Remove cached content used for updates from state directory.\n"
 		"\n"
 		"Options:\n"
 		"   --all                   Remove all the content including recent metadata\n"
 		"   --dry-run               Just print files that would be removed\n"
-		"   -S, --statedir DIR      Specify alternate swupd state directory\n"
-		"   -h, --help              Display this help message\n"
 		"\n");
+	global_print_help();
 }
 
 static struct {
@@ -56,35 +55,29 @@ static struct timespec now;
 static const struct option prog_opts[] = {
 	{ "help", no_argument, 0, 'h' },
 	{ "all", no_argument, &options.all, 1 },
-	{ "statedir", required_argument, 0, 'S' },
 	{ "dry-run", no_argument, &options.dry_run, 1 },
-	{ 0, 0, 0, 0 }
+};
+
+static const struct global_options opts = {
+	prog_opts,
+	sizeof(prog_opts) / sizeof(struct option),
+	NULL,
+	print_help,
 };
 
 static bool parse_options(int argc, char **argv)
 {
-	int opt;
-	while ((opt = getopt_long(argc, argv, "hS:", prog_opts, NULL)) != -1) {
-		switch (opt) {
-		case '?':
-			return false;
-		case 'h':
-			print_help();
-			exit(0);
-		case 'S':
-			if (!optarg || !set_state_dir(optarg)) {
-				fprintf(stderr, "Invalid --statedir argument\n\n");
-				return false;
-			}
-			break;
-		case 0:
-			/* Handle options that don't have shortcut. */
-			break;
-		default:
-			fprintf(stderr, "Error: unrecognized option: -%c,\n\n", opt);
-			return false;
-		}
+	int optind = global_parse_options(argc, argv, &opts);
+
+	if (optind < 0) {
+		return false;
 	}
+
+	if (argc > optind) {
+		fprintf(stderr, "Error: unexpected arguments\n\n");
+		return false;
+	}
+
 	return true;
 }
 
@@ -387,6 +380,7 @@ static void clean_deinit(void)
 int clean_main(int argc, char **argv)
 {
 	if (!parse_options(argc, argv)) {
+		print_help();
 		return EXIT_FAILURE;
 	}
 
--- a/src/clr_bundle_add.c
+++ b/src/clr_bundle_add.c
@@ -37,163 +37,65 @@
 
 static char **bundles;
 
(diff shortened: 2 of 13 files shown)


src/search.c
```diff
@@ -341,7 +341,7 @@ static void print_help()
 	fprintf(stderr, "   -l, --library           Search paths where libraries are located for a match\n");
 	fprintf(stderr, "   -b, --binary            Search paths where binaries are located for a match\n");
 	fprintf(stderr, "   -s, --scope=[query type] 'b' or 'o' for first hit per (b)undle, or one hit total across the (o)s\n");
-	fprintf(stderr, "   -t, --top=[NUM]         Only display the top NUM results for each bundle\n");
+	fprintf(stderr, "   -T, --top=[NUM]         Only display the top NUM results for each bundle\n");
```
matthewrsj: Technically a breaking change to the interface, but since this is a not-often-scripted subcommand maybe we can get away with *not* bumping the MAJOR version according to semver. Thoughts?

src/clean.c
```diff
@@ -30,16 +30,15 @@ static void print_help()
 {
 	fprintf(stderr,
 		"Usage:\n"
-		"	swupd clean [options]\n"
+		"	swupd clean [OPTION...]\n\n"
```
castulo: the usage message for swupd check-update is using spaces to indent, in this case you are using a tab, it would be nice to be consistent in indentation across all the usage messages.

src/clean.c
```diff
@@ -30,16 +30,15 @@ static void print_help()
...
 		"\n"
 		"Remove cached content used for updates from state directory.\n"
 		"\n"
 		"Options:\n"
 		"   --all                   Remove all the content including recent metadata\n"
 		"   --dry-run               Just print files that would be removed\n"
-		"   -S, --statedir DIR      Specify alternate swupd state directory\n"
-		"   -h, --help              Display this help message\n"
 		"\n");
```
castulo: isn't this \n going to cause the options to be separated by an empty line? Like this?

```
--all                   Remove all the content including recent metadata
--dry-run               Just print files that would be removed

-S, --statedir DIR      Specify alternate swupd state directory
-h, --help              Display this help message
```

src/clr_bundle_add.c
```diff
@@ -37,163 +37,65 @@
 
 static char **bundles;
 
-static void print_help(const char *name)
+static void print_help()
```
castulo: in the other function declaration of print_help you used _void_ as argument and here you are leaving it empty, I would prefer we use one way or the other consistently.

src/clr_bundle_add.c
```diff
@@ -37,163 +37,65 @@
...
-	{ "max-parallel-downloads", required_argument, 0, 'D' },
 	{ "skip-diskspace-check", no_argument, &skip_diskspace_check, 1 },
-	{ 0, 0, 0, 0 }
+};
+
+static bool parse_opt(int opt, UNUSED_PARAM char *optarg)
+{
+	switch (opt) {
+	case 'l':
+		fprintf(stderr, "error: [-l, --list] option is deprecated, use\n"
```
castulo: I see sometimes you are using _error_ like in this case and some times you are using _Error_ like in line 77 of clean.c, again I'd prefer if we start using these in a consistent manner, I think this kind of small inconsistencies change the perception of quality for the end user.

src/globals.c
```diff
@@ -559,6 +559,7 @@ static const struct option global_opts[] = {
 	{ "time", no_argument, 0, 't' },
 	{ "url", required_argument, 0, 'u' },
 	{ "versionurl", required_argument, 0, 'v' },
+	{ "max-parallel-downloads", required_argument, 0, 'D' },
```
castulo: Maybe add a TODO here with the comment that D was deprecated and we should remove this option from the base code in X months

src/globals.c
```diff
@@ -584,10 +585,15 @@ static bool global_parse_opt(int opt, char *optarg)
 		set_content_url(optarg);
 		return true;
 	case 'P':
-		if (sscanf(optarg, "%ld", &update_server_port) != 1) {
+		if (sscanf(optarg, "%ld%c", &update_server_port, &tmp) != 1) {
```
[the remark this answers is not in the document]
otaviobp: Because I'd like to check if there's something else trailing in the string. If I used '/0' the scanf would also return 1 if the user added a value like "123abc"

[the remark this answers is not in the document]
otaviobp: but scanf isn't a good way on checking for overflows. I'll remove this call and use the strtoint_err funciton

src/clr_bundle_add.c
otaviobp: I'll use the "Error" because it's more used in swupd in general. Thanks for noticing that.

src/clean.c
otaviobp: true

otaviobp:
Pushed new version addressing reported problems.
I removed the commit that really sanitize the parameters because I'll stop using scanf and I'll send that in a new pull request.

otaviobp:
I pushed the sanitize commit back because I realized I could do it in this PR. It will get better after #722.

PR #715, opened 2018-11-20
Move the MoM from bundles dir to /var/tmp/swupd

### LINKED ISSUE #632: bundles in bash completion stop working after a verfiy --picky --fix
Swupd copies the most current MoM to /usr/share/clear/bundles/.MoM so we can use it for bash completion. The problem is that this directory is tracked by clearlinux and because of that "swupd verify --picky" will list this file as an extra file. And "swupd verify --fix --picky" will remove that file.

Fix: Move this file to a directory not tracked by swupd. The ideal would be /var/lib/swupd/, but regulars users can't read that dir, so it won't work (#490)

castulo:
swupd copies the latest MoM to /usr/share/clear/bundles/.MoM so we
can use it for bash completion. The problem is that this directory
is tracked by clearlinux and because of that "swupd verify --picky"
will list this file as an extra file. And "swupd verify --fix --picky"
will remove that file.

This commit fixes the issue by changing the location where the .MoM
is copied to being this an untracked location, so the file won't be
removed.

Closes #632

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,5 @@ test/functional/*/*/test.log
 test/functional/*/*/debug.log
 test/functional/*/*/state
 test/functional/*/*/lines-output
-test/functional/*/*/target-dir/usr/share/clear/bundles/.MoM
 test/functional/*/*.log
 verifytime
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -594,13 +594,21 @@ verify_mom:
 	/* Make a copy of the Manifest for the completion code */
 	if (latest) {
 		char *momcopy;
-		rm_bundle_file(".MoM");
-		string_or_die(&momcopy, "/bin/cp \"%s\" \"%s/%s/.MoM\" 2>/dev/null",
-			      filename, path_prefix, BUNDLES_DIR);
+		char *momdir;
+		char *momfile;
+
+		string_or_die(&momdir, "%s/var/tmp/swupd", path_prefix);
+		string_or_die(&momfile, "%s/.MoM", momdir);
+		swupd_rm(momfile);
+		mkdir_p(momdir);
+		string_or_die(&momcopy, "/bin/cp \"%s\" \"%s\" 2>/dev/null",
+			      filename, momfile);
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wunused-result"
 		(void)system(momcopy); /* If it fails it fails */
 #pragma GCC diagnostic pop
+		free_string(&momdir);
+		free_string(&momfile);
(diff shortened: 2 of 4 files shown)


configure.ac
```diff
@@ -284,6 +284,7 @@ AH_TEMPLATE([LOCK_DIR], [Directory for lock file])
 AH_TEMPLATE([BUNDLES_DIR], [Directory to use for bundles])
 AH_TEMPLATE([CERT_PATH], [Location of update certificate])
 AH_TEMPLATE([MOTD_FILE], [motd file path])
+AH_TEMPLATE([MOM_DIR], [Directory to copy the MoM used for the completion code])
```
otaviobp: If you set this to anything different from /var/tmp/swupd/ the bash script will fail. I would let this hardcoded on swupd because otherwise we would need to generate a bash script with that info.

src/manifest.c
```diff
@@ -594,13 +594,17 @@ struct manifest *load_mom(int version, bool latest, bool mix_exists, int *err)
 	/* Make a copy of the Manifest for the completion code */
 	if (latest) {
 		char *momcopy;
-		rm_bundle_file(".MoM");
-		string_or_die(&momcopy, "/bin/cp \"%s\" \"%s/%s/.MoM\" 2>/dev/null",
-			      filename, path_prefix, BUNDLES_DIR);
+		char *momdir;
+		string_or_die(&momdir, "%s/%s", path_prefix, MOM_DIR);
+		swupd_rm(momdir);
```
otaviobp: We don't need to remove the directory. this is going to fail if there's something else in there. Can't we just remove the .MoM file?

castulo: actually it won't fail because swupd_rm will delete the directory and its content. The .MoM file is the only file in that directory at the moment. Having said that it is probably better to only remove the file as you suggests in case we want to add more stuff to that directory in the future. I'll make the change

configure.ac
castulo: ok

src/manifest.c
otaviobp: Oh, I didn't know that. Nice. But anyway it's better to delete just the file.

.gitignore
```diff
@@ -58,6 +58,5 @@ test/functional/*/*/test.log
 test/functional/*/*/debug.log
 test/functional/*/*/state
 test/functional/*/*/lines-output
-test/functional/*/*/target-dir/usr/share/clear/bundles/.MoM
```
phmccarty: You also removed this line in #716... just a warning that either this PR or that one will need a rebase :)

PR #717, opened 2018-11-21
sys: Add run_command() function to replace some usages of system()

otaviobp:
System API is a very convenient way on executing external commands, but it has
some problems for some usages. System always launches a shell to execute the
desired command, so we can be affected by environment variables (like PATH) and
shell globbing.

We can avoid most of the environment variables affecting the swupd result by
using the full path of the program being executed, but making sure the shell
won't expand all wildcards is tricky and specially dangerous for some commands
that we need to run, like cp, mkdir, etc.

So to make sure no globbing will be performed we can use fork+execve to run
commands instead of system.

### DIFF
--- a/src/sys.h
+++ b/src/sys.h
@@ -7,6 +7,31 @@ extern "C" {
 
 extern long get_available_space(const char *path);
 
+/* run_command_quiet: runs a command redirecting the standard and error outputs
+ * redirected to /dev/null. */
+#define run_command_quiet(...) run_command_full("/dev/null", "/dev/null", __VA_ARGS__)
+
+/* run_command: runs a command with standard and error output (stdout and stderr) set */
+#define run_command(...) run_command_full(NULL, NULL, __VA_ARGS__)
+
+/* run_command_full: Run command cmd with parameters informed in a NULL
+ * terminated list of strings.
+ *
+ * Return a negative number (-errno) if we are unable to create a new thread to
+ * run the command and a positive exit code (0-255) if the command was executed.
+ * If there's an error in the execution of the program 255 may also be returned
+ *
+ * Notes:
+ * - This function doesn't execute on a shell like system() so we won't have any
+ *   issue with parameters globbing.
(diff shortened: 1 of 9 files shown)


castulo:
Very nice

src/sys.h
```diff
@@ -7,6 +7,31 @@ extern "C" {
 
 extern long get_available_space(const char *path);
 
+/* run_command: runs a command redirecting the standard and error outputs
```
evil-brain: Minor nit, but this comment probably means to say "run_command_quiet: runs a command...."

src/sys.h
```diff
@@ -7,6 +7,31 @@ extern "C" {
...
+
+/* run_command: runs a command with standard and error output (stdout and stderr) set */
+#define run_command(...) run_command_full(NULL, NULL, __VA_ARGS__)
+
+/* run_command_full: Run command cmd with parameters informed in a NULL
+ * terminated list of strings.
+ *
+ * Return a negative number (-errno) if we are unable to create a new thread to
+ * run the command and a positive exit code (0-255) if the command was executed.
+ * If there's an error in the execution of the program in 255 may also be returned
```
evil-brain: excess 'in' before 255

src/sys.h
```diff
@@ -32,6 +32,12 @@ extern long get_available_space(const char *path);
  */
 int run_command_full(const char *stdout_file, const char *stderr_file, const char *cmd, ...);
 
+/* copy_all: Runs cp -a [src] [dst] using run_command_quiet */
+int copy_all(const char *src, const char *dst);
+
+/* copy_all: Runs cp [src] [dst] using run_command_quiet */
```
evil-brain: excess _all after copy

evil-brain:
HUGE readability improvement, very nice! Couple minor nits with comments, but this is great.

otaviobp:
rebased and fixed the reported problems

PR #722, opened 2018-11-27
helpers: Check for trailing characters in strtoi_err

otaviobp:
Besides checking for overflow, also checks if the number in the string
doesn't have any trailing character. For example "123abc" will return
an error and "123 abc" won't.

Also:
 - Don't return EOVERFLOW anymore, because from the user of strtoi_err
perspective it doesn't matter if the overflow is in the int size or in
the long size. I would expect the same return code.


src/helpers.c
```diff
@@ -622,17 +627,44 @@ int strtoi_err(const char *str, char **endptr, int *value)
...
+{
+	char *endptr;
+	int err = strtoi_err_endptr(str, &endptr, value);
+
+	if (err) {
+		return err;
+	}
+
+	if (*endptr != '\0' && !isspace(*endptr)) {
+		return -ERANGE;
```
jwakre: Is this supposed to be -EINVAL?

otaviobp: Good point. It makes more sense to use EINVAL. I'll change that

Issue #669, opened 2018-11-01
If clearlinux default server goes down, swupd won't update from alternative mirror


otaviobp:
If you set the mirror using "swupd mirror --set" command and then tries to update your system, swupd checks if the mirror isn't very outdated. In this process, if the upstream mirror is down, swupd will fail to update from the alternative mirror.

And swupd check-update will work, so versions printed are divergent.

castulo:
I cannot reproduce the bug.

If the upstream content server is down I get a negative value for _central_version_ here -> https://github.com/clearlinux/swupd-client/blob/master/src/mirror.c#L169) which causes a negative diff which will cause the mirror to be used, so everything works fine.

otaviobp:
That's true.

I realized that the problem is when the server is not responding. In that case we got stuck on that line and as the connection was already established, curl won't timeout. We need to investigate if there's a flag we could use on curl to timeout if no data was received in a period of time.

otaviobp:
Sorry. It looks like I did something wrong in my tests. Curl is timing out, so the server will fail after curl timeout. The problem is only the one you reported.

castulo:
I was finally able to recreate the original problem you were describing. The problem is being caused by two things:
1. When running any swupd operation one of the first thing swupd does is to run soup_init to initialize itself. One of the things swupd_init does is to test the connection with the server, however it does not set any timeout for curl before attempting the test, therefore if the server is not responding curl hangs forever.
2. If a mirror is set and responding, and the server is hung, we should really allow swupd to continue since it should not need the server since the mirror is set.

castulo:
yeah, I have a patch ready, the only problem is that the tests are built on top of some PRs that have not merged. I will submit it like this and will rebase when those ones are merged.

PR #724, opened 2018-11-28
Fixes a bug in update when using mirrors

castulo:
There are a couple of problems when the user has a mirror set up,
and tries to update. The first one is that in the case where the
user is using the -p option and a mirror, the version of the mirror
was failing to be retrieved since the path_prefix was being added
to the path twice. The second problem was that when the mirror being
used by the user was stale, the mirror was being unset but some values
had already being cached, causing the update function to fail.

This PR fixes the issues and add a few tests to validate the correct behavior.

### DIFF
--- a/test/functional/verify/verify-fix.bats
+++ b/test/functional/verify/verify-fix.bats
@@ -143,16 +143,18 @@ test_setup() {
 		.deleted
 		--picky removing extra files under .*/target-dir/usr
 		REMOVING /usr/untracked_file3
-		Inspected 19 files
+		REMOVING /usr/share/defaults/swupd/versionurl
+		REMOVING /usr/share/defaults/swupd/contenturl
+		Inspected 21 files
 		  2 files were missing
 		    2 of 2 missing files were replaced
 		    0 of 2 missing files were not replaced
 		  2 files did not match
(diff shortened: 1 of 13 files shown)


castulo:
These bugs where found while trying to reproduce bug #669

test/functional/verify/verify-fix.bats
```diff
@@ -143,16 +143,18 @@ test_setup() {
 		.deleted
 		--picky removing extra files under .*/target-dir/usr
 		REMOVING /usr/untracked_file3
-		Inspected 19 files
+		REMOVING /usr/share/defaults/swupd/versionurl
```
otaviobp: All tests with changes on this commit should be in the 1st commit of this series, right?

castulo: > All tests with changes on this commit should be in the 1st commit of this series, right?

ooh you are right, I'll move those to the first commit

castulo:
yes @otaviobp we don't have tests for those flags, it'll be nice to add those, and I prefer to have explicit tests for them than "testing" them as a side effect of using them anyway

castulo:
@otaviobp I moved the changes on the affected tests to the first commit as requested

PR #728, opened 2018-11-29
Update verify to include purpose for cert check

vaclave:
Newer versions of openssl require a valid purpose so
pass this in for the cert check.

Signed-off-by: David Klimesh <david.j.klimesh@intel.com>


matthewrsj:
@vaclave @otaviobp just so you know this shows up on master branch as a commit authored by @clrbuilder because @vaclave's email address must be associated with that account. Fortunately the commit author and signoff are correct, looks like just a github UI oddity.

Issue #676, opened 2018-11-07
Show information that a bundle is experimental in swupd operations


otaviobp:
Prepend a '*' everytime we print the name of an experimental bundle.

Commands affected:
 - bundle-*
 - search

And in bundle-add, shows a warning saying that we are installing an experimental bundle.

Depends on: https://github.com/clearlinux/mixer-tools/issues/497

castulo:
Update: 
We should have a header "status" in the bundle manifest
if _status: experimental_ then the  bundle is experimental
if there's no status or status != from experimental, it isn't experimental

castulo:
Change of plans... having the _status: experimental_ in the header of each bundle's manifest would cause clients to have to download all the manifests only to figure out which ones are experimental. A much efficient approach is to have a new modifier in the MoM for that bundle.

Example:
`M..e 4b91c3122e7e32f1e3edb0a705a6168ca8e66f597c6f89ff32cfffd977afb5e5 10 test-bundle`

So we no longer need to have that in each bundle's manifest but only in the MoM.

castulo:
Once the 4 PRs above get merged this issue can be closed

PR #729, opened 2018-11-29
Add ability to parse "experimental" bundles in MoM

castulo:
Mixer will soon start using a new modifier for bundles in a MoM that are marked
as experimental. These bundles will include a "e" in the 4th position of the
modifiers of a manifest, like shown here:

`M..e 4b91c3122e7e32f1e3edb0a705a6168ca8e66f597c6f89ff32cfffd977afb5e5 10 test-bundle`

This PR adds the mechanism to parse that element from the MoM and store it in
memory. In the future this will allow swupd to treat this experimental bundles
differently if required.


castulo:
This PR is part of the implementation of #676

src/manifest.c
```diff
@@ -299,6 +299,8 @@ static struct manifest *manifest_from_file(int version, char *component, bool he
 		} else if (c[3] == 'm') {
 			file->is_mix = 1;
 			manifest->is_mix = 1;
+		} else if (c[3] == 'e') {
```
otaviobp: is this already defined with mixer-tools team? If so, I think we're good to merge this

castulo: > is this already defined with mixer-tools team? If so, I think we're good to merge this

@tmarcu @reaganlo can you guys confirm this? ^

reaganlo: @castulo Not yet. Will keep you posted when done.

matthewrsj: It is fine that it is not implemented yet, this change can still be pushed in to swupd-client as long as the mixer team agrees it is the correct approach.

matthewrsj: This approach has a +1 from me fwiw.

castulo: @reaganlo it does not need to be implemented yet, I was only asking if that is the way it is going to work. We can merge this without the implementation on the mixer side, it should not affect swupd client at all.

reaganlo: @castulo aaa I see. It looks fine. But can you please hold on to merging it until tomorrow. I want to first test and ensure that it does not break anything. Currently working on some high priority Coverity tasks.

matthewrsj: It will not break anything mixer side. This is just fixing the client to not fail if it encounters the flag.

castulo:
ups, actually I forgot to add the test to the Makefile  :sweat_smile:

src/manifest.c
reaganlo: @matthewrsj I was not referring to testing this swupd-client change. What I meant was testing the corresponding implementation in mixer.

castulo:
There it is

src/manifest.c
matthewrsj: I don't think we need to wait. Adding flags to manifests is not complicated. It is something we've done before many times.

PR #730, opened 2018-11-29
Distinguish experimental bundles in a bundle list

castulo:
This PR adds the ability to show an identifier in bundles that are
marked as experimental when listing available bundles or installed
bundles:
```
swupd bundle-list --all
swupd bundle-list
```

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -1130,10 +1130,27 @@ clean_and_exit:
  */
 int list_local_bundles()
 {
+	char *name;
 	char *path = NULL;
 	struct list *bundles = NULL;
 	struct list *item = NULL;
+	struct manifest *MoM = NULL;
+	struct file *bundle_manifest = NULL;
+	int current_version;
+	bool mix_exists;
 
+	current_version = get_current_version(path_prefix);
+	if (current_version < 0) {
+		goto skip_mom;
+	}
+
+	mix_exists = (check_mix_exists() & system_on_mix());
+	MoM = load_mom(current_version, false, mix_exists, NULL);
+	if (!MoM) {
+		printf("Warning: Could not determine which installed bundles are experimental\n");
+	}
+
+skip_mom:
 	string_or_die(&path, "%s/%s", path_prefix, BUNDLES_DIR);
 
 	errno = 0;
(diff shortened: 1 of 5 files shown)


castulo:
This PR is built on top of PR #729 , once that one merges this will have to be rebased.
This is part of the work required for #676

src/bundle.c
```diff
@@ -65,8 +65,9 @@ int list_installable_bundles()
 	while (list) {
 		file = list->data;
 		list = list->next;
-		printf("%s\n", file->filename);
+		printf("%s%s\n", file->is_experimental ? "*" : "", file->filename);
```
matthewrsj: instead of the ambigous `*` with the legend below why not just do this inline:
`bundlename (experimental)`

castulo: @matthewrsj sounds good to me, I was using the * because that is what it was on #676 but I like it better the way you are suggesting it

castulo: I'll make the change

castulo:
@matthewrsj I did the requested change.

castulo:
I just realized there is a bit more work that needs to be done for this PR. Please hold on on reviewing it until further notice.

castulo:
ok, now is ready to be reviewed

src/bundle.c
otaviobp: It's a lot better with the " (experimental)" string

castulo:
Just rebased the code now that #729 was merged.

src/bundle.c
```diff
@@ -1108,12 +1108,30 @@ static int lex_sort(const void *a, const void *b)
...
+	current_version = get_current_version(path_prefix);
+	if (current_version < 0) {
+		goto skip_mom;
+	}
+
+	mix_exists = (check_mix_exists() & system_on_mix());
+	MoM = load_mom(current_version, false, mix_exists, NULL);
+	if (!MoM) {
+		printf("Warning: Could not determine which installed bundles are experimental\n");
+		goto skip_mom;
```
otaviobp: you don't need this goto

castulo: true, the code look different before and it needed it, but at some point it changed and forgot to remove it. Thanks for noticing it.

src/clr_bundle_ls.c
```diff
@@ -188,22 +188,19 @@ int bundle_list_main(int argc, char **argv)
...
+	 * listing locally installed bundles (with the limitation of not showing what
+	 * bundles are experimental) */
+	if (ret != 0 && !cmdline_local) {
 		fprintf(stderr, "Error: Failed updater initialization. Exiting now\n");
 		return ret;
 	}
 
-	if (cmdline_option_deps != NULL) {
+	if (cmdline_local) {
+		set_path_prefix(NULL);
```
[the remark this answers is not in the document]
castulo: I think you are right, this is being already set by swupd_init... I'll remove it

PR #733, opened 2018-11-30
Warn user when adding an experimental bundle

castulo:
This commit adds the ability of warning the user if he/she tries
to install an experimental bundle.


castulo:
This PR is built on top of PR #729 , once that one merges this will have to be rebased.
This is part of the work required for #676

castulo:
Just rebased the code now that #729 was merged.

PR #738, opened 2018-12-07
Reorg

otaviobp:


castulo:
What about helpers.c? shouldn't that be moved to src/lib/ too?

otaviobp:
> What about helpers.c? shouldn't that be moved to src/lib/ too?

Eventually yes. But there are several functions in helpers that are aren't generic and are tightly coupled with other swupd code (like verify_fix_path, is_compatible_format, is_current_version...). And even some functions that looks generic, like untar_full_download, are not.

PR #740, opened 2018-12-10
Setting upstream server in client cert tests

castulo:
The tests that attempt to verify the client certificate were using the
swupd -u option to set the url of the "upstream server". Recently a
function was added to the test library to set the default upstream
server.

This commit modifies these tests so they used the new function to set
the upstream server instead of using the -u option. This will provide
the benefit of the test environment being more similar to a prod
environment, also makes the test setup less convoluted and the tests
easier to read.

This PR is built on top of PR #739, once that PR merges this will have to be rebased.

### DIFF
--- a/test/functional/bundleadd/add-client-certificate.bats
+++ b/test/functional/bundleadd/add-client-certificate.bats
@@ -29,6 +29,11 @@ global_setup() {
 	create_trusted_cacert "$server_pub"
 
 	start_web_server -c "$client_pub" -p "$server_pub" -k "$server_key"
+
+	# Set the web server as our upstream server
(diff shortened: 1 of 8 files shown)


jwakre:
I don't think there is still a reason for the check-update tests to be different than the other client certificates tests. Does it make sense for the check-update tests to follow this new test format?

castulo:
> I don't think there is still a reason for the check-update tests to be different than the other client certificates tests. Does it make sense for the check-update tests to follow this new test format?

You are absolutely right. Very good catch. I guess I missed them because I most have searched for the SWUPD_OPTS_HTTP variable and those ones aren't using it. I'll update that test in a new commit. Thanks for noticing that.

PR #745, opened 2018-12-12
docs: Add swupd-alias.7.rst to repository

otaviobp:
We have defined that we are committing to the repository the rst files,
so I'm adding the alias rst file


bryteise:
Normally I don't see autoconf generated content checked into the repo but I guess that isn't a particularly bad for configure and the like from make dist so +1.

Issue #746, opened 2018-12-12
Misleading error message when swupd falls back on pre-built trust store


gtkramer:
On Clear Linux, if the following commands are run:

```
rm -rf /var/cache/ca-certs
swupd check-update
```

The following error message is seen:
```
Error: unable to verify server SSL certificate.
```

But if the following command is run:

```
clrtrust generate
```

The error message goes away.

Swupd should fall back on the stateless trust store (/usr/share/ca-certs/.prebuilt-store) if the stateful trust store (/var/cache/ca-certs) does not exist or cannot be used for whatever reason.  Upon proper working of the fallback, the above error message would not be present.

gtkramer:
After more testing, it was discovered that this is a misleading error message.  Swupd is able to finish an update or a verify install.  This means it's falling back on the prebuilt trust store.  What should really be happening is swupd should emit a warning that it is falling back on the pre-built store when validation with the stateful store fails.  Only when validation with the stateless store fails should swupd error out and not proceed.

PR #747, opened 2018-12-12
curl: Don't warn users on non critical SSL errors

otaviobp:
If we have an ssl error we try to use a fallback capath directory. Don't warn
users for errors unless we have failed on connecting using all available
fallback paths.

Fixes: #746


otaviobp:
tests were updated and we're passing everything now

PR #749, opened 2018-12-13
Timeout for testing connection in swupd init

### LINKED ISSUE #669: If clearlinux default server goes down, swupd won't update from alternative mirror

castulo:
As part of the swupd initialization, curl is initialized, which means it
is configured and the connection is tested. When configuring curl for
this test, no timeout was being configured, which was causing curl to
hang waiting for a response from the server if the server was unresponsive.

This commit configures the curl timeouts the same way they are being
configured when downloading files, so in case we get an unresponsive
server (or mirror) it doesn't get stuck forever.

Closes #669

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -91,11 +91,34 @@ exit:
 	return curl_ret;
 }
 
+static CURLcode swupd_curl_set_timeouts(CURL *curl)
+{
+	CURLcode curl_ret;
+
+	curl_ret = curl_easy_setopt(curl, CURLOPT_CONNECTTIMEOUT, SWUPD_CURL_CONNECT_TIMEOUT);
+	if (curl_ret != CURLE_OK) {
+		goto exit;
+	}
+
+	curl_ret = curl_easy_setopt(curl, CURLOPT_LOW_SPEED_LIMIT, SWUPD_CURL_LOW_SPEED_LIMIT);
+	if (curl_ret != CURLE_OK) {
+		goto exit;
+	}
+
+	curl_ret = curl_easy_setopt(curl, CURLOPT_LOW_SPEED_TIME, SWUPD_CURL_RCV_TIMEOUT);
+	if (curl_ret != CURLE_OK) {
+		goto exit;
+	}
+
+exit:
+	return curl_ret;
+}
+
 static int check_connection(const char *test_capath)
 {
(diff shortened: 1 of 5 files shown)


src/curl.c
```diff
@@ -105,6 +128,11 @@ static int check_connection(const char *test_capath)
 		return -1;
 	}
 
+	curl_ret = swupd_curl_set_timeouts(curl);
```
otaviobp: I was reviewing curl.c source code and I think it's better to call swupd_curl_set_basic_options()  on check_connection to use all other configurations for curl as any other connection

castulo: _swupd_curl_set_basic_options()_ receives a url as parameter. Since this is part of the swupd init I can use version_url or content_url.

castulo:
The slow-server test is failing to initialize curl (getting error ECURL_INIT, cannot initialize curl agent)

src/curl.c
```diff
@@ -116,6 +139,11 @@ static int check_connection(const char *test_capath)
 		return -1;
 	}
 
+	curl_ret = swupd_curl_set_basic_options(curl, content_url);
```
otaviobp: Use version_url here so you can remove the line 121 (set CURLOPT_URL).
You can remove swupd_curl_set_optional_client_cert() call too, because this is also set by swupd_curl_set_basic_options()

castulo: you are right, actually there were another couple more options that were already set in swupd_curl_set_basic_options(), I removed those too

otaviobp:
Merging is always nice, but not sure if it's an easy task. 
It looks like a copy and paste from do_HEAD method from SimpleTest to support that in TestServer. if this is the case, I would go with the 2nd alternative

otaviobp:
oops, I didn't realize SImpleTest depended on hang_server. It looks easier to go with the 1st alternative then

castulo:
> oops, I didn't realize SImpleTest depended on hang_server. It looks easier to go with the 1st alternative then

what do you mean SimpleTest depended on hang_server?

PR #754, opened 2018-12-14
Add space to separate 'mixin build' command

jwakre:
The run_command calls /usr/bin/mixinbuild, but it should call
/usr/bin/mixin build.

### DIFF
--- a/src/update.c
+++ b/src/update.c
@@ -270,9 +270,7 @@ version_check:
 				fclose(verfile);
 			}
 
-			if (run_command("/usr/bin/mixin"
-					"build",
-					NULL) != 0) {
+			if (run_command("/usr/bin/mixin", "build", NULL) != 0) {
 				fprintf(stderr, "ERROR: Could not execute mixin\n");
 				ret = EXIT_FAILURE;
 				goto clean_curl;


src/update.c
```diff
@@ -270,7 +270,7 @@ static int main_update()
 				fclose(verfile);
 			}
 
-			if (run_command("/usr/bin/mixin"
+			if (run_command("/usr/bin/mixin "
```
phmccarty: Hmm, sorry, I reviewed too quickly. I think the intention was for `"/usr/bin/mixin"` and `"build"` to be separate arguments (because `run_command_full` accepts varargs).

Does the code work with the separate arguments (i.e. `run_command("/usr/bin/mixin", "build", ...)`)

jwakre: Yes, you are right. I will update this.

src/update.c
```diff
@@ -270,7 +270,7 @@ static int main_update()
 				fclose(verfile);
 			}
 
-			if (run_command("/usr/bin/mixin"
+			if (run_command("/usr/bin/mixin",
```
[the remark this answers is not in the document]
jwakre: Sure, this is done now.

PR #755, opened 2018-12-15
Sync mixin version with upstream during update

jwakre:
When updating to a new upstream version, the mixin version should be
reset so that it is based on the new upstream version. Removing the
'.valid-mix' file before calling the 'mixin build' command will reset
the mixin version to the new upstream version.

### DIFF
--- a/src/update.c
+++ b/src/update.c
@@ -220,6 +220,7 @@ static int main_update()
 	bool mix_exists;
 	bool re_update = false;
 	bool versions_match = false;
+	char *mixin_verfile = NULL;
 
 	srand(time(NULL));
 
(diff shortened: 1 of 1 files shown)


jwakre:
This PR addresses the following issue from mixer:

https://github.com/clearlinux/mixer-tools/issues/367

It is much easier to implement the solution on swupd

src/update.c
```diff
@@ -270,6 +271,15 @@ static int main_update()
...
+			/* The .valid-mix file contains the mixin version. When updating
+			 * to a new upstream version, erase this file so that mixin can
+			 * recreate it with a version based on the new upstream. */
+			string_or_die(&mixin_verfile, "%s/%s", MIX_DIR, ".valid-mix");
+			if (remove(mixin_verfile)) {
+				fprintf(stderr, "ERROR: Unable to remove %s\n", mixin_verfile);
+			}
+			free_string(&mixin_verfile);
+
 			if (run_command("/usr/bin/mixin"
```
otaviobp: If we have a problem in this line (power goes down, there's any problem in mixin execution, the user ctrl-c swupd) we won't have a .valid-mix file in the system anymore. Isn't this a problem when trying to run swupd again?

jwakre: You're right about power loss and killing swupd causing a problem here. They would have to run mixin build again to re-create the .valid-mix file.

PR #757, opened 2018-12-17
Add swupd development guide

otaviobp:
Add a simple development guide for swupd. We can maintain that as long as we
add new development rules to swupd project.

### DIFF
--- a/docs/how_to_contribute.md
+++ b/docs/how_to_contribute.md
@@ -1,0 +1,53 @@
+# How to contribute
+
+Swupd is open for code contributions. Nevertheless, if you are planning to propose a new feature, it's suggested that you file an issue first so we can discuss the feature idea and the proposed implementation. This way we can avoid future conflicts with other planned features.
+
+When your code is ready, open a pull request to have your patches reviewed. Patches are only going to be merged after being reviewed and approved by at least 2 maintainers (with some exceptions for trivial patches), all functional tests passes and enough tests are added to test the proposed functionality or bug fix.
+
+You should also add tests to the feature or bug fix you are proposing. That's how we ensure we won't have software regressions or we won't break any functionality in the future. For more information about function tests, take a look at the [functional test documentation](test/functional/README.md)
+
+## Error handling
+
+Always check and handle errors. And if possible prefer to have a fallback than to fail swupd execution. Never use abort()(unless on out of memory errors) or assert(), on errors swupd should always print the corresponded error and return an appropriated error code documented on swupd-error.h.
+
+## Swupd Code Stye
+
+Swupd code style is defined on .clang-format and checked with "make compliant". Make sure you are complying with the code style before creating a pull request.
+
+### Comments
+
+You are going to see different comment styles in swupd codebase. We are working on improving the quality of function comments on header files, so for new code, always comment all new exported (non static) functions you create in the header file. It's also well appreciated if you add comments to undocumented functions you have changed.
+
+ - Use /* */ for multiline or single line header comments. It's ok to use // for comments in the code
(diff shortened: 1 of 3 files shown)


docs/how_to_contribute.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Error handling
+
+Always check and handle errors. And if possible prefer to have a fallback than to fail swupd execution. Never use abort()(unless on out of memory errors) or assert(), on errors swupd should always print the corresponded error and return an appropriated error code documented on swupd-error.h.
+
+## Swupd Code Stye
+
+Swupd code style is defined on .clang-format and checked with "make compliant". Make sure you are complying with the code style before creating a pull request.
+
+### Comments
```
otaviobp: we haven't discussed that a lot, so please, give me feedback about this topic. We can change the proposed style if we want that.

docs/how_to_contribute.md
```diff
@@ -0,0 +1,53 @@
+# How to contribute
+
+Swupd is open for code contributions. Nevertheless, if you are planning to propose a new feature, it's suggested that you file an issue first so we can discuss the feature idea and the proposed implementation. This way we can avoid future conflicts with other planned features.
+
+When code your code is ready, open a pull request to have your patches reviewed. Patches are only going to be merged after being reviewed and approved by at least 2 maintainers (with some exceptions for trivial patches), all functional tests passes and enough tests are added to test the proposed functionality or bug fix.
```
castulo: I think there is an extra "code" word in this phrase: 'When code your code is ready,"

docs/how_to_contribute.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Line Length
+
+Avoid extremely long lines. Don't use lines longer than 80~100 characters.
+
+## Swupd generic library
+
+Library on [src/lib](src/lib) was created to group the implementation of
+generic data types, helpers and system interaction used by swupd, but without
+any swupd specific code or swupd.h dependency. Try to place any
```
castulo: This paragraph is incomplete.

docs/how_to_contribute.md
```diff
@@ -0,0 +1,53 @@
...
+char *strdup_or_die(const char *const str);
+
+...
+return NO_ERRORS; // No errors reported
+...
+```
+
+### Function names
+
+Always use meaningful function names, specially non-static functions. Avoid plurals and use imperative verbs. Add a prefix of the component/module to what this function refers to (usually the name of the header file). For example, bundle_sort_files() is a better name than sorts_bundle_files() or sort_files_in_bundle().
```
castulo: I'm ok with this, but it is not the way it is now, is it?

castulo:
Also, maybe you can add something about printing messages to the screen. For example if they should use capital letters for _warning: some message_ or _Error: Some message_ and perhaps also when to use printf vs fprintf...

docs/how_to_contribute.md
otaviobp: oops

docs/how_to_contribute.md
otaviobp: there are some files that uses this (curl, download, list, hashmap), but most code that handle core operations aren't following any standard. I don't think we should try to fix everything, but at least we should try to follow a standard for new code

otaviobp:
> Also, maybe you can add something about printing messages to the screen. For example if they should use capital letters for warning: some message or Error: Some message and perhaps also when to use printf vs fprintf...

We have one ticket to define that (#294). We can add this as soon as we have a better definition on what will we do.

otaviobp:
Merged with proposed fixes.

PR #758, opened 2018-12-17
Log - curl downloads

otaviobp:
I have a local code with prints I use to debug some download bugs. I didn't commit that before because it's too verbose to print it always.

This series introduce a new log API and 2 new flags (--debug and --quiet) and a commit using that to print download debug information. As those flags are experimental and are not completely finished (--quiet is far from quiet in current master) I didn't added that to the documentation or help messages.

### DIFF
--- a/src/lib/log.h
+++ b/src/lib/log.h
@@ -1,0 +1,40 @@
+#ifndef __SWUPD_LOG__
+#define __SWUPD_LOG__
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define LOG_ALWAYS 0
+#define LOG_ERROR 2
+#define LOG_WARN 4
+#define LOG_INFO 6
+#define LOG_DEBUG 8
+
+/*
+ * Set the minimum priority log to be printed by log functions.
+ */
+void set_log_level(int log_level);
+
+/*
+ * Print a message to 'out' if 'log_level' is greater or equal the log level
+ * set by set_log_level().
+ */
+void log_full(int log_level, FILE *out, const char *file, int line, const char *label, const char *format, ...);
+
+/*
+ * Helpers for logging: print(), error(), warn(), info(), debug()
+ */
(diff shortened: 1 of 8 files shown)


src/lib/log.h
```diff
@@ -0,0 +1,44 @@
...
+void log_full(int log_level, FILE *out, const char *str, ...);
+
+/*
+ * Helpers for logging: print(), error(), warn(), info(), debug()
+ */
+#define print(_fmt, ...) log_full(LOG_ALWAYS, stdout, _fmt, ##__VA_ARGS__);
+#define error(_fmt, ...) log_full(LOG_ERROR, stderr, "Error: " _fmt, ##__VA_ARGS__);
+#define warn(_fmt, ...) log_full(LOG_WARN, stderr, "Warning: " _fmt, ##__VA_ARGS__);
+#define info(_fmt, ...) log_full(LOG_INFO, stdout, _fmt, ##__VA_ARGS__);
+#define debug(_fmt, ...) log_full(LOG_DEBUG, stdout, "DEBUG (" __FILE__ ":" STR(__LINE__) "): " _fmt, ##__VA_ARGS__);
```
castulo: What about adding the time to the DEBUG output? couldn't that be useful?

otaviobp: Nice idea. I had a different idea now about the log names. I'll rename info() to msg() and debug() to info(). and print "INFO ()...." instead of debug. And rename --debug to --verbose. What do you think?

So --quiet will show print() and error()
Standard will show print(), error(), warn(), msg()
--verbose will show print(), error(), warn() , msg(), info()

Is this better?

castulo: well aren't debug, info, warning, error more standard?

otaviobp: ok. I'll leave the way it it then.
--quiet shows print() and error()
standard shows print(), error(), warn(), info()
--debug shows print(), error(), warn(), info(), debug()

otaviobp:
Sent a new version adding the current timestamp. Also printing now the timestamp, file and line on every log when --debug is used.

PR #759, opened 2018-12-18
Hashmap bugs

otaviobp:


src/lib/hashmap.c
```diff
@@ -47,7 +47,7 @@ size_t hashmap_hash_from_string(const char *key)
 
 static inline struct list **get_hashmap_list(struct hashmap *hashmap, const void *data)
 {
-	return &hashmap->map[hashmap->hash(data) || HASH_MASK(hashmap->mask_bits)];
+	return &hashmap->map[hashmap->hash(data) & HASH_MASK(hashmap->mask_bits)];
```
jwakre: Is the commit message backwards ("Logical and instead of binary or was used")? Binary or was used and you replaced it with logical and.

otaviobp: you are right and wrong. My message is wrong, but the correct is: "Logical OR instead of a binary AND was used". I'll fix that

PR #760, opened 2018-12-18
Makefile: Only run shellcheck on scripts that have changed in current…

otaviobp:
… branch

Use origin/master branch to check which bash and bats scripts have changed and
run shellcheck only on them. This intends to make travis check faster.


.travis.yml
```diff
@@ -28,6 +28,7 @@ install:
         - sudo apt-get install clang-format-6.0
         - sudo apt-get install shellcheck
         - sudo ln -s /usr/share/docutils/scripts/python3/rst2man /usr/bin/rst2man.py
+        - git fetch origin master:refs/remotes/origin/master #Download origin/master for shelcheck
```
castulo: no big deal but you have a typo in the comment shelcheck instead of shellcheck

PR #762, opened 2018-12-19
Do not retry files download if there is no more space on disk

### LINKED ISSUE #752: If download fails on a full disk, abort instead of retrying
When debugging #751 we realized that when the disk gets full on a file download, curl return an error code and we schedule a retry for that download. In this case the best approach would be to abort the operation, because next retry will fail for sure.

castulo:
When a disk gets full on a file download during an update, we are currently scheduling up to five retries for that download, which are going to fail for sure since there is no more space on disk.
This PR changes this behavior so it only tries to download the file once if we receive an error about the disk being full so we don't waste time.

The PR includes these commits:

- A commit with the necessary updates in the test library so we can create test environments that are limited in space.
- A commit that includes tests to validate the different scenarios where the disk can get filled up during an update.
- The changes to the swupd code to abort download retries when the disk is full.

Closes #752 

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -166,11 +166,6 @@ BATS = \
 	test/functional/checkupdate/chk-update-version-match.bats \
 	test/functional/completion/completion-basic.bats \
 	test/functional/hashdump/hashdump-file-hash.bats \
-	test/functional/verify/verify-install.bats \
-	test/functional/verify/verify-install-bad.bats \
-	test/functional/verify/verify-install-multiple.bats \
-	test/functional/verify/verify-install-no-fullfile-fallbacks.bats \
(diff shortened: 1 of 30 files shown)


otaviobp:
nice commit. with this feature we can also add tests for bundle-add with large bundles.
I will just wait for a test using it to get this merged, to make sure it's really working fine.

castulo:
I'm working on having this working for bundle-add too...

castulo:
There are some strange failures on the tests included in test/functional/search/search-no-disk-space.bats. I'll look into it, in the mean time please hold on reviewing this PR. 
**WIP**

castulo:
The PR is now ready for review

castulo:
Rebased and added one more improvement to the search test scripts.

PR #764, opened 2018-12-19
CURL: Fix http error code handling

otaviobp:
Fix http error code handling for async downloads.

WIP: Still missing tests for this PR. I'm working on that, but I sent that to speed up review.


otaviobp:
I'm closing this pull request. I realized there's a bug on download resume, so it's better to have ppl reviewing this later, when the problem is fixed.

Issue #562, opened 2018-09-05
Parallel downloads is broken for large files


otaviobp:
When swupd is downloading large files in parallel and one package is taking to long to be extracted, downloads can get timed out and operation fails.

This bug was introduced on 3.17.13.

Subtasks:
 - Perform downloads in parallel using multiple threads
 - Create a command line option to control the number of simultaneous downloads so we can have a fallback if the parallel download fails for some other reason

evil-brain:
Otavio and I had a quick discussion about how to do this.

Downloads are already handled by curl_mutli - so items are already downloaded in parallel.

Decompression is what we're going to be threading. This basically means we will enqueue work to a threadpool and return immediately in the curl callback. This will necessitate an extra check for errors from each thread, and retry if something fails.

By running decompression in threads, swupd will spend less time waiting, and this will prevent curl from timing out.

otaviobp:
fixed by #767

PR #767, opened 2018-12-21
Http bug fixes

### LINKED ISSUE #598: Slow internet -> unclear if swupd just wastes my bandwidth and retries entire download several times
My internet sucks. It's so slow that almost any swupd bundle add, even smaller ones, cause swupd to timeout, even though the downloads progress. As a result, the output suggests that swupd throws the partial files out and starts from scratch, wasting bandwidth and time.

```
# sudo swupd bundle-add keepassxc 
Downloading packs...
	...100%
Error for /var/lib/swupd/pack-keepassxc-from-0-to-24550.tar download: Response 200 - Transferred a partial file
Starting download retry #1 for https://cdn.download.clearlinux.org/update//24550/pack-keepassxc-from-0.tar
Error for /var/lib/swupd/pack-keepassxc-from-0-to-24550.tar download: Response 206 - Transferred a partial file
Starting download retry #2 for https://cdn.download.clearlinux.org/update//24550/pack-keepassxc-from-0.tar
Error for /var/lib/swupd/pack-keepassxc-from-0-to-24550.tar download: Response 206 - No error
Starting download retry #3 for https://cdn.download.clearlinux.org/update//24550/pack-keepassxc-from-0.tar
Error for /var/lib/swupd/pack-keepassxc-from-0-to-24550.tar download: Response 206 - No error
Starting download retry #4 for https://cdn.download.clearlinux.org/update//24550/pack-keepassxc-from-0.tar
Error for /var/lib/swupd/pack-keepassxc-from-0-to-24550.tar download: Response 206 - Transferred a partial file
Starting download retry #5 for https://cdn.download.clearlinux.org/update//24550/pack-keepassxc-from-0.tar
Error for /var/lib/swupd/pack-keepassxc-from-0-to-24550.tar download: Response 206 - Timeout was reached
	...100%
Installing bundle(s) files...
	...100%
Successfully installed 1 bundle
```

This is a bug: If the download wasn't stalled, but is just slow, it shouldn't cause the download to interrupt and error, instead it should just continue downloading since data is readily flowing.

Also, the download should be retried if possible at the point it gave an error. It's entirely unclear if it did.

A resumed download is a warning at best, as well.

otaviobp:
fixes some problems in http downloads:
 - There were some http codes that aren't processed the way they should
 - Fix problem in download resume
 - Inform users when a download is resuming (request on #598)
 - Reduce the number of downloads in parallel when there are connection errors (like timeout) in downloads. Retry using 27, 9, 3, 1, 1, 1 downloads in parallel (requested on #598) 

Fixes #598 #562


src/curl.c
```diff
@@ -311,6 +311,99 @@ CURLcode swupd_download_file_complete(CURLcode curl_ret, struct curl_file *file)
...
+enum download_status process_curl_error_codes(int curl_ret, CURL *curl_handle, int *err)
+{
+	char *url;
+	if (curl_easy_getinfo(curl_handle, CURLINFO_EFFECTIVE_URL, &url) != CURLE_OK) {
+		url = "<not available>";
+	}
+
+	if (curl_ret == CURLE_OK || curl_ret == CURLE_HTTP_RETURNED_ERROR) {
+		long response = 0;
+		if (curl_ret == CURLE_OK || curl_ret == CURLE_HTTP_RETURNED_ERROR) {
```
jwakre: This condition is identical to the one on line 324, so it will always evaluate to true.

otaviobp: True. this is a left over. I'll remove this conditional

src/curl_async.c
```diff
@@ -549,6 +565,9 @@ int swupd_curl_parallel_download_end(void *handle, int *num_downloads)
 				list_free_item(l, NULL);
 				l = next;
 
+				// Retry was probably scheduled because of network problems, so
+				// reevaluate the number of parallel downloads
+				reevaluate_number_of_parallel_downloads(h, file->retries);
```
jwakre: I'm just getting familiar with the download code, so I may be wrong here.

It looks like there can be a list of failed downloads that can each call reevaluate_number_of_parallel_downloads, even though they failed with the same MAX_XFER value. This would allow a few download failures to drop the MAX_XFER value from 27 -> 1 instead of incrementally.

src/curl.c
```diff
@@ -311,6 +311,99 @@ CURLcode swupd_download_file_complete(CURLcode curl_ret, struct curl_file *file)
...
+			return DOWNLOAD_STATUS_COMPLETED;
+		case 403:
+			fprintf(stderr, "Curl: Download failed: Forbidden (403) - '%s'\n", url);
+			*err = -EPERM;
+			return DOWNLOAD_STATUS_FORBIDDEN;
+		case 404:
+			fprintf(stderr, "Curl: Download failed: File not found (403) - '%s'\n", url);
+			*err = -ENOENT;
+			return DOWNLOAD_STATUS_NOT_FOUND;
+		default:
```
castulo: For consistency, in this case you can also print a generic error message like "Curl: Download failed."

src/curl.c
```diff
@@ -311,6 +311,99 @@ CURLcode swupd_download_file_complete(CURLcode curl_ret, struct curl_file *file)
...
+		case CURLE_COULDNT_RESOLVE_PROXY:
+			fprintf(stderr, "Curl: Could not resolve proxy\n");
+			*err = -1;
+			return DOWNLOAD_STATUS_ERROR;
+		case CURLE_COULDNT_RESOLVE_HOST:
+			fprintf(stderr, "Curl: Could not resolve host - '%s'\n", url);
+			*err = -1;
+			return DOWNLOAD_STATUS_ERROR;
+		case CURLE_COULDNT_CONNECT:
+			fprintf(stderr, "Curl: Could not connect to host or proxy\n");
```
castulo: you can also print the url in this message similar to the way you do it in the CURLE_COULDNT_RESOLVE_HOST

src/curl.c
```diff
@@ -311,6 +311,99 @@ CURLcode swupd_download_file_complete(CURLcode curl_ret, struct curl_file *file)
...
+			*err = -ENONET;
+			return DOWNLOAD_STATUS_ERROR;
+		case CURLE_FILE_COULDNT_READ_FILE:
+			*err = -ENOENT;
+			return DOWNLOAD_STATUS_NOT_FOUND;
+		case CURLE_PARTIAL_FILE:
+			fprintf(stderr, "Curl: File incompletely downloaded from '%s''\n", url);
```
castulo: you have an extra ' here

src/curl.c
```diff
@@ -311,6 +311,99 @@ CURLcode swupd_download_file_complete(CURLcode curl_ret, struct curl_file *file)
...
+			return DOWNLOAD_STATUS_ERROR;
+		case CURLE_SSL_CERTPROBLEM:
+			fprintf(stderr, "Curl: Problem with the local client SSL certificate\n");
+			*err = -1;
+			return DOWNLOAD_STATUS_ERROR;
+		case CURLE_RANGE_ERROR:
+			fprintf(stderr, "Range command not supported by server, download resume disabled.\n");
+			*err = -ERANGE;
+			return DOWNLOAD_STATUS_RANGE_ERROR;
+		default:
```
castulo: Same comment, you can add a generic error message.

src/curl.c
```diff
@@ -386,103 +478,22 @@ static int swupd_curl_get_file_full(const char *url, char *filename,
...
-		case CURLE_RANGE_ERROR:
-			fprintf(stderr, "Range command not supported by server, download resume disabled.\n");
 
-			//Reset variables
-			ret = 0;
-			err = -1;
-			memset(&local, 0, sizeof(local));
+	status = process_curl_error_codes(curl_ret, curl, &err);
+	if (status == DOWNLOAD_STATUS_RANGE_ERROR) {
+		//Reset variable
```
castulo: missing space

src/swupd-curl-internal.h
```diff
@@ -31,6 +31,21 @@ extern CURLcode swupd_download_file_complete(CURLcode curl_ret, struct curl_file
  */
 extern CURLcode swupd_curl_set_basic_options(CURL *curl, const char *url);
 
+/* Download status code */
+enum download_status {
+	DOWNLOAD_STATUS_COMPLETED =0,
```
castulo: missing space

test/functional/update/update-slow-server.bats
```diff
@@ -40,6 +40,7 @@ test_teardown() {
 		Downloading packs...
 		Curl: Partial file downloaded from '.*pack-test-bundle-from-10.tar'
 		Starting download retry #1 for .*100/pack-test-bundle-from-10.tar
```
castulo: I know this is not part of this review, but we should consider adding the ' ' around the url from the message that goes like "Starting download retry ....", all other messages use it, so it would look more consistent.

src/curl_async.c
```diff
@@ -34,7 +34,7 @@
 #include <unistd.h>
```
castulo: Nice, this is a much better name for this file

src/curl_async.c
otaviobp: In reevaluate_number_of_parallel_downloads I save the last_retry, so we won't decrement the max_xfer in all calls. Only when the first download is retried by last_retry+1 times.

As we remove and add files to the failed list while we are trying to download, there will be files in the failed list in different retry numbers. Setting a last_retry variable was the only way I could think of on how to avoid the problem you commented.

src/curl.c
```diff
@@ -340,6 +340,102 @@ CURLcode swupd_download_file_complete(CURLcode curl_ret, struct curl_file *file)
 	return curl_ret;
 }
 
+// err is only needed because retry code is spread all over swupd. Remove this error code return when
```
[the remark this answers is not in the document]
otaviobp: good. I'll add that

src/curl_async.c
castulo: so what is your opinion on this @otaviobp ?

PR #769, opened 2018-12-27
Makeifile: Check all files under src/ for code style

otaviobp:
Glob ** will only recurse directories if globstar is set. To avoid skipping
important files run clang-format on src/ and src/lib


jwakre:
Nit: s/Makeifile/Makefile

PR #770, opened 2019-01-04
curl: swupd_curl_query_content_size is leaking file content

otaviobp:
When CURLOPT_NOBODY is set to true, curl shouldn't get the body of a file,
we should get only headers. This works fine for all tested scenarios using
http and https. It's also working in most scenarios that uses file://, but
it fails on travis environment. So, add a dummy function to ignore body data
if present.

There was a hack in search tests to ignore leaked file content. Removing that
hack.

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -220,9 +220,9 @@ void swupd_curl_deinit(void)
 	curl_global_cleanup();
 }
 
-static size_t filesize_from_header_cb(void UNUSED_PARAM *func, size_t size, size_t nmemb, void UNUSED_PARAM *data)
+static size_t dummy_write_cb(void UNUSED_PARAM *func, size_t size, size_t nmemb, void UNUSED_PARAM *data)
 {
-	/* Drop the header, we just want file size */
+	/* Drop the content */
 	return (size_t)(size * nmemb);
 }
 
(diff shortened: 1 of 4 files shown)


test/functional/search/search-content-check-negative.bats
```diff
@@ -38,11 +38,11 @@ global_teardown() {
 	run sudo sh -c "$SWUPD search $SWUPD_OPTS fake-file"
 
 	assert_status_is 0
-	assert_in_output "Searching for 'fake-file'"
 	# there is going to be a whole lot of content within the line
```
castulo: you can now remove this comment also

src/curl.c
```diff
@@ -253,6 +253,11 @@ double swupd_curl_query_content_size(char *url)
 		return -1;
 	}
 
+	curl_ret = curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, dummy_write_cb);
+	if (curl_ret != CURLE_OK) {
+		return -1;
+	}
+
```
castulo: nice!

test/functional/search/search-content-check-negative.bats
otaviobp: true

castulo: same for all other similar tests

otaviobp:
fixed.

PR #773, opened 2019-01-08
Testlib: Fixing symlinks of test bundles

### LINKED ISSUE #772: The test library is adding bad links to bundles when the bundle contains symbolic links
When creating a test bundle that contains a symlink like this:
`create_bundle -L -n my_bundle -f /my_file -l /my_link my_test_environment`
A test bundle called _my_bundle_ is created in _my_test_environment_, this bundle has one link called _my_link_ and this link points to a file that is also added to the bundle's manifest. The problem is that during the creation of the bundle the symbolic link is created to point to an existing file but later in the process the file is renamed, causing the symlink to become dangling. This dangling link is then added to the bundle's manifest. If we run "swupd verify --fix" the installed bundle will always show a hash mismatch and the good symlink will be replaced with the dangling symlink.

This issue had not appeared so far because we are not using symlinks in a lot of our tests.

castulo:
When creating a test bundle that contains a symlink like this:
create_bundle -L -n my_bundle -f /my_file -l /my_link my_test_environment
A test bundle called my_bundle is created in my_test_environment,
this bundle has one link called my_link and this link points to a
file that is also added to the bundle's manifest. The problem is
that during the creation of the bundle the symbolic link is created
to point to an existing file but later in the process the file is
renamed, causing the symlink to become dangling. This dangling link
is then added to the bundle's manifest.  If we run "swupd verify
--fix" the installed bundle will always show a hash mismatch and the
good symlink will be replaced with the dangling symlink.

This commit changes the way symlinks are being created when creating a
bundle so they point to the correct file.

Closes #772 

### DIFF
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -1783,6 +1783,9 @@ create_bundle() { # swupd_function
 	local manifest
 	local local_bundle=false
 	local experimental=false
+	local pfile
+	local pfile_name
+	local pfile_path
 
 	# If no parameters are received show help
 	if [ $# -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


test/functional/testlib.bash
```diff
@@ -1888,23 +1891,33 @@ create_bundle() { # swupd_function
...
-		add_to_manifest "$manifest" "$bundle_link" "$val"
-		# Add the file pointed by the link to the zero pack of the bundle
-		pfile=$(basename "$(readlink -f "$bundle_link")")
+		# first create the file that will be pointed by the symlink
+		pfile=$(create_file "$files_path")
+		pfile_name=$(generate_random_name file_)
+		pfile_path="$fdir"/"$pfile_name"
+		# temporarily rename the file to the name it will have in the file system
+		# so we can create the symlink with the correct name
+		sudo mv "$pfile" "$(dirname "$pfile")"/"$pfile_name"
```
phmccarty: Perhaps `"$(dirname "$pfile")"/"$pfile_name"` can be stored in a local variable, and then referenced three times? I think this would make these three lines easier to read.

otaviobp:
merged by other PR

PR #774, opened 2019-01-09
Verify --fix skip files that are unsafe to delete without warning the user

### LINKED ISSUE #765: 'swupd verify --fix' reports about files it cannot safely delete
In latest versions of Clear Linux OS, there are hundreds of files rooted at `/usr/share/mime` that are marked for deletion, so `swupd verify --fix` will try to delete them.

However, deleting that content is problematic: Those deletions were the result of a directory->symlink type change (i.e. `/usr/share/mime` used to be a directory, but it is now a symlink). SWUPD rightfully refuses to delete the content marked to be deleted, because all paths in that directory tree now contain a symlink component, the `/usr/share/mime` symlink. The verify subcommand explicitly checks for this condition to prevent swupd from following symlinks to delete content.

As seen in commit afce2631d64f5abdc06354d4fc054d7abbc86576, I added a FIXME that suggests `swupd verify --fix` should check if a file/directory/symlink can be "safely deleted", and if so, delete it. However, my current thinking is that while the "safe to delete" check is still mission critical, the `"Not safe to delete: <FILE>"` error message is not helpful for the end user; verify can just skip the file and not increment the not-deleted file count.

For reference, the swupd output in question looks like the following:

```
Verifying version 26800
Verifying files
...100%
...100%
Adding any missing files

Missing file: /usr/share/icons/Paper-Mono-Dark/icon-theme.cache

fixed
...60%
Missing file: /usr/share/icons/Paper/icon-theme.cache

fixed
...100%
Fixing modified files
...100%
File that should be deleted: /usr/share/mime/x-epoc/x-sisx-app.xml
Not safe to delete: /usr/share/mime/x-epoc/x-sisx-app.xml
File that should be deleted: /usr/share/mime/x-epoc
Not safe to delete: /usr/share/mime/x-epoc
File that should be deleted: /usr/share/mime/x-content/win32-software.xml
Not safe to delete: /usr/share/mime/x-content/win32-software.xml
File that should be deleted: /usr/share/mime/x-content/video-vcd.xml
Not safe to delete: /usr/share/mime/x-content/video-vcd.xml
File that should be deleted: /usr/share/mime/x-content/video-svcd.xml

.
.
.

File that should be deleted: /usr/share/mime/application/annodex.xml
Not safe to delete: /usr/share/mime/application/annodex.xml
File that should be deleted: /usr/share/mime/application/andrew-inset.xml
Not safe to delete: /usr/share/mime/application/andrew-inset.xml
File that should be deleted: /usr/share/mime/application
Not safe to delete: /usr/share/mime/application
File that should be deleted: /usr/share/mime/aliases
Not safe to delete: /usr/share/mime/aliases
File that should be deleted: /usr/share/mime/XMLnamespaces
Not safe to delete: /usr/share/mime/XMLnamespaces
Inspected 415705 files
2 files were missing
2 of 2 missing files were replaced
0 of 2 missing files were not replaced
817 files found which should be deleted
0 of 817 files were deleted
817 of 817 files were not deleted
none
```

castulo:
When running verify --fix, if there are files in the system that are
marked as deleted in the manifest, then they are removed from the
system. However, if a component of the path of the file is a
symbolic link, the file is determined to be unsafe to be removed and
it is left in the system, the user is warned, and the not-deleted
file count is incremented.

This commit modifies this behavior so verify just skips the file, the
user is not warned with the "Not safe to delete: <FILE>" error message,
and so the not-deleted file count is not incremented since it is really
not useful for the end user.

Closes #765

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -690,13 +690,9 @@ int get_dirfd_path(const char *fullname)
 	}
 
 	if (strcmp(real_path, dir) != 0) {
-		/* FIXME: Should instead check to see if real_path is marked
-		 * non-deleted in the consolidated manifest. If it is
-		 * non-deleted, then we must not delete the file (and also not
-		 * flag an error); otherwise, it can be safely deleted. Until
-		 * that check is implemented, always skip the file, because we
-		 * cannot safely determine if it can be deleted. */
-		ret = -1;
+		/* The path to the file contains a symlink, skip the file,
+		 * because we cannot safely determine if it can be deleted. */
+		ret = -2;
 		close(fd);
 	} else {
 		ret = fd;
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -1783,6 +1783,9 @@ create_bundle() { # swupd_function
 	local manifest
 	local local_bundle=false
 	local experimental=false
+	local pfile
+	local pfile_name
+	local pfile_path
 
 	# If no parameters are received show help
 	if [ $# -eq 0 ]; then
@@ -1888,23 +1891,33 @@ create_bundle() { # swupd_function
 				bundle_dir=$(create_dir "$files_path")
 				add_to_manifest -p "$manifest" "$bundle_dir" "$fdir"
 			fi
+		else
+			fdir=""
 		fi
-		bundle_link=$(create_link "$files_path")
-		sudo tar -C "$files_path" -rf "$version_path"/pack-"$bundle_name"-from-0.tar --transform "s,^,staged/," "$(basename "$bundle_link")"
-		add_to_manifest "$manifest" "$bundle_link" "$val"
-		# Add the file pointed by the link to the zero pack of the bundle
-		pfile=$(basename "$(readlink -f "$bundle_link")")
+		# first create the file that will be pointed by the symlink
+		pfile=$(create_file "$files_path")
+		pfile_name=$(generate_random_name file_)
+		pfile_path="$fdir"/"$pfile_name"
+		# temporarily rename the file to the name it will have in the file system
+		# so we can create the symlink with the correct name
+		sudo mv "$pfile" "$(dirname "$pfile")"/"$pfile_name"
+		bundle_link=$(create_link "$files_path" "$(dirname "$pfile")"/"$pfile_name")
+		sudo mv "$(dirname "$pfile")"/"$pfile_name" "$pfile"
+		# add the pointed file to the zero pack and manifest
 		sudo tar -C "$files_path" -rf "$version_path"/pack-"$bundle_name"-from-0.tar --transform "s,^,staged/," "$(basename "$pfile")"
+		add_to_manifest "$manifest" "$pfile" "$pfile_path"
+		# add the symlink to zero pack and manifest
+		sudo tar -C "$files_path" -rf "$version_path"/pack-"$bundle_name"-from-0.tar --transform "s,^,staged/," "$(basename "$bundle_link")"
+		add_to_manifest -s "$manifest" "$bundle_link" "$val"
 		if [ "$DEBUG" == true ]; then
 			echo "link -> $bundle_link"
-			echo "file pointed to -> $(readlink -f "$bundle_link")"
+			echo "file pointed to -> $pfile"
 		fi
 		if [ "$local_bundle" = true ]; then
 			sudo mkdir -p "$target_path$(dirname "$val")"
 			# if local_bundle is enabled copy the link to target-dir but also
 			# copy the file it points to
-			pfile_path=$(awk "/$(basename $pfile)/"'{ print $4 }' "$manifest")
-			sudo cp "$files_path"/"$pfile" "$target_path$pfile_path"
+			sudo cp "$pfile" "$target_path$pfile_path"
 			sudo ln -rs "$target_path$pfile_path" "$target_path$val"
 		fi
 	done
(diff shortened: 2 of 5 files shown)


src/helpers.c
```diff
@@ -696,7 +696,7 @@ int get_dirfd_path(const char *fullname)
 		 * flag an error); otherwise, it can be safely deleted. Until
 		 * that check is implemented, always skip the file, because we
 		 * cannot safely determine if it can be deleted. */
```
phmccarty: Can you remove this FIXME comment, or rewrite it as a normal code comment that describes the special handling? As currently written, I think it is irrelevant now :-)

test/functional/testlib.bash
```diff
@@ -1754,14 +1755,14 @@ create_bundle() { # swupd_function
 		# this file is added in every bundle by default, it would add too much overhead
 		# for most tests
 		fdir=$(dirname "${val%:*}")
-		if [ ! "$(sudo cat "$manifest" | grep -x "D\\.\\.\\..*$fdir")" ] && [ "$fdir" != "/usr/share/clear/bundles" ] \
+		if [ ! "$(sudo cat "$manifest" | grep -x "[D|L]\\.\\.\\..*$fdir")" ] && [ "$fdir" != "/usr/share/clear/bundles" ] \
```
phmccarty: Should be `[DL]` instead of `[D|L]`, correct? If you did need alternation (in this case, it's not needed), then `-E` would be required option to grep, and the alternation would be written as `(D|L)` instead.

test/functional/testlib.bash
```diff
@@ -1754,14 +1755,14 @@ create_bundle() { # swupd_function
...
 		&& [ "$fdir" != "/" ]; then
 			bundle_dir=$(create_dir "$files_path")
 			add_to_manifest -p "$manifest" "$bundle_dir" "$fdir"
 			# add each one of the directories of the path if they are not in the manifest already
 			while [ "$(dirname "$fdir")" != "/" ]; do
 				fdir=$(dirname "$fdir")
-				if [ ! "$(sudo cat "$manifest" | grep -x "D\\.\\.\\..*$fdir")" ]; then
+				if [ ! "$(sudo cat "$manifest" | grep -x "[D|L]\\.\\.\\..*$fdir")" ]; then
```
phmccarty: See previous comment about the grep regex.

test/functional/testlib.bash
```diff
@@ -1850,7 +1858,50 @@ create_bundle() { # swupd_function
...
+	# 3) Create the requested link(s) to directories in the bundle
+	for val in "${dlink_list[@]}"; do
+		if [[ "$val" != "/"* ]]; then
+			val=/"$val"
+		fi
+		# if the directory the link is doesn't exist,
+		# add it to the bundle (except if the directory is "/")
+		fdir=$(dirname "$val")
+		if [ "$fdir" != "/" ]; then
+			if [ ! "$(sudo cat "$manifest" | grep -x "[D|L]\\.\\.\\..*$fdir")" ]; then
```
phmccarty: See previous comment about the grep regex.

test/functional/testlib.bash
```diff
@@ -1884,32 +1935,42 @@ create_bundle() { # swupd_function
 		# add it to the bundle (except if the directory is "/")
 		fdir=$(dirname "$val")
 		if [ "$fdir" != "/" ]; then
-			if [ ! "$(sudo cat "$manifest" | grep -x "D\\.\\.\\..*$fdir")" ]; then
+			if [ ! "$(sudo cat "$manifest" | grep -x "[D|L]\\.\\.\\..*$fdir")" ]; then
```
phmccarty: See previous comment about the grep regex.

test/functional/testlib.bash
```diff
@@ -1918,10 +1979,12 @@ create_bundle() { # swupd_function
...
```
phmccarty: See previous comment about the grep regex.

src/helpers.c
castulo: but isn't this FIXME still valid? We are only skipping the file if we find it to be not-safe to be deleted. But the FIXME talks about checking to see if real_path is marked non-deleted in the consolidated manifest, and deleting it if it can be safely deleted. After all the fix in this commit is only for not flagging an error.

test/functional/testlib.bash
castulo: I'll fix that, thanks

src/helpers.c
phmccarty: No, after some recent discussions, I don't think my original thinking behind this FIXME was correct... But fair point that the intent of your PR is to not flag an error, so the FIXME can be addressed separately.

castulo: I'm ok with changing the message, I just wanted to understand the reasoning why :)

phmccarty: Okay, this is my current thinking.

If `verify` detects the "not safe to delete" condition, the canonical absolute path either resides (a) outside the /usr tree (or under /usr/local), or (b) within the /usr tree (excluding /usr/local). For case (a), the FIXME is not relevant, because the canonical absolute path will not be present in the consolidated manifest due to the path not being tracked by swupd. For case (b), the FIXME is also not relevant, because the canonical absolute path *will* be present in the consolidated manifest, and `verify` will therefore operate on it in the normal course of iterating through the consolidated manifest. There is no need to check the consolidated manifest "out of order", as the FIXME implies, because `verify` will already fix everything it needs to.

PR #775, opened 2019-01-10
verify: Enable verifying a single bundle

bryteise:
Allow users to verify just a single bundle rather than the entire
system. This also enables an alternative "installation" with a verify
--fix --bundle $foo where foo doesn't include os-core which is useful
for installing 3rd party content that lives outside of /usr.


bryteise:
@otaviobp I'd want to put that in now before it is merged as making a change to command line flags is not very friendly and having a bundle and bundles flag is pretty unpleasant too. That said the implementation could get unpleasant unless we force `--bundles a,b,c` rather than `--bundles a b c`. Any thoughts about this?

otaviobp:
> That said the implementation could get unpleasant unless we force --bundles a,b,c rather than --bundles a b c. Any thoughts about this?

I prefer

     --bundles a b c

Because we it looks like

    swupd bundle-add a b c

test/functional/verify/verify-bundle.bats
```diff
@@ -0,0 +1,78 @@
...
+
+	create_test_environment "$TEST_NAME"
+	create_bundle -L -n test-bundle1 -f /foo/test-file1 "$TEST_NAME"
+	create_bundle -L -n test-bundle2 -f /bar/test-file2 "$TEST_NAME"
+        sudo rm -f "$TARGETDIR"/foo/test-file1
+        sudo rm -f "$TARGETDIR"/bar/test-file2
+
+}
+
+@test "VER046: Verify bundle ignores not specified missing bundle" {
```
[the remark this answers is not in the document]
otaviobp: we should add a check for that in 'make compliant'

otaviobp:
>  I'd want to put that in now before it is merged as making a change to command line flags is not very friendly and having a bundle and bundles flag is pretty unpleasant too. 

By later I mean 'later, but before a swupd release'. We also need to add that to the manpage document.

src/verify.c
```diff
@@ -557,6 +560,13 @@ static bool parse_opt(int opt, char *optarg)
 		}
 		cmdline_option_picky_whitelist = optarg;
 		return true;
+	case 'B':
+		if (!optarg) {
+			fprintf(stderr, "Invalid --bundle argument\n\n");
```
castulo: The other options of the _verify_ command have a text like this:

> "Missing --picky-tree argument\n\n"

https://github.com/clearlinux/swupd-client/blob/master/src/verify.c#L544

So I'd prefer if this message stays consistent with those messages by changing "Invalid" with "Missing"

test/functional/verify/verify-bundle.bats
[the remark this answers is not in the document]
castulo: Nice tests by the way 👌

src/verify.c
bryteise: Updating.

test/functional/verify/verify-bundle.bats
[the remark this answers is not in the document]
bryteise: @castulo Hrm likely I did something stupid copy paste wise in my test writing but I thought when I ran the create script it gave me 46 X(. Fixing.

src/verify.c
bryteise: Actually I'm curious about this. This error should never happen as getopt catches this condition and raises its own error. Is there a reason we have these checks? Or am I misreading the man page?

bryteise:
Updated based on feedback, though had to stick with the --bundles flag taking a comma separated list of bundles. Man page for verify updated and a new test for multiple bundles added.

test/functional/verify/verify-bundle.bats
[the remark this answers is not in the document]
castulo: Yeah, the code that automatically gets the next available test ID is pretty dumb, it only counts how many tests exist at the moment in a given directory (by searching for the _@test_ directive within all files in the dir), so if for some reason you had another dummy test around at the moment that would have increased the counter by 1 causing the results you had. So until that code is updated to become "smarter" it is always a good idea to double check the autogenerated ID by manually inspecting the list of tests.

src/verify.c
otaviobp: you're right. this check doesn't make sense. You can remove that.

bryteise: Alrighty pushing an update.

PR #783, opened 2019-01-16
Removing check for optarg for parameters with required arguments

otaviobp:
According to getopt_long documentation optarg is always set when argument is
required. It can be empty when argument is optional, but we don't have that
case in swupd.

So we can remove the extra check for empty optarg on opt parse functions.


bryteise:
Good cleanup, LGTM!

PR #787, opened 2019-01-17
Curl - Cosmetic changes

otaviobp:


src/curl.c
```diff
@@ -164,6 +164,11 @@ int swupd_curl_init(void)
 	int ret;
 	struct stat st;
 
+	if (curl) {
+		error("curl has already been initialized.\n");
+		return -1;
+	}
```
castulo: what is the purpose of returning a -1 if curl was already initialized? shouldn't it just return with a 0 if it is already initialized and ready to be used? with a -1 there will be no way to distinguish between an already initialized curl and one that failed to initialize (code-wise).

src/curl.c
```diff
@@ -232,6 +237,7 @@ double swupd_curl_query_content_size(char *url)
 	double content_size;
 
 	if (!curl) {
+		error("curl hasn't been initialized.\n");
```
castulo: you are still using a dot at the end 😂

src/curl.c
otaviobp: Hmmm. That's a good question. I was considering multiple inits as an error, but we support multiple deinits too we can return 0 here. I'll change that.

src/curl.c
otaviobp: Doh

src/curl.c
```diff
@@ -165,19 +165,19 @@ int swupd_curl_init(void)
...
 	if (curl) {
-		error("curl has already been initialized.\n");
+		error("curl has already been initialized\n");
 		return -1;
 	}
 
 	curl_ret = curl_global_init(CURL_GLOBAL_ALL);
 	if (curl_ret != CURLE_OK) {
-		fprintf(stderr, "Error: failed to initialize CURL environment\n");
+		error("failed to initialize CURL environment\n");
```
castulo: to be consistent with all other curl messages you could change this to:
"curl - failed to initialize environment\n"

src/curl.c
```diff
@@ -165,19 +165,19 @@ int swupd_curl_init(void)
...
 		return -1;
 	}
 
 	curl = curl_easy_init();
 	if (curl == NULL) {
-		fprintf(stderr, "Error: failed to initialize CURL session\n");
+		error("failed to initialize CURL session\n");
```
castulo: same comment as above about consistency

src/curl.c
```diff
@@ -363,75 +363,75 @@ enum download_status process_curl_error_codes(int curl_ret, CURL *curl_handle, i
...
+			error("curl - partial file downloaded from '%s'\n", url);
 			*err = -1;
 			return DOWNLOAD_STATUS_PARTIAL_FILE;
 		case 200:
 		case 0:
 			*err = 0;
 			return DOWNLOAD_STATUS_COMPLETED;
 		case 403:
-			debug("Curl: Download failed: Forbidden (403) - '%s'\n", url);
+			debug("Curl: download failed - forbidden (403) - '%s'\n", url);
```
castulo: there is some inconsistency in the message here too, I'd suggest using "curl - ..." over "Curl: ..."

src/curl.c
```diff
@@ -363,75 +363,75 @@ enum download_status process_curl_error_codes(int curl_ret, CURL *curl_handle, i
...
 			*err = -EPERM;
 			return DOWNLOAD_STATUS_FORBIDDEN;
 		case 404:
-			debug("Curl: Download failed: File not found (404) - '%s'\n", url);
+			debug("Curl: download failed - file not found (404) - '%s'\n", url);
```
castulo: same as above

src/curl.c
```diff
@@ -478,7 +478,7 @@ static int swupd_curl_get_file_full(const char *url, char *filename,
 		local.path = filename;
 
 		if (resume_ok && resume_download_supported && lstat(filename, &stat) == 0) {
-			fprintf(stderr, "Curl: Resuming download for '%s'\n", url);
+			print("Curl: resuming download for '%s'\n", url);
```
castulo: same here

src/curl.c
```diff
@@ -517,7 +517,7 @@ static int swupd_curl_get_file_full(const char *url, char *filename,
 		goto exit;
 	}
 
-	debug("CURL - start sync download: %s -> %s\n", url, in_memory_file ? "<memory>" : filename);
+	debug("Curl: start sync download: %s -> %s\n", url, in_memory_file ? "<memory>" : filename);
```
castulo: and here

src/curl.c
```diff
@@ -526,7 +526,7 @@ static int swupd_curl_get_file_full(const char *url, char *filename,
 	}
 
 	status = process_curl_error_codes(curl_ret, curl, &err);
-	debug("CURL - complete sync download: %s -> %s, status=%d\n", url, in_memory_file ? "<memory>" : filename, status);
+	debug("Curl: complete sync download: %s -> %s, status=%d\n", url, in_memory_file ? "<memory>" : filename, status);
```
castulo: one more

src/curl_async.c
```diff
@@ -173,7 +173,7 @@ static void reevaluate_number_of_parallel_downloads(struct swupd_curl_parallel_h
 		h->max_xfer = 1;
 	}
 
-	printf("Reducing number of parallel downloads to %ld\n", h->max_xfer);
+	print("Reducing number of parallel downloads to %ld\n", h->max_xfer);
```
castulo: looks like most other messages use lower case for the first letter

src/curl_async.c
```diff
@@ -258,7 +258,7 @@ static int perform_curl_io_and_complete(struct swupd_curl_parallel_handle *h, in
 		 * completing the download encounters further problems. */
 		curl_ret = swupd_download_file_close(msg->data.result, &file->file);
 		file->status = process_curl_error_codes(curl_ret, handle, &err);
-		debug("CURL - complete ASYNC download: %s -> %s, status=%d\n", file->url, file->file.path, file->status);
+		debug("Curl: complete ASYNC download: %s -> %s, status=%d\n", file->url, file->file.path, file->status);
```
castulo: The messages in this section now use "Curl: ..." consistently, should we stick to this format here or should we go with the "curl - ..." format?

src/curl.c
otaviobp: For debug and error messages I agree with you, but not for this one. Print doesn't add any prefix. So error/debug messages are going to be printed like:

    Error: curl - some error message
    2019-01-18 18:34:52 (src/curl.c:513)              Debug: curl - debug message

and print are like:

    Curl: this is a message

I think that this would be weird:

    curl - this is a message

What do you think?

src/curl_async.c
otaviobp: I am unsure about that. Right now it is using lower case when there's a prefix, but I think we should change that. What do you prefer?

    Error: download failed
    Error: Download failed

src/curl.c
castulo: ooh ok, it makes sense, carry on then... :)

src/curl_async.c
castulo: I think I like "Error: Download failed" better

src/curl_async.c
otaviobp: Me too. I'll fix that.

otaviobp: oops, I commented in the wrong question

src/curl_async.c
otaviobp: I prefer that too. I'll fix that.

otaviobp:
I updated all error messages according to what we have discussed

PR #790, opened 2019-01-18
Making swupd consistent in return codes

### LINKED ISSUE #610: Swupd is inconsistent in return codes

castulo:
This PR includes the following things:

- [x] Rename the file from swupd-error.h to swupd_exit_codes.h so we can include “successful” codes if needed.
- [x] Rename existing error codes following these rules:
  - All codes should start with SWUPD_ so we never get confused what code comes from swupd
  - Always separate codes that contain multiple words using an underscore SWUPD_MULTI_WORD_ERROR instead of using SWUPDMULTIWORDERROR
  - Code 0 is reserved for SWUPD_SUCCESS (or SWUPD_OK)
  - Code 1 is reserved to be used by commands that return a boolean state (e.g. _swupd autoupdate_)
- [x] When a function returns a boolean state, explicitly return true or false instead of returning 0 or 1
- [ ] Some functions currently mask error codes in a “catch all” error code, for example bundle-add which only uses EBUNDLE_INSTALL or bundle-remove which only uses EBUNDLE_REMOVE. Remove the catch all return codes from these functions so we can have a more meaningful error.
- [ ] Regardless of what codes are used internally, swupd should always exit with a code defined in swupd-codes.h
- [x] Update the documentation about swupd codes on swupd.1 and swupd.1.rst
- [ ] Add the guidelines mentioned in this ToDo list and these ones below to the developer's guide so contributors follow these project standards in the future (no need to rework these ones at the moment with the existing code base since it would be a lot of work).
  - Internally, swupd can use errors that comply with these rules:
  - it can use errors defined in any included file as long as these don’t get propagated to the end user. For example errors from:
    - /usr/include/asm-generic/errno-base.h
    - /usr/include/asm-generic/errno.h
    - /usr/include/curl/curl.h
  - The errors used internally should preferably be negative values. For example prefer using -ENOENT instead of ENOENT.

**WIP - Do NOT merge**

Closes #610


src/swupd_exit_codes.h
```diff
@@ -1,29 +1,36 @@
...
-#define ENOSWUPDSERVER 16       /* no net connection to swupd server */
-#define EFILEDOWNLOAD 17	/* file download problem */
-#define EBUNDLE_INSTALL 18      /* Cannot install bundles */
-#define EREQUIRED_DIRS 19       /* Cannot create required dirs */
-#define ECURRENT_VERSION 20     /* Cannot determine current OS version */
-#define ESIGNATURE 21		/* Cannot initialize signature verification */
-#define EBADTIME 22		/* System time is bad */
-#define EDOWNLOADPACKS 23       /* Pack download failed */
-#define EBADCERT 24		/* unable to verify server SSL certificate */
+typedef enum {
```
otaviobp: ++ for creating an enum for that. It's a lot better than returning just an int. 
But there's one thing I would change. We don't use a lot of typedefs in swupd code. I noticed that they use that for enums on telemetry.c and xattrs.c, but as we don't use that for structs I prefer to avoid using that for enums.

So I prefer:

    enum swupd_code {

And

    enum swupd_code my_function()

src/swupd_exit_codes.h
```diff
@@ -1,29 +1,36 @@
...
+	SWUPD_INVALID_OPTION,		/* 13 invalid command option */
+	SWUPD_SERVER_CONNECTION_ERROR,  /* 14 no network connection to swupd server */
+	SWUPD_COULDNT_DOWNLOAD_FILE,    /* 15 file download problem */
+	SWUPD_COULDNT_INSTALL_BUNDLE,   /* 16 cannot install bundles */
+	SWUPD_COULDNT_CREATE_DIRS,      /* 17 cannot create required directories */
+	SWUPD_CURRENT_VERSION_UNKNOWN,  /* 18 cannot determine current OS version */
+	SWUPD_COULDNT_VERIFY_SIGNATURE, /* 19 cannot initialize signature verification */
+	SWUPD_BAD_TIME,			/* 20 system time is bad */
+	SWUPD_COULDNT_DOWNLOAD_PACK,    /* 21 pack download failed */
+	SWUPD_BAD_CERT			/* 22 unable to verify server SSL certificate */
```
otaviobp: Let's try to keep the same code for constants that didn't change. We're not sure if there are customers already using this in scripts.

src/swupd_exit_codes.h
castulo: my idea is that functions that need to return a documented code are defined to explicitly return a _swupd_code_, for example:

`swupd_code my_function() { ... }`

I think in order to do that we need to have the enum defined as a typedef.

otaviobp: we can do it adding the prefix enum. take a look at src/swupd_curl_internal.h

src/swupd_exit_codes.h
castulo: ok, I'll move them around

src/swupd_exit_codes.h
castulo: so do you prefer to use: `enum swupd_code my_func()` over `swupd_code my_func()`? If so, why? I'm not complaining just curious.

castulo:
It is finally finished. It can now be fully reviewed.

PR #796, opened 2019-01-25
globals: Use path_prefix for cert_path

bryteise:
The certificate path should be looked up based on the path_prefix as
the configuration of the content under the path_prefix may be
different than the base system's.


bryteise:
Closing as this solution carries problems and needs a rethink. Will use work arounds of always passing the certificate file for now.

PR #798, opened 2019-01-28
Unit tests

otaviobp:
Introducing Unit tests in swupd.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -101,6 +101,8 @@ verifytime_SOURCES = src/verifytime.h \
 	src/verifytime.c \
 	src/verifytime_main.c
 
+verifytime_LDADD =
+
 bin_PROGRAMS += verifytime
 
 EXTRA_DIST += \
@@ -122,16 +124,15 @@ systemdunit_DATA = \
 	data/swupd-update.timer \
 	data/verifytime.service
 
-TEST_EXTENSIONS = .bats .prereq
+TEST_EXTENSIONS = .bats .prereq .test
 
 if ENABLE_TESTS
 tap_driver = env AM_TAP_AWK='$(AWK)' $(SHELL) \
 	$(top_srcdir)/tap-driver.sh
 
-LOG_DRIVER = $(tap_driver)
 BATS_LOG_DRIVER = $(tap_driver)
 
-TESTS = $(dist_check_SCRIPTS)
+TESTS = $(dist_check_SCRIPTS) $(UNIT_TESTS)
 
 BATS = \
 	test/functional/bundleadd/add-alias-basic.bats \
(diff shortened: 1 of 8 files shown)


src/helpers.c
```diff
@@ -585,68 +585,6 @@ enum swupd_code swupd_init(void)
...
-int strtoi_err(const char *str, int *value)
-{
-	char *endptr;
-	int err = strtoi_err_endptr(str, &endptr, value);
-
-	if (err) {
-		return err;
-	}
-
-	if (*endptr != '\0' && !isspace(*endptr)) {
```
jwakre: Can we remove #include <ctype.h> now? It was added to support isspace(*endptr) which is removed.

otaviobp: I removed now ctype.h and types.h includes from helpers.c
Thanks.

src/lib/strings.h
```diff
@@ -23,6 +23,34 @@ void free_string(char **s);
 /* Join strings from string list separated by the separator */
 char *string_join(const char *separator, struct list *string);
 
+/*
+ * strtoi_err: Safely convert and string to integer avoiding overflows
```
castulo: Isn't this suppose to say "Safely convert a~~nd~~ string to integer avoiding overflows."

src/lib/strings.h
```diff
@@ -23,6 +23,34 @@ void free_string(char **s);
...
+ *
+ * This function returns negative error codes based on the errno table:
+ * -ERANGE is returned when the string is out of range for int value
+ * -EINVAL is returned when the string isn't a valid number or has any invalid
+ * trailing character.
+ */
+int strtoi_err(const char *str, int *value);
+
+/*
+ * strtoi_err: Safely convert and string to integer avoiding overflows.
```
castulo: wrong function name in the comment strtoi_err -> strtoi_err_endptr (and same typo as above)

Makefile.am
```diff
@@ -261,13 +262,19 @@ BATS = \
 	test/functional/verify/verify-picky-whitelist.bats \
 	test/functional/verify/verify-skip-scripts.bats
 
+UNIT_TESTS = \
+	test/unit/test_strings.test
```
castulo: This is not necessary for this PR, but I would really like a way of discovering these in the future. It is something I've been meaning to do for functional tests too so we don't have to add/remove every test to/from the list every time.

test/unit/test_strings.c
```diff
@@ -0,0 +1,266 @@
...
+	// test EINVAL errors
+	char *einval_errors[] = { "string", "123.456", "123a", "a123a", "456b123", "0x12345", "4f"};
+
+	for (i = 0; i < sizeof(einval_errors) / sizeof(char *); i++) {
+		err = strtoi_err(einval_errors[i], &value);
+		check(err == -EINVAL);
+	}
+}
+
+static void test_strtoi_enderr()
```
castulo: shouldn't this be _test_strtoi_err_endptr_ instead of _test_strtoi_enderr_?

Makefile.am
otaviobp: +1 for this too, but I don't know how to do it. haha

otaviobp:
fixed

PR #800, opened 2019-01-30
Configurable download retries

### LINKED ISSUE #211: Retry code rework

### LINKED ISSUE #322: configurable download retries

castulo:
This PR tackles two main issues in download retries:

- Sync download retries were handled all over the place in the base code. This PR removes the retries from all those places and adds a single place for retrying a download, making the code more consistent and a lot easier to follow and maintain.

- Download retries where hardcoded to 3 in the case of sync downloads and to 5 in the case of async downloads. This PR add options to allow users to configure how many download retries they want swupd to do, and how long to wait between retries. This will be helpful for users in networks with unreliable connections, and will also help developers testing the code that were facing long wait times when simulating network problems.

Closes #211 
Closes #322


src/curl.c
```diff
@@ -539,7 +526,91 @@ static int swupd_curl_get_file_full(const char *url, char *filename,
 		unlink(filename);
 	}
 
-	return err;
+	return -status;
```
otaviobp: You could return status here and change the return of this function from in to enum download_status

test/functional/usability/usa-download-retries.bats
```diff
@@ -0,0 +1,91 @@
...
+	port=$(get_web_server_port "$TEST_NAME")
+	# set the versionurl and contenturl to point to the web server
+	set_upstream_server "$TEST_NAME" http://localhost:"$port"
+
+}
+
+@test "USA005: Retries for failed downloads can be configured" {
+
+	# Users should be able to configure the retry of a failed download using
+	# two options --max-retries and --retry-delay
```
otaviobp: trailing space here

src/curl.c
```diff
@@ -539,7 +526,107 @@ static int swupd_curl_get_file_full(const char *url, char *filename,
...
+
+	for (;;) {
+
+		/* download file */
+		ret = swupd_curl_get_file_full(url, filename, in_memory_file, resume_ok);
+
+		if (ret == DOWNLOAD_STATUS_COMPLETED) {
+			/* operation successful */
+			break;
+		} else {
```
otaviobp: It's easier to read if you take this else out because break is going to get out of the for.

src/curl.c
```diff
@@ -539,7 +526,107 @@ static int swupd_curl_get_file_full(const char *url, char *filename,
...
+			/* operation failed, determine retry strategy (only if the content
+			 * URL is not local). The value of ret on error will be < 0 */
+			current_retry++;
+			if (content_url_is_local) {
```
otaviobp: you can move this to determine_strategy()

src/curl.c
```diff
@@ -539,7 +526,107 @@ static int swupd_curl_get_file_full(const char *url, char *filename,
...
+			}
+			switch (strategy) {
+			case DONT_RETRY:
+				/* don't retry just return the failure */
+				return ret;
+			case RETRY_NOW:
+				/* if we have reached the retry limit just return the failure,
+				 * if not try again immediately */
+				if (max_retries) {
+					if (current_retry <= max_retries) {
```
otaviobp: current_retry < max_retries.

current_retry starts with 0. So in the first retry it will be 1. if max_retries is 1 we shouldn't retry.

src/curl.c
```diff
@@ -539,7 +526,107 @@ static int swupd_curl_get_file_full(const char *url, char *filename,
...
+				strategy = determine_strategy(-ret);
+			}
+			switch (strategy) {
+			case DONT_RETRY:
+				/* don't retry just return the failure */
+				return ret;
+			case RETRY_NOW:
+				/* if we have reached the retry limit just return the failure,
+				 * if not try again immediately */
+				if (max_retries) {
```
otaviobp: You don't need an exception for this case. If max_retries is 0 we just won't retry.

src/curl.c
```diff
@@ -539,7 +526,107 @@ static int swupd_curl_get_file_full(const char *url, char *filename,
...
+			case RETRY_WITH_DELAY:
+				/* if we have reached the retry limit just return the failure,
+				 * if not, wait for the delay and try again */
+				if (max_retries) {
+					if (current_retry <= max_retries) {
+						if (sleep_time) {
+							fprintf(stderr, "Waiting %d seconds before retrying the download\n", sleep_time);
+						}
+						fprintf(stderr, "Retry #%d downloading from %s\n", current_retry, url);
+						sleep(sleep_time);
```
otaviobp: We have too much code duplication here. Instead of using a switch we can use an if just to set the sleep time:

    if (strategy == DONT_RETRY) {
       ....
       return;
    }
    
    if (current_retry <= max_retries) {
        fprintf(stderr, "Retry #%d downloading from %s\n", current_retry, url);
        if (strategy == RETRY_WITH_DELAY) {
            sleep(sleep_time);
            ...
        }
        continue;
....

src/curl.c
```diff
@@ -539,7 +526,107 @@ static int swupd_curl_get_file_full(const char *url, char *filename,
...
+		return RETRY_NOW;
+	case DOWNLOAD_STATUS_ERROR:
+	case DOWNLOAD_STATUS_TIMEOUT:
+		return RETRY_WITH_DELAY;
+	default:
+		return RETRY_NOW;
+	}
+}
+
+static int retry_download(const char *url, char *filename, struct curl_file_data *in_memory_file, bool resume_ok)
```
otaviobp: I would change the name of this function. It's not just retrying. it's trying and then retrying. First time I read this name I though that this function would be called only when we had a download failed.

src/curl.c
```diff
@@ -539,7 +526,107 @@ static int swupd_curl_get_file_full(const char *url, char *filename,
...
+			current_retry++;
+			if (content_url_is_local) {
+				strategy = DONT_RETRY;
+			} else {
+				strategy = determine_strategy(-ret);
+			}
+			switch (strategy) {
+			case DONT_RETRY:
+				/* don't retry just return the failure */
+				return ret;
```
otaviobp: return an error code from errno instead of a negative status code.

src/curl_async.c
```diff
@@ -194,7 +191,7 @@ void *swupd_curl_parallel_download_start(size_t max_xfer)
 
 	h->max_xfer = max_xfer;
 	h->curl_hashmap = hashmap_new(SWUPD_CURL_HASH_BUCKETS, file_hash_cmp, file_hash_value);
-	h->timeout = RETRY_TIMEOUT;
+	h->timeout = retry_delay;
```
otaviobp: Can you rename h->timeout too? Because it's not a timeout

test/functional/usability/usa-download-retries.bats
castulo: fixed

src/curl.c
castulo: you are right, I don't remember the reason why I decided to have that check outside of the determine_strategy function, but based on how the code looks now it totally makes sense. I'll fix that.

src/curl.c
castulo: I'm not sure I understand what you mean on this one

castulo: can you elaborate a little more please?

src/curl.c
castulo: current_retry is being pre incremented before getting to that if, so it is really as if it started at 1, not 0, so if the user set up a max_retry of 1, if there is a download failure it will retry once and give up. The functional tests added verify the correct behavior of this check them out.

src/curl.c
castulo: yes, the if is only so I could print a message saying that retries are disabled.

src/curl.c
castulo: yes, but I just thought that using a switch would look cleaner than using ifs, but if you think having less lines of code is better I can change that code to use ifs instead of a switch

src/curl.c
castulo: ok, I named it like that because the implementation is based on the "retry" pattern. Let me think on a different name then....

src/curl_async.c
castulo: Good idea

src/curl.c
otaviobp: No. You just don't need this else here and it's easier to read because the code won't have an extra indentation

    for (;;) {
             ret = swupd_curl_get_file_full(url, filename, in_memory_file, resume_ok);
            if (ret == DOWNLOAD_STATUS_COMPLETED) {
                break; //Or you can add a return here instead of the break
            }
             /*code that is in the else come here*
    }

src/curl.c
otaviobp: Sorry, I didn't notice that. You're right.

src/curl.c
otaviobp: Well I think it's better to print a message saying that it's disabled. You can leave this here.

src/curl_async.c
castulo: Done

src/curl.c
otaviobp: It's not about saving lines, but in reducing code duplication. printfs for example are hard to maintain if they are duplicated.

src/curl.c
castulo: what about _retry_download_loop_ ? or what do you suggest?

otaviobp: good

castulo: done

src/curl.c
castulo: done

src/curl.c
castulo: The problem here is that the **_process_curl_error_codes_** function which is called by **_swupd_curl_get_file_full_** used to return a negative errno number, and with this refactoring those errnos were removed from the function so it now returns a positive **_enum download_status_**. Most functions that make use of this function only check for values !=0 to determine if the file was gotten successfully by curl, but some functions (like _get_latest_version_ rely on the negative value that the function used to return). So bottom line is that to make this change we will need to scan many other functions to make sure they won't break if we change the return value from negative to positive.

src/curl.c
castulo: This is related to the same comment I made above.

src/curl.c
otaviobp: The problem with the suggested approach is that it is inconsistent. It's not a good idea to use different return codes in  the same function. For example, in this case, if curl hasn't been initialized, or if there's any error in curl api calls, swupd_curl_get_file_full returns -1. When you use this value by doing determine_strategy(-ret), we assume the status code is 1 (DOWNLOAD_STATUS_PARTIAL_FILE), what's incorrect.

You don't need to propagate the return value all over. The only user of swupd_curl_get_file_full is retry_download, so it's ok to change the return code from download_status.

And I would change retry_download to download a negative errno code, to be compatible with other functions in the API.

src/curl.c
castulo: Done

src/curl.c
castulo: Done

castulo:
@otaviobp all your comments have been addressed. 👍

PR #803, opened 2019-02-01
Return non-0 code if search does not find entries

### LINKED ISSUE #731: return non-0 code if search does not find entries
Nice-to-have: swupd search returns `0` both when `needle` is found and when it isn't found. It would be nice to have a different exit code if search completed but `needle` is not found.

castulo:
swupd search returns 0 both when something is found and when nothing
is found.

This commit makes a change so it returns 0 (SWUPD_OK) if search found
somthing and 1 (SWUPD_NO) if it didn't.

Closes #731

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -551,7 +551,7 @@ static void report_find(char *bundle, char *file, char *search_term, bool is_exp
  * Description: Perform a lookup of the specified search string in all Clear manifests
  * for the current os release.
  */
-static void do_search(struct manifest *MoM, char search_type, char *search_term)
+static enum swupd_code do_search(struct manifest *MoM, char search_type, char *search_term)
 {
 	struct list *list;
 	struct list *sublist;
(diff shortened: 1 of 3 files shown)


src/search.c
```diff
@@ -810,6 +814,7 @@ static enum swupd_code download_manifests(struct manifest **MoM)
 enum swupd_code search_main(int argc, char **argv)
 {
 	int ret = SWUPD_OK;
+	int search = SWUPD_NO;
```
otaviobp: Why not using ret instead of creating a new variable?

src/search.c
```diff
@@ -850,12 +855,16 @@ enum swupd_code search_main(int argc, char **argv)
...
+	search = do_search(MoM, search_type, search_string);
 
 clean_exit:
 	if (MoM) {
 		free_manifest(MoM);
 	}
 	swupd_deinit();
-	return ret;
+	/* if we had a failure previous to the search, return that */
+	if (ret) {
```
otaviobp: when --init is used and the function is successful, ret is set to SWUPD_OK that is 0. so search is returned, and search value is SWUPD_NO.

castulo: so do you think when using --init we should also return SWUPN_NO (1) then?

src/search.c
castulo: Done

src/search.c
castulo: fixed

PR #808, opened 2019-02-08
Telemetry: include `bytes transferred` wherever applicable.

ahkok:
Display the bytes transferred after each update in the message
that shows "took N seconds" by appending the info.

This includes total transfer bytes for verify, update, check,
bundle-add and bundle-remove.

The curl error handling function is the perfect location to
capture the bytes for each transfer, and add them to a global
counter. The global counter is then included in each telemetry
event.

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -53,6 +53,8 @@
 
 static CURL *curl = NULL;
 
+curl_off_t total_curl_sz = 0;
+
 /* alternative CA Path */
 static char *capath = NULL;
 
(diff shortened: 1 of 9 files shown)


src/curl.c
```diff
@@ -360,6 +362,13 @@ enum download_status process_curl_error_codes(int curl_ret, CURL *curl_handle)
 		url = "<not available>";
 	}
 
+	/*
+	 * retrieve bytes transferred, errors or not
+	 */
+	curl_off_t curl_sz;
+	curl_easy_getinfo(curl_handle, CURLINFO_SIZE_DOWNLOAD_T, &curl_sz);
```
otaviobp: Culr documentation say we can't trust the result value if this function returs anything other than CURLE_OK. I'm ok on just checking that and ignoring the result if it's not CURLE_OK.

otaviobp:
I made a wrong assumption about the broken travis build. It's a build time problem. Maybe we can fix that by building a newer curl, but not sure.

PR #810, opened 2019-02-09
WIP: Use manifest-deltas in swupd on update

otaviobp:
Tested with a minimal mix generated using PR: https://github.com/clearlinux/mixer-tools/pull/543

TODO:
 - Test update with a mix with more than 1 bundle
 - Test verify going forward
 - Create function tests for this feature

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -396,8 +396,66 @@ void free_manifest(struct manifest *manifest)
 	free(manifest);
 }
 
+static int try_manifest_delta_download(int from, int to, char *component)
+{
+	char *from_manifest = NULL;
+	char *to_manifest = NULL;
+	char *manifest_delta = NULL;
+	char *url = NULL;
+	int ret = 0;
+
+	if (from <= 0 || from > to) {
+		// We don't have deltas for going back
+		return -1;
+	}
+
+	if (strcmp(component, "MoM") == 0) {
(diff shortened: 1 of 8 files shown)


src/manifest.c
```diff
@@ -897,7 +959,7 @@ void link_submanifests(struct manifest *m1, struct manifest *m2, struct list *su
 
 /* if component is specified explicitly, pull in submanifest only for that
  * if component is not specified, pull in any tracked component submanifest */
-struct list *recurse_manifest(struct manifest *manifest, struct list *subs, const char *component, bool server, int *err)
+struct list *recurse_manifest(int current_version, struct manifest *manifest, struct list *subs, const char *component, bool server, int *err)
```
jwakre: This requires that delta manifests for every bundle for 'N' previous versions are generated for every version, even when only one bundle was updated. This is currently done in clearlinux/mixer-tools#543, but I think it would be better to reference previous versions to get manifest deltas for bundles that were not changed in the 'to' version which is how the old Swupd implementation of manifest deltas worked.

bryteise: Agreed, I've updated the mixer-tools code to align with that. Sorry for the churn.

otaviobp: Nice. Point. I'll fix that

PR #811, opened 2019-02-11
docs: Fix documentation on --force option

otaviobp:
--force/-x is specific for verify command, but documentation was stating
that it was global. Removing all references from a global --force and
adding that information to the help page of verify command.

Also moves the force variable from global to verify.


otaviobp:
> IIRC, the original intent of this option was to enable it for the other subcommands as appropriate. But because that plan never took shape, I'm fine with this change.

Yes, we can move it back if we use force for something else, but I don't see this happening in a close future.

PR #814, opened 2019-02-11
update: Be consistent in exit codes when trying to update to the same version

otaviobp:
Use similar output and same exit code used when you are already in the
    last version in the case where you are updating to the same version you
    already are.
    
    i.e. if you are in version 100, both commands are going to have the same
    exit code:
    
     - swupd update
     - swupd update -m 100

### DIFF
--- a/test/functional/update/update-to-same-version.bats
+++ b/test/functional/update/update-to-same-version.bats
@@ -1,0 +1,62 @@
+#!/usr/bin/env bats
+
+# Author: Otavio Pontes
+# Email: otavio.pontes@intel.com
+
+load "../testlib"
+
+test_setup() {
+
+	create_test_environment "$TEST_NAME"
+	create_version -p "$TEST_NAME" 20 10
+
+}
+
+@test "UPD050: Update to the same version using -m" {
+
+	run sudo sh -c "$SWUPD update $SWUPD_OPTS -m 10"
+
+	assert_status_is 0
+	expected_output=$(cat <<-EOM
+		Update started.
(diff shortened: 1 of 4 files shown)


test/functional/update/update-to-same-version.bats
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bats
+
+load "../testlib"
+
```
castulo: please consider adding your contact info at the top of the test for consistency with the rest of the other tests:

```
# Author: Otavio Pontes
```
We started doing this as a good practice in tests so it is easier to find the author and ask questions if needed.

otaviobp: Fixed that, but do we really need this? Git keeps the history of that file and git blame is better than this info to find who to talk to.

castulo: well, consider this example; when I migrated all the tests from the old framework to the new, all old tests were deleted so if you look at all new tests they show as if I'm the author (which in a way I guess I am) but all those tests were actually designed by someone else, I just ported them to the new framework. So in that scenario we lost traceability of the author of all those tests. When I was porting the tests I had a lot of questions about them and when I looked at GitHub to see who the author was I ran into the problem that sometimes the person who had submitted the test was not the author or things like that so it was very difficult to find who to ask for each test.
So I guess is just extra info that could be useful, and it doesn't harm if it is there.

(If you use the generate_test function from the testilb.bash to automatically create a new blank test it will already add your contact info to the test)

otaviobp:
Fixes #821

PR #816, opened 2019-02-12
update: Print total bytes downloaded in MB

### LINKED ISSUE #271: Report bytes downloaded during update
Feature request:

It would be nice to see how many bytes were downloaded and reported with the current statistics that show how many files changed/removed/added/etc.

As per #576 - printing the download speed would be nice also. A live report and also an average speed in the final statistics.

otaviobp:
Fixes #271

### DIFF
--- a/src/update.c
+++ b/src/update.c
@@ -474,8 +474,8 @@ clean_curl:
 		  total_curl_sz);
 
 	if (server_version > current_version) {
-		printf("Update took %0.1f seconds, %ld bytes transferred\n", delta,
-		       total_curl_sz);
+		printf("Update took %0.1f seconds, %ld MB transferred\n", delta,
+		       total_curl_sz / 1000 / 1000);
 	}
 	timelist_print_stats(global_times);
 


src/update.c
```diff
@@ -474,8 +474,8 @@ static enum swupd_code main_update()
 		  total_curl_sz);
 
 	if (server_version > current_version) {
-		printf("Update took %0.1f seconds, %ld bytes transferred\n", delta,
-		       total_curl_sz);
+		printf("Update took %0.1f seconds, %ld MB transferred\n", delta,
+		       total_curl_sz / 1000 / 1000);
```
castulo: I always get confused in these cases if we should use the Base10 (1 MB = 1000000 bytes) or Base2 (1 MB = 1048576 bytes) for these conversions. What is more commonly used in Linux when referring to storage?

otaviobp: 1 MB = 10^6
1 MiB = 1024*1024

I used MB because that's what search uses.

phmccarty: Both abbreviations are pretty commonly used these days, though sometimes `MB` (megabyte) is still used for cases when `MiB` (mebibyte) is technically the correct abbreviation.

otaviobp: So, can I leave it the way it is?

phmccarty: I am fine with using `MB` in this case, because as you said, swupd uses `MB` units elsewhere in the code.

castulo: I guess if search is already using Base10 we should stick to it then, consistency is important

PR #817, opened 2019-02-12
Fix manually installed bundle tracking and add tests

bryteise:
Update bundle tracking to correctly initialize tracked bundle state
and add testing to validate tracking works as expected.


src/bundle.c
```diff
@@ -435,32 +435,47 @@ static void track_installed(const char *bundle_name)
...
 	free_string(&tracking_file);
 	if (fd < 0) {
+		ret = -1;
 		goto out;
 	}
 	close(fd);
 
 out:
 	free_string(&dst);
+	/* Consider checking ret and printing a warning for users
```
otaviobp: Go ahead and add a message using debug() here. It will only be printed if --debug is used.

src/bundle.c
```diff
@@ -760,8 +775,6 @@ static enum swupd_code install_bundles(struct list *bundles, struct list **subs,
 		if (is_tracked_bundle(bundle)) {
 			fprintf(stderr, "Warning: Bundle \"%s\" is already installed, skipping it...\n", bundle);
 			already_installed++;
-			/* track as installed since they tried to install it */
```
otaviobp: Not sure if we should remove this line. And we need better function names.
is_tracked_bundle() is checking if bundle is installed, and not if it's tracked.

src/bundle.c
bryteise: Okay, thanks.

src/bundle.c
bryteise: Ah indeed, I was fooled by the naming X(.

PR #819, opened 2019-02-12
Curl ssl

otaviobp:


src/curl.c
```diff
@@ -150,16 +152,23 @@ static int check_connection(const char *test_capath)
...
 	case CURLE_SSL_CACERT:
-		warn("Curl - Unable to verify server SSL certificate\n");
+		debug("Curl - Unable to verify server SSL certificate\n");
 		return -SWUPD_BAD_CERT;
 	case CURLE_SSL_CERTPROBLEM:
-		warn("Curl - Problem with the local client SSL certificate\n");
+		debug("Curl - Problem with the local client SSL certificate\n");
 		return -SWUPD_BAD_CERT;
 	case CURLE_OPERATION_TIMEDOUT:
+		debug("Curl - Check connection Timed out\n");
```
castulo: "Check connection"? it sounds strange, isn't it better to just use "connection timed out"? also drop the capital "T"

src/curl.c
```diff
@@ -219,6 +228,11 @@ int swupd_curl_init(void)
 		free_string(&str);
 	}
 
+	if (ret != 0) {
+		error("Failed to connect to server using local SSL ceritificates.\n");
```
castulo: you are adding "." to the end of the messages again, we agreed we wouldn't

src/curl.c
otaviobp: I'll change to "check_connection - timed out". It's a debug message and it's good to know where is this from.

otaviobp: No, we don't need that prefix. I'll leave just the string Time out

otaviobp:
Fixed issues and added another commit changing some other messages.

PR #820, opened 2019-02-12
Curl async - parallel downloads

otaviobp:


src/fullfile.c
```diff
@@ -28,7 +28,7 @@
 
 /* hysteresis thresholds */
 //FIXME #562
-#define MAX_XFER 27
+#define MAX_XFER 15
```
castulo: where did this value come from?

otaviobp: There's no easy way on setting that, because it depends a lot on the network speed and file sizes. I did some tests in multiple configurations and this looked like a good number for fullfiles.

For packs, I left 1 MAX_XFER because I couldn't get expressive improvements downloading more than that. In fullfiles we benefit from this because files are smaller and we can download more files while we are extracting a single one.

castulo: thanks for the explanation +1

PR #832, opened 2019-02-25
Consolidating minversion implementations in testlib

### LINKED ISSUE #618: Testlib: Consolidate Minversion Implementation
The update_minversion and bump_format testlib functions both perform a minversion update, but with separate implementations. These implementations should be consolidated.

castulo:
The update_minversion and bump_format testlib functions both perform a minversion update, but with separate implementations. These implementations should be consolidated.

Closes #618

### DIFF
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -474,70 +474,81 @@ create_tar() { # swupd_function
 
 }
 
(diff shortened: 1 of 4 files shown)


otaviobp:
lgtm, but I'll wait on @jwakre

PR #833, opened 2019-02-28
Generating delta packs for all previous versions

### LINKED ISSUE #615: Testlib.bash: Incorrect Delta Packs Generated
Currently, testlib.bash generates delta packs for a given bundle from its most recently changed version to the latest version. This is not correct because the current version may be older than a bundle's most recently changed version. For instance, if an update was performed from version 10 to version 30 and a bundle has been updated in versions 20 and 30, a delta pack will be generated for that bundle from version 20 to version 30. It is expected that a delta pack for that bundle should be generated for an update from version 10 to version 30.

castulo:
Currently, testlib.bash generates delta packs for a given bundle from its most recently changed version to the latest version. This is not sufficient because the current version may be older than a bundle's most recently changed version. For instance, if an update was performed from version 10 to version 30 and a bundle has been updated in versions 20 and 30, a delta pack will be generated for that bundle from version 20 to version 30. It is expected that a delta pack for that bundle should be generated for an update from version 10 to version 30.

This PR generates delta packs for all previous version of a bundle when an update is created.

Closes #615

### DIFF
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -322,7 +322,6 @@ copy_manifest() {
 	local bundle_file
 	local file_tar
 	local format
-	local prev_ver
 
 	from_path="$env_name"/web-dir/"$from_version"
 	to_path="$env_name"/web-dir/"$to_version"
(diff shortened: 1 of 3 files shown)


otaviobp:
lgtm

Issue #834, opened 2019-02-28
Swupd-search shouldn't fallback to old search because it is confusing and requires sudo


nsauzede:
**Is your feature request related to a problem? Please describe.**
It's really painful (insecure ?) to *have* to use sudo for some/most/any swupd search.
The "some/most/any" situation is the worst part : we can never know if this or that search would require sudo, eg: if the bundle in question is in cache (or not), also I understand that search "by file" always need to fill the cache (??)
Today, swupd search is not a local, offline, non-privileged task :
>$ sudo swupd search brand_new_bundle
>Searching for 'brand_new_bundle'
>Downloading Clear Linux manifests
>   62.70 MB total...
>Completed manifests download.
>Search term not found.
>Displaying top 5 file results per bundle

**Describe the solution you'd like**
To me, bundle searches should always use local cache; for all searches that is : by bundle "title" and by "file" (or whatever other search type)
To update the cache, I think this should be reserved to : sudo swupd update
Eg:
>$ swupd search brand_new_bundle
>Searching for 'brand_new_bundle'
>Search term not found.
>Displaying top 5 file results per bundle
>$ sudo swupd update
> ... update, update, oh! found brand_new_bundle! .. update, update
>$ swupd search brand_new_bundle
>Bundle with the best search result:
>     brand_new_bundle                          - Brand New Bundle  (42Mb)

**Describe alternatives you've considered**
The alternative is to always use sudo for such trivial searches, which seems to me very insecure if we want to reduce the potential attack/vuln surface etc..

**Additional context**
N/A

otaviobp:
I agree that there are some commands that require root and they shouldn't. But this case was already fixed with a new implementation. Can you test that in a newer Clear Linux release?

otaviobp:
And I forgot to ask you another feedback on that. Is there any other command that you think we should be able to run as a regular user? Thank you.

nsauzede:
@otaviobp I get this behaviour :
>$ cat /etc/os-release | grep VERSION_ID
>VERSION_ID=28060
>$ swupd search pdflatex
>This program must be run as root..aborting.
>$ swupd search bin/pdflatex
>This program must be run as root..aborting.

I will try to swupd update and the latest version..

otaviobp:
Thank you for the feedback. This is a bug and should be fixed. I'll rename this issue.

otaviobp:
[...]

PR #835, opened 2019-03-01
Running local search when searching

castulo:
swupd-search is a script that searches locally for the best bundle that
matches a search-term.

This commit runs the swupd-search script when users use the "swupd search"
command instead of calling the previous search code.

The previous search functionality will still be available for now using
the "search-legacy" command.


castulo:
There is a bug in swupd-search that causes the script to crash if no argument is provided when running the script. It should be fixed before merging this PR. I have already provided a fix for that issue.

castulo:
This PR should at least fix the problem of falling back to the old search being confusing in this issue #834

PR #841, opened 2019-03-06
Removing search heuristics

castulo:
The search command uses a score system to weight the results it finds,
the purpose of this is to try and show the most relevant results to the
user on top, however since the focus of the search command is changing,
we don't need to guess what the user is looking for and just show all we
find.

This commit removes the score ordering from the search command.

### DIFF
--- a/test/functional/search/search-sort.bats
+++ b/test/functional/search/search-sort.bats
@@ -41,7 +41,7 @@ global_teardown() {
 
 @test "SRH023: search shows the results in alphabetical order" {
 
-	run sudo sh -c "$SWUPD search-legacy $SWUPD_OPTS -T 10 alfa"
+	run sudo sh -c "$SWUPD search-file $SWUPD_OPTS --order=alpha alfa"
 
 	assert_status_is 0
 	expected_output=$(cat <<-EOM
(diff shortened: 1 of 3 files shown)


test/functional/search/search-sort.bats
```diff
@@ -0,0 +1,124 @@
...
+	expected_output=$(cat <<-EOM
+		Bundle alfa-charly	\\(2 MB to install\\)
+		./usr/share/clear/bundles/alfa-charly
+		Bundle alfa-zulu	\\(1 MB to install\\)
+		./alfaromeo/echo
+		./bar/alfa
+		./foo/alfa
+		./usr/share/clear/bundles/alfa-zulu
+		./yankee/alfalfa
+		Bundle victor	\\(3 MB to install\\)
```
otaviobp: shouldn't we order bundle names too?

castulo: what do you mean? bundle names are also ordered alphabetically:
Bundle alfa-charly
Bundle alfa-zulu
Bundle victor

otaviobp: hahhaa. I got confused when I read the bundle names. You are right.

PR #843, opened 2019-03-07
Set signature verification enabled as default

otaviobp:
Signature verification was disabled as a default parameters to make it easier
for people to build the project on other distros. But this can be dangerous
as users may start using a built swupd without signature checking to
install a CL distro.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -62,13 +62,14 @@ AS_IF(
 
 AC_ARG_ENABLE(
   [signature-verification],
-  [AS_HELP_STRING([--enable-signature-verification], [Enable signature check (disabled by default)])],
-  [AC_DEFINE([SIGNATURES], 1, [Enable signature check as default])]
+  [AS_HELP_STRING([--disable-signature-verification], [Disable signature check (enabled by default)])]
 )
-SIGVERIFICATION="no"
+SIGVERIFICATION="yes"
 AS_IF(
-  [test -n "$enable_signature_verification" -a "$enable_signature_verification" = "yes"],
-  [SIGVERIFICATION="$enable_signature_verification"]
+  [test "x$enable_signature_verification" != "xno"],
+  [AC_DEFINE(SIGNATURES, 1, [Enable signature check])
+     SIGVERIFICATION="yes"],
+  [SIGVERIFICATION="no"]
 )
 
 AC_ARG_ENABLE(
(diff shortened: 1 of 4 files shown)


phmccarty:
@otaviobp Can you provide a little more context to justify enabling this by default? I don't quite follow your reasoning. Who is building swupd to install Clear Linux? The installer images ship with the swupd-client that is part of the distro, correct?

otaviobp:
No one is necessary building Swupd to install clear. Build a person could do that. And if he/she uses the default settings we won't check for signatures. I prefer to have a --disable in the case the person doesn't have the required libraries so the person knows that signatures won't be checked.

configure.ac
```diff
@@ -47,10 +47,6 @@ AC_ARG_ENABLE(
   [AS_HELP_STRING([--disable-bzip2], [Do not use bzip2 compression (uses bzip2 by default)])]
 )
 BZIP="yes"
-AS_IF(
-  [test -n "$enable_bzip2" -a "$enable_bzip2" = "yes"],
-  [BZIP="$enable_bzip2"]
-)
```
phmccarty: Removal of this `AS_IF` block is unrelated, right?

configure.ac
```diff
@@ -60,17 +56,21 @@ AS_IF(
      BZIP="no"]
 )
 
+
+
 AC_ARG_ENABLE(
   [signature-verification],
-  [AS_HELP_STRING([--enable-signature-verification], [Enable signature check (disabled by default)])],
-  [AC_DEFINE([SIGNATURES], 1, [Enable signature check as default])]
+  [AS_HELP_STRING([--disable-signature-verification], [Dsdasdasdsodwd by default)])]
```
phmccarty: The help string here should be fixed up :)

configure.ac
```diff
@@ -60,17 +56,21 @@ AS_IF(
...
 )
-SIGVERIFICATION="no"
+SIGVERIFICATION="yes"
 AS_IF(
-  [test -n "$enable_signature_verification" -a "$enable_signature_verification" = "yes"],
-  [SIGVERIFICATION="$enable_signature_verification"]
+  [test "x$enable_signature_verification" != "xno"],
+  [AC_DEFINE(SIGNATURES, 1, [Use bzip2 compression])
```
phmccarty: Can you fix up the doc string here, too?

otaviobp:
thanks for the review. It took me some time to get things right and I forgot to fix the docstrings and check for leftovers.
Fixed now

PR #846, opened 2019-03-08
Create machine readable output for bundle-add

### LINKED ISSUE #294: Be consistent about where output goes.

### LINKED ISSUE #433: Create a machine readable output for swupd commmands
Print a status reporting how many steps swupd-client is supposed to report before the iterations calling ```print_progress()``` starts.

Bonus: It would be nice to report a description for the step.

i.e:
```bash
[step] verifying fix path: 10
.
.
.
```

### LINKED ISSUE #489: Print warnings and errors consistently
We sometimes print warnings like
```
WARNING: ...
```
sometimes like
```
Warning: ...
```
and sometimes just
```
...
```
without the prefix.

We should have `print_warn` and `print_error` functions to make our warnings and errors consistent. They should accept `va_list` type arguments for formatted strings.

### LINKED ISSUE #864: Replace all printf() and fprintf() calls to functions in log.h
In order to make sure all commands will support the json formatter we need to make sure we use only functions on log.h file.

castulo:
This PR adds the `--json-output` option for the bundle-add command in swupd which formats the output generated by swupd to comply with the JSON format. This way the output can be read and parsed easily by other applications that have an interest in real time status of a swupd operation.

These tasks are part of the PR:
 - [x] add the `--json-output` flag to the `bundle-add` command
 - [x] implement a formatter which will grab a normal string literal and will wrap it with a valid JSON format
 - [x] replace current messages in `bundle-add` with the formatted messages
 - [x] add a function to report an exit status of the swupd operation in JSON format to the formatter
 - [x] report the status of the add operation using the `json_status` function.
 - [x] add a `json_progress` function to report progress in JSON format to the formatter
 - [x] report progress using the `json_progress` function throughout the bundle add process, specially during the three steps:
     - Download packs
     - Installing bundle(s) files onto filesystem
     - Download missing files
 - [x] create tests for the new functionality

Closes #433 
Closes #294
Closes #489 
Closes #864


src/lib/log.c
```diff
@@ -19,19 +19,94 @@
...
+	}
+}
+
+void format_to_json(const char *type, const char *msg, va_list args_list)
+{
+	const int MAX_TYPE_LENGTH = 8;
+	char *full_msg;
+	char ltype[MAX_TYPE_LENGTH];
+
+	/* make sure the type is all lower case */
```
otaviobp: It looks like there's no easy way to perform a toLower in a string in c :(
This is one thing that could move to one specific function, like str_lower(). You can create the string using malloc(strlen(string)) and return that, so we don't need to assume there's a max length.

src/lib/log.c
```diff
@@ -19,19 +19,94 @@
...
+	if (vasprintf(&full_msg, msg, args_list) < 0) {
+		abort();
+	}
+
+	if (strcmp(full_msg, "") != 0) {
+		int j = 0;
+		for (unsigned int i = 0; i <= strlen(full_msg); i++) {
+			/* if the message has double quotes replace them with single quotes
+			* since double quotes are not supported in JSON */
+			if (full_msg[i] == '"') {
```
otaviobp: actually " are supported, but we need to escape them (https://www.json.org has a really good chart for json format). I'm ok on changing that to ' though.

src/lib/log.c
```diff
@@ -19,19 +19,94 @@
...
+				full_msg[i] = '\'';
+			}
+			/* remove all '\n' from the message */
+			if (full_msg[i] != '\n') {
```
otaviobp: I would replace \n with a space. So we won't connect words. "this\nword" could be "this word" instead of "thisword"

src/lib/log.c
```diff
@@ -19,19 +19,94 @@
...
+{
+	json_format = 1;
+}
+
+void json_start(const char *op)
+{
+	if (json_format) {
+		fprintf(stderr, "[\n");
+		fflush(stderr);
+		fprintf(stderr, "{ \"%s\" : \"start\" },\n", op);
```
otaviobp: we could use the same format here:

    {type: "start", section: "bundle-add" }

src/lib/log.c
otaviobp: and '\' isn't supported too.

otaviobp:
I also pushed a commit to your branch adding a suggestion to keep json functions in a different file. By doing this it is easier to extend swupd to support more formatters if we need to.

src/lib/log.c
castulo: good idea

src/lib/log.c
castulo: ok, I can make that change

src/lib/log.c
castulo: > and '' isn't supported too.

I didn't get this

otaviobp: Probably I removed the character by accident. I was trying to say that \ isn't a valid json character too because it's used as a escape char. We need to replace all \ to \\ (or just remove \ because probably we aren't even using it).

src/lib/log.c
castulo: ok, done

test/functional/bundleadd/add-basics.bats
```diff
@@ -44,15 +44,24 @@ global_teardown() {
...
+		.*...66%
+		.*...83%
+		.*...100%
+
 		Calling post-update helper scripts.
 		Successfully installed 1 bundle
 	EOM
 	)
-	assert_is_output --identical "$expected_output"
+	assert_regex_is_output --identical "$expected_output"
```
otaviobp: Can't we use assert_regex_in_output() here so we don't need to have all % printed in the test file?

src/swupd.h
```diff
@@ -278,8 +279,8 @@ static inline void account_delta_miss(void)
 
 extern void print_statistics(int version1, int version2);
 
-extern int download_fullfiles(struct list *files, int *num_downloads);
-extern int download_subscribed_packs(struct list *subs, struct manifest *mom, bool required);
+extern int download_fullfiles(struct list *files, int *num_downloads, struct step step);
```
otaviobp: we need a better way to handle this step struct. I don't think that the the fullfiles.c or packs.c should care about the step we are in to.

src/manifest.c
```diff
@@ -729,7 +729,7 @@ struct manifest *load_manifest(int version, struct file *file, struct manifest *
 			retried = true;
 			goto retry_load;
 		}
-		fprintf(stderr, "Failed to load %d %s manifest\n", version, file->filename);
+		info("Failed to load %d %s manifest\n", version, file->filename);
```
otaviobp: "failed" messages are probably error() or warn().
Use error() if we are going to abort and warn() if we are going to try to do something else.

src/lib/formatter_json.c
```diff
@@ -0,0 +1,163 @@
...
+void complete_step(unsigned int current, unsigned int total, char *description)
+{
+	struct step step;
+
+	step.current = current;
+	step.total = total;
+	step.description = description;
+
+	if (json_format) {
+		print_step_progress(step, 1, 1);
```
otaviobp: Maybe we can save the step we are in on this json_formatter file and use it on print_progress(). so we wouldn't need to set this in many areas of the code. We could have a json_formatter_set_step() and json_formatter_get_step() and create a step static in this file.

test/functional/bundleadd/add-json.bats
```diff
@@ -0,0 +1,84 @@
...
+	# command is created as a JSON stream that can be read and parsed by other
+	# applications interested into knowing real time status of the command
+	# (for example installer so it can provide the user a status of the install process)
+
+	run sudo sh -c "$SWUPD bundle-add --json-output $SWUPD_OPTS test-bundle"
+
+	assert_status_is 0
+	expected_output=$(cat <<-EOM
+		[
+		{ "type" : "start", "section" : "bundle-add" },
```
otaviobp: Do we really need that section json? We have only one section for each command and the caller already knows what command we did run.

src/manifest.c
castulo: I didn't do that as part of this PR because of two reasons:
1) that was going to change the swupd output, which was going to break the current tests.
2) we need to consider how we want to handle this, because using the error() function is going to prepend an "Error: " label to the message, but sometimes error messages are cascade from within inner functions, so we might end up having lots of "Error" labels in the output. For example, in this particular case, let's say we are trying to load a manifest with an invalid number of files in it. These are the messages that would be printed now:
```
Error: preposterous (9000000) number of files in test-bundle Manifest, more than 4 million skipping
Warning: Removing corrupt Manifest.test-bundle artifacts and re-downloading...
Error: preposterous (9000000) number of files in test-bundle Manifest, more than 4 million skipping
Failed to load 10 test-bundle manifest
Unable to download manifest test-bundle version 10, exiting now
Failed to install 1 of 1 bundles
```

If we use error() for every error message we are going to end up with something like this:
```
Error: preposterous (9000000) number of files in test-bundle Manifest, more than 4 million skipping
Warning: Removing corrupt Manifest.test-bundle artifacts and re-downloading...
Error: preposterous (9000000) number of files in test-bundle Manifest, more than 4 million skipping
Error: Failed to load 10 test-bundle manifest
Error: Unable to download manifest test-bundle version 10, exiting now
Error: Failed to install 1 of 1 bundles
```

which doesn't look pretty, it would seem like multiple errors happened while it was just one that was printed many times in different parts of the code.

castulo: @otaviobp what do you think we should do about that ^?

test/functional/bundleadd/add-basics.bats
castulo: actually the percentages are normally being filtered out from the output by the test library (look at the `use_ignore_list()` function). But I specifically instructed testlib NOT to strip the percentages out from the output in this case by using the `--identical` flag. I thought it was valuable to have one test that would validate the percentages are being printed out correctly in case a change in the code breaks this output for some reason, otherwise we wouldn't catch that cause we are always filtering that out.

test/functional/bundleadd/add-json.bats
castulo: These sections:

```
[
 { "type" : "start", "section" : "bundle-add" },
```
and 
```
{ "type" : "end", "section" : "bundle-add" }
 ]
```
were added for two reasons:
- so we could have a way to instruct the code to open and close an array in the JSON stream "[ .. ]"
- and most importantly so we could remove the "," from the last item in the array. If we just close the array and there is a colon after the last element, the JSON is invalid. There is no easy way to remove that "," from the last element if we remove this cause we don't know when one element is the last.

src/lib/formatter_json.c
castulo: That could work. I'll work on making that change.

src/swupd.h
castulo: I think this would be solved by implementing your suggestions about the `json_formatter_set_step()` and `json_formatter_get_step()`.

src/manifest.c
otaviobp: I think that all messages that are errors/warnings/fails should use warn() or error() and have the prefix appended. We can review the messages to avoid cascading that much (maybe by using debug() for some).

It's totally ok to let this out of this patch though, because of fixing tests. And fix that later, in another commit.

test/functional/bundleadd/add-json.bats
otaviobp: +1 for the "," problem

src/manifest.c
castulo: > I think that all messages that are errors/warnings/fails should use warn() or error() and have the prefix appended. We can review the messages to avoid cascading that much (maybe by using debug() for some).

Make sense

src/lib/progress.h
```diff
@@ -0,0 +1,65 @@
...
+ * Finishes the reporting of a process (as defined above). If the
+ * format used, needs a special closing statement, this function will
+ * print it.
+ */
+void finish_steps(char *);
+
+/*
+ * Sets a step of the process to be tracked/reported.
+ */
+void set_step(unsigned int, char *);
```
otaviobp: add a prefix to all functions in this file. Like progress_ It's way easier to find where a function come from.

src/lib/progress.h
```diff
@@ -0,0 +1,65 @@
...
+struct step {
+	unsigned int current;
+	unsigned int total;
+	char *description;
+};
+
+/*
+ * available formats the output can be printed
+ */
+enum progress_formats {
```
otaviobp: it would be better to use the same approach on log for that. Instead of having one enum for types we can have a function pointer for a report_progress(). We can have a type set to start, end or progress. And the json could print the start/end messages or progress messages depending on the type.

otaviobp:
I also tested with the golang sample I posted on the bug report and it's working file. Entries are consumed as they are printed and not just at the end. Cool

otaviobp:
Note: We need to review all usages of fprintf in the code. If any fprintf(stderr, ...) is used during the execution, parser will break.

src/lib/progress.h
castulo: ok, I'll do that

src/lib/progress.h
castulo: ok, good idea

castulo:
I have addressed all comments @otaviobp , let me know if there is something else that needs to be done with this PR.

src/lib/formatter_json.c
```diff
@@ -0,0 +1,115 @@
...
+	if (json_format) {
+		fprintf(stderr, "{ \"type\" : \"end\", \"section\" : \"%s\" }\n]\n", op);
+		fflush(stderr);
+	}
+}
+
+void json_status(int status)
+{
+	if (json_format) {
+		fprintf(stderr, "{ \"type\" : \"status\", \"status\" : \"%d\" },\n", status);
```
otaviobp: Use int instead of a string for status in the json output.

    status: 31

instead of

    status: "31"

I would move the status to the end json object:

    { type: "end", section: "bundle-add", status=31 }

castulo: good idea about using int for numbers in the JSON output. I will change that also for the "progress" messages like this:
`{ "type" : "progress", "currentStep" : 1, "totalSteps" : 7, "stepCompletion" : 100, "stepDescription" : "download_bundle_manifests" }`

castulo: about moving status to the `end` json object... don't you think there may be cases where we want to report an intermediate status during a process? so if we merge them then we won't have that flexibility. One possibility would be to include status in the `end` object but also leave the `json_status` function in case we want to use it some time for some use case. What do you think @otaviobp ?

otaviobp: I don't know. Good point. Let it the way it is then

castulo: I will add the status to the json `end` object, I anticipate this is the one we'll use most of the time, but I will also leave the `json_status` function in case we ever need it. We can remove it in the future if we don't see any use for it.

josuedhg: great! I think use int instead string must be useful also for the gnome-software.

PR #856, opened 2019-03-18
Travis improvements to speed up the build

### LINKED ISSUE #852: The build job is exceeding the timeout in Travis
Travis has a timeout of 50 minutes per job in its public infra, at the moment the swupd build has reached the point where it takes longer than that so it is timing out, therefore failing to build.

From the Travis documentation:

> Build Timeouts 
> 
> It is very common for test suites or build scripts to hang. Travis CI has specific time limits for each job, and will stop the build and add an error message to the build log in the following situations:
> 
> When a job produces no log output for 10 minutes.
> When a job on a public repository takes longer than 50 minutes.
> When a job on a private repository takes longer than 120 minutes.

One possibility to explore is to see if we can split the testing, which is the most time consuming into two or more jobs that take less than 50 minutes each, so the build doesn't time out.

> There is no timeout for a build; a build will run as long as needed to complete all the jobs, but will stop immediately if any job hits a timeout limit.

### This is the proposed solution:
- [x] Run `make compliant` and `shellcheck` before running `make` and running the functional tests  so if there is code style issues with the code the build fails right away instead of failing after all tests have finished. This will allow use to fail faster if the build is going to fail anyway.
- [x]	Use the correct number of parallel jobs in make. From the documentation: "See [Virtualization environments](https://docs.travis-ci.com/user/reference/overview/#virtualization-environments) to determine how many CPUs an environment normally has and set the make job parameter to a similar number (or slightly higher if your build frequently waits on disk I/O)". Ubuntu environments have only 2 cores.
- [x] Running functional tests is the most time consuming task in the build process, so if we split them in different jobs instead of running them in a single job we can avoid reaching the timeout.

castulo:
### This is the proposed solution:
- [x] Run `make compliant` and `shellcheck` before running `make` and running the functional tests  so if there is code style issues with the code the build fails right away instead of failing after all tests have finished. This will allow use to fail faster if the build is going to fail anyway.
- [x]	Use the correct number of parallel jobs in make. From the documentation: "See [Virtualization environments](https://docs.travis-ci.com/user/reference/overview/#virtualization-environments) to determine how many CPUs an environment normally has and set the make job parameter to a similar number (or slightly higher if your build frequently waits on disk I/O)". Ubuntu environments have only 2 cores.
- [x] Running functional tests is the most time consuming task in the build process, so if we split them in different jobs instead of running them in a single job we can avoid reaching the timeout.

Closes #852

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,18 @@ language: c
 
 env:
     global:
-        - MAKEFLAGS="-j 32"
+        - MAKEFLAGS="-j 2"
+    matrix:
+        - FUNCTIONAL_TESTS=test/functional/update
+        - FUNCTIONAL_TESTS=test/functional/verify
+        - FUNCTIONAL_TESTS=test/functional/bundleadd
+        - FUNCTIONAL_TESTS=test/functional/bundleremove
+        - FUNCTIONAL_TESTS=test/functional/bundlelist
+        - FUNCTIONAL_TESTS=test/functional/search
+        - FUNCTIONAL_TESTS=test/functional/checkupdate
+        - FUNCTIONAL_TESTS=test/functional/hashdump
+        - FUNCTIONAL_TESTS=test/functional/mirror
+        - FUNCTIONAL_TESTS=test/functional/usability
 
 # Pre-install missing build dependencies:
 # - libcheck 0.9.10 is slightly too old, since 0.9.12 adds TAP support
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,6 +2,10 @@ sudo: required
 dist: xenial
 language: c
 
+env:
+    global:
+        - MAKEFLAGS="-j 32"
```
otaviobp: shouldn't we use 2 here then?

.travis.yml
```diff
@@ -39,13 +50,13 @@ before_script:
...
         - sudo find test/functional -exec chmod g-w {} \;
         - make &&
-          make check &&
           sudo sh -c 'umask 0022 && make install' &&
-          sudo sh -c 'umask 0022 && make install-check'
-
-after_failure: cat test-suite.log
+          sudo sh -c 'umask 0022 && make install-check' &&
+          sudo sh -c 'umask 0022 && make unit-check' &&
+          bats "$FUNCTIONAL_TESTS"
```
otaviobp: can't we use:

    env TESTS=$(FUNCTIONAL_TESTS) make -e check

.travis.yml
```diff
@@ -4,7 +4,18 @@ language: c
 
 env:
     global:
-        - MAKEFLAGS="-j 32"
+        - MAKEFLAGS="-j 2"
+    matrix:
+        - FUNCTIONAL_TESTS=test/functional/update
+        - FUNCTIONAL_TESTS=test/functional/verify
```
otaviobp: This doesn't look to be working :(

otaviobp: my bad. it is working

PR #867, opened 2019-03-21
Manifest

### LINKED ISSUE #858: Manifest parser should support extensions
Currently the Manifest parser aborts when there's any field unexpected in the Manifest. This is a pain every time we are trying to add a new feature to the Manifest. We could abort only when any supported feature is different from what's expected, but continue normally if there new fields or flags.

otaviobp:
Reorganizing manifest parser, adding some improvements and making it easier to extend Manifests without breaking swupd. 

Fixes #858

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -77,314 +77,47 @@ int file_sort_filename_reverse(const void *a, const void *b)
 	return -ret;
 }
 
-static struct manifest *alloc_manifest(int version, char *component)
-{
-	struct manifest *manifest;
-
-	manifest = calloc(1, sizeof(struct manifest));
-	ON_NULL_ABORT(manifest);
-
-	manifest->version = version;
-	manifest->component = strdup_or_die(component);
-
-	return manifest;
-}
-
 static struct manifest *manifest_from_file(int version, char *component, bool header_only, bool is_mix)
 {
-	FILE *infile;
-	char line[MANIFEST_LINE_MAXLEN], *c, *c2;
-	int count = 0;
-	int deleted = 0;
-	int err;
-	struct manifest *manifest;
-	struct list *includes = NULL;
 	char *filename;
 	char *basedir;
-	unsigned long long filecount = 0;
-	unsigned long long contentsize = 0;
-	int manifest_hdr_version;
-	int manifest_enc_version;
+	struct manifest *manifest;
 
 	if (!is_mix) {
 		basedir = state_dir;
 	} else {
 		basedir = MIX_STATE_DIR;
 	}
-	string_or_die(&filename, "%s/%i/Manifest.%s", basedir, version, component);
 
-	infile = fopen(filename, "rbm");
-	if (infile == NULL) {
-		free_string(&filename);
+	string_or_die(&filename, "%s/%i/Manifest.%s", basedir, version, component);
+	manifest = manifest_parse(component, filename, header_only);
+	free(filename);
+
+	if (!manifest) {
 		return NULL;
 	}
-	free_string(&filename);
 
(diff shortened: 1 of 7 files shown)


src/manifest_parser.c
```diff
@@ -0,0 +1,313 @@
+/*
+ *   Software Updater - client side
+ *
+ *      Copyright © 2012-2016 Intel Corporation.
```
castulo: 2016?

src/manifest_parser.c
```diff
@@ -0,0 +1,313 @@
...
+
+	if (strncmp(line, "MANIFEST\t", 9) != 0) {
+		goto err_close;
+	}
+
+	c = &line[9];
+	err = strtoi_err(c, &manifest_enc_version);
+
+	if (manifest_enc_version <= 0 || err != 0) {
+		fprintf(stderr, "Error: Loaded incompatible manifest version\n");
```
castulo: errors should use the error() function

src/manifest_parser.c
```diff
@@ -0,0 +1,313 @@
...
+				 * to be big enough to allow Manifest.Full to
+				 * pass (currently about 450,000 March 18) but
+				 * small enough that when multiplied by
+				 * sizeof(struct file) it fits into
+				 * size_t. For a system with size_t being a 32
+				 * bit value, this constrains it to be less
+				 * than about 6,000,000, but close to infinity
+				 * for systems with 64 bit size_t.
+				 */
+				fprintf(stderr, "Error: preposterous (%llu) number of files in %s Manifest, more than 4 million skipping\n",
```
castulo: same comment for all error messages in this section, use error() function

src/manifest.c
```diff
@@ -77,323 +77,47 @@ int file_sort_filename_reverse(const void *a, const void *b)
...
-	struct manifest *manifest;
-	struct list *includes = NULL;
 	char *filename;
 	char *basedir;
-	unsigned long long filecount = 0;
-	unsigned long long contentsize = 0;
-	int manifest_hdr_version;
-	int manifest_enc_version;
+	struct manifest *manifest;
+	struct list *l;
```
castulo: I would prefer if we use variable names that are more descriptive and longer than one letter, that makes the code more readable.

src/manifest.h
```diff
@@ -0,0 +1,38 @@
...
+#endif
+
+struct manifest {
+	// Header
+	int manifest_version;
+	int version;
+	uint64_t filecount;
+	uint64_t contentsize;
+	struct list *includes; /* list of strings with manifest names */
+	//struct list *actions;     /* list of strings with actions */
```
castulo: leftover?

src/manifest.h
```diff
@@ -0,0 +1,38 @@
...
+	// File list
+	struct list *files;     /* struct file for files */
+	struct list *manifests; /* struct file for possible manifests */
+
+	// Helper data
+	struct list *submanifests; /* struct manifest for subscribed manifests */
+	unsigned int is_mix : 1;
+};
+
+struct manifest *manifest_parse(const char *component, const char *filename, bool header_only);
```
castulo: Most other header files include a brief description of what the function is for, this one doesn't.

src/manifest.c
otaviobp: This list is used as an iterator in a for, so I don't think we need a better name, but I could move it to a proper place, just before it's used.

src/manifest.h
otaviobp: haha, I totally forgot to add comments here.

otaviobp:
Fixed the proposed changes

PR #868, opened 2019-03-21
Adding the --quiet flag to hashdump

castulo:
Currently when calculating the hash from a file two lines are printed:

Calculating hash with xattrs for: my_file
5114e8e3cc667fa80eb5407e5825c618e6668503aba53c8cca7f8daf84cec03f

The first line goes to STDERR while the second one goes to STDOUT. The
test library uses the hash that goes to STDOUT to perform many tasks, so
it is important the output can be filtered to just print the hash.

In the near future, all messages printed with printf or fprintf will be
replaced by logger functions in all the source code which will cause both
lines to be printed to the output which will break the test library.

This commit adds the --quiet flag to the hashdump command so that only
the calculated hashdump is printed on screen.

### DIFF
--- a/src/hashdump.c
+++ b/src/hashdump.c
@@ -33,26 +33,29 @@
 /* outputs the hash of a file */
 
 static bool use_prefix = false;
+static int log_level = LOG_INFO;
 
 static struct option opts[] = {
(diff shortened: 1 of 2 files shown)


castulo:
This PR is related to #864

PR #872, opened 2019-03-26
Verify: Fix tests on test/real_content/

otaviobp:


src/extra_files.c
```diff
@@ -98,7 +98,7 @@ static bool handle(const char *filename, bool is_dir, bool fix)
 
 	if (fix) {
 		string_or_die(&temp, "%s%s", path_prefix, filename);
-		info(
+		print(
```
castulo: Do you really think all these messages have to be printed even when the user chooses the --quiet option? I mean, why would he/she be interested in what files were removed or not, if you are running verify and use --quiet I would expect only to know if the verify succeded or not, not what files were repaired or deleted. Maybe I would only be interested in the total number of files repaired/deleted, not each one... that sounds like more info I need when in --quiet mode.

otaviobp: The ideal place to get the information on if the verify was succeeded or not is the return code. I'd like to be able to do something like this:

    ret=$(swupd verify)
    if [[ "$ret" == "1" ]]; then
        echo "System corrupt"
    fi

There's a discussion on #294 that it's important to have one way on getting a list of files verified so we can use that on scripts. So we can do something like:

    swupd verify > corrupted_files

We can print the number of corrupted files by:

    cat corrupted_files | wc -l #print number of corrupted files

The output is far from perfect though. it's very inconsistent and hard to parse, but I still prefer to print the files then the messages when --quiet is used. I think --quiet needs to print the minimum number of information that makes the command useful.

PR #878, opened 2019-03-27
Enabling the --json-output flag as a global option

### LINKED ISSUE #869: Enabling the --json-output flag as a global option in swupd
--json-output can be enabled for every command, so it make sense to convert it to be a global flag instead of implementing it in every command.

castulo:
The --json-output flag can be enabled for every swupd command.

This commit converts it to be a global flag, and enables it in every
swupd command.

Closes #869


src/hashdump.c
```diff
@@ -56,6 +56,8 @@ static void usage(const char *name)
 	print("\n");
 	print("The filename is the name of a file on the filesystem.\n");
 	print("\n");
+	/* TODO(castulo): add --debug and --quiet to the menu options once we are ready
```
otaviobp: This was already addressed on other PR, right?

castulo: yep, I forgot to remove it

otaviobp:
@castulo Can you rebase that?

castulo:
> @castulo Can you rebase that?

yeah, I'm on it

PR #880, opened 2019-03-29
Remove test ID duplication

otaviobp:


test/functional/usability/usa-tests-consistency.bats
```diff
@@ -0,0 +1,31 @@
...
+}
+
+test_teardown() {
+
+	# do nothing
+	return
+
+}
+
+@test "USA010: Check if there are any duplicated test ID" {
```
castulo: instead of adding this as a test we should add it somewhere at the beginning of the Travis job, maybe right after running "make compliant" and "shellcheck", so we get a faster failure if the IDs are wrong so we don't have to wait for the tests to run.

PR #883, opened 2019-04-05
Report progress based on downloaded content

### LINKED ISSUE #877: Improve progress report when downloading files
When we are downloading files we report progress on each completed file (pack or fullfile). We could instead report progress according to the total download size.

Steps:
 - Query for the total download size before starting the download. 
 - Use curl callback to get how much we downloaded so far and update progress accordingly.

castulo:
When swupd reports progress of content it needs, it does so by counting
how many items it needs to download (fullfiles or packs) and how many it
has already downloaded. This gives a rough estimate of what is the
progress of the overall download, but it can also be very misleading
since some files may be very different in size compared to others. This
is specially true when talking about packs, one pack could be a couple
of Mb big while another one could be a few houndred Mbs.

This PR changes the reporting of progress for packs nd fullfiles so swupd reports
progress based on how much of the total Mbs we have downloaded instead
of how many files, this will give a more accurate feeling of where the overall
process is.

Closes #877


src/packs.c
```diff
@@ -167,21 +167,46 @@ int download_subscribed_packs(struct list *subs, struct manifest *mom, bool requ
...
+	download_handle = swupd_curl_parallel_download_start(get_max_xfer(MAX_XFER));
 	swupd_curl_parallel_download_set_callbacks(download_handle, download_successful, download_error, download_free_data);
+
+	info("Downloading packs (%.2lf Mb) for:\n", download_size / 1000000);
+	iter = list_head(subs);
+	while (iter) {
+		sub = iter->data;
+		iter = iter->next;
+
+		if (!is_installed_bundle(sub->component)) {
```
otaviobp: This looks like a bug fix that isn't related to this patch. If so, it should be in a different commit.

test/functional/bundleadd/add-json.bats
```diff
@@ -24,61 +24,67 @@ test_setup() {
...
-		\{ "type" : "end", "section" : "bundle-add", "status" : 0 \}
-		\]
+		[
+		{ "type" : "start", "section" : "bundle-add" },
+		{ "type" : "progress", "currentStep" : 1, "totalSteps" : 7, "stepCompletion" : 100, "stepDescription" : "load_manifests" },
+		{ "type" : "progress", "currentStep" : 2, "totalSteps" : 7, "stepCompletion" : 100, "stepDescription" : "consolidate_files" },
+		{ "type" : "progress", "currentStep" : 3, "totalSteps" : 7, "stepCompletion" : 100, "stepDescription" : "check_disk_space_availability" },
+		{ "type" : "info", "msg" : "Downloading packs (0.13 Mb) for: " },
+		{ "type" : "info", "msg" : " - test-bundle " },
+		{ "type" : "progress", "currentStep" : 4, "totalSteps" : 7, "stepCompletion" : 12, "stepDescription" : "download_packs" },
```
otaviobp: it's very very very hard to maintain this test. I'd use regexp to check only for the things we are really interested on.

src/packs.c
```diff
@@ -199,7 +223,64 @@ int download_subscribed_packs(struct list *subs, struct manifest *mom, bool requ
...
+	struct file *bundle = NULL;
+	char *targetfile = NULL;
+	char *url = NULL;
+	struct stat stat;
+	int count = 0;
+
+	for (list = list_head(subs); list; list = list->next) {
+		sub = list->data;
+
+		if (sub->oldversion == sub->version) {
```
otaviobp: This code is duplicated from download_pack. This is hard to maintain. If we change the strategy on when to download packs we would need to change in both functions.

src/packs.c
```diff
@@ -199,7 +240,51 @@ int download_subscribed_packs(struct list *subs, struct manifest *mom, bool requ
...
 		}
 	}
-	progress_report(list_length, list_length); /* Force out 100% */
 	info("\n");
+	list_free_list(need_download);
+
 	return swupd_curl_parallel_download_end(download_handle, NULL);
 }
+
+double packs_query_total_download_size(struct list *subs, struct manifest *mom)
```
otaviobp: this functions should be static and not exported

src/swupd.h
```diff
@@ -266,7 +266,9 @@ static inline void account_delta_miss(void)
 extern void print_statistics(int version1, int version2);
 
 extern int download_fullfiles(struct list *files, int *num_downloads);
+extern double fullfile_query_total_download_size(struct list *files);
```
otaviobp: this should be static

src/swupd.h
```diff
@@ -266,7 +266,9 @@ static inline void account_delta_miss(void)
...
 extern int download_subscribed_packs(struct list *subs, struct manifest *mom, bool required);
+extern double packs_query_total_download_size(struct list *subs, struct manifest *mom);
```
otaviobp: this should be static

src/curl_async.c
```diff
@@ -425,6 +460,16 @@ static int process_download(struct swupd_curl_parallel_handle *h, struct multi_c
 	}
 	file->curl = curl;
 
+	curl_ret = curl_easy_setopt(curl, CURLOPT_XFERINFOFUNCTION, progress_callback);
+	if (curl_ret != CURLE_OK) {
+		goto out_bad;
+	}
+
+	curl_ret = curl_easy_setopt(curl, CURLOPT_NOPROGRESS, 0L);
```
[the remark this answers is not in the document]
castulo: done

src/packs.c
```diff
@@ -167,21 +167,46 @@ int download_subscribed_packs(struct list *subs, struct manifest *mom, bool requ
...
-	unsigned int list_length = list_len(subs);
-	unsigned int complete = 0;
 	void *download_handle;
+	double download_size;
 
-	info("Downloading packs...\n");
-	download_handle = swupd_curl_parallel_download_start(get_max_xfer(MAX_XFER));
+	download_size = packs_query_total_download_size(subs, mom);
+	if (download_size < 0) {
+		warn("The size of the packs could not be determined, skipping pack download\n");
```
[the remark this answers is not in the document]
castulo: yeah, you are right, I'll change that

src/swupd.h
castulo: yes you are right, it should.

src/packs.c
castulo: done

src/curl_async.c
```diff
@@ -80,6 +80,8 @@
  */
 #define HASH_TO_KEY(hash) (HASH_VALUE(hash[0]) << 4 | HASH_VALUE(hash[1]))
 
+static double total_download_size = 0;
```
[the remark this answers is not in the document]
castulo: good idea!

src/packs.c
castulo: this duplication was removed in a latter commit

src/packs.c
castulo: yes, this is a fix to a bug that I found while working on this.... since this fix ended up in a section of code that didn't exist in the original code it would be difficult to separate it in a different commit. If you really think I should try to separate it, it would be easier to move it to a whole different PR.

otaviobp: If it's hard you can leave it here. But add a comment in the commit message explaining that you are also fixing this bug.

src/curl_async.c
```diff
@@ -412,6 +419,34 @@ static int poll_fewer_than(struct swupd_curl_parallel_handle *h, size_t xfer_que
...
+		/* new file */
+		dlprev = 0;
+	}
+	increment = dlnow - dlprev;
+	dlprev = dlnow;
+	downloaded += increment;
+
+	/* if more data has been downloaded, report progress */
+	if (dltotal && increment > 0) {
+		progress_report(downloaded, total_download_size);
```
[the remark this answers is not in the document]
castulo: ok, I'll handle this through a progress_callback.

test/functional/bundleadd/add-json.bats
castulo: This problem was fixed in a latter commit.

castulo:
I just found a problem with this PR, @otaviobp please hold on reviewing until further notice 🚫

castulo:
I fixed the issue, it is now ready for review @otaviobp

otaviobp:
I did some tests and I found a bug on download fullfile. I tried this:

    swupd verify --quick --fix -B os-core,vim --path /custom -F 27 --force -m 28100 -u https://cdn.download.clearlinux.org/update/

And the verify command failed. It works fine with master.

src/curl_async.c
```diff
@@ -224,6 +226,20 @@ void swupd_curl_parallel_download_set_callbacks(void *handle, swupd_curl_success
 	h->free_cb = free_cb;
 }
 
+void swupd_curl_parallel_download_set_progress_callbacks(void *handle, swupd_curl_progress_cb progress_cb, struct download_progress *progress)
```
otaviobp: s/struct download_progress *progress/void *data/

src/curl_async.c
```diff
@@ -86,13 +86,15 @@ struct swupd_curl_parallel_handle {
...
-	swupd_curl_free_cb free_cb;       /* Callback to free user data*/
+	CURLM *mcurl;			    /* Curl handle */
+	struct list *failed;		    /* List of failed downloads */
+	struct hashmap *curl_hashmap;       /* Hashmap mentioned above */
+	struct tp *thpool;		    /* Pointer to the threadpool */
+	swupd_curl_success_cb success_cb;   /* Callback to success function */
+	swupd_curl_error_cb error_cb;       /* Callback to error function */
+	swupd_curl_free_cb free_cb;	 /* Callback to free user data */
+	swupd_curl_progress_cb progress_cb; /* Callback to report download progress */
+	struct download_progress *download_progress;
```
otaviobp: You don't use this struct in curl_async. It's more generic to have a void *data here, so we can the user's of the API can define what they need to use.

src/swupd_progress.c
```diff
@@ -0,0 +1,50 @@
...
+
+#include "swupd_progress.h"
+#include "swupd.h"
+
+int swupd_progress_callback(void UNUSED_PARAM(*clientp), curl_off_t dltotal, curl_off_t dlnow, curl_off_t UNUSED_PARAM ultotal, curl_off_t UNUSED_PARAM ulnow)
+{
+	struct download_progress *download_progress;
+	double increment;
+
+	download_progress = clientp;
```
otaviobp: You are using an parameter set as UNUSED

src/fullfile.c
```diff
@@ -125,20 +122,88 @@ static bool download_successful(void *data)
...
+			download_mix_file(file);
+		} else {
+			download_file(download_handle, file);
+		}
+
+		/* fall back for progress reporting when the download size
+		* could not be determined */
+		if (download_progress.total_download_size == 0) {
+			complete++;
+			progress_report(complete, list_length);
```
otaviobp: can you mix the progress_report here with progress report from the swupd_progress_callback()?

test/functional/bundleadd/add-json.bats
```diff
@@ -7,7 +7,6 @@ load "../testlib"
 
 test_setup() {
 
-	skip "This test is unstable and should be improved. Skipping that for now"
```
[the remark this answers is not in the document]
castulo: In fact all the messages that are non deterministic have been stripped out of the tests. All the messages that are related to steps like "download_packs" or "download_fullfiles" are only looking for the message where the step has reached "100%" of completion, which should always be there. Steps that are still showing other percentages in the tests like "install_files" are deterministic because  they are based on the amount of files and not based on downloaded data so they always have the same percentages.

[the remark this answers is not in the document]
otaviobp: oh, I see that now. you are right. We are good to go

src/fullfile.c
castulo: The swupd_progress_callback() will only be used if the progress_cb was set in the swupd_curl_parallel_handle and if the total_download_size was set to something other than 0 in the data sent to the function. This function is called asynchronously by curl: "This function gets called by libcurl with a frequent interval. While data is being transferred it will be called very frequently, and during slow periods like when nothing is being transferred it can slow down to about one call per second.".

The progress_report here on fullfile.c is the fallback when swupd_progress_callback() is not being used, and in this case is used synchronously after queuing each file to be downloaded, so I don't think they can get mixed.

src/swupd_progress.c
castulo: ups, good catch. At first I wasn't using that parameter and I forgot to remove the UNUSED.

src/curl_async.c
```diff
@@ -80,21 +80,21 @@
...
-	swupd_curl_success_cb success_cb; /* Callback to success function */
-	swupd_curl_error_cb error_cb;     /* Callback to error function */
-	swupd_curl_free_cb free_cb;       /* Callback to free user data*/
+	CURLM *mcurl;			    /* Curl handle */
+	struct list *failed;		    /* List of failed downloads */
+	struct hashmap *curl_hashmap;       /* Hashmap mentioned above */
+	struct tp *thpool;		    /* Pointer to the threadpool */
+	swupd_curl_success_cb success_cb;   /* Callback to success function */
+	swupd_curl_error_cb error_cb;       /* Callback to error function */
+	swupd_curl_free_cb free_cb;	 /* Callback to free user data */
```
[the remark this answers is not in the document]
castulo: I think those alignments were set by "make compliant", if I make changes to them and run `make compliant` it changes them back 😕 .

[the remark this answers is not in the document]
otaviobp: :(

src/fullfile.c
otaviobp: ok, if the total_download_size is always zero when curl isn't reporting progress, it's fine.

src/curl_async.c
castulo: will do that

src/curl_async.c
castulo: I will make the change

src/swupd_curl.h
```diff
@@ -45,6 +46,11 @@ typedef bool (*swupd_curl_error_cb)(enum download_status status, void *data);
  */
 typedef void (*swupd_curl_free_cb)(void *data);
 
+/*
+ * Callback called periodically by curl to report on download progress
+ */
+typedef int (*swupd_curl_progress_cb)(void UNUSED_PARAM(*clientp), curl_off_t dltotal, curl_off_t dlnow, curl_off_t UNUSED_PARAM ultotal, curl_off_t UNUSED_PARAM ulnow);
```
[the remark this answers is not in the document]
castulo: > I did some tests and I found a bug on download fullfile. I tried this:
> 
> ```
> swupd verify --quick --fix -B os-core,vim --path /custom -F 27 --force -m 28100 -u https://cdn.download.clearlinux.org/update/
> ```
> And the verify command failed. It works fine with master.

Looks like for some reason it was failing when extracting some tars, there were some changes recently on tar extraction. I rebased the PR and it is working now. I cannot recreate the issue anymore.

castulo:
> I did some tests and I found a bug on download fullfile. I tried this:
> 
> ```
> swupd verify --quick --fix -B os-core,vim --path /custom -F 27 --force -m 28100 -u https://cdn.download.clearlinux.org/update/
> ```
> And the verify command failed. It works fine with master.

Looks like for some reason it was failing when extracting some tars, there were some changes recently on tar extraction. I rebased the PR and it is working now. I cannot recreate the issue anymore.

src/swupd_curl.h
```diff
@@ -45,6 +46,11 @@ typedef bool (*swupd_curl_error_cb)(enum download_status status, void *data);
  */
 typedef void (*swupd_curl_free_cb)(void *data);
 
+/*
+ * Callback called periodically by curl to report on download progress
+ */
+typedef int (*swupd_curl_progress_cb)(void *clientp, int64_t dltotal, int64_t dlnow, int64_t UNUSED_PARAM ultotal, int64_t UNUSED_PARAM ulnow);
```
otaviobp: Remove the UNUSED_PARAM from declarations

castulo: done

src/curl_async.c
```diff
@@ -86,13 +86,15 @@ struct swupd_curl_parallel_handle {
...
+	swupd_curl_progress_cb progress_cb; /* Callback to report download progress */
+	void *data;
```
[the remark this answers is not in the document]
castulo: I can do that, I thought the idea was to leave it generic that is why I had left it like that. I'll rename it.

castulo:
> Overall it looks great. I added some minor comments and I have another suggestion.
> When we finish the download we print a 100% progress, but we still need to extract the last tarball. So the progress get stuck in 100% for some time. And that's a problem hard to fix, but there's an easy workaround. Can we move progress until 99% and only print 100% at the end of the process?
Well the percentage we are showing is strictly related to download progress, so it doesn't really reflect any progress in tarball extraction, so moving the percentage based on that it would be misleading and that was the problem we wanted to fix in the first place. To avoid being misleading maybe we can just show a message like "Finishing package extraction..." after the 100% has been reached, so the user knows download is complete but tars are still being unpacked. The unpacking only take a few extra seconds anyway (at least in all the tests I ran) so users shouldn't really need detailed progress update on that task.

otaviobp:
Yes, you're right. let we keep the way it is for now

PR #885, opened 2019-04-08
bundle-add overwrites old content

### LINKED ISSUE #863: bundle-add fails to install bundle content that should overwrite custom content
**Describe the bug**

If I install custom content on Clear Linux OS to locations that swupd controls (e.g. `/usr/bin`), and I then `swupd bundle-add` a bundle that should overwrite that custom content, `swupd bundle-add` fails to overwrite that content with its own content.

**To Reproduce**

1. Run `swupd bundle-remove keepassxc`
2. Run `touch /usr/bin/keepassxc`
3. Run `swupd bundle-add keepassxc`
4. Observe that `/usr/bin/keepassxc` is still the empty file created in step 2.

**Expected behavior**

Expected is for `swupd bundle-add keepassxc` to install all content for the bundle, also overwriting any custom content that may exist at the install destinations of the content.

 - swupd info
```
Installed version: 28410
Version URL:       https://cdn.download.clearlinux.org/update/
Content URL:       https://cdn.download.clearlinux.org/update/
```

castulo:
When swupd is installing a bundle, it goes through the list of files to
install and removes those files that already exist in the target system
from the list without checking if the file needs to be updated or not.

This commit changes this process so the file is removed from the
list of files to install only if the file already exist in the target
system and the hash matches the hash in the manifest.

Closes #863

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -848,7 +848,7 @@ struct list *filter_out_existing_files(struct list *files)
 		file = list->data;
 
 		filename = mk_full_filename(path_prefix, file->filename);
-		if (verify_file_lazy(filename)) {
+		if (verify_file(file, filename)) {
 			ret = list_free_item(list, NULL);
 		}
 		free_string(&filename);
--- a/Makefile.am
+++ b/Makefile.am
@@ -170,6 +170,7 @@ BATS = \
 	test/functional/bundleadd/add-multiple.bats \
 	test/functional/bundleadd/add-no-disk-space.bats \
 	test/functional/bundleadd/add-no-signature.bats \
+	test/functional/bundleadd/add-overwrite-files.bats \
 	test/functional/bundleadd/add-skip-scripts.bats \
 	test/functional/bundleadd/add-uses-fullfile.bats \
 	test/functional/bundleadd/add-uses-zeropack.bats \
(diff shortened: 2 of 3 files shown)


src/manifest.c
```diff
@@ -848,7 +848,7 @@ struct list *filter_out_existing_files(struct list *files)
 		file = list->data;
 
 		filename = mk_full_filename(path_prefix, file->filename);
-		if (verify_file_lazy(filename)) {
+		if (verify_file(file, filename)) {
```
otaviobp: verify_file() is very slow and calling it here would slow down the bundle-add significantly.
And using verify_file_lazy() is terrible too, because it doesn't make sense. We need to check if the file that is being added isn't present in a current Manifest and not in the system.

castulo: I don't follow, the file will obviously be in a manifest, that is why it is being added to the system with bundle-add. Or do you mean in the manifest of all other bundles already installed in the system?

otaviobp: When we are installing files from a BundleA instead of checking if those files are present and correct in the system we need to check if they are present in any manifest of a bundle that is assumed as installed.

There's a big overlap between files in bundle Manifests. Checking all files that are installed by a bundle is to expensive.

castulo:
This PR is being superseded by #903, I'm closing this PR.

PR #886, opened 2019-04-09
Support "optional" bundles in manifests

### LINKED ISSUE #875: Support header called 'optional in manifests
Support header described in mixer issue: https://github.com/clearlinux/mixer-tools/issues/562

castulo:
This PR adds support in swupd for optional bundles in manifests. An optional bundle is similar to the required bundles (included bundles) but have a couple of differences.

- Optional bundles are installed by default when the main bundle that includes them is installed.
- Users have the option to skip the installation of `Optional` bundles by using the flag `--skip-optional` or `-o` with `bundle-add`.
- Optional bundles can be removed from the system even if the main bundle that installed them is still in the system.

Closes #875

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -170,6 +170,7 @@ BATS = \
 	test/functional/bundleadd/add-experimental.bats \
 	test/functional/bundleadd/add-fall-back-to-fullfile.bats \
 	test/functional/bundleadd/add-include.bats \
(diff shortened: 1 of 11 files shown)


otaviobp:
Overall this looks good, but let's wait on a server implementation to merge this.

castulo:
This PR is the mixer side implementation -> https://github.com/clearlinux/mixer-tools/pull/580

PR #893, opened 2019-04-12
archives: Don't assume tarball includes '/' for directories

otaviobp:
At first I assumed that all tarballs would have a trailing '/' for directories
and this isn't true. But we can't assume that we won't have a trailing '/'. So
archives_check_single_file_tarball() is now ignoring trailing '/' when comparing
files. That is the same behavior of the function that
archives_check_single_file_tarball() replaced.

### DIFF
--- a/src/lib/archives.c
+++ b/src/lib/archives.c
@@ -225,28 +225,48 @@ int archives_check_single_file_tarball(const char *tarfilename, const char *file
 {
 	struct archive *a;
 	struct archive_entry *entry;
-	int r;
-	bool found = false;
+	const char *file_entry;
+	size_t file_len;
+	int ret = 0;
 
-	/* set up read */
-	r = archive_from_filename(&a, tarfilename);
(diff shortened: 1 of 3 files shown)


src/lib/archives.c
```diff
@@ -225,28 +225,47 @@ int archives_check_single_file_tarball(const char *tarfilename, const char *file
...
+	if (ret < 0) {
+		return ret;
+	}
 
-	/* set up read */
-	r = archive_from_filename(&a, tarfilename);
-	if (r < 0) {
-		return r;
+	ret = -ENOENT;
+	file_len = strlen(file);
```
phmccarty: The `file_len` value here appears to be unused, since it is soon overwritten on line 249.

otaviobp: It's a left over. Thanks

otaviobp: And I had the same problem with ret. I fixed that now.

PR #899, opened 2019-04-18
Signature

otaviobp:


test/unit/config_noocsp_critical.cnf
```diff
@@ -0,0 +1,21 @@
+[ ca ]
```
castulo: I wonder if all these resources the unit tests use (like manifests and these cnf config files) should be in a separate directory like `test/unit/data` or something like that, so they don't get mixed with the unit tests so it looks cleaner?

src/lib/sys.c
```diff
@@ -277,3 +277,42 @@ bool systemd_in_container(void)
...
+	char *tmp, *dir;
+
+	tmp = strdup_or_die(path);
+	dir = strdup_or_die(dirname(tmp));
+
+	free(tmp);
+	return dir;
+}
+
+char *sys_path_join(const char *prefix, const char *path)
```
castulo: Nice

test/unit/config_noocsp_critical.cnf
otaviobp: Good point. fixed it/.

PR #901, opened 2019-04-22
Global: More generic swupd_init()

otaviobp:


src/helpers.c
```diff
@@ -528,54 +528,61 @@ void swupd_deinit(void)
 *	- Initialize curl
 *	- Initialize signature checking
 */
-enum swupd_code swupd_init(void)
+enum swupd_code swupd_init(enum swupd_init_config config)
 {
 	int ret = SWUPD_OK;
```
castulo: you changed `int ret` variables in other functions to be `enum swupd_code`, but you left this one. Did you miss it?

src/mirror.c
```diff
@@ -259,6 +249,7 @@ void handle_mirror_if_stale(void)
 enum swupd_code mirror_main(int argc, char **argv)
 {
 	int ret = SWUPD_OK;
```
castulo: should `int ret` here be changed to `enum swupd_code ret`?

src/mirror.c
```diff
@@ -114,13 +114,6 @@ static int unset_mirror_url()
 		goto out;
 	}
 
-	/* we need to set the version and content URLs again if not they will
-	 * remain with the value from the mirror */
-	free_string(&version_url);
-	free_string(&content_url);
-	set_version_url(NULL);
```
castulo: why are these not necessary anymore?

test/functional/checkupdate/chk-update-json.bats
```diff
@@ -25,6 +25,7 @@ test_setup() {
 		[
 		{ "type" : "start", "section" : "check-update" },
 		{ "type" : "info", "msg" : "Current OS version: 10 " },
+		{ "type" : "info", "msg" : "Latest server version: 20 " },
```
castulo: Looks like the update on theses tests are in the wrong commit.

src/helpers.c
otaviobp: yes. fixing it.

src/mirror.c
otaviobp: This function mixes int with enum swupd_code ret. It's not ideal, but as I didn't fix that, I let the type of ret to int.

src/mirror.c
otaviobp: On mirror --unset those variables weren't set because swupd isn't initialized yet. So we don't need/should unset them.
The only place we really need to have this is in handle_mirror_if_stale() function. I moved this to that function then.

otaviobp:
fixed reported problems

src/mirror.c
castulo: ok

src/mirror.c
castulo: ahh make sense

PR #903, opened 2019-04-23
Bundleadd improvements

### LINKED ISSUE #863: bundle-add fails to install bundle content that should overwrite custom content

castulo:
When staging a file using do_staging(), if the path of the file is  missing verify_fix_path() is called to try fix the missing path. verify_fix_path() then removes bad directories (if any), downloads the file and again and stages the new directory using do_staging(), doing a circular reference.
On top of this bundle add is calling verify_fix_path() after running do_staging() again.
All this circular reference makes the code very difficult to understand and may cause swupd to do extra work that is not necessary. This commit simplifies the code by not allowing do_staging() to call verify_fix_path() so we can manually call it later if necessary.

This PR also includes a fix for the following bug:
Closes #863

The PR also adds more comments in the bundle-add code to try to make it easier to follow.


src/bundle.c
```diff
@@ -906,14 +907,15 @@ static enum swupd_code install_bundles(struct list *bundles, struct list **subs,
...
+			/* the file does not exist in the staged directory, it will need
+			 * to be downloaded again */
 			free_string(&hashpath);
 			continue;
 		}
 
+		/* make sure the file is not corrupt */
 		ret = verify_file(file, hashpath);
-		if (!ret) {
+		if (ret == false) {
```
otaviobp: don't mix int and bool. It's very confusing. Use 

    if (!verify_file(...))

or create another bool variable for that.

src/staging.c
```diff
@@ -184,6 +194,17 @@ enum swupd_code do_staging(struct file *file, struct manifest *MoM)
 			ret = SWUPD_COULDNT_RENAME_DIR;
 			goto out;
 		}
+
+		/* add the directory name to the staging member to indicate the
+		 * staging was successful */
+		free_string(&file->staging);
+		string_or_die(&file->staging, "%s%s/%s", path_prefix, rel_dir, base);
```
otaviobp: I didn't like this approach. We use the staging variable to keep the name of a temporary location of the file and this one is already in it's final destination. The problem you reported isn't a problem on update. It looks like we should do a clean up on bundle add to avoid doing this here. Take a look at the update_loop() on update.c.

src/bundle.c
```diff
@@ -989,34 +1010,21 @@ static enum swupd_code install_bundles(struct list *bundles, struct list **subs,
...
-				mom->files = consolidate_files(mom->files);
-				timelist_timer_stop(global_times); // closing: Add tracked bundles
-				full_mom = mom;
-			}
-			file = search_file_in_manifest(full_mom, file->filename);
+			/* the current file struct doesn't have the name of the "staging" file
+			 * since it was staged by verify_fix_path, the staged file is in the
+			 * file struct in the MoM, so we need to load that one instead
+			 * so rename_staged_file_to_final works properly */
+			file = search_file_in_manifest(mom, file->filename);
```
otaviobp: I have a great feeling that this is slowing down bundle_add a lot. Any idea why we need that on bundle_add and not on update?
Can't we use rename_all_files_to_final() instead of this loop?

src/bundle.c
castulo: we were already mixing int and bool cause verify_file returns a bool. But I changed this to make it clearer so will change this back if it makes it more confusing.

src/staging.c
castulo: ok

src/bundle.c
castulo: to be honest I'm not 100% sure why we need to load the file there again.... I guess the it is just so we have the correct metadata for that file since it should have been fixed by verify_fix_path

castulo: I don't think this is that problematic because this code should only run for every file that was fixed by verify_fix_path (if any), so in theory it should not run at all or it should only run a few times.

PR #910, opened 2019-05-01
Set user-agent header

alexjch:
This change sets swupd client User-Agent header to PACKAGE/VERSION. This is
 a standard HTTP header to identify the client that originates a request.

Signed-off-by: Alex Jaramillo <alex.v.jaramillo@intel.com>

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -725,6 +725,9 @@ CURLcode swupd_curl_set_basic_options(CURL *curl, const char *url, bool fail_on_
 		goto exit;
 	}
 
+	curl_easy_setopt(curl, CURLOPT_USERAGENT, PACKAGE "/" VERSION);
+	// No error checking needed, this is not critical information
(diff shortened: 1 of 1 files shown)


src/curl.c
```diff
@@ -725,6 +725,9 @@ CURLcode swupd_curl_set_basic_options(CURL *curl, const char *url, bool fail_on_
 		goto exit;
 	}
 
+        curl_easy_setopt(curl, CURLOPT_USERAGENT, PACKAGE "/" VERSION);
```
castulo: Make sure you run `make compliant` on your patch to make sure you are compliant with the code style being used. Looks like you are using spaces here over tab.

alexjch: I did not realize TAB was still a thing =D. Thanks for educating me though. I'm force pushing a fix to my code style error.

PR #912, opened 2019-05-06
Making verify --install a top-level command

### LINKED ISSUE #250: verify --install should be made a top-level command
the `--install` flag to the `verify` top-level command should be made a top-level command, and decoupled from the `verify` codepath for a couple reasons:
* `--install` is designed to do something completely different (install an OS to a blank partition/directory)
* `--install` *actually* does something completely different in that it doesn't verify the existing files in the target directory, it just assumes it doesn't need to do anything for them. Perhaps it should be verifying those files when it exists, but it currently does not.

When  moved to a top-level command it should probably be named something different to prevent a new user from running
```
swupd install <bundlename>
```
expecting to install a new bundle (like `apt-get install pkgname`) but accidentally installing a whole OS to the current directory (obviously checks would be in place, however)

### LINKED ISSUE #911: Rename "swupd verify --install" to "swupd os-install"
The purpose of `swupd verify --install` is so it can be used to install a new Clear OS in a target system. However since currently it is part of the `swupd verify` command it makes it confusing  as of to what its purpose is.

We should move the `swupd verify --install` option to be its own command `swupd os-install` so its purpose is more straightforward. The new command will still use the verify code to perform the task.

Note: the name `swupd os-install` was chosen over `swupd install` to avoid misleading users to believe they can use `swupd install` to install bundles. 

Tasks:
- [x] Add `swupd os-install` to main menu.
- [x] Add new os-install.c file with the following functions:
    - The command main function
    - Function(s) to parse the command
    - Validation of command options
- [x] Deprecate the `--install` option from verify.c
- [x] Add to the auto-complete script
- [x] Update functional tests
- [x] Update Travis with the new tests
- [x] Update documentation

**Important: This change will require an update in the clr-installer side which uses `swupd verify --installl` to install the OS.**

castulo:
The purpose of swupd verify --install is so it can be used to install
a new Clear OS in a target system. However since currently it is part
of the swupd verify command it makes it confusing as of to what its
purpose is.

This commit moves the swupd verify --install option to be a top level
command (swupd os-install) so its purpose is more straightforward. The
new command still uses the verify code to perform the task underneath.

Closes #250 
Closes #911 


configure.ac
```diff
@@ -1,8 +1,8 @@
 #                                               -*- Autoconf -*-
 # Process this file with autoconf to produce a configure script.
 
-AC_PREREQ([2.66])
-AC_INIT(swupd-client, 3.19.0, otavio.pontes@intel.com)
+AC_PREREQ([2.69])
```
otaviobp: why did you change that? (I'm more curious than concerned)

docs/swupd.1.rst
```diff
@@ -352,12 +376,6 @@ SUBCOMMANDS
 
             Matches nothing, because paths are never empty.
 
-    - `-i, --install`
```
otaviobp: As we discussed, let's keep this here, and add a comment that this is deprecated and os-install should be used.

configure.ac
castulo: I don't think I did.... maybe a rebase did it?

docs/swupd.1.rst
castulo: sure thing

castulo:
@otaviobp I addressed all the points in the feedback.

PR #916, opened 2019-05-09
Making verify --fix a top-level command

### LINKED ISSUE #914: Rename "swupd verify --fix" to a top level command "swupd repair"
In an attempt to make the swupd commands easier to understand, it was decided to move some of the functionality `swupd verify` has through its many flags, to be top level commands where make sense.

`swupd verify --fix` should be made a top level command `swupd repair`

Tasks:
- [x] Add `swupd repair` to main menu.
- [x] Add new repair.c file with the following functions:
  - The command main function
  - Function(s) to parse the command
  - Validation of command options
- [x] Deprecate the `--fix` option from verify.c
- [x] Add to the auto-complete script
- [x] Update functional tests
- [x] Update Travis with the new tests
- [x] Update documentation

castulo:
The purpose of "swupd verify --fix" is to repair local issues relative
to a server manifest. However the verify command currently has many
flags, and some flags are mutually exclusive, this makes the command
very confusing to use for users.

This commit makes "swupd verify --fix" the top level command
"swupd repair". This new command still uses the verify code under the
covers.

Closes #914


castulo:
**NOTE:** This PR is built on top of PR #912, so PR #912 has to be reviewed and merged before this one. Once that has happened this PR can be rebased and reviewed.
🛑

mdhorn:
@castulo @otaviobp 
The installer relies on using `swupd verify --install`, so that command is going away in a future swupd release? We will have to coordinate this change or it will break the DevOps and anyone installing.

gvancuts:
I would also recommend that a more formal and descriptive announcement about this change be sent to the relevant mailing list prior to landing in Clear Linux. This is a command that's fairly widely used by people and referenced in various wikis, it will catch many by surprise so a heads-up would be useful.

docs/swupd.1.rst
```diff
@@ -222,6 +222,93 @@ SUBCOMMANDS
...
+        - ``--bundles xterm,vi``
+
+            Installs bundles os-core and vi, along with os-core (installed by default).
+
+``repair``
+
+    Correct any issues found. This will overwrite incorrect file content,
+    add missing files and do additional corrections, permissions, etc.
+
+    - `-m, --manifest`
```
gvancuts: This sounds like the same meaning than `-V, --version` as used by `os-install`. If true, it would be better (i.e. less error-prone and confusing) to have both `os-install` and `repair` use the same.

otaviobp: yes. you're write. we should have used -m in os-install

otaviobp:
For now we are just introducing the replacements and tuning the parameters. We aren't going to remove verify install in next swupd release and verify will be around for some time. Before removing it completely we will for sure announce it in advance to make sure people won't have scripts broken.

castulo:
> @castulo @otaviobp
> The installer relies on using `swupd verify --install`, so that command is going away in a future swupd release? We will have to coordinate this change or it will break the DevOps and anyone installing.

@mdhorn yes, actually this PR from installer already includes that change https://github.com/clearlinux/clr-installer/pull/372. In any case the old command/options will continue to work for some time with a proper warning so other users have time to update their scripts.

mdhorn:
@castulo @otaviobp Great that backward compatibility will still be there! :)
Some older tools like 'ister.py' which are still in use also use the `verify --install`; so something to capture when you get to the phase of removing that command.

docs/swupd.1.rst
castulo: @otaviobp I changed `-m` in `os-install` to be `-V`/`--version` because IMO it makes more sense. When you are installing an OS you are really thinking about choosing a version of it, not really a manifest of it. Maybe it made more sense for the `verify` command when you were verifying against one specific manifest. I think referencing "manifests" in those commands may cause confusion to users not familiar with swupd & mixer and therefore manifests. Maybe it would be better to also change the -m/--manifest option from the `repair` and `diagnose` commands to also be `-V`.`--version`. This will be a somewhat more disruptive change for users that are used to use `-m`, but since we are doing a disruptive change here I guess now is a good time to do it.

PR #919, opened 2019-05-10
Renaming "swupd verify" to "swupd diagnose"

### LINKED ISSUE #918: Rename "swupd verify" to "swupd diagnose"
In an attempt to make the swupd commands easier to understand, it was decided to move some of the functionality swupd verify has through its many flags, to be top level commands where make sense. Also the `swupd verify` command should  be renamed to `swupd diagnose` since it is more straightforward.

Tasks:
- [x] Add `swupd diagnose` to main menu.
- [x] Rename verify references in verify.c
- [x] Deprecate the `verify` option from verify.c
- [x] Update the auto-complete script
- [x] Update functional tests
- [x] Update Travis with the updated tests
- [x] Update documentation

castulo:
This PR renames the "verify" command to "diagnose" to make it
more straightforward for the user to understand the purpose of the
command.

Closes #918

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
                 - UPDATE_SUBGROUP2="$(find test/functional/update -name *.bats | tail -n $UPDATE_SUBGROUP2_TOTAL | tr '\n' ' ')"
                 - GROUP1="$UPDATE_SUBGROUP1"
                 - GROUP2="$UPDATE_SUBGROUP2 $(find test/functional/{checkupdate,hashdump,mirror,usability} -name *.bats -printf '%p ')"
-                - GROUP3="$(find test/functional/{verify,search,os-install,repair} -name *.bats -printf '%p ')"
+                - GROUP3="$(find test/functional/{diagnose,search,os-install,repair} -name *.bats -printf '%p ')"
                 - GROUP4="$(find test/functional/{bundleadd,bundleremove,bundlelist} -name *.bats -printf '%p ')"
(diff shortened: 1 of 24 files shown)


castulo:
🛑  **NOTE:** This PR is built on top of PRs #912 and #916, so those PRs have to be reviewed and merged before this one (in that order). Once that has happened this PR can be rebased and reviewed.

castulo:
> lgtm. just restore old dos with a warning for deprecation. other than that, it's good to go

I will @otaviobp. One more thing I'm thinking I may need to update (not sure) is other testing besides from the functional tests, like the "real content" tests.

otaviobp:
> I will @otaviobp. One more thing I'm thinking I may need to update (not sure) is other testing besides from the functional tests, like the "real content" tests.

yes we will need to update them.

busykai:
+1 on rearranging commands. However, let's not call `verify --install` `os-install`. It can be used to install bundles and it seems to me just `install` will reflect it's purpose well.

castulo:
> +1 on rearranging commands. However, let's not call `verify --install` `os-install`. It can be used to install bundles and it seems to me just `install` will reflect it's purpose well.
> 
> EDIT: oops, it's seems i'm a little late to the party. this comment is related to #912 rather.

There were some arguments made in the original issue #250 against using `swupd install`, that is why we went with `os-install`. We don't want to mislead users to use `install` to install bundles but `bundle-add` instead.

PR #928, opened 2019-05-21
global: Fallback to system configuration when configuration on path_p…

### LINKED ISSUE #924: swupd verify --install ignores OS defaults
**Describe the bug**
Using swupd to install a new OS from mix fails.

**To Reproduce**
- Build a mix
- Create an image from that mix
- Boot the new image
- Use swupd from booted image to install OS with `swupd verify --install`
- swupd will fail to verify the signature and will be using that last version from clearlinux.org instead of the mix

**Expected behavior**
swupd will use the content and version URL from the /usr/share/defaults/swupd

```
root@clr-live ~ # swupd info
Installed version: 10
Version URL:       http:/myhost.com/mixer/
Content URL:       http://myhost.com/mixer/

root@clr-live ~ # swupd mirror
Installed version: 10
Version URL:       http://myhost.com/mixer/
Content URL:       http://myhost.com/mixer/

root@clr-live ~ # swupd verify --path=/mnt --statedir=/mnt/var/lib/swupd --install -m latest --force --no-boot-update -B os-core-update,openssh-server,telemetrics,tzdata,kernel-native
/mnt//usr/share/clear/versionstamp does not exist!
Verifying version 29480
Certificate verification error: self signed certificate
Certificate verification error: self signed certificate


root@clr-live ~ # swupd -v
swupd-client 3.19.0
   Copyright (C) 2012-2018 Intel Corporation

Compile-time options: +BZIP2 +SIGVERIFY -COVERAGE -BSDTAR -XATTRS -TAR_SELINUX +STATELESS
Compile-time configuration:
mount point                  /
state directory              /var/lib/swupd
bundles directory            /usr/share/clear/bundles
certificate path             /usr/share/clear/update-ca/Swupd_Root.pem
fallback certificate path    /usr/share/ca-certs/.prebuilt-store/anchors
systemd unitdir              /usr/lib/systemd/system
content URL                  https://cdn.download.clearlinux.org/update/
version URL                  https://cdn.download.clearlinux.org/update/
format ID                    27
pre-update hook
post-update hook             /usr/bin/update-helper
```

Note that even though the defaults are set correctly for the mix, and one can do a `swupd bundle-add` or `swupd verify --fix` and the correct URL is used, when doing a `swupd verify --install`, swupd falls back to its compile time defaults.

otaviobp:
…refix

If --path is used swupd looks for version and content url inside the informed chroot path.
This is the expected, but in some cases not enough. When installing a CL we usually point to
an empty chroot. So if there's no config file inside the chroot, fallbacks to system configuration.

Fixes #924

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -177,49 +177,47 @@ static int set_default_value_from_path(char **global, const char *path)
 	return ret;
 }
 
-static int set_url(char **global, char *url, const char *path)
-{
-	int ret = 0;
-
-	if (url) {
-		free_string(global);
-		string_or_die(global, "%s", url);
-	} else {
-		if (*global) {
-			/* option passed on command line previously */
-			return ret;
-		} else {
-			/* no option passed; use the default value */
-			ret = set_default_value_from_path(global, path);
-			return ret;
-		}
-	}
-
-	return ret;
-}
-
 /* Initializes the content_url global variable. If the url parameter is not
  * NULL, content_url will be set to its value. Otherwise, the value is read
  * from the 'contenturl' configuration file.
  */
 int set_content_url(char *url)
 {
+	int ret = 0;
+
 	if (content_url) {
 		/* Only set once; we assume the first successful set is the best choice */
 		return 0;
 	}
 
-	/* check for mirror configuration first */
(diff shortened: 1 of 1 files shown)


clefrks:
@otaviobp when would this be merged?

PR #933, opened 2019-05-24
Fixing false positive in a functional test

castulo:
The update/update-json.bats functional test is giving a false positive
since one hardcoded value (time) was taking longer to run in a local
environment.
This commit fixes the issue by using a regex to accept any time value.


castulo:
> Cool. This one was breaking in my system sometimes too. I looks fixed now. I'll ping you if it breaks again

I saw there is one more test breaking, but it is a unit test `test/unit/test_signature.c`

PR #943, opened 2019-06-03
Fix download progress for fullfiles

### LINKED ISSUE #939: swupd shows incorrect progress report when doing repair
When running "swupd repair" to fix mismatching files in the system, at some point swupd needs to download fullfiles of the files that need to be repaired, when this is happening, swupd shows the progress of the download to the user. This progress is being reported incorrectly, it goes above 100%.
 
It looks like this:
```
Diagnosing version 29690
Verifying files
	...100%
	...101850%
Adding any missing files
	...100%
Repairing modified files
```

This bug can be reproduced by corrupting a system file then running `swupd clean --all` followed by `swupd repair`.

**Expected behavior**
When downloading the fullfiles swupd should show the progress going from 0% to 100%

 - Clear Linux OS Version: 29690
 - swupd version: 3.20.0

castulo:
When the download progress of fullfiles was being calculated it was
being done considering as if the files were being downloaded serially
not in parallel. This was causing an error in the calculation of the
download progress.

This commit fixes the issue by considering the files are being downloaded
in parallel when calculating the download progress.

Closes #939

### DIFF
--- a/src/curl_async.c
+++ b/src/curl_async.c
@@ -112,6 +112,7 @@ struct multi_curl_file {
 
 	void *data;     /* user's data */
 	bool cb_retval; /* return value from callback */
+	struct file_progress *progress;
 };
 
 /*
@@ -157,6 +158,8 @@ static void free_curl_file(struct swupd_curl_parallel_handle *h, struct multi_cu
 		curl_easy_cleanup(curl);
 		file->curl = NULL;
 	}
+
+	free(file->progress);
 	free(file);
 }
 
@@ -435,7 +438,7 @@ static int process_download(struct swupd_curl_parallel_handle *h, struct multi_c
 	}
 	file->curl = curl;
 
-	if (h->progress_cb && h->data) {
+	if (h->progress_cb && file->progress) {
 
 		curl_ret = curl_easy_setopt(curl, CURLOPT_XFERINFOFUNCTION, h->progress_cb);
 		if (curl_ret != CURLE_OK) {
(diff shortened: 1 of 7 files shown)


src/curl_async.c
```diff
@@ -157,6 +158,11 @@ static void free_curl_file(struct swupd_curl_parallel_handle *h, struct multi_cu
 		curl_easy_cleanup(curl);
 		file->curl = NULL;
 	}
+
+	if (file->progress != NULL) {
```
otaviobp: you don't need to check for null before free. if pointer is null free() doesn't fail.

src/fullfile.c
```diff
@@ -153,6 +153,25 @@ int download_fullfiles(struct list *files, int *num_downloads)
 		free_string(&targetfile);
 	}
 
+	/* remove tar duplicates from the list */
+	need_download = list_sort(need_download, file_sort_hash);
```
otaviobp: Sorting a list is very expensive. We prevent a download to be performed twice by using the hashmap in curl_async. 
Do we need to filter files in this stage?

src/manifest.c
```diff
@@ -77,6 +77,27 @@ int file_sort_filename_reverse(const void *a, const void *b)
...
+	struct file *A, *B;
+	int ret;
+	A = (struct file *)a;
+	B = (struct file *)b;
+
+	ret = strcmp(A->hash, B->hash);
+	if (ret) {
+		return ret;
+	}
+	if (A->is_deleted > B->is_deleted) {
```
otaviobp: we just add files to the list if if_deleted == 0. Why do we need this here?

otaviobp:
lgtm

src/curl_async.c
castulo: nice! thanks for the tip, I'll remove the if

src/fullfile.c
castulo: yes, otherwise those files that are not actually going to be downloaded because use the same tar are going to be counted when we determine the total download size, then when we are reporting download progress we are never going to reach 100%

src/manifest.c
castulo: we don't, I copied this function from `file_sort_filename` and forgot to remove that part

castulo:
@otaviobp I addressed your comments, let me know if there is something else missing

PR #947, opened 2019-06-06
Add `zsh` completion

lebensterben:
I wrote a `zsh` completion function for `clearlinux/swupd-client`.

- It shows both available arguments/options and their description
- It provides URL and PATH completions for certain options
- It provides BUNDLE completion for `swupd diagnose -b|--bundles`,
  `swupd repair -b|--bundles` and `swupd bundle-remove`. The list
  of available bundle is dynamic, that is to say if you've already
  specified a bundle, it won't appear in the list.
- Those labelled as `deprecated` are not presented in the completion.
  (But you may add them.)

I also suggest the following:

- `swupd.zsh` shall be placed under `/usr/share/zsh/site-functions` and
  named as `_swupd` according to `zsh` convention, if `zsh` is installed.
  It will be loaded by `zsh` because this directory is in `zsh`'s `$fpath`.
- `swupd.bash` currently is located in `/usr/share/defaults/etc/profile.d`,
  but we put all other `bash` completions in
  `/usr/share/bash-completion/completions` directory. And I suggest to move
  `swupd.bash` here, which would be loaded by `bash` without needing to

### DIFF
--- a/swupd.zsh
+++ b/swupd.zsh
@@ -31,7 +31,7 @@ local installed # cache for installed bundles
     elif [ -r /var/lib/swupd/version ] && [ -r /var/lib/swupd/"$(cat /var/lib/swupd/version)"/Manifest.MoM ]; then
       MoM=/var/lib/swupd/"$(cat /var/lib/swupd/version)"/Manifest.MoM
     fi
-    if [ -n $"MoM" ]; then
+    if [ -n "$MoM" ]; then
       bundles="$(sed '1,/^$/d; s/^.*\t/ /; /.*\.I\..*/d' "$MoM" | LC_ALL=C sort )"
       bundles='('"$(sed 's/\n/ /g' <<< $bundles)"')'
     fi
--- a/swupd.zsh
+++ b/swupd.zsh
@@ -1,0 +1,281 @@
+#compdef swupd
+# -----------------------------------------------------------------------
+#   Software Updater - autocompletion script
+#
+#   Author: Lucius Hu - http://github.com/lebensterben
+#
+#   This program is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation, version 2 or later of the License.
+#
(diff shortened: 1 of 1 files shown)


miguelinux:
LGTM

@castulo @otaviobp any feedback?

swupd.zsh
```diff
@@ -0,0 +1,281 @@
...
+local installed # cache for installed bundles
+
+(( $+functions[_swupd_bundle_add] )) ||
+  _swupd_bundle_add () {
+    local bundles
+    local MoM
+    if [ -r /var/tmp/swupd/Manifest.MoM ]; then
+      MoM=/var/tmp/swupd/Manifest.MoM
+    elif [ -r /var/lib/swupd/version ] && [ -r /var/lib/swupd/"$(cat /var/lib/swupd/version)"/Manifest.MoM ]; then
+      MoM=/var/lib/swupd/"$(cat /var/lib/swupd/version)"/Manifest.MoM
```
otaviobp: This file is only accessible as root. I didn't test this, isn't this going to print errors when running as a regular user?

lebensterben: The `[ -r foo ]` test reruns true if the file is readable by current user, from which a list of bundles are added as candidates for completion.

The bottom line is, even if none of those are readble, it shall not report error while users will just not seeing any completion candidates here.

otaviobp: You're right about that. But there's another problem.

It looks like [ -n $"MoM" ] is true when running as a regular user in the case /var/tmp/swupd doesn't exist. And I'm seeing the error: "sed: can't read : No such file or directory"

lebensterben: I fixed the typo. It should be `[ -n "$MoM" ]`

otaviobp: Nice. Works for me now

lebensterben: If there is any change to `swupd` command, please let me know and I will update this `zsh` completion.

PR #948, opened 2019-06-07
Checking mirror status only if it is set

### LINKED ISSUE #922: swupd update does not get content from the desired mirror
**Describe the bug**
swupd update does not get content from the desired mirror but from the default mirror with timeout. 

swupd info
Installed version: 28910
Version URL:       http://10.xxx.xx.xx/pub/mirrors/clearlinux/update
Content URL:       http://10.xxx.xx.xx/pub/mirrors/clearlinux/update

root@clr-node ~ # swupd update
Update started.
Error: Curl - Communicating with server timed out - 'https://cdn.download.clearlinux.org/update//version/format27/latest'
Waiting 10 seconds before retrying the download
Retry #1 downloading from https://cdn.download.clearlinux.org/update//version/format27/latest

 - Clear Linux OS Version: 28910
 - Platform: Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz

castulo:
When doing an update, we check to see if a mirror is stale, this only
needs to be done if there is a mirror set.

This commit checks to see if a mirror is set and only then it checks to
see if it is stale. This gives us the advantage of adding more detailed
messages when checking for the mirror status without polluting the output
with those extra messages when a mirror is not set.

Closes #922

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -123,13 +123,18 @@ exit:
 	return curl_ret;
 }
 
-static int check_connection(const char *test_capath)
+int check_connection(const char *test_capath, char *url)
 {
 	CURLcode curl_ret;
 	long response = 0;
 
-	debug("Curl - check_connection url: %s\n", version_url);
-	curl_ret = swupd_curl_set_basic_options(curl, version_url, false);
+	if (!url) {
+		debug("Please provide the url to use to test the connection\n");
+		return -1;
+	}
+
+	debug("Curl - check_connection url: %s\n", url);
+	curl_ret = swupd_curl_set_basic_options(curl, url, false);
 	if (curl_ret != CURLE_OK) {
 		return -1;
 	}
@@ -150,6 +155,9 @@ static int check_connection(const char *test_capath)
 
 	switch (curl_ret) {
 	case CURLE_OK:
+		/* Note: when using http and the user is behind a proxy, we may get
+		 * a valid response here from the proxy even if the http server is
+		 * down, allow it to continue for now, it will fail further in the process */
 		return 0;
 	case CURLE_SSL_CACERT:
 		debug("Curl - Unable to verify server SSL certificate\n");
@@ -173,7 +181,7 @@ static int check_connection(const char *test_capath)
 	}
 }
 
-int swupd_curl_init(void)
+int swupd_curl_init(char *url)
 {
 	CURLcode curl_ret;
 	char *str;
@@ -203,15 +211,17 @@ int swupd_curl_init(void)
 	/* enforce the use of https or file */
 	if (!is_url_allowed(version_url) ||
 	    (strcmp(content_url, version_url) != 0 && !is_url_allowed(content_url))) {
+		swupd_curl_deinit();
 		return -1;
 	}
 
-	ret = check_connection(NULL);
+	ret = check_connection(NULL, url);
 	if (ret == 0) {
 		return 0;
 	} else if (ret == -CURLE_OPERATION_TIMEDOUT) {
 		error("Curl - Communicating with server timed out\n");
-		return -1;
+		ret = -1;
+		goto exit;
 	}
 
 	if (FALLBACK_CAPATHS[0]) {
@@ -225,7 +235,7 @@ int swupd_curl_init(void)
 			}
 
 			debug("Curl - Trying fallback CA path %s\n", tok);
-			ret = check_connection(tok);
+			ret = check_connection(tok, url);
 			if (ret == 0) {
 				capath = strdup_or_die(tok);
 				break;
@@ -234,13 +244,16 @@ int swupd_curl_init(void)
 		free_string(&str);
 	}
 
+exit:
 	if (ret != 0) {
+		/* curl failed to initialize */
 		error("Failed to connect to update server: %s\n", version_url);
 		info("Possible solutions for this problem are:\n"
 		     "\tCheck if your network connection is working\n"
 		     "\tFix the system clock\n"
 		     "\tRun 'swupd info' to check if the urls are correct\n"
 		     "\tCheck if the server SSL certificate is trusted by your system ('clrtrust generate' may help)\n");
+		swupd_curl_deinit();
 	}
 
 	return ret;
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -518,6 +518,16 @@ void swupd_deinit(void)
 	dump_file_descriptor_leaks();
 }
 
+void remove_trailing_slash(char *url)
+{
+	int len = strlen(url);
+
+	while (len > 0 && url[len - 1] == '/') {
+		len--;
+		url[len] = 0;
+	}
+}
+
 /* this function is intended to encapsulate the basic swupd
 * initializations for the majority of commands, that is:
 * 	- Make sure root is the user running the code
@@ -579,7 +589,7 @@ enum swupd_code swupd_init(enum swupd_init_config config)
 	}
 
 	if ((config & SWUPD_NO_NETWORK) == 0) {
-		if (swupd_curl_init() != 0) {
+		if (swupd_curl_init(version_url) != 0) {
 			ret = SWUPD_CURL_INIT_FAILED;
 			goto out_close_lock;
 		}
--- a/src/mirror.c
+++ b/src/mirror.c
@@ -114,6 +114,13 @@ static int unset_mirror_url()
 		goto out;
 	}
 
+	/* we need to also unset the mirror urls from the cache and set it to
+	 * the central version */
+	free_string(&version_url);
+	free_string(&content_url);
+	set_default_version_url();
+	set_default_content_url();
+	get_latest_version("");
 out:
 	free_string(&content_path);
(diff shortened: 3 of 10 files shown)


src/mirror.c
```diff
@@ -200,11 +200,34 @@ static enum swupd_code set_mirror_url(char *url)
...
 }
 
+static bool mirror_is_set(void)
+{
+	bool mirror_set;
+	char *version_path;
+	char *content_path;
+
+	version_path = mk_full_filename(path_prefix, MIRROR_VERSION_URL_PATH);
+	content_path = mk_full_filename(path_prefix, MIRROR_CONTENT_URL_PATH);
```
otaviobp: Content url isn't used by handle_mirror_if_stale, so it doesn't make sense to test mirror if version_path isn't set.

src/mirror.c
```diff
@@ -221,10 +244,14 @@ void handle_mirror_if_stale(void)
 	 * will be used due to the diff being negative. */
 	int diff = central_version - mirror_version;
 	if (diff > MIRROR_STALE_UNSET || mirror_version == -1) {
```
otaviobp: error can be any number < 0 and not just -1

src/mirror.c
```diff
@@ -221,10 +244,14 @@ void handle_mirror_if_stale(void)
 	 * will be used due to the diff being negative. */
 	int diff = central_version - mirror_version;
```
otaviobp: counting in this diff to decide if we get in or not is very fragile.
If we have errors on central_version and mirror_versions we're just getting into the if if mirror_version error is -1. But in this case I don't think it makes sense to unset the mirror, because we can have a connection problems.

So if the central_version is negative we should just ignore.
If the mirror_version has error or it's outdated and central_version works, we should unset the mirror.

src/mirror.c
castulo: `handle_mirror_if_stale` only uses the version_url to see if the mirror is stale, but when a mirror is set, both version_url and content_url are set, so to actually claim that a valid mirror is set both of them would need to be set, wouldn't them? That is why I included both in the `mirror_is_set` function. Do you really think I should get rid of the content_url check in this function?

src/mirror.c
castulo: ok I'll refactor that

castulo: done

castulo:
Hey @otaviobp I think I addressed all your concerns with this update in the PR

src/curl.c
```diff
@@ -150,6 +151,16 @@ static int check_connection(const char *test_capath)
 
 	switch (curl_ret) {
 	case CURLE_OK:
+		if (curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &response) != CURLE_OK) {
+			return -1;
+		}
+		/* when using http an invalid url will still return CURL_OK, we can consider
+		 * responses < 400 valid responses since they mean the request was received
+		 * and understood */
+		if (response >= 400) {
```
otaviobp: We don't require that the server responds for requests to / so we can have a 404 here and it's fine.
This function is only checking if the server is up and if we can communicate with the server using ssl (in the case of https).
If it responds anything we're good to go. So don't worry about errors here.

src/curl.c
```diff
@@ -173,7 +184,7 @@ static int check_connection(const char *test_capath)
 	}
 }
 
-int swupd_curl_init(void)
+int swupd_curl_init(char *url)
```
otaviobp: ++ for that.

src/curl.c
```diff
@@ -123,13 +123,14 @@ static CURLcode swupd_curl_set_timeouts(CURL *curl)
 	return curl_ret;
 }
 
-static int check_connection(const char *test_capath)
+static int check_connection(const char *test_capath, char *url)
 {
 	CURLcode curl_ret;
 	long response = 0;
+	char *connection_url = url ? url : version_url;
```
otaviobp: It's a good idea to try to remove dependencies from swupd.h in this file. So take the version_url out. We can always use url. If url == NULL returns an error.

src/helpers.c
```diff
@@ -579,7 +612,44 @@ enum swupd_code swupd_init(enum swupd_init_config config)
 	}
 
 	if ((config & SWUPD_NO_NETWORK) == 0) {
-		if (swupd_curl_init() != 0) {
+
+		bool curl_init = true;
+
+		/* try to init curl with whatever server is set up */
+		if (swupd_curl_init(NULL) != 0) {
```
otaviobp: use version_url here.

src/helpers.c
```diff
@@ -579,7 +612,44 @@ enum swupd_code swupd_init(enum swupd_init_config config)
...
+			/* the server is not responding */
+			if (mirror_is_set()) {
+				/* the server is a mirror, check to see if the upstream server is working */
+				swupd_curl_deinit();
```
otaviobp: if curl_init fails it shouldn't be in a state where it's inited. this is a bug in curl_init. we need to run denit in culr_init if there's any errors.

src/helpers.c
```diff
@@ -579,7 +612,44 @@ enum swupd_code swupd_init(enum swupd_init_config config)
...
+					curl_init = false;
+					free_string(&upstream_version_url);
+				}
+			} else {
+				/* the upstream server is not responding and there is no mirror */
+				curl_init = false;
+			}
+		}
+
+		if (!curl_init) {
```
otaviobp: this can move inside the if above because curl_init is never set to false outside of that if

src/helpers.c
```diff
@@ -579,7 +612,44 @@ enum swupd_code swupd_init(enum swupd_init_config config)
...
+		if (swupd_curl_init(NULL) != 0) {
+			/* the server is not responding */
+			if (mirror_is_set()) {
+				/* the server is a mirror, check to see if the upstream server is working */
+				swupd_curl_deinit();
+				char *upstream_version_url = NULL;
+				get_upstream_version_url(&upstream_version_url);
+				if (upstream_version_url && swupd_curl_init(upstream_version_url) == 0) {
+					/* the upstream server is up, unset the mirror */
+					warn("The mirror '%s' is not responding\n", version_url);
```
otaviobp: If the mirror is offline we should hard fail, so we don't need to unset_mirror_url() here.
If the mirror is online, but outdated (or errors on downloading last version) we should unset the mirror and continue

src/curl.c
castulo: good idea

src/curl.c
castulo: good point. as long as we received a response we are good to go 👌

castulo: ooh no, wait, I remember now why I added this....
If you add a fake http mirror, lets say "http://fakemirror.org/update", during check_connection it is going to respond with curl_ret = CURLE_OK, and response 404. But the server doesn't even exist, so that is why I think we need to check for valid responses.

castulo: turns out that the proxy I'm connected to is the one responding instead of "http://fakemirror.org/update", this would be solved if using https, so we can basically ignore this at this point and allow it to continue, the command will fail further down the line anyway

castulo: I'll remove that response check

src/helpers.c

src/helpers.c
castulo: true, I'll fix that

src/helpers.c
castulo: I'll change that

src/helpers.c
castulo: That is exactly how it was before... 😖 I guess there was a confusion somewhere

castulo:
@otaviobp ok, I made some changes, I think it now works as expected

src/swupd.h
```diff
@@ -388,6 +389,9 @@ extern void verify_set_picky_tree(char *picky_tree);
 /* repair.c */
 extern regex_t *compile_whitelist(const char *whitelist_pattern);
 
+/* mirror.c */
+extern bool mirror_is_set(void);
```
otaviobp: mirror_is_set() can now be a static method.

src/mirror.c
```diff
@@ -206,46 +213,85 @@ static enum swupd_code set_mirror_url(char *url)
...
+
 	fullpath = mk_full_filename(path_prefix, DEFAULT_VERSION_URL_PATH);
 	int ret = get_value_from_path(&ret_str, fullpath, true);
 	if (ret != 0 || ret_str == NULL) {
 		/* no versionurl file here, might not exist under --path argument */
 		goto out;
 	}
+
+	/* before trying to get the latest version let's make sure the central version is up */
+	if (!content_url_is_local && !check_connection(NULL, ret_str)) {
```
otaviobp: check_connection returns an error int so the correct is
check_connection() != 0.

src/swupd.h
castulo: true

src/mirror.c
castulo: done

castulo:
@otaviobp I addressed your comments in the last update

PR #951, opened 2019-06-10
Fixes a bug when calculating the download size

castulo:
To calculate the total download size of packs and fullfiles, swupd is
relying in a function that uses curl to request the server for the
download size of each file. However, when a file is not found in the
server, a size of 162 is returned by curl, which was causing the total
download size calculation to be wrong.

This commit fixes the issue by counting the file size returned by curl
only if the file exists.

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -262,6 +262,7 @@ double swupd_curl_query_content_size(char *url)
 {
 	CURLcode curl_ret;
 	double content_size;
+	long response = 0;
 
 	if (!curl) {
 		error("Curl hasn't been initialized\n");
@@ -310,6 +311,25 @@ double swupd_curl_query_content_size(char *url)
 
 	curl_ret = curl_easy_perform(curl);
 	if (curl_ret != CURLE_OK) {
+		return -1;
+	}
+
+	/* if the file doesn't exist in the remote server, the server will respond
+	 * with its "404 Not Found" page, which has a size itself (162 bytes for
+	 * NGINX which is the default content server used by clear). So if the file is
+	 * not found we need to return a size of 0 instead, otherwise the download size
+	 * calculation will be wrong */
+	curl_ret = curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &response);
+	if (curl_ret != CURLE_OK) {
+		return -1;
+	}
+	/* only acceptable responses at this point are 0, 200 and 404 */
+	if (response == 404) {
+		/* file not found, size zero */
+		return 0;
+	} else if (response != 200 && response != 0) {
+		/* some other error happened with the request, this may cause the
+		 * download size to be wrong, so just return an error */
 		return -1;
 	}
 


src/curl.c
```diff
@@ -281,7 +282,7 @@ double swupd_curl_query_content_size(char *url)
 		return -1;
 	}
 
-	curl_ret = curl_easy_setopt(curl, CURLOPT_HEADER, 0L);
+	curl_ret = curl_easy_setopt(curl, CURLOPT_HEADER, 1L);
```
otaviobp: Wont this include the size of the header in the calculation of the total size? is this expected?

src/curl.c
```diff
@@ -313,6 +314,16 @@ double swupd_curl_query_content_size(char *url)
 		return -1;
 	}
 
+	/* if the file doesn't exist in the remote server return a size of 0
+	 * otherwise a value of 162 will be returned for a non existent file
```
otaviobp: Why 162?

src/curl.c
castulo: yeah it looks like this is not necessary, I was under the impression that we needed to include the headers because the file size was actually retrieved from the headers, but I see now that even when changing this to 0L the file size is still retrieved correctly. I will leave the value as 0L (which is the default).

src/curl.c
castulo: Because when querying the content server for a resource, and the resource is not found, the server will respond with a "404 Not Found" page which has a size itself. In the case of the default content server, it is serving the content using an nginx server, and the size of the 404 page from nginx is 162 bytes.
```
$ curl -I "https://cdn.download.clearlinux.org/update/29750/files/fake"
HTTP/1.1 200 Connection established

HTTP/2 404
server: nginx
content-type: text/html
content-length: 162
vary: Accept-Encoding
date: Mon, 24 Jun 2019 22:25:33 GMT
```
 
This value however can be different with other servers:
```
$ curl -I "http://somefakepage.org"
HTTP/1.1 404 Not Found
Cache-Control: no-cache
Pragma: no-cache
Content-Type: text/html; charset=utf-8
Content-Length: 662
Via: 1.1 jfintpr02
Proxy-Connection: Keep-Alive
Connection: Keep-Alive
Date: Mon, 24 Jun 2019 22:32:27 GMT
```

I will update the comment so it is better explained

castulo:
@otaviobp I just pushed an update addressing all your comments

src/curl.c
```diff
@@ -313,6 +314,18 @@ double swupd_curl_query_content_size(char *url)
...
 
+	/* if the file doesn't exist in the remote server, the server will respond
+	 * with its "404 Not Found" page, which has a size itself (162 bytes for
+	 * NGINX which is the default content server used by clear). So if the file is
+	 * not found we need to return a size of 0 instead, otherwise the download size
+	 * calculation will be wrong */
+	curl_ret = curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &response);
+	if (curl_ret != CURLE_OK) {
+		return -1;
+	} else if (response == 404) {
```
phmccarty: What if the response is a `403` or any other error code?

Should we set `CURLOPT_FAILONERROR` instead so that it's simpler to implement error handling for server error responses?

castulo: That is a good point @phmccarty we can either set `CURLOPT_FAILONERROR` as you suggest or just check for a return code of 200, either way it is a good idea

castulo: hmm actually as a second thought, I think it would be better to do something like this:
```
curl_ret = curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &response);
if (curl_ret != CURLE_OK) {
	return -1;
} else if (response == 404) {
	return 0;
} else if (response != 200) {
	return -1;
} 
```

this function is intended to only provide the size of the file to be downloaded so the total download size is calculated. If the download actually succeeds or not is handled in a different function. If the file does not exist (which is something that should never happen with full files but happens all the time with delta packs) is the only conditions that matters at this point, that is not an error we should just return 0. If another error happens which prevents us from retrieving the file size then we should return a failure to discard the download size calculation all together.

If we use `CURLOPT_FAILONERROR` then 404 is going to be treated as an error, and we don't want that.

src/curl.c
otaviobp: ok, now it makes sense

src/curl.c
otaviobp: We handle 404 differently so we need to have fail_on_error == false as castulo said.

PR #959, opened 2019-06-21
Require a force flag to continue with insecure URL

### LINKED ISSUE #953: Inform users that http is insecure and require a force flag to continue
If users try to use a server using http inform users that this is insecure and don't proceed unless flag --force-insecure-http is used. There's one tricky problem on adding this flag. We need a way to inform that on autoupdate. So we need to have a way to do it using a config file too.

**Expected behavior** 

 - [x] User uses -u (or equivalent) with http: print a warning message and require --force-insecure-http to proceed
 - [x] User tries to run any command and the default server (mirror or server in /usr/share/default) is a http server: print a warning message and require --force-insecure-http to proceed
 - [x] User runs "swupd mirror" command to set a new mirror pointing to an http server. Warn the user and require --force-insecure-http to proceed. If flag is used create a configuration file to always use that flag when a swupd command is run.
 - [x] Users run any command pointing to an http server and using the --force-insecure-http flag: Warns that this is insecure and proceed.
 - [ ] Users have a mirror or default server pointing to a http server and no configuration file forcing that flag: We need to notify him that autoupdate is disabled on system logins with instructions on what to do about that.

Depends on #952

castulo:
Currently users can set content and version urls based on http or https
protocols. This pose a security risk if users decide to use http.

This commit blocks swupd from working with http unless is specifically
allowed by using the --allow-insecure-http flag.

Closes #953

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -33,6 +33,7 @@
 #include "config.h"
 #include "lib/log.h"
 #include "swupd.h"
+bool allow_insecure_http = false;
 bool allow_mix_collisions = false;
 bool migrate = false;
 bool sigcheck = true;
@@ -578,6 +579,7 @@ static const struct option global_opts[] = {
 	{ "max-retries", required_argument, 0, 'r' },
 	{ "retry-delay", required_argument, 0, 'd' },
 	{ "json-output", no_argument, 0, 'j' },
+	{ "allow-insecure-http", no_argument, 0, 'E' },
 	{ 0, 0, 0, 0 }
 };
 
@@ -666,6 +668,9 @@ static bool global_parse_opt(int opt, char *optarg)
 	case 'j':
 		set_json_format();
 		return true;
+	case 'E':
+		allow_insecure_http = true;
+		return true;
 	default:
 		return false;
 	}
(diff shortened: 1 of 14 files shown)


castulo:
This PR is still missing this part:

> Users have a mirror or default server pointing to a http server and no configuration file forcing that flag: We need to notify him that autoupdate is disabled on system logins with instructions on what to do about that.

This will be added as soon as PR #958 merges

src/helpers.c
```diff
@@ -1015,3 +1015,31 @@ void print_regexp_error(int errcode, regex_t *regexp)
 
 	    free(error_buffer);
 }
+
+bool is_url_allowed(char *url)
+{
+	bool insecure = false;
+
+	if (url) {
```
otaviobp: I know that this is used in other parts of the code, but using a condition == NULL to select a different behavior of a function is confusing. Instead you could create a is_url_allowed() that returned true or false if the url is valid or not. and use that with content_url and verion_url when approprite.

src/globals.c
```diff
@@ -578,6 +579,7 @@ static const struct option global_opts[] = {
 	{ "max-retries", required_argument, 0, 'r' },
 	{ "retry-delay", required_argument, 0, 'd' },
 	{ "json-output", no_argument, 0, 'j' },
+	{ "allow-insecure-http", no_argument, 0, 'E' },
```
otaviobp: I wouldn't add a short argument for this.

PR #969, opened 2019-06-25
Prevent swupd from running interactive commands. 

busykai:
Replace stdin with /dev/null for children. This addresses the issue with
hangs when running update hooks which for whatever reason may expect
interactive input.

### DIFF
--- a/src/lib/sys.c
+++ b/src/lib/sys.c
@@ -58,7 +58,6 @@ int run_command_full_params(const char *stdout_file, const char *stderr_file, ch
 {
 	int pid, ret, child_ret;
 	const char *cmd = params[0];
-	;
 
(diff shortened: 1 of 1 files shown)


thiagomacieira:
May require an extra step, to disassociate with the termrinal:
```c
    /* detach from the controlling TTY */
    tty = open("/dev/tty", O_RDWR | O_NOCTTY);
    if (tty != -1) {
        ioctl(tty, TIOCNOTTY);
        close(tty);
    }
```

busykai:
> May require an extra step, to disassociate with the termrinal:

I though about that, but after some testing confirmed this was enough. `systemctl` does not spawn any children (polkit or `systemd-tty-ask-password`).

PR #970, opened 2019-06-26
More tests to validate signature checking

otaviobp:


Makefile.am
```diff
@@ -304,6 +304,13 @@ BATS = \
 	test/functional/usability/usa-completion-basic.bats \
 	test/functional/usability/usa-download-retries.bats \
 	test/functional/usability/usa-external-modules.bats
+	test/functional/signature/cert-chain.bats \
+	test/functional/signature/corrupted-certificate.bats \
+	test/functional/signature/invalid-certificate.bats \
+	test/functional/signature/permission-incorrect.bats \
+	test/functional/signature/corrupted-signature.bats \
+	test/functional/signature/no-signature.bats
+
```
castulo: you missed adding a  '\' at the end of `test/functional/usability/usa-external-modules.bats` so your tests are never gonna run. Also the list of tests is sorted alphabetically, I'd prefer if it stayed that way, it makes easier to find tests in the list.

castulo: Also you forgot to add this test `test/functional/signature/key-rotation.bats` to the list in Makefile.am

otaviobp:
Fixed. thanks.

test/functional/signature/cert-chain.bats
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/env bats
```
castulo: I can tell you are not using the `generate_test [relative_path]/<test_name>` command from testlib to generate all these new tests since they are missing the header boilerplate:
```
# Author: Otavio Pontes
```
You may want to consider using it in the future, it saves you time by creating a new test file with the appropriate test ID, just saying... 😉

test/functional/signature/cert-chain.bats
```diff
@@ -0,0 +1,158 @@
...
+
+load "../testlib"
+
+test_setup() {
+
+	create_test_environment "$TEST_NAME"
+	create_bundle -n test-bundle -f /test-file "$TEST_NAME"
+
+	key_generation_cmd=$(cat <<-EOM
```
castulo: I'm not very familiar with these functions, but don't we already have functions to generate these in testlib?
`generate_certificate` and `create_trusted_cacert`, maybe they are not for the same purpose, I'm just pointing them out in case you missed them.

test/functional/signature/cert-chain.bats
```diff
@@ -0,0 +1,158 @@
...
+			openssl smime -sign -binary -in "$WEBDIR"/10/Manifest.MoM -signer "$TEST_NAME/intermediate_self_signed.pem" -inkey "$TEST_NAME/intermediate.key" -out "$WEBDIR"/10/Manifest.MoM.sig_self_signed -outform DER
+			# Create a certificate store with both root and self signed intermediate
+			cat "$TEST_NAME/root.pem" > "$TEST_NAME/trust.pem"
+			cat "$TEST_NAME/intermediate_self_signed.pem" >> "$TEST_NAME/trust.pem"
+	EOM
+	)
+
+	run sudo sh -c "$key_generation_cmd"
+	assert_status_is 0
+
```
castulo: I don't think you need to use the `run` command to execute those commands, as far as I understand run needs to be used to execute the command you want to test cause it will capture the output and will prevent the script from failing if the command being run fails (which sometimes is expected for tests). Test prerequisites like these can just be executed as in any other bash script, and if the command fails it will fail the whole test script. Which is expected. That way you would not have to use the assertion there (again since that is supposed to be used with the thing you want to test).

test/functional/testlib.bash
```diff
@@ -407,9 +407,10 @@ set_env_variables() { # swupd_function
 	# different options for swupd
```
castulo: you also need to update this function ->https://github.com/clearlinux/swupd-client/blob/master/test/functional/testlib.bash#L2985 since you added a new test group called "SIG"

test/functional/signature/cert-chain.bats
```diff
@@ -0,0 +1,158 @@
...
+	assert_in_output "$expected_output"
+	assert_file_not_exists "$TARGETDIR"/test-file
+
+}
+
+@test "SIG011: MoM signed with intermediate key and root cert is used" {
+
+	run sudo sh -c "$SWUPD bundle-add $SWUPD_OPTS_NO_CERT -C $TEST_NAME/root.pem test-bundle"
+
+	assert_status_is 0
```
castulo: 0 is fine, but you could also use $SWUPD_OK

test/functional/signature/cert-chain.bats
```diff
@@ -0,0 +1,158 @@
...
+	)
+	assert_in_output "$expected_output"
+	assert_file_exists "$TARGETDIR"/test-file
+
+}
+
+@test "SIG012: MoM signed with self-signed cert, but using root to validate" {
+
+	# replace sig file to one signed with intermediate_self_signed.pem
+	run sudo sh -c "cp "$WEBDIR"/10/Manifest.MoM.sig_self_signed "$WEBDIR"/10/Manifest.MoM.sig"
```
castulo: same comment from above about using "run" to execute a test prerequisite, if you use run and that command fails you are just gonna mask the error, and makes the test less straightforward since it makes you think you are testing that action.

test/functional/signature/cert-chain.bats
```diff
@@ -0,0 +1,158 @@
...
+	)
+	assert_in_output "$expected_output"
+	assert_file_not_exists "$TARGETDIR"/test-file
+
+}
+
+@test "SIG013: MoM signed with self_signed and using correct cert to validate" {
+
+	# replace sig file to one signed with intermediate_self_signed.pem
+	run sudo sh -c "cp "$WEBDIR"/10/Manifest.MoM.sig_self_signed "$WEBDIR"/10/Manifest.MoM.sig"
```
castulo: same here

test/functional/signature/corrupted-certificate.bats
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bats
+
+load "../testlib"
+
+test_setup() {
+
+	create_test_environment "$TEST_NAME"
+	create_bundle -n test-bundle -f /test-file "$TEST_NAME"
+	run sudo sh -c "echo invalid > $TEST_NAME/cert"
```
castulo: misuse of run again. Also testlib has the function `write_to_protected_file [-a] <file> <stream>` that can be used to write to a sudo owned file instead.

test/functional/signature/cert-chain.bats
otaviobp: I'll fix that, but copy and paste is a way easier way to create tests than the script.
Btw, I don't like having people names in files, why do we really needs it?
Because we use git as the source control and the information on who created and changed the file is always available in git.

otaviobp: And only 40% of our tests have author information in it. So I don't think that this information is going to be useful

test/functional/signature/cert-chain.bats
otaviobp: I'm using some parameters that are specific for this test. We don't support them in testlib and I don't think it make sense to support, because it's just going to be used in this test.
We are basically checking here if swupd works with this specifics certificates.
But in the key-rotation test you are right. I probably can use the function in testlib. I'll recheck that.

test/functional/signature/cert-chain.bats
otaviobp: I was unsure about that too. I'll rake the runs out. I prefer that too.

test/functional/testlib.bash
otaviobp: thanks. I'll update

test/functional/signature/cert-chain.bats
castulo: > I'll fix that, but copy and paste is a way easier way to create tests than the script.

True, but it is easier to copy mistakes that existed in the previous tests too, that way you always start from scratch with a template. But that is just my opinion, if you like copy/paste by all means go ahead 😁 


> Btw, I don't like having people names in files, why do we really needs it?

Yeah, I don't like having names on source file either, but the idea of having them in tests was just because sometimes tests can be confusing since some tests are meant to validate corner cases or weird scenarios difficult to understand by reading the code. So it was an easier way to contact the author in case you need help with the tests. But we can get rid of the name on tests if we do a good job documenting what the test is attempting to test within the test itself.
I know we can use git blame but some times it is difficult to see the real author when many people have modified the test file. But maybe is just me not being too familiar with that command. 😝 

Bottom line, adding the name is optional and we can get rid of names all together if you think they are not useful.

test/functional/signature/cert-chain.bats
castulo: make sense, thanks for the explanation, and yeah, I'm really not familiar with those functions related to certificates, so I was shooting in the dark....

otaviobp:
Required changes fixed

PR #974, opened 2019-06-28
docs: Add document explaining how we do signature verification on swupd

otaviobp:

### DIFF
--- a/docs/swupd_signature_verification.md
+++ b/docs/swupd_signature_verification.md
@@ -1,0 +1,114 @@
+# Swupd Signature Verification
+
+## Checking the integrity of a file
+
+Swupd contains multiple subcommands that install files in the system, including but not limited to update, bundle-add, verify. Before installing any file to the system root swupd needs to check the downloaded file integrity.
+
+Swupd uses a chain of files to validate the content and attributes of any file to be installed. A simplified description of the process to install a file F, that belongs to bundle B in version V of Clear Linux is:
+Download (if needed) the MoM (Manifest of Manifests) of version V of Clear Linux and its correspondent signature
+Check if MoM signature was signed with swupd root certificate installed on the system.
+Download (if needed) the Manifest of bundle B and calculate its hash. Check if the listed hash for bundle B in MoM is correct.
+Download (if needed) file F and calculate its hash. Check if the listed hash for file F in bundle B Manifest is correct.
+Install the file.
+
+The hash used for that operation is a SHA256 of the content and attributes of the file in question (F or Manifest of bundle B). Swupd hashdump command can be used to calculate the hash for debugging purposes.
+
+## MoM signature verification
+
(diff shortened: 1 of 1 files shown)


docs/swupd_signature_verification.md
```diff
@@ -0,0 +1,113 @@
...
+
+ - Change certificate path (-C, --certpath)
+Change in runtime to PATH the default pem file to be used on MoM signature verification. The pem file on PATH will be used instead of the default pem file defined in compile time.
+
+ - Disable signature verification (-n, --nosigcheck)
+Proceed even if signature verification fails. Warns the user about that if signature is invalid.
+**Warning**: This is not recommended and all swupd operations that install files are going
+to be insecure.
+
+
```
castulo: should we also include a note saying these same values can be defined in the config file (at either /etc/swupd/config or /run/swupd/config or /usr/share/defaults/swupd/config)?

docs/swupd_signature_verification.md
```diff
@@ -0,0 +1,113 @@
...
+## Examples
+
+Examples on how to create and keep keys to be used with swupd. This command lines are examples only and we suggest you to check openssl documentation on how to create keys with the level of security expected for your use case.
+
+All examples listed below uses a 1 day expiry date which is unrealistic for a real use case. The only purpose of this documentation is to explain what’s supported by swupd and shouldn’t be used as a guide on how to sign Clear Linux based distributions.
```
castulo: I think the word "uses" here "All examples listed below uses a 1 day expiry" should be use since referring to many examples.

docs/swupd_signature_verification.md
```diff
@@ -0,0 +1,113 @@
...
+
+# Check if the signature is correct using the same root key
+openssl smime -verify -in Manifest.MoM.sig -inform DER -content Manifest.MoM -purpose any -CAfile Swupd_Root.pem
+```
+
+## Certificate Revocation
+
+### Certificate Revocation List
+
+Usage of Certificate Revocation list is implemented in src/signature.c, but isn’t currently used on swupd. To enable that calls to signature_init() should set the crl parameter pointing to a file with the Certificate Revocation List.
```
castulo: To enable that calls to signature_init() -> To enable that, calls to signature_init()

docs/swupd_signature_verification.md
```diff
@@ -0,0 +1,113 @@
...
+
+### OCSP (Online Certificate Status Protocol)
+
+OCSP isn’t yet supported on swupd. If a certificate with OCSP set and flagged as critical is used swupd will abort with a warning saying that OCSP isn’t supported.
```
castulo: If a certificate with OCSP set and flagged as critical is used swupd will abort  ->  If a certificate with OCSP set and flagged as critical is used, swupd will abort

PR #987, opened 2019-07-11
also-add: Update, diagnose and repair shoudn't consider the also-add …

otaviobp:
…bundles

In the case that we have a bundle installed as also-add and removed later,
repair shouldn't reinstall that bundle. The same is valid for diagnose and update.
Os-install is the only exception. We should always install all bundles that are
listed as also-add in os-install.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -218,6 +218,7 @@ BATS = \
 	test/functional/checkupdate/chk-update-no-target-content.bats \
 	test/functional/checkupdate/chk-update-slow-server.bats \
 	test/functional/checkupdate/chk-update-version-match.bats \
+	test/functional/diagnose/diagnose-also-add.bats \
 	test/functional/diagnose/diagnose-basics.bats \
 	test/functional/diagnose/diagnose-boot-file.bats \
 	test/functional/diagnose/diagnose-client-certificate.bats \
@@ -232,6 +233,7 @@ BATS = \
 	test/functional/mirror/mirror-createdir-negative.bats \
 	test/functional/mirror/mirror-createdir.bats \
 	test/functional/mirror/mirror-json.bats \
+	test/functional/os-install/install-also-add.bats \
 	test/functional/os-install/install-bad.bats \
 	test/functional/os-install/install-basics.bats \
 	test/functional/os-install/install-json.bats \
@@ -252,6 +254,7 @@ BATS = \
 	test/functional/repair/repair-format-mismatch-overdrive.bats \
 	test/functional/repair/repair-format-mismatch.bats \
 	test/functional/repair/repair-ghosted.bats \
+	test/functional/repair/repair-ignore-also-add.bats \
 	test/functional/repair/repair-json.bats \
 	test/functional/repair/repair-missing-file.bats \
 	test/functional/repair/repair-no-disk-space.bats \
@@ -282,6 +285,7 @@ BATS = \
 	test/functional/update/update-client-certificate.bats \
 	test/functional/update/update-download.bats \
 	test/functional/update/update-fail-to-get-mom.bats \
+	test/functional/update/update-ignore-also-add.bats \
 	test/functional/update/update-include-old-bundle-with-tracked-file.bats \
 	test/functional/update/update-include-old-bundle.bats \
 	test/functional/update/update-include.bats \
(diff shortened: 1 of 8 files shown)


src/verify.c
```diff
@@ -700,6 +700,12 @@ enum swupd_code verify_main(void)
 		goto clean_args_and_exit;
 	}
 
+	/* Unless we are installing a new bundle we shoudn't include optinal
+	* bundles to the bundle list. */
+	if (!cmdline_option_install) {
+		skip_optional_bundles = true;
+	}
+
```
castulo: small typo, "optional" -> "optional"

test/functional/diagnose/diagnose-also-add.bats
```diff
@@ -0,0 +1,73 @@
...
+		Inspected 5 files
+		  3 files were missing
+		Use "swupd repair" to correct the problems in the system
+		Diagnose successful
+
+	EOM
+	)
+	assert_regex_is_output "$expected_output"
+
+}
```
castulo: so if I understand this correctly, if an also-add bundle has never been removed explicitly by the user, a "swupd diagnose" should find it as "missing", right? Because also-add bundles should be added by default. Once a user removes it, a future "swupd diagnose" should not report it as missing, correct?

test/functional/repair/repair-ignore-also-add.bats
```diff
@@ -0,0 +1,85 @@
...
+		Calling post-update helper scripts
+		Repair successful
+	EOM
+	)
+	assert_regex_is_output "$expected_output"
+	assert_file_exists "$TARGETDIR"/core
+	assert_file_exists "$TARGETDIR"/foo/test-file1
+	assert_file_not_exists "$TARGETDIR"/bar/test-file2
+
+}
```
castulo: since test-bundle3 was an also-add of test-bundle1, wasn't it supposed to be installed during the `swupd repair`?

test/functional/update/update-ignore-also-add-flag.bats
```diff
@@ -0,0 +1,55 @@
...
+	# add dependencies
+	add_dependency_to_manifest -o "$WEBDIR"/100/Manifest.test-bundle1 test-bundle2
+
+}
+
+@test "UPD057: Update a system where nested dependencies are added in the newer version" {
+
+	run sudo sh -c "$SWUPD update $SWUPD_OPTS"
+
+	assert_status_is 0
```
castulo: 0 -> "$SWUPD_OK"

Makefile.am
```diff
@@ -282,6 +285,7 @@ BATS = \
 	test/functional/update/update-client-certificate.bats \
 	test/functional/update/update-download.bats \
 	test/functional/update/update-fail-to-get-mom.bats \
+	test/functional/update/update-ignore-also-add-flag.bats \
```
castulo: I think the name of this test is not correct, "update-ignore-also-add-flag", also-add is not a flag.

test/functional/diagnose/diagnose-also-add.bats
otaviobp: There's no way for us in a diagnose to know if a bundle was removed by the user or if it wasn't added by a bundle-add.

If I run:

    $ swupd bundle-add --skip-optional bundle_name
    $ swupd diagnose

I would expect diagnose to be successful. Do this make sense?

test/functional/repair/repair-ignore-also-add.bats
otaviobp: No. swupd repair shouldn't install also-add bundles. I didn't add a check here to make sure bundle3 was not installed because I created it for the next test. do you think I should?

Makefile.am
otaviobp: you're right. I'll rename that

otaviobp:
Sent new versions with updated changes

test/functional/diagnose/diagnose-also-add.bats
castulo: yes, my bad, for some reason I though test-bundle1 had been added as optional dependency to os-core, but I just realized it is an include, that is why it was added. 😊

test/functional/repair/repair-ignore-also-add.bats
castulo: no, if it is included in the other test that is fine 👍

PR #988, opened 2019-07-12
Adding flag validation to the swupd build process

castulo:
swupd supports global flags and supports local flags that are specific
to subcommands. Flags that are specific to subcommands can be reused in
a different subcommand. Global flags should be unique and should not be
reused as local flags to avoid conflicts.

This commit adds a bash script that checks that the flags currently used
in swupd are valid (have no conflicts with other flags from the same
command or with the global flags). This script will be run as part of
the build process so in case an invalid flag is found the build will be
stopped. Alternatively, the script can be used by developers to validate
a new flag during the implementation time.

### DIFF
--- a/docs/swupd.1.rst
+++ b/docs/swupd.1.rst
@@ -168,7 +168,7 @@ SUBCOMMANDS
         are available. It will return 0 with succeeded and a different value
         of 0 with failed.
 
-    - `-d, --deps={bundle}`
+    - `--deps={bundle}`
 
         Lists all bundle dependencies of the passed bundle, including
         recursively included bundles.
@@ -281,7 +281,7 @@ SUBCOMMANDS
         <path>/etc/swupd/mirror_contenturl and
         <path>/etc/swupd/mirror_versionurl
 
-    - `-u, --unset`
+    - `-U, --unset`
 
         Remove the content and version URL configuration by removing
         <path>/etc/swupd
@@ -378,7 +378,7 @@ SUBCOMMANDS
 
             Only runs the repair operation on the os-core and vi bundles.
 
-``search {string}``
+``search-file {string}``
 
     Search for matching paths in manifest data. The specified {string}
     is matched in any part of the path listed in manifests, and all
@@ -396,7 +396,7 @@ SUBCOMMANDS
 
         Restrict search to designated dynamic shared library paths.
 
-    - `-b, --binary`
+    - `-B, --binary`
 
         Restrict search to designated program binary paths.
 
@@ -433,7 +433,7 @@ SUBCOMMANDS
         available on the version url server, and what version number is
         available.
 
-    - `-d, --download`
+    - `--download`
 
         Do not perform an update, instead download all resources needed
         to perform the update, and exit.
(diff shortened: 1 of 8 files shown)


docs/swupd.1.rst
```diff
@@ -378,7 +378,7 @@ SUBCOMMANDS
 
             Only runs the repair operation on the os-core and vi bundles.
 
-``search {string}``
+``search-file {string}``
```
otaviobp: +1 for this too

docs/swupd.1.rst
```diff
@@ -433,7 +433,7 @@ SUBCOMMANDS
         available on the version url server, and what version number is
         available.
 
-    - `-d, --download`
+    - `-D, --download`
```
otaviobp: This option isn't going to be used very frequent, so I would remove the short option

docs/swupd.1.rst
```diff
@@ -168,7 +168,7 @@ SUBCOMMANDS
         are available. It will return 0 with succeeded and a different value
         of 0 with failed.
 
-    - `-d, --deps={bundle}`
+    - `-i, --deps={bundle}`
```
otaviobp: This option isn't going to be used very frequent, so I would remove the short option

docs/swupd.1.rst
castulo: sounds good

docs/swupd.1.rst
castulo: I'll do that then

castulo: btw, I had chosen "i" because the "deps" are actually the **i**ncludes

castulo: if we are getting rid of the short option for --deps, we might as well get rid of the short option for --has-dep, don't you think? both are very similar so it would look odd to have a short option for one but not have it for the other. What do you think @otaviobp ?

otaviobp: hnaahah. I didn't get that. 
I don't think --has-deps should have a short option, but I wouldn't remove it to avoid breaking compatibility. --deps breaks it, but it's needed. Let's try to avoid at lest adding more short options for all flags

castulo: 😩 ok, that kind of visually unattractive things bother me, but only because I'm very anal haha, so ok, I will just remove one short option.

src/clr_bundle_ls.c
```diff
@@ -54,14 +56,14 @@ static void print_help(void)
...
-	print("   -d, --deps=[BUNDLE]     List bundles included by BUNDLE\n");
 	print("   -D, --has-dep=[BUNDLE]  List dependency tree of all bundles which have BUNDLE as a dependency\n");
+	print("   --deps=[BUNDLE]         List bundles included by BUNDLE\n");
 	print("\n");
 }
 
 static const struct option prog_opts[] = {
 	{ "all", no_argument, 0, 'a' },
-	{ "deps", required_argument, 0, 'd' },
+	{ "deps", required_argument, 0, DEPS_FLAG },
```
otaviobp: maybe we should port (later, not in this PR) all cases that we have longopts only to use this strategy. it's way better for the config loader. What do you think?
I though it's way cleaner

castulo: yes, we can do that

PR #992, opened 2019-07-12
 autoupdate: Don't report progress on autoupdate 

otaviobp:


config
```diff
@@ -76,6 +76,9 @@
 # Print extra information to help debugging problems (boolean value)
 #debug=<true/false>
 
+# Don't report progress (boolean value)
+#no-progress=<true/false>
```
castulo: you should use underscore here: no_progress, it will be converted to no-progress at parsing time.

docs/swupd.1.rst
```diff
@@ -125,6 +125,10 @@ used to modify the core behavior and resources that swupd uses.
 
    Set the maximum number of parallel downloads
 
+- ``--no-progress``
+
+   Don't report progress
```
castulo: I think this should say something a little more explanatory for example "Don't report **download** progress", because just saying "Don't report progress" I think can be confused with running a command "quietly", the output from a command is still some kind of progress report.

otaviobp:
Fixed reported problems

PR #995, opened 2019-07-15
Flags

otaviobp:


docs/swupd.1.rst
```diff
@@ -382,6 +382,14 @@ SUBCOMMANDS
 
             Only runs the repair operation on the os-core and vi bundles.
 
+``search``
```
castulo: 👍

src/globals.c
```diff
@@ -625,8 +625,6 @@ static const struct option global_opts[] = {
 	{ "versionurl", required_argument, 0, 'v' },
 	{ "quiet", no_argument, &log_level, LOG_ERROR },
 	{ "debug", no_argument, &log_level, LOG_DEBUG },
-	//TODO: -D option is deprecated. Remove that on a Major release
-	{ "", required_argument, 0, 'D' },
```
castulo: nice

src/os_install.c
```diff
@@ -31,6 +31,7 @@ static const struct option prog_opts[] = {
 	{ "force", no_argument, 0, 'x' },
 	{ "bundles", required_argument, 0, 'B' },
 	{ "version", required_argument, 0, 'V' },
+	{ "manifest", required_argument, 0, 'm' },
```
castulo: Why are you adding -m to os-install? we are using -V for that. Since it is a "new" command I don't think we need to support an old flag. If people want to use the old -m flag then they can use "verify --install -m"

otaviobp: I think that diagnose, repair and os-install should be consistent. -m was added to diagnose and repair. So I added it here too.
They aren't listed in docs, like diagnose and repair.

PR #1000, opened 2019-07-16
Progress overlap

castulo:
Fix format of do_staging when reporting progress

The function do_staging is regularly used within a loop to stage all the files for a given bundle-add or update, repair, etc. However when the loop contains a progress report (with a percentage), if there are errors in the do_staging function, they get printed in the same line as the percentage, messing up the output. Like this:

    ...22%Warning: Update target directory does not exist: some_file...

This commit fixes that output by adding a carriage return at the beginning of each message do_staging could print so it would overlap with the percentage value effectively deleting it.

### DIFF
--- a/src/lib/log.c
+++ b/src/lib/log.c
@@ -38,6 +38,14 @@ static void log_internal(FILE *out, const char *file, int line, const char *labe
 	struct tm *timeinfo;
 	int printed;
 
+	/* if the message starts with '\n' add a line break before the message */
+	while (strncmp(format, "\n", 1) == 0 || strncmp(format, "\r", 1) == 0) {
+		if (out == stdout || out == stderr) {
+			fprintf(out, "%c", format[0]);
+		}
+		format++;
+	}
+
 	// In debug mode, print more information
 	if (cur_log_level == LOG_DEBUG) {
 		time_t currtime;
@@ -55,6 +63,7 @@ static void log_internal(FILE *out, const char *file, int line, const char *labe
 		fprintf(out, "%s: ", label);
 	}
 	vfprintf(out, format, args_list);
+	fflush(out);
 }
 
 void log_set_level(int log_level)
--- a/src/lib/progress.c
+++ b/src/lib/progress.c
@@ -113,7 +113,7 @@ void progress_report(double count, double max)
 			if (progress_function) {
 				progress_function(step.description, step.current, step.total, percentage);
 			} else {
-				info("\r\t...%d%%", percentage);
+				info("\t...%d%%", percentage);
 				if (!isatty(fileno(stdout))) {
 					/* if not in a tty add new lines so every percentage
 					 * is in its own line */
@@ -121,6 +121,8 @@ void progress_report(double count, double max)
 				} else if (percentage == 100) {
 					/* add a line break once we reach 100% */
 					info("\n");
+				} else {
+					info("\r");
 				}
 			}
 			fflush(stdout);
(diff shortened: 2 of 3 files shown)


src/staging.c
```diff
@@ -97,13 +97,13 @@ enum swupd_code do_staging(struct file *file, struct manifest *MoM)
 	ret = stat(targetpath, &s);
 	if ((ret == -1) && (errno == ENOENT)) {
 		if (MoM) {
-			warn("Update target directory does not exist: %s. Trying to fix it\n", targetpath);
+			warn("\rUpdate target directory does not exist: %s. Trying to fix it\n", targetpath);
```
otaviobp: If we use \r won't the percentages be truncated with the test? Won't we have:

    22% rning: Update ....

And \r are not graceful when printing to a file. journal will probably consider this a blob instead of a text. If it's not bad to use \n I prefer?

src/lib/log.c
```diff
@@ -37,6 +37,14 @@ static void log_internal(FILE *out, const char *file, int line, const char *labe
 	char time_str[50];
 	struct tm *timeinfo;
 	int printed;
+	const char *msg;
```
otaviobp: we don't need a msg here. Just use the format variable.
It's ok to do a format++;

src/lib/log.c
```diff
@@ -37,6 +37,14 @@ static void log_internal(FILE *out, const char *file, int line, const char *labe
...
+
+	/* if the message starts with '\n' add a line break before the message */
+	msg = format;
+	while (strncmp(msg, "\n", 1) == 0 || strncmp(msg, "\r", 1) == 0) {
```
otaviobp: I like this patch, but I'm still unsure about the \r

src/staging.c
castulo: I have tested this locally and it works as expected. It overwrites the progress percentage and the next value of the percentage is printed t the bottom, as expected. Actually we are already using this same approach for printing output when diagnose find mismatches https://github.com/clearlinux/swupd-client/commit/7d295f3ec5d92f7d0b382a6b3a64026180be5068

Using \n leaves traces of the previous percentage which looks ugly, like this:

```
Warning: Update target directory does not exist: some_dir. Trying to fix....
Path /some_dir is missing on the file system ... fixing
	...39%
Warning: Update target directory does not exist: /usr/share/man/fr/man1. Trying to fix it
Path /usr/share/man/fr is missing on the file system ... fixing
Path /usr/share/man/fr/man1 is missing on the file system ... fixing
	...42%
Warning: Update target directory does not exist: /usr/share/man/man5. Trying to fix it
Path /usr/share/man/man5 is missing on the file system ... fixing
etc...
```

src/lib/log.c
castulo: ok I wasn't sure if modifying that pointer was ok, so I wanted to play it safe. I'll use the `format` pointer then

src/staging.c
otaviobp: That's true. We won't have the problem I said because of the \n at the end of the line.
the only problem now is when output is redirected to files.

castulo: I submitted an update that I think addresses that problem

castulo:
> A suggestion on how to fix the \r problem when redirecting to a file.
> Instead of adding the \r to the start of functions that are printed during progress we could always print a \r at the end of the progress report. So if a new message is printed it will print over the progress bar. In that funciton we have information if it is a file or tty, so we can prevent printing \r if it is a file.

That is a good idea

src/lib/progress.c
```diff
@@ -113,7 +113,7 @@ void progress_report(double count, double max)
 			if (progress_function) {
 				progress_function(step.description, step.current, step.total, percentage);
 			} else {
-				info("\r\t...%d%%", percentage);
+				info("\t...%d%%\r", percentage);
```
otaviobp: Instead of printing this here I would move that to the else of the if (!isatty) so we wont print it to files

PR #1002, opened 2019-07-16
Use lowercase URL comparison for is_url_allowed()

jwakre:
The URL value must be set to lowercase to prevent support for mixed case
or uppercase HTTP strings without the --allow-insecure-http flag.

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -1028,7 +1028,7 @@ void print_regexp_error(int errcode, regex_t *regexp)
 
 bool is_url_allowed(char *url)
 {
-	if (strncmp(url, "http://", 7) == 0) {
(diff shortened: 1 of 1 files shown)


src/helpers.c
```diff
@@ -1028,7 +1028,10 @@ void print_regexp_error(int errcode, regex_t *regexp)
 
 bool is_url_allowed(char *url)
 {
-	if (strncmp(url, "http://", 7) == 0) {
+	char *url_lower;
+
+	url_lower = str_tolower(url);
```
otaviobp: Really good catch.
But you're leaking memory here.
And you can use strncasecmp() instead of str_tolower() + strncmp()

PR #1003, opened 2019-07-17
Adding --wait-for-scripts flag

### LINKED ISSUE #909: Post update triggers are not blocking
**Describe the bug**
After running `swupd verify --install` on a chroot that has touched bundle files in it to perform an install of Clear Linux, post update triggers are run in a non-blocking mode.  This prevents automation from creating the official Docker image for Clear Linux, or validation from running that checks whether the clrtrust post update trigger has run without resorting to adding a `sleep 1m` after the chroot is created.

**To Reproduce**
Use the following script:
```
#!/bin/bash
set -xe
TMP_ROOT="$(mktemp -d)"
mkdir -p "${TMP_ROOT}/usr/share/clear/bundles"
touch "${TMP_ROOT}/usr/share/clear/bundles/"{os-core,os-core-update}
swupd verify --install -p "${TMP_ROOT}" -b -m "$(< /usr/share/clear/version)"
# Missing a sleep 1m here, so the following command will sometimes fail
[ -d "${TMP_ROOT}/var/cache/ca-certs" ]
rm -rf "${TMP_ROOT}"
```

**Expected behavior**
After a `swupd verify --install`, only when all of the post update triggers have finished running does the swupd command exit.  While this could be exposed as a new command line option, it would probably be better to have this blocking behavior for post update triggers enabled by default to ensure all users/consumers of swupd get this behavior, namely DevOps, the clr-installer, and QA.

 - Clear Linux OS Version: 29100 (N/A)
 - Platform: Baremetal, VM, anywhere swupd is used

castulo:
When running tasks that require to run post update scripts in swupd,
for most users we don't want to wait for the scripts to finish running,
those can be run in the background so we can finish the task faster.
However for some users like DevOps, clr-installer, and QA, it is useful
to have a way to instruct swupd to wait until these scripts have finish
running.

This commit adds a way of doing this by introducing the
--wait-for-scripts flag.

Closes #909

### DIFF
--- a/swupd.bash
+++ b/swupd.bash
@@ -23,7 +23,7 @@ _swupd()
     # $1 is the command being completed, $2 is the current word being expanded
     local opts IFS=$' \t\n'
     local -i i installed
-    local global="--help --url --contenturl --versionurl --port --path --format --nosigcheck --ignore-time --statedir --certpath  --time --no-scripts --no-boot-update --max-parallel-downloads --max-retries --retry-delay --json-output  --allow-insecure-http --debug --quiet --no-progress "
+    local global="--help --url --contenturl --versionurl --port --path --format --nosigcheck --ignore-time --statedir --certpath  --time --no-scripts --no-boot-update --max-parallel-downloads --max-retries --retry-delay --json-output  --allow-insecure-http --debug --quiet --no-progress --wait-for-scripts "
     COMPREPLY=()
     for ((i=COMP_CWORD-1;i>=0;i--))
     do case "${COMP_WORDS[$i]}" in
--- a/config
+++ b/config
@@ -79,6 +79,8 @@
 # Don't report progress (boolean value)
 #no_progress=<true/false>
 
+# Wait for the post-update scripts to complete (boolean value)
+#wait_for_scripts=<true/false>
 
 
 #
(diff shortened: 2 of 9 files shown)


swupd.bash
```diff
@@ -23,7 +23,7 @@ _swupd()
     # $1 is the command being completed, $2 is the current word being expanded
     local opts IFS=$' \t\n'
     local -i i installed
-    local global="--help --url --contenturl --versionurl --port --path --format --nosigcheck --ignore-time --statedir --certpath  --time --no-scripts --no-boot-update --max-parallel-downloads --max-retries --retry-delay --json-output  --allow-insecure-http --debug --quiet --no-progress "
+    local global="--hlp --url --contenturl --versionurl --port --path --format --nosigcheck --ignore-time --statedir --certpath  --time --no-scripts --no-boot-update --max-parallel-downloads --max-retries --retry-delay --json-output  --allow-insecure-http --debug --quiet --no-progress --wait-for-scripts "
```
otaviobp: I think there's a typo here:

    --hlp?

otaviobp: btw, github++ for showing that. I wouldn't see that in a git diff

castulo: ups, I must have pressed the delete key without noticing it. Good catch

castulo: fixed 😊

PR #1008, opened 2019-07-17
Config - Cleanup 2

otaviobp:
And also stopping using static variable for the config_parser () callls. With this we can use the config parser to parse the os-release file intestead of having other parser implementation on version.c


src/config_loader.c
```diff
@@ -24,40 +24,31 @@
 #define is_from_global_section(_section) (strcmp(_section, "global") == 0)
 #define is_from_command_section(_section, _command) (strcmp(_section, _command) == 0)
 
-static struct config_loader_data {
```
castulo: Nice, good idea. 👍

src/globals.c
```diff
@@ -551,43 +556,36 @@ static bool global_parse_opt(int opt, char *optarg)
 			set_json_format(true);
 		}
```
castulo: Maybe you should update options "j", "FLAG_NO_PROGRESS" and "FLAG_WAIT_FOR_SCRIPTS" so their if is consistent with the ones below. For example:
`wait_for_scripts = optarg ? strtobool(optarg) : true;`

src/clr_bundle_add.c
```diff
@@ -53,13 +56,19 @@ static void print_help(void)
...
-	{ "skip-diskspace-check", no_argument, &skip_diskspace_check, 1 },
+	{ "skip-optional", no_argument, 0, FLAG_SKIP_OPTIONAL },
+	{ "skip-diskspace-check", no_argument, 0, FLAG_SKIP_DISKSPACE_CHECK },
 };
 
 static bool parse_opt(int opt, UNUSED_PARAM char *optarg)
 {
 	switch (opt) {
+	case FLAG_SKIP_OPTIONAL:
+		skip_optional_bundles = optarg ? strtobool(optarg) : true;
```
castulo: since these options are local they don't need that logic,`skip_optional_bundles = true` should be enough since we are ignoring values of `false` from the config_loader

src/swupd.h
```diff
@@ -169,9 +169,6 @@ struct file_counts {
 extern bool verify_esp_only;
 extern bool have_network;
 extern bool verify_bundles_only;
-extern bool ignore_config;
-extern bool ignore_state;
-extern bool ignore_orphans;
 extern bool no_scripts;
```
castulo: nice!

src/globals.c
```diff
@@ -201,35 +201,35 @@ static bool is_valid_integer_format(char *str)
...
 
-		/* Prevent some disasters: since the state dir can be destroyed and
-		 * reconstructed, make sure we never set those by accident and nuke the
-		 * system. */
-		if (!strcmp(path, "/") || !strcmp(path, "/var") || !strcmp(path, "/usr")) {
-			error("Refusing to use '%s' as a state dir because it might be erased first\n", path);
-			return false;
-		}
+	if (path[0] != '/') {
+		error("state dir must be a full path starting with '/', not '%c'\n", path[0]);
```
castulo: Upper case for the first letter

src/globals.c
```diff
@@ -287,70 +287,48 @@ static bool set_default_format_string()
...
 	}
 
+	free(path_prefix);
+	path_prefix = new_path;
 	return true;
+
+error:
+	error("Bad path_prefix %s (%s), cannot continue\n",
+	      path_prefix, strerror(errno));
+	free(new_path);
```
castulo: remember you were using free_string, no big deal but it is good to be consistent 😌

src/helpers.c
```diff
@@ -1041,3 +1041,91 @@ bool is_url_allowed(char *url)
 
 	return true;
 }
+
```
castulo: 👍 for this, looks a lot better here

src/globals.c
otaviobp: well, I was unsure what was the best way to do it to be honest. hahaha.
So I didn't change what was already in place. I'm thinking that maybe it's useful to create a opt_to_bool() that returns true on NULL or return strtobool(optarg). What do you think?

src/clr_bundle_add.c
otaviobp: true

src/globals.c
castulo: haha, also I had suggested the same thing in my previous review but ended up deleting the comment, but I do think is a good idea. Or since this is only used in globals.c even a #define could work.

otaviobp: hahaha. nice. I'll create a define in another commit

otaviobp:
Fixed the repoted problems

otaviobp:
Except that I'm using free instead of free_string in some cases. There was some memory leaks when using free_string. I'm going to check that to find out what's going on. But for now I'll keep using free for local variables.

src/clr_bundle_add.c
```diff
@@ -53,13 +56,19 @@ static void print_help(void)
...
 };
 
 static bool parse_opt(int opt, UNUSED_PARAM char *optarg)
 {
 	switch (opt) {
+	case FLAG_SKIP_OPTIONAL:
+		skip_optional_bundles = true;
+		return true;
+	case FLAG_SKIP_DISKSPACE_CHECK:
+		skip_diskspace_check = optarg ? strtobool(optarg) : true;
```
castulo: you forgot to also change this one to:
skip_diskspace_check = true;

otaviobp: oops. fixed

PR #1011, opened 2019-07-18
Display distribution on swupd info

litcoder:
Display distribution information on swupd info which is from PRETTY_NAME filed
in os-release file to support distinguishable version information display
in down streams.

Signed-off-by: Brandon Hong <brandon.hong@intel.com>

### DIFF
--- a/src/info.c
+++ b/src/info.c
@@ -29,12 +29,14 @@ enum swupd_code print_update_conf_info()
 	int current_version = get_current_version(path_prefix);
 	bool dist_string_found = get_distribution_string(path_prefix, dist_string);
 
-	if (current_version < 0 || !dist_string_found) {
+	// Set distribution string 'unknown' when not found.
+	info("Distribution:      %s\n", dist_string_found ? dist_string : "unknown");
+
+	if (current_version < 0) {
 		error("Unable to determine current OS version\n");
 		info("Installed version: unknown\n");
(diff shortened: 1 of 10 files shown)


src/info.c
```diff
@@ -25,13 +25,16 @@
 enum swupd_code print_update_conf_info()
 {
 	enum swupd_code ret = SWUPD_OK;
-
+	char dist_string[LINE_MAX];
 	int current_version = get_current_version(path_prefix);
-	if (current_version < 0) {
+	bool dist_string_found = get_distribution_string(path_prefix, dist_string);
+
+	if (current_version < 0 || !dist_string_found) {
```
otaviobp: If !dist_string_found you have the information about the version to display. So instead you can print

    Distribution: unknown
    Installed version: 12345

And if current_version < 0

    Distribution: Clear Linux OS
    Installed version: unknown

otaviobp:
There was a patch that was merged that may help you when running tests too. I recommend rebasing it again

litcoder:
I updated some of test cases to reflect the new swupd info output however I'm still seeing some failures in usa-config-files.bats when I run local testing. Since those errors were still observed without the patch, I'm thinking this might not directly related to my patch.

litcoder:
After the rebase, most of the issues gone but still have a issue with the test case USA010 and this still fails without the patch in my local testing.

otaviobp:
merged. thanks

PR #1014, opened 2019-07-18
Changes in diagnose flags: --picky and --extra-files-only

### LINKED ISSUE #932: Some modified files show in swupd verify but not swupd verify --picky
**Describe the bug**
/usr/lib/modules is part of the default picky-whitelist so they don't get targeted with a `swupd verify --picky` but they do get targeted with a `swupd verify`.

Additionally, `swupd verify --picky --fix` fixes those files even though they didn't show in `swupd verify --picky` 

**To Reproduce**
Have custom kernel modules in use. Run `swupd verify` vs `swupd verify --picky`.

**Expected behavior**
I'd expect `swupd verify` to always be a subset of `swupd verify --picky`

 - Clear Linux OS Version: 29620

castulo:
Things in this PR:

- Makes `diagnose --picky` a superset of `diagnose`. Running `swupd diagnose --picky` used to result in only identifying the extra files in the system, but it wouldn't run the other standard checks that a `swupd diagnose` would do, like checking for missing files and corrupt files. `swupd diagnose --picky` now does everything `swupd diagnose` do plus searching for the extra files.
- Adds the `--extra-files-only` flag for `diagnose` and `repair` so there is still the option to only check or remove the extra files in a system.
- The PR also refactors the verify function to simplify its workflow and remove code duplication
- It also adds some validations for flags that are mutually exclusive in `verify`, `diagnose` and `repair`.
- Minor updates to the documentation

Closes #932


src/repair.c
```diff
@@ -86,9 +92,27 @@ static bool parse_opt(int opt, char *optarg)
 		return true;
 	case 'q':
 		cmdline_option_quick = true;
+		/* mutually exclusive flags */
+		if (cmdline_option_picky) {
+			error("You cannot use --quick and --picky together\n\n");
```
otaviobp: why we can't use --quick and --picky together?

castulo: because the purpose of --quick is to only add missing files...  And the process of a verify is this:

1) Add missing files
(with --quick) the process is finished here
2) Fix corrupt files
3) Remove files that were deleted in manifests but are still in the system
4) remove files that are not in any manifest

So with --quick and --picky nothing would happen cause the code would never reach the picky part anyway, so it is confusing.

castulo: it is actually how it works today, but it is wrong:
```
$ sudo swupd diagnose --quick --picky
Diagnosing version 30440

Inspected 673920 files

Diagnose successful
```

otaviobp: In my mind quick was suppose to do both. Reading the doc looks like you are correct. Merging this now then.

PR #1015, opened 2019-07-19
Fix logger format when printing debug messages

### LINKED ISSUE #1009: `--debug` flag should also suppress progress
There's a new flag `--no-progress` and I was checking whether it's mutually exclusive with `--debug` flag.

It seems to me that when there is `--debug`, it'd be better to suppress the progress as if `--no-progress` is also specified. Below are the output for adding `gvim` bundle, where the first one only has `--debug` and looks messy.

```
❯ sudo swupd bundle-add gvim --debug                          
2019-07-17 20:20:15 (src/curl.c:136)              Debug: Curl - check_connection url: https://cdn.download.clearlinux.org/update
2019-07-17 20:20:15 (src/bundle.c:777)            Loading required manifests...
2019-07-17 20:20:16 (src/packs.c:248)             No packs need to be downloaded
2019-07-17 20:20:16 (src/fullfile.c:158)          No extra files need to be downloaded
2019-07-17 20:20:16 (src/bundle.c:935)            Installing bundle(s) files...
2019-07-...100%2019-07-17 20:20:16 (src/lib/progress.c:123)      
2019-07-17 20:20:16 (src/scripts.c:146)           Calling post-update helper scripts
2019-07-17 20:20:41 (src/bundle.c:1077)           Successfully installed 1 bundle
```

```
❯ sudo swupd bundle-add gvim --debug --no-progress                    
2019-07-17 20:21:00 (src/curl.c:136)              Debug: Curl - check_connection url: https://cdn.download.clearlinux.org/update
2019-07-17 20:21:00 (src/bundle.c:777)            Loading required manifests...
2019-07-17 20:21:01 (src/packs.c:248)             No packs need to be downloaded
2019-07-17 20:21:01 (src/fullfile.c:158)          No extra files need to be downloaded
2019-07-17 20:21:01 (src/bundle.c:935)            Installing bundle(s) files...
2019-07-17 20:21:01 (src/scripts.c:146)           Calling post-update helper scripts
2019-07-17 20:21:26 (src/bundle.c:1077)           Successfully installed 1 bundle
```

castulo:
Debug messages need to be treated a little differently from other
messages, since they include more information, like the time, file and
line of code. So certain things like having carriage returns and
messages with only a line break should not be allowed since they would
mess up the debug output.

This commit fixes that issue

Closes #1009

### DIFF
--- a/src/lib/log.c
+++ b/src/lib/log.c
@@ -1,7 +1,7 @@
 /*
  *   Software Updater - client side
  *
- *      Copyright © 2012-2016 Intel Corporation.
+ *      Copyright © 2012-2019 Intel Corporation.
  *
  *   This program is free software: you can redistribute it and/or modify
  *   it under the terms of the GNU General Public License as published by
@@ -20,50 +20,86 @@
 #define _GNU_SOURCE
 
 #include <ctype.h>
+#include <limits.h>
 #include <stdarg.h>
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
+#include <unistd.h>
 
 #include "log.h"
+
 #define DEBUG_HEADER_SIZE 50
+#define LINE_MAX _POSIX2_LINE_MAX
 
 static void log_internal(FILE *out, const char *file, int line, const char *label, const char *format, va_list args_list);
 static int cur_log_level = LOG_INFO;
 static log_fn_t log_function = log_internal;
 
-static void log_internal(FILE *out, const char *file, int line, const char *label, const char *format, va_list args_list)
+static void print_debug_info(FILE *out, const char *file, int line)
 {
 	char time_str[50];
 	struct tm *timeinfo;
 	int printed;
+	time_t currtime;
 
-	/* if the message starts with '\n' add a line break before the message */
-	while (strncmp(format, "\n", 1) == 0 || strncmp(format, "\r", 1) == 0) {
-		if (out == stdout || out == stderr) {
+	time(&currtime);
+	timeinfo = localtime(&currtime);
+	strftime(time_str, sizeof(time_str), "%Y-%m-%d %H:%M:%S", timeinfo);
+	printed = fprintf(out, "%s (%s:%d)", time_str, file, line);
+	if (printed < DEBUG_HEADER_SIZE) {
+		fprintf(out, "%*c", 50 - printed, ' ');
+	}
+}
+
+static void log_internal(FILE *out, const char *file, int line, const char *label, const char *format, va_list args_list)
+{
+
+	char msg[LINE_MAX];
+
+	/* do not print the carriage return found at the beginning of the message if
+	 * in debug mode or if printing to a file, that would mess up the output */
+	while (strncmp(format, "\r", 1) == 0) {
+		if (cur_log_level != LOG_DEBUG && isatty(fileno(out))) {
 			fprintf(out, "%c", format[0]);
 		}
 		format++;
 	}
 
+	/* print the line breaks found at the beginning of the message */
+	while (strncmp(format, "\n", 1) == 0) {
+		if (cur_log_level == LOG_DEBUG) {
+			print_debug_info(out, file, line);
+		}
+		fprintf(out, "%c", format[0]);
+		format++;
+	}
+
+	/* if we don't have anything else to print, finish here */
+	if (strcmp(format, "") == 0) {
+		return;
+	}
+
 	// In debug mode, print more information
 	if (cur_log_level == LOG_DEBUG) {
-		time_t currtime;
-		time(&currtime);
-		timeinfo = localtime(&currtime);
-		strftime(time_str, sizeof(time_str), "%Y-%m-%d %H:%M:%S", timeinfo);
-
-		printed = fprintf(out, "%s (%s:%d)", time_str, file, line);
-		if (printed < DEBUG_HEADER_SIZE) {
-			fprintf(out, "%*c", 50 - printed, ' ');
-		}
+		print_debug_info(out, file, line);
 	}
 
 	if (label) {
 		fprintf(out, "%s: ", label);
 	}
-	vfprintf(out, format, args_list);
-	fflush(out);
+
+	/* all debug messages need to finish with a line break, even if the
+	 * user didn't add one to the message, otherwise the output will be
+	 * messed up
+	 * Note that the "\n" could have been provided as an argument to the
+	 * string, as in this example: info("example%s", flag ? "" : "\n")
+	 * so we need to pass the string to a variable first to inspect it */
+	vsnprintf(msg, LINE_MAX, format, args_list);
+	fprintf(out, "%s", msg);
+	if ((cur_log_level == LOG_DEBUG) && (msg[strlen(msg) - 1] != '\n')) {
+		fprintf(out, "\n");
+	}
 }
 
 void log_set_level(int log_level)
--- a/src/lib/progress.c
+++ b/src/lib/progress.c
@@ -113,16 +113,12 @@ void progress_report(double count, double max)
 			if (progress_function) {
 				progress_function(step.description, step.current, step.total, percentage);
 			} else {
-				info("\t...%d%%", percentage);
-				if (!isatty(fileno(stdout))) {
-					/* if not in a tty add new lines so every percentage
-					 * is in its own line */
-					info("\n");
-				} else if (percentage == 100) {
-					/* add a line break once we reach 100% */
-					info("\n");
+				if (isatty(fileno(stdout))) {
+					info("\t...%d%%%s", percentage, percentage != 100 ? "\r" : "\n");
 				} else {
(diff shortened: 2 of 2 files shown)


src/lib/log.c
```diff
@@ -57,12 +66,22 @@ static void log_internal(FILE *out, const char *file, int line, const char *labe
...
+		if (format[strlen(format) - 1] != '\n') {
+			format_end = "\n";
+		}
 	}
 
 	if (label) {
 		fprintf(out, "%s: ", label);
 	}
 	vfprintf(out, format, args_list);
+	fprintf(out, "%s", format_end);
```
otaviobp: I like the idea, but I'd move the if here so you won't need to have another variable for that.

src/lib/log.c
```diff
@@ -57,12 +66,22 @@ static void log_internal(FILE *out, const char *file, int line, const char *labe
...
+
+exit:
 	fflush(out);
```
otaviobp: This was added by a previous patch, but adding a fflush() is usually not a good idea because it can slow down swupd when there's too much logs.
What is the problem to remove this?

src/lib/log.c
castulo: Well the if cannot be moved because the time(file:line) has to be printed before the message. I thought of adding another if (cur_log_level == LOG_DEBUG) {} here but using a variable seemed cleaner. But I can use another if instead if you think is better.

src/lib/log.c
castulo: I don't remember why I added it in a previous patch, I remember when it wasn't there the output was not correctly printed. But we could remove it and see how it goes. And add it back if we see is necessary.

otaviobp: yes. it would be nice to test that

src/lib/log.c
otaviobp: I prefer to have another if. And then we won't call fprintf again.

castulo:
Fixed all the reported issues

src/lib/log.c
```diff
@@ -38,14 +38,22 @@ static void log_internal(FILE *out, const char *file, int line, const char *labe
 	struct tm *timeinfo;
 	int printed;
 
-	/* if the message starts with '\n' add a line break before the message */
+	/* if the message starts with '\n' add a line break before the message
+	 * unless we are printing in debug mode, or to a file because that would
+	 * mess up the output, if that is the case remove those characters */
 	while (strncmp(format, "\n", 1) == 0 || strncmp(format, "\r", 1) == 0) {
-		if (out == stdout || out == stderr) {
+		if ((out == stdout || out == stderr) && (cur_log_level != LOG_DEBUG)) {
```
otaviobp: shouldn't we print '\n' even if we are using --debug?

src/lib/log.c
```diff
@@ -38,14 +38,22 @@ static void log_internal(FILE *out, const char *file, int line, const char *labe
...
 			fprintf(out, "%c", format[0]);
 		}
 		format++;
 	}
 
+	/* if the message was empty or had only line breaks or carriage return
+	 * finish here, otherwise we could print empty debug info or labels */
+	if (strcmp(format, "") == 0) {
```
otaviobp: Is this a real case? Are we using empty prints?

src/lib/log.c
otaviobp: the only command that is printing \r now is the progress report. I'd change it to print \n instead if debug is used.
And stop handling \r here.

castulo: > shouldn't we print '\n' even if we are using --debug?

What I was thinking is that the "\n" at the beginning of the lines are mostly used to space messages between each other to improve the looks of the output. When using --debug you are printing more information, so the output doesn't really need that extra spacing and printing line break would result in debug messages that look like this:
```
2019-07-23 17:46:27 (src/curl.c:570)              Debug: some debug message
2019-07-23 17:46:27 (src/curl.c:438)              Debug: more messages
2019-07-23 17:46:27 (src/curl.c:579)              
2019-07-23 17:46:27 (src/bundle.c:777)            Bla bla bla
```
But maybe it is ok to print those empty lines even in debug mode. What do you think?

castulo: > the only command that is printing \r now is the progress report. I'd change it to print \n instead if debug is used.
> And stop handling \r here.

The problem with handling it there instead of here is that here it would allow for more flexibility. I like the idea of being able to use \n and \r with the logger if necessary so you can format your output as you please. It is not very common that you would need to use the \r thing, but if you do, it would be supported.

src/lib/log.c
castulo: We are not using empty strings, but with this implementation, if someone is printing to a file and the code has for example a `info("\n")`, since the \n is going to be stripped from the message, the resulting message would be an empty string.

src/lib/log.c
otaviobp: I think we should always print something if the user of that lib calls the print function. For example if the user calls info("\n") I prefer to print an empty line than not printing anything. If we are debugging the output looking for a bug we are going to see prints in code that aren't printed on the output. And that makes things harder. Imagine one output like

    2019-07-23 17:46:27 (src/curl.c:570)              Debug: some debug message
    2019-07-23 17:46:27 (src/curl.c:438)              Debug: more messages
    2019-07-23 17:46:27 (src/curl.c:579)              
    Segmentation fault

We know that segfault occurred after that call and not before.

castulo:
> I tested the output and it looks good, with the exception of empty lines. I think we should print them in debug mode. Debug mode isn't suppose to be used unless you are debugging a problem, so it's better to have more than less.
> I'm convinced that it's a good thing to handle the \r on log too.

I'll include those in debug mode then.

src/lib/log.c
```diff
@@ -20,50 +20,88 @@
...
+
+static void log_internal(FILE *out, const char *file, int line, const char *label, const char *format, va_list args_list)
+{
+
+	char msg[LINE_MAX];
+
+	/* do not print the carriage return found at the beginning of the message if
+	 * in debug mode or if printing to a file, that would mess up the output */
+	while (strncmp(format, "\r", 1) == 0) {
+		if (cur_log_level != LOG_DEBUG && (out == stdout || out == stderr)) {
```
otaviobp: I just test this again and this isn't working. '\r' are being outputted to a file. If you redirect the output to a file out is still going to be stdout or stderr. In progress we check if we are outputting to a tty using isatty().
I still prefer to remove this code and handle that only on progress. But as I said I'm ok to leave it here if you prefer. But check for tty instead.

src/lib/log.c
```diff
@@ -20,50 +20,88 @@
...
+		}
+		format++;
+	}
 
-	/* if the message starts with '\n' add a line break before the message */
-	while (strncmp(format, "\n", 1) == 0 || strncmp(format, "\r", 1) == 0) {
+	/* do not print the line breaks found at the beginning of the message
+	 * only if printing to a file */
+	while (strncmp(format, "\n", 1) == 0) {
 		if (out == stdout || out == stderr) {
```
otaviobp: same here. For me I would remove this if

src/lib/log.c
castulo: you are right, I will change this

src/info.c
```diff
@@ -90,6 +90,13 @@ enum swupd_code info_main(int UNUSED_PARAM argc, char UNUSED_PARAM **argv)
 	}
 	progress_init_steps("info", steps_in_info);
 
+	warn("\r");
```
otaviobp: I think you forgot this test here. haha

src/lib/log.c
```diff
@@ -20,50 +20,89 @@
...
-	vfprintf(out, format, args_list);
-	fflush(out);
+
+	/* all debug messages need to finish with a line break, even if the
+	 * user didn't add one to the message, otherwise the output will be
+	 * messed up
+	 * Note that the "\n" could have been provided as an argument to the
+	 * string, as in this example: info("example%s", flag ? "" : "\n")
+	 * so we need to pass the string to a variable first to inspect it */
+	vsnprintf(msg, LINE_MAX, format, args_list);
```
otaviobp: when we replaced vfprintf() with vsnprintf() + fprintf()?

src/info.c
castulo: 😫

src/lib/log.c
otaviobp: oh, I got it. to check the last \n

castulo: I added an explanation why I did that. If we do something like this:
`info("example%s", flag ? "" : "\n")`

We need that so we can inspect the string to see if it already contains a "\n" at the end. If we remove that part we are going to be printing an extra blank line in debug mode in those cases, so it is not a big deal, if you prefer to print the extra line I can remove that and go back to using vfprintf()

otaviobp: no, it's fine the way it was. I didn't get that at first

castulo:
Removed the extra code I was using for testing.

Issue #941, opened 2019-05-31
Swupd mirror with --versionurl and --contenturl can be confusing


otaviobp:
**Describe the bug**
If I run

    sudo swupd mirror --versionurl=https://myserver/ --contenturl=https://myserver

Swupd prints:

    Installed version: xxxxxx
    Version URL:       https://myserver
    Content URL:       https://myserver

And this may look like we are setting the mirror. But to set a mirror we need to use the flags --set.

**Possible solutions**

1. Set the mirror if --versionurl or --contenturl is used.

or

2. Warn the user that --set is required to set the mirror.

We need to evaluate which one is better

dorileo:
I vote up for checking if ```--set``` or ```--unset``` is not present.

PR #1023, opened 2019-07-23
Mirror: fix bug #941

otaviobp:


test/functional/mirror/mirror-set-unset.bats
```diff
@@ -0,0 +1,182 @@
...
+global_teardown() {
+
+	destroy_test_environment "$TEST_NAME"
+
+}
+
+@test "MIR011: Set/Unset a full mirror" {
+
+	run sudo sh -c "$SWUPD mirror --set https://example.com/swupd-file $SWUPD_OPTS"
+	assert_status_is 0
```
castulo: It is better to use $SWUPD_OK

test/functional/mirror/mirror-set-unset.bats
```diff
@@ -0,0 +1,182 @@
...
+
+	expected_output=$(cat <<-EOM
+		Mirror url configuration removed
+		Distribution:      Swupd Test Distro
+		Installed version: 10
+		Version URL:       .*web-dir
+		Content URL:       .*web-dir
+	EOM
+	)
+	assert_regex_is_output "$expected_output"
```
castulo: just as FYI, instead of using:
`Content URL:       .*web-dir` with `assert_regex_is_output "$expected_output"`

you could use:
`Content URL:       $WEBDIR` with `assert_is_output "$expected_output"`

I generally try to avoid regex as possible when comparing the output, it makes it harder to find errors when the test fails, because the line with regex characters is always gonna show up in the diff.

src/mirror.c
```diff
@@ -121,7 +121,6 @@ static int unset_mirror_url()
 	free_string(&content_url);
 	set_default_version_url();
 	set_default_content_url();
-	get_latest_version("");
```
castulo: the call to get_latest_version("") is necessary to reload the defaults in the mirror, after unsetting it... so I do think this belongs to the unset_mirror_url() function.

It is true that at the moment handle_mirror_if_stale() is the only function that really continues doing more stuff after unsetting the mirror, but what if in the future we need to unset the mirror as part of other process, we will have to reload the defaults again.

src/mirror.c
```diff
@@ -191,34 +209,48 @@ static enum swupd_code set_mirror_url(char *url)
...
+		ret = set_new_url(content_url_new, MIRROR_CONTENT_URL_PATH);
+		if (ret != SWUPD_OK) {
+			return ret;
+		}
 	}
 
-	/* if mirror is http, warn user about needing to set
-	 * allow_insecure_http=true in order for auto-update to continue working */
-	if (is_url_insecure(url)) {
+	if (is_url_insecure(content_url_new) || is_url_insecure(version_url_new)) {
```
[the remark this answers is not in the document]
otaviobp: No. if url_new is NULL we use the value set to content_url/version_url, so they can be different. We are supporting now:

    swupd mirror --set --content_url https://my_url

test/functional/mirror/mirror-set-unset.bats
otaviobp: $WEBDIR is pointinting to web-dir directory but I need the full url. Is there any variable exporting that value? The url is something like:

    file:///home/otaviobp/swupd-client/mirror-set-unset_1/web-dir

src/mirror.c
otaviobp: yes. I didn't have a better idea for now. I'd be more concerned if this function was exported, but as it's a local function we have more control on where it's used.

test/functional/mirror/mirror-set-unset.bats
castulo: ohh, in that case I think you can use `$TEST_DIRNAME/web-dir`

src/mirror.c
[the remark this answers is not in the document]
castulo: you are right

otaviobp:
Fixed proposed changes

PR #1027, opened 2019-07-26
Changes to avoid misuse of "swupd verify -B"

### LINKED ISSUE #981: swupd repair --picky -B bundle deletes critical system files
**Describe the bug**
As the title says, running `sudo swupd repair --picky -B <bundle>` deletes system files that should clearly not be deleted and definitely do not belong to the bundle specified. I am no longer able to boot the system, however, `/home` is still intact, as well as `/opt`.

**To Reproduce**
I first ran `sudo swupd repair -B vlc` which reported 14 or so files that should not be there, but could not be removed. Since the `picky` option removes files that should not exist, I then ran `sudo swupd repair --picky -B vlc`, which deleted literally the complete system.

**Expected behavior**
I expected the 14 files to be removed that were previously reported to be incorrect.

As I am no longer able to boot the system, I can't run `swupd info`, but the system had just auto-updated and therefore was probably on the latest version.
 - Platform: bare metal (installed on usb)

**EDIT**
Apparently, files were removed from `/bin`, `/lib`, `/lib64`, `/sbin`, and `/usr`. `/bin` and `/sbin` contain only vlc related files.

castulo:
`swupd verify -B` is a confusing option which is often misused by users which end up deleting system files they did not intended to. The `-B` option was mainly designed to work along with `--install` to install a Clear OS that included some bundles, but the flag was never restricted to also work with other flags like `--fix` or `--picky`. If used carefully, `-B` has its use case when combined with `--fix` or `--picky`, but since the option is often misunderstood it causes confusion with users which end up doing unintended actions on their system. To mitigate this problem this PR implements the following actions:
- Removes the `-B`/`--bundle` option from `swupd diagnose` and `swupd repair`, so it is only available at `os-install` where it is safe to use it. It is also left available at `verify` for backward compatibility, but `verify` in general has been superseded and we discourage users from using `verify`.
- Changes the `verify` help to try to make it clearer what the `-B` option is for. 
- Limits the use of `-B` within `swupd verify` so it can only be used along with the `--fix` flag (and the other flags supported with --fix).
- When using `-B` a warning is printed explaining the possible side effects of using the `-B` option and the other flags used with it.
- Updates the man documentation to reflect this changes.

Closes #981


castulo:
Rebased it

lebensterben:
There's one problem if this is removed, in [tutorial on installing NVDIA driver](
https://github.com/clearlinux/clear-linux-documentation/blob/master/source/clear-linux/tutorials/nvidia.rst), there's a command `swupd repair --quick --bundles=lib-opengl`.

Since this is removed, I'd like to know whether there is any alternative methods to perform the same thing.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

castulo:
we are going to add it back to diagnose and repair as soon as we implement it in a safe way. Check #1035

PR #1031, opened 2019-07-26
Multiple updates to the man pages

### LINKED ISSUE #1020: Add information about where the config file is in swupd man page
We don't have information on where swupd looks for its config file in the man page.
We can also print that in the allow_insecure_http output

castulo:
This PR includes multiple updates to the man page including explanation about the config file.

Closes #1020


docs/swupd.1.rst
```diff
@@ -566,6 +627,43 @@ SUBCOMMANDS
...
+    swupd command except for that one overwritten in the command section.
+
+    A sample swupd configuration file can be found at this location (this file should not
+    be modified):
+    /usr/share/defaults/swupd
+
+    To use it, it should be copied to one of these locations where swupd reads the
+    configuration from (in that order):
+    - /etc/swupd
+    - /run/swupd
```
otaviobp: /run is a tmpfs mounted fs on Clear so it's not a good place for a user to copy the config file. We read the config from /run, but shouldn't be used unless it's something temporary. we added that more for compatibility reasons.

castulo: I got those files from this comment -> https://github.com/clearlinux/swupd-client/pull/958#issuecomment-504493148, so do you think we should only use /etc/swupd there?

otaviobp: I think so. I wouldn't recommend users to copy config files to /run

PR #1041, opened 2019-08-01
Enabling --bundles for diagnose and repair

### LINKED ISSUE #1035: Implementing an option for diagnosing and repairing specific bundles safely
The `os-install` command (and the legacy command `verify --install`) have a `-B`/`--bundles` option that can be used to install a Clear Linux OS that includes the list of bundles specified with that option.

Users have been requesting a similar option that can be used with `diagnose` and `repair` to repair only specific bundles.

This option needs to be implemented in a way that it doesn't mess up other installed bundles in the system.

Related to #981

castulo:
The PR addresses the following things:
- This PR enables the `-B` / `--bundles` option to diagnose or repair only the specified list of bundles in a safe way.
- When using `--bundles` and a file is being marked as deleted for one of the specified bundles, it makes sure it is not needed by any of the other installed bundles before deleting it.
- Disallow the use of `--bundles` together with `--picky` or `--extra-files-only` for `diagnose` or `repair`. That combination of options is only allowed for `verify` for backward compatibility.

Closes #1035

### DIFF
--- a/docs/swupd.1.rst
+++ b/docs/swupd.1.rst
@@ -437,6 +437,17 @@ SUBCOMMANDS
         Like --picky, but it only removes extra files. It omits checking
         hash values, and for missing files, directories and/or symlinks.
 
+    - `--bundles`
+
+      Forces swupd to only repair the (comma separated) list of bundles
+      provided.
+
+      Examples:
+
+        - ``--bundles os-core,vi``
+
+            Repairs only bundles os-core and vi.
+
 ``search``
 
     Swupd search functionality is provided by swupd-search binary available
--- a/docs/swupd.1.rst
+++ b/docs/swupd.1.rst
@@ -303,6 +303,17 @@ SUBCOMMANDS
         Like --picky, but it only looks for extra files. It omits checking
         hash values, and for missing files, directories and/or symlinks.
 
+    - `--bundles`
+
+      Forces swupd to only diagnose the (comma separated) list of bundles
+      provided.
+
+      Examples:
+
+        - ``--bundles os-core,vi``
+
+            Diagnoses only bundles os-core and vi.
+
 ``hashdump {path}``
 
     Calculates and print the Manifest hash for a specific file on disk.
--- a/src/fullfile.c
+++ b/src/fullfile.c
@@ -30,6 +30,26 @@
 //FIXME #562
 #define MAX_XFER 15
 
+static int compare_fullfile(const void *a, const void *b)
+{
+	struct file *file1 = (struct file *)a;
+	struct file *file2 = (struct file *)b;
+	int comp;
+
+	comp = strcmp(file1->hash, file2->hash);
+	if (comp != 0) {
+		return comp;
+	}
+
+	/* they have the same hash, now let's check the version */
+	if (file1->last_change == file2->last_change) {
+		/* we have a match */
+		return 0;
+	}
+
+	return -1;
+}
+
 static void download_mix_file(struct file *file)
 {
 	char *url, *filename;
@@ -197,22 +217,11 @@ int download_fullfiles(struct list *files, int *num_downloads)
 	if (list_length < MAX_FILES) {
 		/* remove tar duplicates from the list first */
 		need_download = list_sort(need_download, file_sort_hash);
-		for (iter = list_head(need_download); iter; iter = iter->next) {
-			if (iter->next == NULL) {
-				break;
-			}
-			struct file *file1, *file2;
-			file1 = iter->data;
-			file2 = iter->next->data;
 
-			/* different directories may need the same tar, in those cases it needs
-			 * to be downloaded only once, the tar for each file is downloaded from
-			 * <last_change>/files/<hash>.tar */
-			if (strcmp(file1->hash, file2->hash) == 0 && (file1->last_change == file2->last_change)) {
-				iter = list_free_item(iter, NULL);
-			}
-		}
-		need_download = list_head(iter);
+		/* different directories may need the same tar, in those cases it needs
+		 * to be downloaded only once, the tar for each file is downloaded from
+		 * <last_change>/files/<hash>.tar */
+		need_download = list_deduplicate(need_download, compare_fullfile, NULL);
 
 		download_progress.total_download_size = fullfile_query_total_download_size(need_download);
 		if (download_progress.total_download_size > 0) {
--- a/src/lib/list.c
+++ b/src/lib/list.c
@@ -338,3 +338,46 @@ struct list *list_deduplicate(struct list *list, comparison_fn_t comparison_fn, 
 
 	return list;
 }
+
+struct list *list_filter_common_elements(struct list *list1, struct list *list2, comparison_fn_t comparison_fn, list_free_data_fn_t list_free_data_fn)
+{
+	struct list *iter1, *iter2 = NULL;
+	struct list *preserver = NULL;
+	void *item1, *item2 = NULL;
+	int comp;
+
+	preserver = iter1 = list_head(list1);
+	iter2 = list_head(list2);
+
+	while (iter1 && iter2) {
+
+		item1 = iter1->data;
+		item2 = iter2->data;
+
+		/* if the criteria has not been met, there is nothing to be done,
+		 * just move the pointers to the next elements appropriately */
+		comp = comparison_fn(item1, item2);
+		if (comp < 0) {
+			iter1 = iter1->next;
+			continue;
+		} else if (comp > 0) {
+			iter2 = iter2->next;
+			continue;
+		}
+
+		/* the condition was met */
+
+		/* preserver was pointing to the first element of
+		 * the list, if we are removing it, redirect it so it
+		 * points to the new first element of the list */
+		if (preserver == iter1) {
+			preserver = iter1->next;
+		}
+		iter1 = list_free_item(iter1, list_free_data_fn);
+		/* the iter1 pointer was already moved by list_free_item
+		 * so only move iter2 to the next element in the list */
+		iter2 = iter2->next;
+	}
+
+	return preserver;
+}
--- a/src/lib/list.c
+++ b/src/lib/list.c
@@ -318,3 +318,23 @@ int list_strcmp(const void *a, const void *b)
 {
 	return strcmp((const char *)a, (const char *)b);
 }
+
+struct list *list_deduplicate(struct list *list, comparison_fn_t comparison_fn, list_free_data_fn_t list_free_data_fn)
+{
+	struct list *iter = NULL;
+	void *item1, *item2;
+
+	list = list_head(list);
+
+	for (iter = list; iter && iter->next; iter = iter->next) {
+
+		item1 = iter->data;
+		item2 = iter->next->data;
+
+		if (comparison_fn(item1, item2) == 0) {
+			list_free_item(iter->next, list_free_data_fn);
+		}
+	}
+
+	return list;
+}
--- a/src/repair.c
+++ b/src/repair.c
@@ -35,7 +35,8 @@ static bool cmdline_option_quick = false;
 static int cmdline_option_version = 0;
 static char *cmdline_option_picky_tree = NULL;
 static const char *cmdline_option_picky_whitelist = picky_whitelist_default;
-static bool cmdline_extra_files_only = false;
+static bool cmdline_option_extra_files_only = false;
+static struct list *cmdline_bundles = NULL;
 
 /* picky_whitelist points to picky_whitelist_buffer if and only if regcomp() was called for it */
 static regex_t *picky_whitelist;
@@ -50,6 +51,7 @@ static const struct option prog_opts[] = {
 	{ "picky-whitelist", required_argument, 0, 'w' },
 	{ "quick", no_argument, 0, 'q' },
 	{ "extra-files-only", no_argument, 0, FLAG_EXTRA_FILES_ONLY },
+	{ "bundles", required_argument, 0, 'B' },
 };
 
 static void print_help(void)
@@ -63,6 +65,7 @@ static void print_help(void)
 	print("   -V, --version=[VER]     Compare against version VER to repair\n");
 	print("   -x, --force             Attempt to proceed even if non-critical errors found\n");
 	print("   -q, --quick             Don't compare hashes, only fix missing files\n");
+	print("   -B, --bundles=[BUNDLES] Forces swupd to only repair the specified BUNDLES. Example: --bundles=os-core,vi\n");
 	print("   -Y, --picky             Also remove files which should not exist\n");
 	print("   -X, --picky-tree=[PATH] Selects the sub-tree where --picky and --extra-files-only  looks for extra files. Default: /usr\n");
 	print("   -w, --picky-whitelist=[RE] Directories that match the regex get skipped. Example: /var|/etc/machine-id\n");
@@ -105,7 +108,19 @@ static bool parse_opt(int opt, char *optarg)
 		cmdline_option_picky_whitelist = strdup_or_die(optarg);
 		return true;
 	case FLAG_EXTRA_FILES_ONLY:
-		cmdline_extra_files_only = optarg_to_bool(optarg);
+		cmdline_option_extra_files_only = optarg_to_bool(optarg);
+		return true;
+	case 'B':
+		/* if we are parsing a list from the command line we don't want to append it to
+		 * a possible existing list parsed from a config file, we want to replace it, so
+		 * we need to delete the existing list first */
+		list_free_list(cmdline_bundles);
+		cmdline_bundles = NULL;
+		cmdline_bundles = string_split(",", optarg);
+		if (!cmdline_bundles) {
+			error("Missing required --bundles argument\n\n");
+			return false;
+		}
 		return true;
 	default:
 		return false;
@@ -139,14 +154,25 @@ static bool parse_options(int argc, char **argv)
 			error("--quick and --picky options are mutually exclusive\n");
 			return false;
 		}
-		if (cmdline_extra_files_only) {
+		if (cmdline_option_extra_files_only) {
 			error("--quick and --extra-files-only options are mutually exclusive\n");
 			return false;
 		}
 	}
-	if (cmdline_extra_files_only) {
+	if (cmdline_option_extra_files_only) {
 		if (cmdline_option_picky) {
 			error("--extra-files-only and --picky options are mutually exclusive\n");
+			return false;
+		}
+	}
+
+	if (cmdline_bundles) {
+		if (cmdline_option_picky) {
+			error("--bundles and --picky options are mutually exclusive\n");
+			return false;
+		}
+		if (cmdline_option_extra_files_only) {
+			error("--bundles and --extra-files-only options are mutually exclusive\n");
 			return false;
 		}
 	}
@@ -202,7 +228,8 @@ enum swupd_code repair_main(int argc, char **argv)
 	verify_set_option_version(cmdline_option_version);
 	verify_set_picky_whitelist(picky_whitelist);
 	verify_set_picky_tree(cmdline_option_picky_tree);
-	verify_set_extra_files_only(cmdline_extra_files_only);
+	verify_set_extra_files_only(cmdline_option_extra_files_only);
+	verify_set_option_bundles(cmdline_bundles);
 
 	/* run verify --fix */
 	ret = verify_main();
--- a/src/update.c
+++ b/src/update.c
@@ -411,8 +411,7 @@ version_check:
 	}
 
 	/* consolidate the current collective manifests down into one in memory */
-	current_manifest->files = files_from_bundles(current_manifest->submanifests);
-	current_manifest->files = consolidate_files(current_manifest->files);
+	current_manifest->files = consolidate_files_from_bundles(current_manifest->submanifests);
 	latest_subs = list_clone(current_subs);
 	set_subscription_versions(server_manifest, current_manifest, &latest_subs);
 	link_submanifests(current_manifest, server_manifest, current_subs, latest_subs, false);
@@ -442,8 +441,7 @@ version_check:
 	}
 
 	/* consolidate the new collective manifests down into one in memory */
-	server_manifest->files = files_from_bundles(server_manifest->submanifests);
-	server_manifest->files = consolidate_files(server_manifest->files);
+	server_manifest->files = consolidate_files_from_bundles(server_manifest->submanifests);
 	set_subscription_versions(server_manifest, current_manifest, &latest_subs);
 	link_submanifests(current_manifest, server_manifest, current_subs, latest_subs, true);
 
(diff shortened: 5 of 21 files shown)


src/lib/log.c
```diff
@@ -35,6 +35,12 @@
 static void log_internal(FILE *out, const char *file, int line, const char *label, const char *format, va_list args_list);
 static int cur_log_level = LOG_INFO;
 static log_fn_t log_function = log_internal;
+static bool mute = false;
+
+void log_mute(bool state)
```
otaviobp: I don't like the idea on supporting a log mute. This looks like a problem on how the log is called and that the function that is using it could be improve to print log correctly.

src/verify.c
```diff
@@ -930,7 +1001,16 @@ enum swupd_code verify_main(void)
...
 
-	ret = add_included_manifests(official_manifest, &subs);
+	ret = add_included_manifests(official_manifest, &all_subs);
+	if (cmdline_option_bundles) {
+		/* when using --bundles, the subs lists will be browsed
+		 * and validated twice, once for all subs and once for the
+		 * specific bundles subs, to avoid printing double warning
+		 * we need to mute the log during those operations */
+		log_mute(true);
+		ret = add_included_manifests(official_manifest, &bundles_subs);
```
otaviobp: what real logging are you trying to avoid here?

castulo: add_included_manifests calls add_subscriptions which is going to print this warning if a bundle is not found in the mom:
`warn("Bundle \"%s\" is invalid, skipping it...\n", bundle);`

So if we use a command like `swupd diagnose --bundles valid_bundle,invalid_bundle`

We are going to get this message twice:
```
Bundle "invalid_bundle" is invalid, skipping it...
Bundle "invalid_bundle" is invalid, skipping it...
```

So in some cases I think is useful to call a function without printing any of its output, specially if you are already printing something similar at a higher level, so there is where muting the log comes handy. Is like having a parameter for the function to enable it to run silently.

castulo:
@otaviobp I figured out a way of avoiding printing the double warning without muting the log. It should be fine now.

src/repair.c
```diff
@@ -105,8 +108,22 @@ static bool parse_opt(int opt, char *optarg)
 		cmdline_option_picky_whitelist = strdup_or_die(optarg);
 		return true;
 	case FLAG_EXTRA_FILES_ONLY:
-		cmdline_extra_files_only = optarg_to_bool(optarg);
+		cmdline_option_extra_files_only = optarg_to_bool(optarg);
 		return true;
+	case 'B': {
+		char *arg_copy = strdup_or_die(optarg);
+		char *token = strtok(arg_copy, ",");
```
otaviobp: Don't duplicate code here. We can do it in a function. 
And I'm unsure about this implementation. In the case we set the bundles property in a config file it will be concatenated with the bundles from the command line. Is this expected?

otaviobp:
I also found a memory leak when I run:

    $ swupd clean
    $ swupd repair -B vim

src/repair.c
castulo: hmm didn't think of that, I don't think we should concatenate them , they should be overwritten

castulo: nice catch by the way

src/verify.c
```diff
@@ -693,19 +694,11 @@ static bool parse_options(int argc, char **argv)
 		}
 
 		if (cmdline_option_bundles) {
```
otaviobp: Instead of having an extra variable can't we just have a check if bundles != NULL?

src/globals.c
```diff
@@ -688,3 +688,19 @@ int global_parse_options(int argc, char **argv, const struct global_options *opt
...
+bool global_parse_csv(const char *args, struct list **parsed_args)
+{
+	char *arg_copy = strdup_or_die(args);
+	char *token = strtok(arg_copy, ",");
+	while (token) {
+		*parsed_args = list_prepend_data(*parsed_args, strdup_or_die(token));
+		token = strtok(NULL, ",");
+	}
+	free(arg_copy);
+	if (!*parsed_args) {
```
otaviobp: if global_parse_args() is always returning if *parsed_args is NULL or not we can just return that value. So we can have:

    struct list *global_parse_csv(const char *args);
    ...
    my_list = global_parse_csv(my_string);
    if (my_list) {
         //it's ok
    } else {
        print("error");
    }

And then you can even move that function to string.h/string.c so it can be reused by other code. string_parse_csv(), or something like that.

otaviobp: it's the inversion of string_join(). To make it more generic you can even get the "," as argument too

src/verify.c
```diff
@@ -755,6 +758,43 @@ static enum swupd_code deal_with_extra_files(struct manifest *manifest, bool fix
...
+		next = bundles_files->next;
+		file1 = bundles_files->data;
+		/* bundles_list is a subset of the all_files list
+		 * and both lists should be already sorted, so we
+		 * can continue the search for the next file where
+		 * we left off from the previous file */
+		for (; all_files; all_files = all_files->next) {
+			file2 = all_files->data;
+			if (strcmp(file2->filename, file1->filename) == 0) {
+				sanitized_files = list_append_data(sanitized_files, file2);
```
otaviobp: In this function you are adding some elements of list 'all_files' to the sanitized_files list. By doing that you don't know when you are freeing the all_files list which element you should free and which you shouldn't. In current code you aren't cleaning any element, so we are having memory leaks.

One solution would be to duplicate all elements that you are appending to the sanitized_files. It's the easiest one, but not ideal. Instead, it looks like you aren't using all_files anywhere else after calling this function. So instead of creating a new list with the files that are 'sanitized' you can filter the all_files list removing the files that doesn't make sense to be there (and freeing the data). So we don't need to duplicate the list and we know who to free at the end.

src/verify.c
castulo: not with the current implementation because at some point we remove everything from bundles. We could make some changes so we don't do this.

castulo:
WIP

src/verify.c
castulo: done

src/verify.c
```diff
@@ -755,6 +758,43 @@ static enum swupd_code deal_with_extra_files(struct manifest *manifest, bool fix
...
+			}
+		}
+	found:
+		if (next == NULL) {
+			break;
+		}
+		bundles_files = next;
+	}
+
+	return sanitized_files;
```
otaviobp: you are returning just last element of the list here

src/verify.c
otaviobp: Well, actually the data from all_files aren't cleared too. Because this files are kept in bundles and are all freed over there. So there's no need to be concerned about memory leaks here. My comment was incorrect.
But I'm still concerned about this function being O(n^2) while it could be O(n).

castulo:
@otaviobp this is still a WIP, I'm going to modify the `sanitize_files` so it filters files from all_files instead of creating a new list.

src/verify.c
```diff
@@ -621,25 +621,18 @@ static bool parse_opt(int opt, char *optarg)
...
-			cmdline_option_bundles = list_prepend_data(cmdline_option_bundles, strdup_or_die(token));
-			token = strtok(NULL, ",");
-		}
-		free(arg_copy);
+	case 'B':
+		/* if we are parsing a list from the command line we don't want to append it to
+		 * a possible existing list parsed from a config file, we want to replace it, so
+		 * we need to delete the existing list first */
+		list_free_list(cmdline_option_bundles);
+		cmdline_option_bundles = NULL;
```
otaviobp: You're setting the same variable twice. you don't need this line

src/verify.c
```diff
@@ -993,20 +1089,42 @@ enum swupd_code verify_main(void)
...
+	all_files = consolidate_files(all_files);
+	official_manifest->files = all_files;
+	if (cmdline_option_bundles) {
+		bundles_files = files_from_bundles(bundles_submanifests);
+		bundles_files = consolidate_files(bundles_files);
+		/* we need to make sure the files from the selected bundles
+		 * are compared against the full list of consolidated files
+		 * from all bundles so we don't end up deleting a file that
+		 * is needed by another bundle */
+		sanitized_files = sanitize_files(bundles_files, all_files);
```
otaviobp: You don't use the variable sanitized_files anywhere else so you don't need to keep it.

src/verify.c
```diff
@@ -1164,7 +1282,18 @@ enum swupd_code verify_main(void)
 	/* this concludes the critical section, after this point it's clean up time, the disk content is finished and final */
 
 clean_and_exit:
+	if (official_manifest) {
+		official_manifest->files = all_files;
```
otaviobp: why do you need to recover official_manifest->files at the end of the function? Can't you free all_files when you don't need it anymore?

src/verify.c
```diff
@@ -959,28 +1004,79 @@ enum swupd_code verify_main(void)
...
 	 * continue only if --force was used since the bundles could be removed */
-	if (ret) {
-		if (ret & add_sub_BADNAME) {
+	if (ret || invalid_bundle) {
+		if ((ret & add_sub_BADNAME) || invalid_bundle) {
 			if (cmdline_option_force) {
 				if (cmdline_option_picky && cmdline_option_fix) {
-					warn("One or more installed bundles that are not "
+					warn("\nOne or more installed bundles that are not "
 					     "available at version %d will be removed\n",
```
otaviobp: We're not removing a bundle if the bundle name is incorrect.

src/verify.c
```diff
@@ -959,28 +1004,79 @@ enum swupd_code verify_main(void)
 		}
 	}
 
-	ret = add_included_manifests(official_manifest, &subs);
+	/* get the list of subscribed (installed) bundles */
+	read_subscriptions(&all_subs);
+	selected_subs = all_subs;
+	if (cmdline_option_bundles) {
```
otaviobp: you can group this block in a function to validate the bundle names and print all needed messages returning just a boolean to say if we should continue or abort.
we can free any resource that we don't need anymore and keep all tmp variables inside the function

src/verify.c
```diff
@@ -781,9 +826,18 @@ enum swupd_code verify_main(void)
 {
 	struct manifest *official_manifest = NULL;
 	int ret;
-	struct list *subs = NULL;
+	struct list *all_subs = NULL;
```
otaviobp: Do we really need that many pointers?
Instead of saving everything can't we free data when we don't need anymore? And group tasks into functions so we can keep some variables local

src/verify.c
```diff
@@ -993,20 +1089,42 @@ enum swupd_code verify_main(void)
...
+		official_manifest->submanifests = bundles_submanifests;
+	}
 	progress_complete_step();
 	timelist_timer_stop(globals.global_times); // closing: Load manifests
 
 	timelist_timer_start(globals.global_times, "Consolidate files from bundles");
 	progress_set_step(4, "consolidate_files");
-	official_manifest->files = files_from_bundles(official_manifest->submanifests);
-	official_manifest->files = consolidate_files(official_manifest->files);
+	all_files = files_from_bundles(all_submanifests);
```
otaviobp: this block of calls files_from_bundles() + consolidate_files() can be consolidated in a function like consolidated_files_from_bundles().

src/verify.c
otaviobp: And the only reason we keep the bundle_files is to free that later. We can change sanitize files to be a filter that frees the items we don't need anymore. So we don't need to keep the pointer to free it later. And probably we can free all_files here too.

otaviobp: I would try to simplify this a bit like:

    official_manifest->files = ...
    if (cmdline_option_bundles) {
        bundles_files = consolidated_files_from_bundles(bundles_submanifests);
        //get needed items from all_files and free the ones we don't need. Don't change the bundle_files list.
        official_manifest->files = filter_sanitize_files(official_manifest->files, bundle_files);
        free_list(bundle_files);
    }

src/verify.c
```diff
@@ -770,6 +770,51 @@ static enum swupd_code deal_with_extra_files(struct manifest *manifest, bool fix
...
+	struct file *filea, *fileb;
+
+	/* the list of all_files will always include all files in bundles_files
+	 * and both lists are sorted by filename */
+	while (list_all) {
+
+		if (!list_bundles) {
+			/* we have reached the last element of the list of bundle files
+			 * no need to keep going */
+			return list_head(sanitized);
```
otaviobp: It's better just to save the first element of the list to return so we don't need to run through the list again to find the first element.

src/verify.c
```diff
@@ -770,6 +770,51 @@ static enum swupd_code deal_with_extra_files(struct manifest *manifest, bool fix
 	return ret;
 }
 
+static struct list *sanitize_files(struct list *bundles_files, struct list *all_files)
```
otaviobp: That's another function that is a candidate to be a generic list function.

    // return a list with the elements of listA that are present in listB
    // listA and listB are unchanged
    struct list *filter_list(struct list *listA, struct list *listB, comp_fn_t comp_fn);

So you can use it like this:

    listC = filter_list(listA, listB, cmp_fn);
    free_list_and_data(listB, free_fn);
    // and just use listC
    free_list_and_data(listA, free_fn);
    free_list(listC, free_fn);

Or you can implement it in a way that it changes the listA and returns it. But in this case you need to free elements and data that aren't used:

    void filter_list(struct list *listA, struct list *listB, comp_fn_t comp_fn, list_free_data_fn_t free_fn);
    filter_list(listA, listB, cmp_fn, free_fn);
    free_list_and_data(listB, free_fn);
    // and just use listA
    free_list_and_data(listA, free_fn);

I think I prefer the second (I suggested that in other comments too) but the first one is good too.

otaviobp:
I merged one commit "Don't allow --fix or --install with diagnose" that is independent from other commits and it was good to go

src/fullfile.c
```diff
@@ -210,6 +210,12 @@ int download_fullfiles(struct list *files, int *num_downloads)
 			 * <last_change>/files/<hash>.tar */
 			if (strcmp(file1->hash, file2->hash) == 0 && (file1->last_change == file2->last_change)) {
 				iter = list_free_item(iter, NULL);
+				/* if we removed an element, let's make sure we haven't reached the
```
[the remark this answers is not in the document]
castulo: I like it!

src/update.c
```diff
@@ -193,14 +193,12 @@ int add_included_manifests(struct manifest *mom, struct list **subs)
 	/* Pass the current version here, not the new, otherwise we will never
 	 * hit the Manifest delta path. */
 	ret = add_subscriptions(subbed, subs, mom, true, 0);
+	list_free_list(subbed);
 	if (ret & (add_sub_ERR | add_sub_BADNAME)) {
-		ret = -ret;
+		return ret;
 	} else {
```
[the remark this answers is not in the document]
castulo: ups, I missed that.

src/verify.c
castulo: haha, dah! ☺️

src/verify.c
castulo: when this condition happens it means we found a bundle that was available in an older MoM but is no longer available in the new MoM, it is not really an incorrect bundle name. So this warning is to let the user know that the bundle is going to be removed if he/she --force the update using --picky.

castulo: btw, I didn't implement that part in this PR, I'm just making the output prettier

src/verify.c
castulo: good idea

src/verify.c
castulo: I think the purpose of this function is being misunderstood. It's purpose is not to get the elements of listA that are present in listB, its purpose is only to make sure the files in listA marked as deleted, are not needed by another installed bundle. I don't see it being useful as a generic function.

otaviobp: this function does a loop in the list, applies a condition and if condition is true we add that element to the returned list. The condition in this case is what you described. And this code is repeated several times in swupd with different implementations. I suggested having a generic list function (similar to python filter https://docs.python.org/3/library/functions.html#filter) to execute that.

otaviobp: Sorry, I pointed the wrong function. Not sure if python lists have one implementation on what I thought, but I know we could use that in filter_out_existing_files() too

castulo: I think I found a way to make it generic, I'll implement that

src/verify.c
castulo: Nice, I'll do that

src/verify.c
castulo: Done

src/verify.c
castulo: this won't be needed in the new implementation

src/verify.c
castulo: it is not that easy in this case, the loop not only validate the names of the bundles provided but it also creates the subscription in the appropriate list, so I could extract this block t a function but the function would be like:
`validate_provided_bundles(cmdline_option_bundles, official_manifest, all_subs, bundles_subs);`
I don't see much benefit out of it, we would still need to keep all variables in the verify_main function and we won't free any resource at this point anyway.
Ohh and the function could not return a bool, since it'd need to return 2 values:
- the boolean go/no go result
- a boolean value to indicate there was at least one invalid bundle specified by the user, we need that value later in the code

Bottom line I don't see this saving any resources, or even making the code more readable, IMO it just makes it more complex

src/verify.c
castulo: It is hard to reduce the number of pointers in this case, we could do it if we try hard but it would make the code even harder to read.

This verify_main function has many possible paths, so it is already a confusing function, and the main reason for this is that it includes the paths from 4 different commands, repair, diagnose and os-install, and the deprecated verify.

Maybe we should really consider splitting this function into many smaller functions so each of those commands can have their own <command>_main function instead and consume all the pieces performed in this function.

I made changes to the code in the last update to free the data as soon as we are done with it though, so at least that was improved.

castulo: > It is hard to reduce the number of pointers in this case, we could do it if we try hard but it would make the code even harder to read.
> 
> This verify_main function has many possible paths, so it is already a confusing function, and the main reason for this is that it includes the paths from 4 different commands, repair, diagnose and os-install, and the deprecated verify.
> 
> Maybe we should really consider splitting this function into many smaller functions so each of those commands can have their own _main function instead and consume all the pieces performed in this function.
> 
> I made changes to the code in the last update to free the data as soon as we are done with it though, so at least that was improved.

@otaviobp I think that except from this ^ I finished addressing all your comments in this last update. Let me know what you think.

src/lib/list.c
```diff
@@ -335,3 +335,64 @@ struct list *list_deduplicate(struct list *list, comparison_fn_t comparison_fn,
...
 }
+
+struct list *list_filter_common_elements(struct list *list1, struct list *list2, comparison_fn_t comparison_fn, list_free_data_fn_t list_free_data_fn)
+{
+	/* is mandatory that list1 and list2 are sorted */
+	struct list *iter1, *iter2 = NULL;
+	struct list *preserver = NULL;
+	void *item1, *item2 = NULL;
+	int comp;
+	enum comparison_results {
```
otaviobp: We can't have a comparison_fn_f function with different meanings depending on the list function that uses it. We need to be consistent. I'm still trying to understand why we need 4 different return codes and not just 3.

src/verify.c
```diff
@@ -770,6 +770,29 @@ static enum swupd_code deal_with_extra_files(struct manifest *manifest, bool fix
 	return ret;
 }
 
+static int find_unsafe_to_delete(const void *a, const void *b)
```
otaviobp: Can you check if this works for you?

        static int find_unsafe_to_delete(const void *a, const void *b)
        {
                struct file *A, *B;
                int ret;
        
                A = (struct file *)a;
                B = (struct file *)b;
        
                ret = strcmp(A->filename, B->filename);
                if (ret) {
                        return ret < 0 ? -1 : 1;
                }
        
                if (!A->is_deleted) {
                        return -1;
                }
        
                if (B->is_deleted) {
                        return 1;
                }
        
                return 0;
        }

castulo: 🤔 I think I get your idea now... so when both lists are pointing to a matched element, but the condition we want is not met (in this case `A->is_deleted && !B->is_deleted`), we don't necessarily have to move both pointers to the next element, we can just move either one to the next element, and that will force the other pointer to move forward also in the next loop cycle. That way you still only need 3 return codes and not 4, very clever. I couldn't think of a way of making it work without needing 4 RCs. I'll give this a try! 😎

src/lib/list.c
```diff
@@ -318,3 +318,20 @@ int list_strcmp(const void *a, const void *b)
...
+struct list *list_deduplicate(struct list *list, comparison_fn_t comparison_fn, list_free_data_fn_t list_free_data_fn)
+{
+	struct list *iter = NULL;
+	void *item1, *item2;
+	list = list_head(list);
+
+	for (iter = list; iter && iter->next; iter = iter->next) {
+		item1 = iter->data;
+		item2 = iter->next->data;
+		if (comparison_fn(item1, item2)) {
```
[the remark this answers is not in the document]
castulo: > btw, that's somthing worth adding to the comment too. in comparison_fn_t definition.

I'll do that

castulo:
@otaviobp here is a new update. It is ready to be reviewed again.

src/fullfile.c
```diff
@@ -30,6 +30,14 @@
 //FIXME #562
 #define MAX_XFER 15
 
+static int list_fullfilecmp(const void *a, const void *b)
+{
+	struct file *file1 = (struct file *)a;
+	struct file *file2 = (struct file *)b;
+
+	return (strcmp(file1->hash, file2->hash) == 0 && (file1->last_change == file2->last_change));
```
karthikprabhuvinod: is it enough to do one? preferrrably just hash match?

src/lib/list.c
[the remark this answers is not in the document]
otaviobp: this is still pending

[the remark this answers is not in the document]
castulo: what is?

docs/swupd.1.rst
```diff
@@ -303,6 +303,17 @@ SUBCOMMANDS
...
+    - `--bundles`
+
+      Forces swupd to only diagnose the (comma separated) list of bundles
+      provided.
+
+      Examples:
+
+        - ``--bundles xterm,vi``
+
+            Diagnoses only bundles os-core and vi.
```
karthikprabhuvinod: vi and xterm?

docs/swupd.1.rst
```diff
@@ -426,6 +437,17 @@ SUBCOMMANDS
...
         hash values, and for missing files, directories and/or symlinks.
 
+    - `--bundles`
+
+      Forces swupd to only repair the (comma separated) list of bundles
+      provided.
+
+      Examples:
+
+        - ``--bundles xterm,vi``
```
karthikprabhuvinod: same as above?

src/lib/list.c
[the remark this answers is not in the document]
castulo: done

docs/swupd.1.rst
castulo: 🤣 good catch @karthikprabhuvinod , it should be xterm and vi

docs/swupd.1.rst
castulo: fixed

castulo:
> nice! looks good now

Finally! 🕺

PR #1042, opened 2019-08-02
Fixing a bug using ignore-ilst files in tests

castulo:
When using testlib to create tests, ignore lists can be used which
include lines of output that should be ignored when running tests.

The ignore lists can exist at three different levels:
- an ignore file that applies to all tests
- an ignore file that applies to all files of one theme
- an ignore file that applies to all tests within the same file

There was a bug in this last one that was causing ignore-list files to
be unrecognized unless the test number was also used as part of the
ignore-list name. This was causing issues because the test number
can change depending on how you run the test.

This commit fixes the issue.


castulo:
Ups, I forgot to update one test. Updated it now.

Issue #1017, opened 2019-07-19
os-install: Add downloadonly flag


jwakre:
**Problem**
It would be useful for the os-install command to populate the statedir without installing content. This would enable offline installs from the cached content.

**Describe the solution you'd like**
* Add `--downloadonly` flag to `os-install` command
* When the `--downloadonly` flag is set, the statedir should be updated, but no content should be installed.

otaviobp:
I just suggest to call this flag --download to be compatible with the update command

jwakre:
> I just suggest to call this flag --download to be compatible with the update command

Earlier I wasn't aware that `swupd update` had this capability. I will add the `--download` flag to the `os-install` and `verify` commands.

castulo:
I think we can close this issue now

PR #1046, opened 2019-08-05
os-install: Add download flag

jwakre:
When the --download flag is set for the os-install command, update
content will be downloaded to the statedir and no content will be
installed.

### DIFF
--- a/test/functional/os-install/install-download.bats
+++ b/test/functional/os-install/install-download.bats
@@ -1,0 +1,40 @@
+#!/usr/bin/env bats
+
+# Author: John Akre
+# Email: john.w.akre@intel.com
+
+load "../testlib"
+
+test_setup() {
+
+	create_test_environment -e "$TEST_NAME" 10
+	create_bundle -n os-core -f /core "$TEST_NAME"
+
+}
+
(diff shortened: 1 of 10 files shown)


test/functional/os-install/install-download.bats
```diff
@@ -0,0 +1,40 @@
...
+
+	assert_status_is 0
+	expected_output=$(cat <<-EOM
+		Installing OS version 10 (latest)
+		Downloading packs for:
+		 - os-core
+		Finishing packs extraction...
+		Checking for corrupt files
+		No extra files need to be downloaded
+		Installation successful
```
castulo: Please change this message so it shows something like "Installation files downloaded successfully" or something like that when `--download` is used

castulo:
> This change adds the support for the download feature described in #1017. I only need the `--download` flag for the `os-install` command, so I decided not to add this flag to `verify`. If it is desirable to add `--download` to `verify`, I can update the PR.

+1 for only adding this to `os-install`, it is not really needed for any other command, so there is no point on adding it, it would just cause confusion.

jwakre:
@castulo I added bash/zsh completions, a config entry, and updated the print messages.

PR #1049, opened 2019-08-07
Multiple configuration files

castulo:
Enable the ability to provide many paths for the configuration file and read configuration from all of them. If the same option is defined in more than one configuration file it will just be overwritten by the one last read.

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -42,8 +42,6 @@
 #define FLAG_DEBUG 1003
 #define FLAG_ALLOW_INSECURE_HTTP 1004
 
-#define optarg_to_bool(_optarg) (_optarg ? strtobool(_optarg) : true)
-
 struct globals globals = {
 	.sigcheck = true,
 	.timecheck = true,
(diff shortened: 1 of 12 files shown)


src/globals.c
```diff
@@ -607,8 +608,8 @@ void load_flags_in_config(char *command, struct option *opts_array, const struct
 		config_found = config_parse(config_file, config_loader_set_opt, &loader_data);
 		free_string(&config_file);
 		if (config_found) {
-			debug("Using configuration file %s\n", tok);
-			break;
+			debug("Loading from configuration file %s\n", tok);
+			continue;
```
otaviobp: you don't need a continue here, it's the last command of the for

otaviobp: btw, This line isn't being printed for me when I run

Swupd diagnose --debug

castulo: > you don't need a continue here, it's the last command of the for

true ☺️

castulo: > btw, This line isn't being printed for me when I run
> 
> Swupd diagnose --debug

oh you are right, that is not going to show because regardless of if the user has chosen --debug or not, we haven't parsed that value yet. So I guess there is no point on adding those messages if they are never going to be printed. I'll remove them.

otaviobp: that's a chicken egg problem

castulo:
@otaviobp I addressed your comments

otaviobp:
The continue is still in the code. And I noticed now that you are overloading the config_found variable so if there's no /etc/swupd/config we will print "No configuration file was found".
Actually I would get rid of the print, but if you want that we need to return the correct value on config_found.

castulo:
> The continue is still in the code

haha ups, forgot to remove the continue



> And I noticed now that you are overloading the config_found variable so if there's no /etc/swupd/config we will print "No configuration file was found".


You are right

PR #1055, opened 2019-08-08
Show the real latest version using check-update

### LINKED ISSUE #482: "swupd check-update" should show very latest version based on very latest format available
Today, "swupd update -s" only shows you the latest available version for the current format the system is on.  It doesn't show the very latest available version based on the very latest format.  I think this is misleading because when an update is performed, it actually brings the system to the very latest version based on the very latest format.  

For example:
If the user's system is on 21010, "swupd update -s" shows the latest version as 21070 - based on format 23.  So when the user performs the update, they think they are getting 21070, but in fact it's going all the way to 24130 (for example) based on format 24.  

Improvement suggestion is to provide more information saying that you're on a version based on a specific format.  The very latest available version is X and its format is Y...

castulo:
When running the check-update command we get the latest version for the
format we are currently in, but we really want to get the latest version
regardless of the format.

This commit fixes the issue.

Closes #482

### DIFF
--- a/src/check_update.c
+++ b/src/check_update.c
@@ -47,7 +47,7 @@ enum swupd_code check_update()
 	int current_version, server_version;
 	enum swupd_code ret;
 
-	ret = read_versions(&current_version, &server_version, globals.path_prefix);
+	ret = read_versions(&current_version, &server_version, globals.path_prefix, true);
 
 	if (current_version > 0) {
 		info("Current OS version: %d\n", current_version);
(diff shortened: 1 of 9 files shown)


src/check_update.c
```diff
@@ -47,7 +47,7 @@ enum swupd_code check_update()
 	int current_version, server_version;
 	enum swupd_code ret;
 
-	ret = read_versions(&current_version, &server_version, globals.path_prefix);
+	ret = read_versions(&current_version, &server_version, globals.path_prefix, true);
```
otaviobp: boolean parameters to modify how a function work is usually hard to understand. 
We can just use the get_current_version() and get_absolute_latest_version() here.

PR #1056, opened 2019-08-08
Don't check network when listing bundles locally

### LINKED ISSUE #895: offline bundle-list
I understand why bundle-list -a would require an internet connection, but bundle-list with no args first tries to connect... and reconnect... waiting about 60 seconds total before spitting out the results for bundles installed on system.

It'd also be nice to have the file sizes of each bundle.

castulo:
When listing local bundles swupd doesn't really require network to
list them, it only requires it to report which of those local bundles are
experimental.

This commit allows users to run the command to list local bundles
without checking for network connectivity.

Closes #895

### DIFF
--- a/src/clr_bundle_ls.c
+++ b/src/clr_bundle_ls.c
@@ -116,6 +116,7 @@ static bool parse_options(int argc, char **argv)
 enum swupd_code bundle_list_main(int argc, char **argv)
 {
 	int ret;
+	int init_option = SWUPD_ALL;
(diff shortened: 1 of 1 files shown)


otaviobp:
Some changes on curl reinit were made and this PR will need to be reworked.
Now we don't initialize curl on init. We just need to make sure we will fail fast when there's no network

castulo:
Since there were many changes in the curl code, I think is best to close this PR and submit a new one.

PR #1057, opened 2019-08-09
bundle.c: Return included manifest loading errors

jwakre:
Previously, errors when adding included manifests were thrown away which
resulted in swupd operations that could succeed, even when failing to
load an included manifest. Now when an included manifest fails to load,
swupd will return a failure.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -732,21 +732,13 @@ int add_subscriptions(struct list *bundles, struct list **subs, struct manifest 
 		}
 
 		if (manifest->includes) {
-			int r = add_subscriptions(manifest->includes, subs, mom, find_all, recursion + 1);
-			if (r & add_sub_ERR) {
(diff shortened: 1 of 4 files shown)


src/bundle.c
```diff
@@ -732,21 +732,13 @@ int add_subscriptions(struct list *bundles, struct list **subs, struct manifest
...
 
 		if (manifest->includes) {
-			int r = add_subscriptions(manifest->includes, subs, mom, find_all, recursion + 1);
-			if (r & add_sub_ERR) {
-				free_manifest(manifest);
-				goto out;
-			}
-			ret |= r; /* merge in recursive call results */
+			/* merge in recursive call results */
+			ret |= add_subscriptions(manifest->includes, subs, mom, find_all, recursion + 1);
```
jwakre: This was added in #194 which doesn't seem relevant anymore.

PR #1058, opened 2019-08-09
Improvements for bundle-remove

### LINKED ISSUE #674: Composite bundle operations print confusing messages
I ran `swupd bundle-remove joe xemacs` and received the following messages:

```
Removing bundle: xemacs
Error: bundle requested to be removed is required by the following bundles:
format:
 # * is-required-by
 #   |-- is-required-by
 # * is-also-required-by
 # ...

   * desktop-apps
     |-- desktop
          |-- desktop-autostart
Removing bundle: joe
Error: bundle requested to be removed is required by the following bundles:
 # * is-required-by
 #   |-- is-required-by
 # * is-also-required-by
 # ...

   * editors
```

So reran the command with the editors bundle appended `swupd bundle-remove joe xemacs editors`:
```
Removing bundle: xemacs
Error: bundle requested to be removed is required by the following bundles:
format:
 # * is-required-by
 #   |-- is-required-by
 # * is-also-required-by
 # ...

   * desktop-apps
     |-- desktop
          |-- desktop-autostart
Removing bundle: joe
Error: bundle requested to be removed is required by the following bundles:
 # * is-required-by
 #   |-- is-required-by
 # * is-also-required-by
 # ...

   * editors
Removing bundle: editors
Deleting bundle files...
Total deleted files: 1
Failed to remove 2 of 3 bundles
```

At this point I was confused because:
a) the editors bundle was requested for removal, so the is-also-required-by seemed contradictory
b) swupd reports that 2 of 3 bundles failed to be removed, but a further call of `swupd remove-bundles joe` fails becuase the joe bundle isn't installed.

It would be more user friendly if swupd would take into account the full list of passed bundles and perform a composite operation with appropriate reporting, rather than executing each bundle operation in turn and reporting output on the individual operations along with a count of reported errors.

### LINKED ISSUE #732: telemetry: swupd bundle-remove redundant critical notification
Scenario: I install bundle A that also (implicitly) needs bundle B.
Then I try to remove bundle B, I then get an error message bundle B is required by A.

So this is normal and expected behavior. But this also results in telemetry critical error 
swupd-client/bundleremove notification. I don't see this as a critical error. I think swupd/telemetry should differentiate between various errors due to swupd bundle-remove. This particular one should probably not be reported by telemetry at all.

### LINKED ISSUE #891: Improve swupd bundle-remove
swupd bundle-remove prints an excessively large dependency tree and its difficult to actually remove bundles at times.

For reference, I tried to remove 'c-basic' from a machine that has 586 bundles installed (many devpkgs, desktop bundles, -dev bundles, etc). This resulted in bundle-remove spewing 24830 lines into my terminal. Sorting through those lines, there's actually only ~360 bundles that depend on it.

With high bundle counts it may be wise to not print the entire tree, and instead print a simple dependency list, preferably with leaf nodes printed first, followed by those only depending on the leaf nodes, and so on.

It would also be nice to be able to remove a bundle and anything that depends on it in one command. If I actually wanted to remove c-basic from my system right now, the way I'd do it is 'swupd bundle-remove <those 360 bundles>' in a loop. Its not too pretty but it would work eventually.

castulo:
This PR includes a couple of improvements for bundle-remove>

 - When listing the dependencies of a given bundle, swupd uses a tree view
that includes all bundles that depend on the specified bundle, plus all
the bundles that depend on that dependendant bundle, and so on.
Some times for some bundles the list can grow quite large (over 1000
lines) which make the output useless. This PR fixes the issue by simplifying
the list of dependencies if the list is big so it is easier to read.
 - A --force flag was added to bundle-remove to be used in cases where
a user wants to remove a bundle that is required by other installed bundles
without having to delete one by one.
- Removes non critical notifications being sent to telemetry when user was doing
user mistakes while removing

Closes #891 
Closes #732 
Closes #674 


castulo:
This PR is part 1 of 2 to fix issue #891 but it is still a **work in progress**

test/functional/bundleremove/remove-basics.bats
```diff
@@ -55,12 +55,13 @@ global_teardown() {
...
 	assert_file_exists "$STATEDIR"/bundles/test-bundle2
 	expected_output=$(cat <<-EOM
+		Removing bundle: test-bundle1
 		Deleting bundle files...
 		Total deleted files: 6
 		Successfully removed 1 bundle
 	EOM
 	)
-	assert_is_output --identical "$expected_output"
+	assert_is_output "$expected_output"
```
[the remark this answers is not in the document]
castulo: because I am now printing progress report when deleting the files like this:
```
Removing bundle: test-bundle1

Deleting bundle files...
	...16%
	...33%
	...50%
	...66%
	...83%
	...100%
Total deleted files: 6

Successfully removed 1 bundle
```
And a limitation our test library has is that when comparing percentages in a test we need to use "assert_regex_is_output" or "assert_regex_in_output" because "assert_in_output" will always find  a mismatch with the "tab" in front of the percentage, it cannot detect it.

src/bundle.c
```diff
@@ -601,26 +608,47 @@ enum swupd_code remove_bundles(char **bundles)
...
-		if (reqd_by != NULL) {
+		struct list *tree_reqd_by = NULL;
+		struct list *simple_reqd_by = NULL;
+
+		required_by(&simple_reqd_by, bundle, current_mom, 0, TREE_OFF);
+		if (simple_reqd_by != NULL) {
+			list_str_deduplicate(simple_reqd_by);
+			int number_of_reqd = list_len(simple_reqd_by);
+			/* let's try to build a dependency tree */
+			required_by(&tree_reqd_by, bundle, current_mom, 0, TREE_ON);
```
otaviobp: We need now to execute required_by() function twice for every bundle that we want to install just to decide if we will print the full list or not. Not sure if it's something that we want.
What about only printing the list of dependent bundles and printing the full list only when --verbose is used?

src/bundle.c
```diff
@@ -179,15 +183,18 @@ static void required_by(struct list **reqd_by, const char *bundle_name, struct m
 
 			if (strcmp(name, bundle_name) == 0) {
 				char *bundle_str = NULL;
+
 				indent = (recursion - 1) * 4;
-				if (recursion == 1) {
+				if (!tree_view) {
+					string_or_die(&bundle_str, "%s\n", bundle->component);
```
otaviobp: Instead of storing info to be printed later is way better to print what we need to print here so we don't need to keep everything in memory to print later. I'd try to refactor the calls of this function to prevent that.

src/helpers.c
```diff
@@ -1170,3 +1170,87 @@ bool check_mix_exists(void)
...
+ * free all files found in files1 that happens to be duplicated on files2
+ *
+ * Is _mandatory_ to have both files lists ordered by filename before calling
+ * this function
+ *
+ * This function removes all duplicated files matching those in files2 from
+ * files1, but it doesn't free the data.
+ *
+ */
+void deduplicate_files(struct list **files1, struct list *files2)
```
otaviobp: same here. this could be a generic list function.

src/bundle.c
```diff
@@ -135,26 +99,22 @@ bool is_installed_bundle(const char *bundle_name)
...
 }
 
-/* When loading all tracked bundles into memory, they happen
- * to be hold in subs global var, for some reasons it is
- * needed to just pop out one or more of this loaded tracked
- * bundles, this function search for bundle_name into subs
- * struct and if it found then free it from the list.
- */
-static enum swupd_code unload_tracked_bundle(const char *bundle_name, struct list **subs)
+static enum swupd_code unload_manifest(const char *bundle_name, struct list **bundles_installed, struct list **bundles_to_remove)
```
otaviobp: This a generic list operation too. You are removing one searching one element from one list, removing it and moving to another one. We could have a list_remove(list, comparison_fn) and then:

    item = list_remove(bundles_installed, manifest_cmp_fn);
    bundles_to_remove = list_prepend_data(bundles_to_remove, item);

otaviobp:
Merged commit: "Always print "Removing bundle" when removing"

otaviobp:
I reverted that actually. I didn't noticed that the tests were not fixed in the same commit

test/functional/bundleremove/remove-basics.bats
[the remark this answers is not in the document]
karthikprabhuvinod: Okay

src/bundle.c
```diff
@@ -517,161 +501,222 @@ enum swupd_code remove_bundles(char **bundles)
...
+				}
 
-		/* Now that we have the consolidated list of all files, load bundle to be removed submanifest */
-		ret = load_bundle_manifest(bundle, subs, current_version, &bundle_manifest);
-		if (ret != 0 || !bundle_manifest) {
-			error("Cannot load %s sub-manifest (ret = %d)\n", bundle, ret);
-			bad++;
-			goto out_free_mom;
-		}
+				list_free_list_and_data(simple_reqd_by, free);
```
karthikprabhuvinod: Shouldn't we be freeing bundles_iter as well.

src/lib/list.c
```diff
@@ -318,3 +318,28 @@ int list_strcmp(const void *a, const void *b)
 {
 	return strcmp((const char *)a, (const char *)b);
 }
+
+void list_str_deduplicate(struct list *strs)
```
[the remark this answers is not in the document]
castulo: what do you mean is not generic? this is a generic list function, that is why I added it to list.c. Or do you mean is not generic because it is specific for string data types? would you prefer it if it was `list_deduplicate(struct list *list, comparison_fn_type cmd_fn)` instead of `list_str_deduplicate(struct list *list)`?

src/bundle.c
castulo: that is something we can do..... now that we have the --verbose flag, instead of sowing one way or the other based on number of tree branches we can just always default for the simple way and only show the tree view if the user uses --verbose. I like that idea

src/lib/list.c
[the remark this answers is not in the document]
otaviobp: list_deduplicate(struct list *list, comparison_fn_type cmd_fn) would be better so we can use it do deduplicate lists of other elements too, like files and manifests.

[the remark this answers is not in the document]
otaviobp: but it's fine to have a str_deduplicate and implement that when needed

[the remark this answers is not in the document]
castulo: ok, I'll make the function generic of data types then 👍

src/helpers.c
otaviobp: You can use the same function that is proposed on #1041 here

src/lib/list.c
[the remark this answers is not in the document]
castulo: done

src/bundle.c
castulo: bundles_iter is just a pointer to a list (which is pointed by bundles too), I'm using this pointer just so I don't loose the pointer during the while loop.... but as a second thought,I don't even need bundles_iter, I can just use a for loop instead of a while so I don't loose the pointer

src/helpers.c
castulo: done, we'll need to merge PR #1041 first then for this to work

castulo:
This is still no ready to be reviewed again, it needs a couple more changes based on review feedback

src/bundle.c
castulo: good idea!

castulo: done

castulo:
Ready for review again, the PR went through a major refactor

Makefile.am
```diff
@@ -341,6 +341,7 @@ BATS = \
 UNIT_TESTS = \
 	test/unit/test_signature.test \
 	test/unit/test_strings.test \
+	test/unit/test_list.test \
```
castulo: nice 👌

src/lib/list.c
```diff
@@ -305,9 +305,11 @@ bool list_longer_than(struct list *list, int count)
 
 void *list_search(struct list *list, const void *item, comparison_fn_t comparison_fn)
 {
-	for (; list; list = list->next) {
-		if (comparison_fn(list->data, item) == 0) {
-			return list->data;
+	struct list *heystack = list_head(list);
```
otaviobp: Usually list implementations consider the current pointer the begging of the list. The list implementation on swupd has an unusual (and inefficient) behavior of always assuming that the user can point to any element of the list. Because of that in several functions we don't care where we are pointing, what I consider a bad habit. Instead of going to the begging of the list on every single operation I prefer to try to remove the list_head() calls as much as possible and try to use the list in a saner way.

By saying that this is not a bug in the search function. Maybe the documentation should be updated to say that the search function starts to look for the item in the pointed element. And caller is in charge on pointing to correct item.

src/lib/list.c
```diff
@@ -324,18 +324,23 @@ int list_strcmp(const void *a, const void *b)
...
 {
 	struct list *iter = NULL;
-	void *item1, *item2;
+	void *item1, *item2 = NULL;
 
-	list = list_head(list);
+	iter = list = list_head(list);
 
-	for (iter = list; iter && iter->next; iter = iter->next) {
+	while (iter && iter->next) {
```
otaviobp: +1

Can you send bug fixes that aren't related to this PR in separate PRs, so we can merge them faster?

src/helpers.c
```diff
@@ -1180,3 +1180,33 @@ bool check_mix_exists(void)
...
+			 * files.
+			 * Note: If a file didn't exist it will still be counted as deleted,
+			 * this is a limitation */
+			deleted--;
+		}
+		free_string(&fullfile);
+		count++;
+		progress_report(count, total);
+	}
+	info("Total deleted files: %i\n", deleted);
```
otaviobp: Don't print this here. Return an int in this function with the total of deleted files and print where appropriate.

src/helpers.c
```diff
@@ -1180,3 +1180,33 @@ bool check_mix_exists(void)
...
+	iter = list_head(files);
+	while (iter) {
+		file = iter->data;
+		iter = iter->next;
+		string_or_die(&fullfile, "%s/%s", globals.path_prefix, file->filename);
+		if (swupd_rm(fullfile) == -1) {
+			/* if a -1 is returned it means there was an issue deleting the
+			 * file or directory, in that case decrease the counter of deleted
+			 * files.
+			 * Note: If a file didn't exist it will still be counted as deleted,
```
otaviobp: Can't you check for that? Why do we have this limitation?
Is there a problem on using the following?

    if (swupd_rm(fullfile) < 0) {

src/lib/list.h
```diff
@@ -139,4 +142,23 @@ struct list *list_deduplicate(struct list *list, comparison_fn_t comparison_fn,
  */
 struct list *list_filter_common_elements(struct list *list1, struct list *list2, comparison_fn_t comparison_fn, list_free_data_fn_t list_free_data_fn);
 
+/**
+ * @brief sorts and removes all duplicates from a list of strings
+ */
+struct list *list_str_deduplicate(struct list *list);
+
+/**
+ * @brief removes the specified item if found, it returns a pointer to the data removed from the list as long as the
```
otaviobp: Fix: Removes the first occurrence of an item.

src/lib/list.c
```diff
@@ -388,3 +388,50 @@ struct list *list_filter_common_elements(struct list *list1, struct list *list2,
...
+		if (comp == 0) {
+			if (*list == iter) {
+				/* iter should always point to the head of the list,
+				 * if that is the element being removed, we need to
+				 * point to the new head */
+				*list = list_free_item(iter, list_free_data_fn);
+			} else {
+				list_free_item(iter, list_free_data_fn);
+			}
+			/* if the data was freed, it will return NULL */
```
otaviobp: This is incorrect. Free don't set the value to null. REturn null if list_free_data_fn == NULL

src/lib/list.c
```diff
@@ -388,3 +388,50 @@ struct list *list_filter_common_elements(struct list *list1, struct list *list2,
...
+
+	/* item not found */
+	return NULL;
+}
+
+void list_move_item(void *item_to_move, struct list **list1, struct list **list2, comparison_fn_t comparison_fn)
+{
+	void *item;
+
+	item = list_remove(item_to_move, list1, comparison_fn, NULL);
```
otaviobp: I would put this into a loop to move all occurrences of item to the new list.

src/verify.c
```diff
@@ -773,7 +773,7 @@ static int find_unsafe_to_delete(const void *a, const void *b)
 	B = (struct file *)b;
 
 	ret = strcmp(A->filename, B->filename);
-	if (ret) {
+	if (ret != 0) {
```
otaviobp: This doesn't look like a bug to me, just a style change. What was the incorrect behavior? Do we need tests for that?

src/bundle.c
```diff
@@ -478,9 +490,33 @@ static void track_installed(const char *bundle_name)
 	free_string(&dst);
 }
 
+static int find_unsafe_to_delete(const void *a, const void *b)
```
otaviobp: Avoid having so similar functions with the same name. it took me some time to figure out this function was different than the other.

src/bundle.c
```diff
@@ -132,66 +103,129 @@ bool is_installed_bundle(const char *bundle_name)
...
-					string_or_die(&bundle_str, "%*s|-- %s\n", indent, "", bundle->component);
+
+				/* this bundle has *bundle_name as a dependency */
+
+				/* if the bundle being looked at is in the list of exclusions
+				 * then don't consider it as a dependency, the user added it to
+				 * the list of bundles to be removed too, but we DO want to
+				 * consider its list of includes */
+				if (list_search(exclusions, bundle->component, list_strcmp)) {
+					goto check_includes;
```
otaviobp: Avoid gotos unless you're checking for errors or going to the end of a function.

src/lib/list.c
castulo: but what is the harm with making sure you are at the beginning of the list by doing a list_head? if you are already at the beginning it won't do anything so it won't have performance impacts, I agree we should follow some good practices when using lists, but if for some reason we are not I'd like functions to be robust to catch it. Otherwise we could end up having bugs that would be very difficult to debug

otaviobp: The major harm is that we need to add list_head() calls everywhere. In any point of your code that you receive a list as an argument you need to call to list_head() because you never know if the list is in the begging or at the end. The problem that you are trying to fix is that the called of list_search is pointing to an incorrect position of this list and not that list_search is not working.
If instead of list search you were calling other function that works with a list you would have the same problem.

src/helpers.c
castulo: I'm not sure, that part of the code (and the comment) already existed so I just copy/pasted it

castulo: I just assumed it was true

otaviobp: oh, got it. I though it was a new code. you can leave the way it is then

src/lib/list.c
castulo: ok, I will submit a new PR cherry-picking those

src/lib/list.c
castulo: yes, I understand that, but still don't see any harm on making sure. But I'll remove this commit if you think is not necessary

src/verify.c
castulo: I thought I remember in another PR you had added comments mentioning that when the variable was an integer we needed to compare it to an integer and not just use the `variable`, `!variable`, that was valid for pointers and NULL pointers. But I may be getting confused, if this is not the case then we can remove this commit

src/lib/list.c
otaviobp: Ok, for now let's leave it here, but let's try to make sure we are always returning the head of the list on functions too. To remove all unnecessary calls to list_head() sometime in the future

src/verify.c
otaviobp: I'm fine with both. You can change that if you prefer, but it's not a bug.
One thing we need to avoid is mixing ints, bools, enums, pointers in comparison. That's important to be careful

castulo: I'll remove the commit, I bought there had been a discussion about that before but I'm probably mistaken

src/lib/list.c
castulo: ok

src/helpers.c
castulo: yeah, as I mentioned, this function already existed here in manifest.c https://github.com/clearlinux/swupd-client/blob/master/src/manifest.c#L1049 and I made it a little bit more generic by receiving a list as an argument instead of a manifest, but I agree that now that it is more generic it is more appropriate to print the message from the place where the function is called. I'll make that change.

src/lib/list.c
castulo: Oh, I thought it did

src/lib/list.h
castulo: nice

src/lib/list.c
castulo: I can do that but we are going to have a problem when the data is being freed, because then list_remove is going to return NULL which would be the same result we would get if the item was not found in the list.... I can implement this but I would have to remove the option of freeing the data, so it would only remove the data from the list. You would manually have to free it if you want to clean it up

castulo: ooh wait, never mind haha my bad, I'll do this change

castulo: done

src/bundle.c
castulo: yeah, I used the same name since in both cases the function is for the exact same purpose and they were static functions, so there was no conflicts, so I thought each file could have its implementation... but I can see how that can be confusing, I'll think of a different name for this function.

src/bundle.c
castulo: ok, I'll replace this goto with an if 👌

castulo:
> But I noticed that this function is way slower when I ask it to remove only bundles that can't be removed.

Removing multiple valid bundles is now way faster.

With master:
```
$ time sudo ./swupd bundle-remove ansible blender xterm
Removing bundle: ansible
Deleting bundle files...
Total deleted files: 9490
Removing bundle: blender
Deleting bundle files...
Total deleted files: 2083
Removing bundle: xterm
Deleting bundle files...
Total deleted files: 36
Successfully removed 3 bundles

real	0m10.684s
user	0m10.134s
sys	0m0.493s
```

With this PR:
```
$ time sudo ./swupd bundle-remove ansible blender xterm

Removing bundle: ansible

Removing bundle: blender

Removing bundle: xterm

Deleting bundle files...
	...100%
Total deleted files: 12120

Successfully removed 3 bundles

real	0m3.856s
user	0m3.496s
sys	0m0.319s
```

So it took about 1/3 of the time than the previous implementation.

When testing it with bundles that cannot be removed (due to dependencies), in my test I still saw a very slight time improvement with this PR:

With master:
```
$ time sudo ./swupd bundle-remove python3-basic c-basic
...
...
real	0m4.739s
user	0m4.316s
sys	0m0.414s
```

With PR:
```
real	0m4.688s
user	0m4.431s
sys	0m0.247s
```

The only case where I saw the performance worsen was when only using invalid bundle names:

With master:
```
$ time sudo ./swupd bundle-remove fake1 fake2 fake3 fake4
Warning: Bundle "fake1" is not installed, skipping it...
Warning: Bundle "fake2" is not installed, skipping it...
Warning: Bundle "fake3" is not installed, skipping it...
Warning: Bundle "fake4" is not installed, skipping it...
Failed to remove 4 of 4 bundles

real	0m0.014s
user	0m0.009s
sys	0m0.006s
```
With PR:
```
$ time sudo ./swupd bundle-remove fake1 fake2 fake3 fake4

Warning: Bundle "fake1" is invalid, skipping it...

Warning: Bundle "fake2" is invalid, skipping it...

Warning: Bundle "fake3" is invalid, skipping it...

Warning: Bundle "fake4" is invalid, skipping it...

Failed to remove 4 of 4 bundles

real	0m3.709s
user	0m3.438s
sys	0m0.263s
```

This is because I switched the order of the checks with do with the provided bundles. In the original code we first check to see if the bundle is installed in the system, if it is not we skip it. In the submitted code it first checks to see if the bundle is valid (is in the MoM). Looking for a tracking file is faster than looking through the MoM, but the new implementation has the advantage that now the message we show is more accurate. In master if you enter an invalid bundle, it will tell you the bundle is not installed (which may be misleading). And besides that, the only moment you will see the performance worsen is if the user enters a bunch of invalid bundles in the same command, which is highly unlikely.

otaviobp:
> Removing multiple valid bundles is now way faster.

This is really really nice. Can you gather some data on that so we can add this info in next release notes? Removing one bundle and a list of big bundles to compare the time spent on remove operation and check how much in % we are better

otaviobp:
> $ time sudo ./swupd bundle-remove python3-basic c-basic
...
...
real	0m4.739s
user	0m4.316s
sys	0m0.414s

This is very very slow. Are you running that with libasan or in a slow machine? if so, can you test it with default CL CFlags?

src/bundle.c
```diff
@@ -517,159 +555,155 @@ enum swupd_code remove_bundles(char **bundles)
...
-			  bundle,
-			  current_version,
-			  ret,
-			  total_curl_sz);
-		/* if at least one of the bundles fails to be removed, exit with a failure */
-		if (ret) {
-			ret_code = ret;
-		}
+	/* get the list of the files contained in the bundles to be removed */
+	files_to_remove = consolidate_files_from_bundles(bundles_to_remove);
```
otaviobp: I found the performance problem I was complaining about. It's not a big deal, but we don't need to calculate that if there's no bundle to be installed.

castulo:
> > $ time sudo ./swupd bundle-remove python3-basic c-basic
> > ...
> > ...
> > real	0m4.739s
> > user	0m4.316s
> > sys	0m0.414s
> 
> This is very very slow. Are you running that with libasan or in a slow machine? if so, can you test it with default CL CFlags?

yeah I'm using libasan, I can test with default CFlags. Bu I'll take a look at your suggestion about not consolidating files from bundles if there is nothing to install. 👌

src/bundle.c
castulo: Fixed it

castulo:
I tested removing 20 bundles and the time was improved by almost 90%.

Released swupd:
```
$ time sudo swupd bundle-remove -p target1 -S state1 ansible blender xterm automotive-tools cassandra cockpit django games kamera jupyter krita nginx openqa productivity vnc-server xscreensaver desktop wireshark vinagre unison
Removing bundle: ansible
Deleting bundle files...
Total deleted files: 9490
Removing bundle: blender
Deleting bundle files...
Total deleted files: 2070
Removing bundle: xterm
Deleting bundle files...
Total deleted files: 34
Removing bundle: automotive-tools
Deleting bundle files...
Total deleted files: 29
Removing bundle: cassandra
Deleting bundle files...
Total deleted files: 252
Removing bundle: cockpit
Deleting bundle files...
Total deleted files: 1057
Removing bundle: django
Deleting bundle files...
Total deleted files: 728
Removing bundle: games
Deleting bundle files...
Total deleted files: 1056
Removing bundle: kamera
Deleting bundle files...
Total deleted files: 89
Removing bundle: jupyter
Deleting bundle files...
Total deleted files: 4426
Removing bundle: krita
Deleting bundle files...
Total deleted files: 1191
Removing bundle: nginx
Deleting bundle files...
Total deleted files: 26
Removing bundle: openqa
Deleting bundle files...
Total deleted files: 1417
Removing bundle: productivity
Deleting bundle files...
Total deleted files: 333
Removing bundle: vnc-server
Deleting bundle files...
Total deleted files: 1
Removing bundle: xscreensaver
Deleting bundle files...
Total deleted files: 180
Removing bundle: desktop
Deleting bundle files...
Total deleted files: 21273
Removing bundle: wireshark
Deleting bundle files...
Total deleted files: 419
Removing bundle: vinagre
Deleting bundle files...
Total deleted files: 147
Removing bundle: unison
Deleting bundle files...
Total deleted files: 6
Successfully removed 20 bundles

real	0m15.485s
user	0m14.779s
sys	0m0.581s
```

swupd from PR:
```
$ time sudo ./swupd bundle-remove -p target2 -S state2 ansible blender xterm automotive-tools cassandra cockpit django games kamera jupyter krita nginx openqa productivity vnc-server xscreensaver desktop wireshark vinagre unison

Removing bundle: ansible

Removing bundle: blender

Removing bundle: xterm

Removing bundle: automotive-tools

Removing bundle: cassandra

Removing bundle: cockpit

Removing bundle: django

Removing bundle: games

Removing bundle: kamera

Removing bundle: jupyter

Removing bundle: krita

Removing bundle: nginx

Removing bundle: openqa

Removing bundle: productivity

Removing bundle: vnc-server

Removing bundle: xscreensaver

Removing bundle: desktop

Removing bundle: wireshark

Removing bundle: vinagre

Removing bundle: unison

Deleting bundle files...
	...100%
Total deleted files: 51801

Successfully removed 20 bundles

real	0m1.323s
user	0m0.987s
sys	0m0.330s
```

castulo:
Can you guys take another look at this PR? @otaviobp @karthikprabhuvinod

src/lib/list.h
```diff
@@ -141,4 +144,23 @@ struct list *list_deduplicate(struct list *list, comparison_fn_t comparison_fn,
...
 struct list *list_filter_common_elements(struct list *list1, struct list *list2, comparison_fn_t comparison_fn, list_free_data_fn_t list_free_data_fn);
 
+/**
+ * @brief sorts and removes all duplicates from a list of strings
+ */
+struct list *list_str_deduplicate(struct list *list);
+
+/**
+ * @brief removes the first occurrence of an item, it returns a pointer to the data removed from the list as long as the
+ * data was not freed, it returns NULL if the data was freed or nothing was removed (item not found)
```
karthikprabhuvinod: Why are we removing the first occurence only? is that because we run list_str_deduplicate first on it?

castulo: that is just how the function is designed to work, that way it can return a pointer to the data being removed from the list

src/clr_bundle_rm.c
```diff
@@ -89,7 +108,17 @@ enum swupd_code bundle_remove_main(int argc, char **argv)
 	}
 	progress_init_steps("bundle-remove", steps_in_bundleremove);
 
-	ret = remove_bundles(bundles);
+	remove_set_option_force(cmdline_option_force);
+
+	/* move the bundles provided in the command line into a
+	 * list so it is easier to handle them */
+	for (; *bundles; ++bundles) {
```
karthikprabhuvinod: Can building a new list just for providing as arg for remove_bundles be alright? Can this potentially be a bottle neck? I know we increase time_complexity only by o(n) but is it worth to avoid that?

castulo: we need to manipulate that list in the bundle removal process, that is why I moved this to a list, I don't see how this could be a bottle neck. Maybe I'm not getting your question correctly.

karthikprabhuvinod: I mean we are re-creating a list by traversing a pointer to a char ..so it takes an additional space and requires 2 * n traversing of the items. But since you manipulate the list & require it, thats perfect

castulo: yeah, the bundles provided by the user are not in a list, is just an array of chars with all bundles in the same array: "bundle1 bundle2 bundle3 etc", so I'm passing each one of those bundles into a list so we can manipulate them easier. The data in the list still points to each one of those elements in the original array. So basically I'm just adding a pointer to each one of those names inside the array.

Issue #1016, opened 2019-07-19
os-install: Add flag for statedir cache directory


jwakre:
**Problem**
When using `swupd os-install` to install content from one partition to another partition, swupd will not take advantage of hardlinks which will lead to more disk space usage. If the statedir exists inside the target partition, hardlinks can be used to reduce disk space usage. Currently, there is no workaround to use the statedir on one partition to install to another partition while taking advantage of hardlinks.

**Describe the solution you'd like**
* Add `--statedir-cache` flag to `os-install` command
* When a bundle does not exist in the statedir, the statedir-cache directory will be checked.
  * When the statedir-cache has the desired bundle, the cached content will be copied to the statedir.
  * When the statedir-cache does not have the desired bundle, the bundle will be downloaded over the network.

By specifying a populated state directory on one partition with `--statedir-cache` and a directory on a second partition with `--statedir`, swupd can be used to install content from one partition to another while taking advantage of the hardlink optimization.

**Example:**
`sudo swupd os-install --statedir=chroot/var/lib/swupd --statedir-cache=/var/lib/swupd --path chroot`

jwakre:
Feature merged in #1063

PR #1063, opened 2019-08-14
Add Support for statedir-cache

jwakre:
The `--statedir-cache` flag specifies a secondary statedir that will be checked before downloading files over the network. This flag is only supported for the `os-install` command.

The primary reason to use the statedir-cache is to enable the `os-install`
command to take advantage of hardlinks when installing content across
partitions. In this case, update content stored in the statedir-cache
(source partition) will be copied to the statedir (target partition) before
installing content into the target partition. Since the update content
will exist on the target partition before installation, hardlinks can be
used instead of copies.

NOTE: This PR depends on #1061 and the offline functional tests will fail due to this dependency.

### DIFF
--- a/config
+++ b/config
@@ -306,6 +306,10 @@
 # option (string value)
 #version=V
 
+# After checking for content in the statedir, check the statedir-cache before
+# downloading it over the network
+#statedir-cache=[PATH]
+
 # Download all content, but do not actually install it (boolean value)
 #download=<true/false>
 
--- a/docs/swupd.1.rst
+++ b/docs/swupd.1.rst
@@ -356,6 +356,10 @@ SUBCOMMANDS
 
             Installs bundles os-core and vi, along with os-core (installed by default).
 
+    - `-s, --statedir-cache=[PATH]`
+
+        After checking for content in the statedir, check the statedir-cache before downloading it over the network.
+
     - `--download`
 
         Do not perform an os-install, instead download all resources needed
--- a/src/fullfile.c
+++ b/src/fullfile.c
@@ -109,6 +109,34 @@ static double fullfile_query_total_download_size(struct list *files)
 	return total_size;
 }
 
+static int get_cached_fullfile(struct file *file)
+{
+	char *targetfile;
+	char *targetfile_cache;
+	struct stat stat;
+	int ret = 0;
+
+	/* Check the statedir and statedir-cache for the expected fullfile. When the
+	 * fullfile exists in the statedir-cache, it is copied to the statedir. */
+	string_or_die(&targetfile, "%s/staged/%s", globals.state_dir, file->hash);
+	if (lstat(targetfile, &stat) != 0 || !verify_file(file, targetfile)) {
+		ret = 1;
+
+		if (globals.state_dir_cache != NULL) {
+			string_or_die(&targetfile_cache, "%s/staged/%s", globals.state_dir_cache, file->hash);
+			if (lstat(targetfile_cache, &stat) == 0 && verify_file(file, targetfile_cache)) {
+				if (link_or_copy_all(targetfile_cache, targetfile) == 0) {
+					ret = 0;
+				}
+			}
+			free_string(&targetfile_cache);
+		}
+	}
+	free_string(&targetfile);
+
+	return ret;
+}
+
 /*
  * Download fullfiles from the list of files.
  *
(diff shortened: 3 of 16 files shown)


src/manifest.c
```diff
@@ -298,25 +298,35 @@ static void set_untracked_manifest_files(struct manifest *manifest)
 }
 
 /* Removes the extracted Manifest.<bundle> and accompanying tar file, cache file, and
- * the signature file */
+ * the signature file from the statedir and statedir-cache */
 static void remove_manifest_files(char *filename, int version, char *hash)
 {
 	char *file;
+	char *state_dirs[] = { globals.state_dir, globals.state_dir_cache };
```
jwakre: When the statedir contains a corrupted manifest, the statedir-cache will also be wiped, so it cannot be used as a fallback. It would be better for the statedir-cache to be used as a fallback in this case, but the current fallback code for manifests doesn't work well for more than 1 fallback method.

The manifest fallback code could be refactored to better support the statedir-cache. The `retrieve_manifests` function could validate manifest hashes instead of only checking for the existence of correctly named manifests (this is how the fullfile fallback code works). This would make it easy to find the desired manifest by checking the statedir, checking the statedir-cache, or downloading over the network.

src/fullfile.c
```diff
@@ -139,15 +139,30 @@ int download_fullfiles(struct list *files, int *num_downloads)
...
 		if (file->is_deleted || file->do_not_update) {
 			continue;
 		}
 
+		/* Check the statedir and statedir-cache for the expected fullfile. When the
+		 * fullfile exists in the statedir-cache, it is copied to the statedir. */
 		string_or_die(&targetfile, "%s/staged/%s", globals.state_dir, file->hash);
 		if (lstat(targetfile, &stat) != 0 || !verify_file(file, targetfile)) {
-			need_download = list_append_data(need_download, file);
+			bool download = true;
```
otaviobp: this boolean here makes the code a bit hard to read. Can you group all this block (including the if lstat) in a function and return if we need to download or not?

    if (!fullfile_exists_or_cached()) {
        need_download = list_append_data(need_download, file);
    }

src/packs.c
```diff
@@ -237,7 +238,20 @@ int download_subscribed_packs(struct list *subs, struct manifest *mom, bool requ
 		/* make sure the file is not already in the client system */
 		string_or_die(&targetfile, "%s/pack-%s-from-%i-to-%i.tar", globals.state_dir, sub->component, sub->oldversion, sub->version);
 		if (lstat(targetfile, &stat) != 0 || stat.st_size != 0) {
-			need_download = list_append_data(need_download, sub);
+			bool download = true;
```
otaviobp: same here

src/fullfile.c
```diff
@@ -109,6 +109,33 @@ static double fullfile_query_total_download_size(struct list *files)
 	return total_size;
 }
 
+static int get_cached_fullfile(struct file *file) {
```
jwakre: @otaviobp Since in some cases I am copying content to the statedir, I chose `get_cached_fullfile` instead of `fullfile_exists_or_cached`.

jwakre:
This PR closes #1016

config
```diff
@@ -306,6 +306,10 @@
 # option (string value)
 #version=V
 
+# After checking for content in the statedir, check the statedir-cache before
+# downloading it over the network
+#statedir-cache=[PATH]
```
castulo: I know this is late but this should have been `stated_cache` (underscore), we are only using words with underscore in the config file. We have code to convert that _ to - before using it in swupd. Since this has already merged I will submit a PR to change this.

docs/swupd.1.rst
```diff
@@ -356,6 +356,10 @@ SUBCOMMANDS
 
             Installs bundles os-core and vi, along with os-core (installed by default).
 
+    - `-s, --statedir-cache=[PATH]`
```
castulo: same here

src/globals.c
```diff
@@ -205,11 +205,45 @@ static bool set_state_dir(char *path)
...
+
+	return true;
+}
+
 static void set_default_state_dir()
 {
 	string_or_die(&globals.state_dir, "%s", STATE_DIR);
 }
 
+static void set_default_state_dir_cache()
```
castulo: you missed adding a void as argument

otaviobp: good catch. The PR is already merged, @castulo can you add that in a new commit?

PR #1072, opened 2019-08-21
Values in config file should use underscores

castulo:

### DIFF
--- a/config
+++ b/config
@@ -236,20 +236,20 @@
 
 # Selects the sub-tree where --picky looks for extra files (string value)
 # Default: /usr (string value)
-#picky-tree=[PATH]
+#picky_tree=[PATH]
 
 # Any path completely matching the POSIX extended regular expression is ignored
 # by --picky, matched directories get skipped (string value)
 # Example: /var|/etc/machine-id
 # Default: /usr/lib/modules|/usr/lib/kernel|/usr/local|/usr/src
-#picky-whitelist=[RE]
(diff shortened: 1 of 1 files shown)


config
```diff
@@ -236,20 +236,20 @@
...
-#picky-whitelist=[RE]
+#picky_whitelist=[RE]
 
 # Diagnose if BUNDLES are installed correctly (string value)
 # Example: --bundles=os-core,vi
 #bundles=[BUNDLES]
 
 # Like --picky, but it only searches for extra files
-#extra-files-only=<true/false>
+#extra_files_only=<true/false>
```
otaviobp: Just one question. Why we aren't supporting '-' in options in config files?

castulo: we do support "-", it is just a matter of preference.

At the time I wrote the code for the config parser I was more used to seeing values in config files like this "some_key = value" than this "some-key = value", so I decided to go with the underscore in the config file (maybe cause I had worked more with python projects that tend to use snake_case). But if a user uses one or the other it will still work the same since "_" are covered to "-" when parsing.

But since at this moment, all other keys in the config file use "_" instead of "-" I think it could be confusing to have mixed values in the config we provide as a template. So I'd say we should either replace all "_" with "-" if you prefer using "-", or keep using only "_" in the values we provide.

otaviobp: I wasn't sure if we were supporting both. So we're good. I prefer _ on config files too/

PR #1077, opened 2019-08-29
Fail fast when testing connection with server

### LINKED ISSUE #895: offline bundle-list

castulo:
When curl is being initialized, we test the connection with the server.
curl is being initialized (if not already) on demand every time we need
to use curl to get a file. In some cases swupd performs an
alternate path if there is no network connectivity, and in those cases
we don't want to wait too long before taking that path so we don't
impact respose time from swupd.

This commit modifies the timeout value when testing for connectivity
during curl initialization so it fails fast in case the connectivity
test is unsuccessful.

Closes #895

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -50,6 +50,7 @@
 #define SWUPD_CURL_LOW_SPEED_LIMIT 1
 #define SWUPD_CURL_CONNECT_TIMEOUT 30
 #define SWUPD_CURL_RCV_TIMEOUT 120
+#define SWUPD_CURL_INIT_CONNECT_TIMEOUT 1
 
 static CURL *curl = NULL;
 
@@ -118,7 +119,7 @@ exit:
 
 static int check_connection_capath(const char *test_capath, const char *url)
 {
-	CURLcode curl_ret;
+	CURLcode curl_ret, ret;
 	long response = 0;
 
 	if (!url) {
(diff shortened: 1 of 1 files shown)


castulo:
> I'm unsure if we should change the timeout on check_connection(). If the user's network is slow we may be timing out even if there is a connection available.
> 
> If there's no network available curl should fail fast. I couldn't reproduce a case where it get stuck trying to check the connection. Do you know what was the case where we are having that?

No, I never really tested that. That is a good point, maybe it cannot even be reproduced anymore.

otaviobp:
I have reproduced the problem here in the case where a proxy is used. Not sure yet if it's a good idea to reduce the timeout

castulo:
what do you suggest then?

karthikprabhuvinod:
@otaviobp : Did yours timeout for around 60 secs like in the #895 ? 

@castulo I could not reproduce this with bundle-list no args as well.

castulo:
We are closing this PR, swupd is only hanging when there is a proxy set up and there is no network connection with it, and we cannot do anything about that. In that case it will only wait for 30 seconds and timeout.

PR #1081, opened 2019-09-09
Progress bar for manifest download

karthikprabhuvinod:


castulo:
Again, one general comment is that I think you should squash your commits into one, and remove the "WIP" from the first commit. You don't necessarily need to always use one commit in a PR, you can use as many as you want as long as it make sense to have them separately. Usually we try to keep the commits that add or fix tests for a new functionality in the same commit, unless you are doing TDD in which case the tests would be pushed first than the implementation.

castulo:
Remember you can (and should) use "closes #PR_Number" in the PR description if there is a related issue so it is closed when this PR is merged. Pleas try to add a detailed description of what your PR is implementing (unless is very obvious), that make the PR so much easier to review (by the way I think this is one of those cases where that is not necessary cause it is very obvious).

karthikprabhuvinod:
@castulo :  In some cases like this example, if you trace the code with recurse_manifest while running diagnose,

the problem is that recurse_manifest does not get called before add_subscriptions and add_subscriptions directly calls load_manifest. So the manifests are downloaded even before recurse_manifest.

castulo:
> @castulo : In some cases like this example, if you trace the code with recurse_manifest while running diagnose,
> 
> the problem is that recurse_manifest does not get called before add_subscriptions and add_subscriptions directly calls load_manifest. So the manifests are downloaded even before recurse_manifest.

I would implement the download manifest progress thing for both functions then. The manifest download progress ideally should be a function that could be called from recurse_manifest or add_subscriptions or whatever other function we use that loads many manifests... but if a generic function can't be implemented for those cases, I guess we would need to implement that separately.

otaviobp:
We have a problem here that I haven't thought about when I suggested that change. There's no way for us to know how many manifests we need to download or load before calling the recurse_manifest() function. And we need to download manifests to know what's the dependency list. We just know the total number of manifests we need to download at the end of the process.
@castulo am I missing something here? Can you think in a way on estimating that?

karthikprabhuvinod:
Closing one as I created a PR

castulo:
> We have a problem here that I haven't thought about when I suggested that change. There's no way for us to know how many manifests we need to download or load before calling the recurse_manifest() function. And we need to download manifests to know what's the dependency list. We just know the total number of manifests we need to download at the end of the process.
> @castulo am I missing something here? Can you think in a way on estimating that?

No @otaviobp you are right, we need the manifests first to get our dependency list 😢

PR #1083, opened 2019-09-11
Showing indirect bundles installed by bundle-add

### LINKED ISSUE #679: swupd bundle-add needs better reporting on installed metrics
Reported by @ikeyd 

It seems we're not counting dependents - intentional?
Perhaps "2 bundles (N total)" would be more appropriate?

```
ikey@ikey-clr-nuc ~ $ sudo swupd bundle-add  os-clr-on-clr-dev
Downloading packs...

Extracting os-clr-on-clr-dev pack for version 26120
        ...3%
Extracting editors-dev pack for version 26130
        ...6%
Extracting go-basic-dev pack for version 25990
        ...10%
Extracting kvm-host-dev pack for version 26130
        ...13%
Extracting os-clr-on-clr pack for version 26090
        ...16%
Extracting c-basic pack for version 26130
        ...20%
Extracting dev-utils-dev pack for version 26130
        ...23%
Extracting go-basic pack for version 26130
        ...26%
Extracting koji pack for version 26130
        ...30%
Extracting web-server-basic pack for version 26130
        ...33%
Extracting kvm-host pack for version 26100
        ...36%
Extracting libX11client pack for version 26120
        ...40%
Extracting mail-utils-dev pack for version 26130
        ...43%
Extracting mail-utils pack for version 26100
        ...46%
Extracting mixer pack for version 26130
        ...50%
Extracting containers-basic pack for version 26130
        ...53%
Extracting os-installer pack for version 26100
        ...56%
Extracting zstd pack for version 25770
        ...60%
Extracting network-basic-dev pack for version 26130
        ...63%
Extracting os-core-update-dev pack for version 26130
        ...66%
Extracting package-builder pack for version 26130
        ...70%
Extracting curl pack for version 26050
        ...73%
Extracting diffutils pack for version 25770
        ...76%
Extracting package-utils pack for version 26130
        ...80%
Extracting python2-basic pack for version 26080
        ...83%
Extracting perl-basic-dev pack for version 26020
        ...86%
Extracting storage-utils-dev pack for version 26130
        ...90%
Extracting storage-utils pack for version 26100
        ...93%
Extracting sysadmin-basic-dev pack for version 26130
        ...96%
Extracting os-core-dev pack for version 26130
        ...100%
        ...100%
Installing bundle(s) files...
        ...100%
Successfully installed 2 bundles
ikey@ikey-clr-nuc ~ $
```

I agree that it would be nice to additionally report number of dependencies installed as a separate metric.

castulo:
When installing a bundle using bundle-add, swupd also installs any
dependency not already installed in the system (if any). When the
bundle-add operation is complete, swupd reports how many of the
requested bundles were successfully installed and how many failed, but
if there are bundles that got installed as a side effect because they
are dependencies of the requested bundle(s), they are not included in
the summary of bundle-add.

This commit adds the number of dependencies installed in the summary of
bundle-add so it provides a more accurate view of the changes in the
system.

Closes #679 

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -753,6 +753,7 @@ static enum swupd_code install_bundles(struct list *bundles, struct list **subs,
 	int bundles_failed = 0;
 	int already_installed = 0;
 	int bundles_installed = 0;
+	int dependencies_installed = 0;
 	int bundles_requested = list_len(bundles);
 	long bundle_size = 0;
 	long fs_free = 0;
@@ -1048,9 +1049,13 @@ out:
 		struct manifest *to_install_manifest;
 		to_install_manifest = iter->data;
 		iter = iter->next;
-		if (string_in_list(to_install_manifest->component, bundles)) {
-			bundles_installed++;
-			track_installed(to_install_manifest->component);
+		if (is_installed_bundle(to_install_manifest->component)) {
+			if (string_in_list(to_install_manifest->component, bundles)) {
+				bundles_installed++;
+				track_installed(to_install_manifest->component);
+			} else {
+				dependencies_installed++;
+			}
 		}
 	}
 
@@ -1063,12 +1068,16 @@ out:
 	} else {
 		bundles_failed = bundles_requested - bundles_installed - already_installed;
 	}
+
 	if (bundles_failed > 0) {
 		print("Failed to install %i of %i bundles\n", bundles_failed, bundles_requested - already_installed);
-	} else if (bundles_installed) {
+	} else if (bundles_installed > 0) {
 		print("Successfully installed %i bundle%s\n", bundles_installed, (bundles_installed > 1 ? "s" : ""));
 	}
-	if (already_installed) {
+	if (dependencies_installed > 0) {
+		print("%i bundle%s\n", dependencies_installed, (dependencies_installed > 1 ? "s were installed as dependencies" : " was installed as dependency"));
+	}
+	if (already_installed > 0) {
 		print("%i bundle%s already installed\n", already_installed, (already_installed > 1 ? "s were" : " was"));
 	}
 
--- a/test/functional/bundleadd/add-also-add-flag.bats
+++ b/test/functional/bundleadd/add-also-add-flag.bats
@@ -43,6 +43,7 @@ test_setup() {
 		Installing bundle(s) files...
 		Calling post-update helper scripts
 		Successfully installed 1 bundle
+		3 bundles were installed as dependencies
 	EOM
 	)
 	assert_is_output "$expected_output"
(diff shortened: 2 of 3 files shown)


src/bundle.c
```diff
@@ -727,7 +729,11 @@ int add_subscriptions(struct list *bundles, struct list **subs, struct manifest
 			}
 		} else {
 			// Just add it to a list if it doesn't exist
-			create_and_append_subscription(subs, bundle);
+			if (recursion > 0) {
```
otaviobp: This isn't going to work.

If BundleA and BundleB aren't installed and bundleA includes bundleB and you run

    swupd bundle-add bundleB bundleA

we are going to try to install bundleA first and add bundleB as a dependency. And then we try to install bundleB, but this is already listed as a dependency.
At the end we print:

    Successfully installed 2 bundles
    1 bundles was installed as dependency

What's is incorrect.

Instead of that, can't we just report the total number of installed bundles at the end?

Successfully installed X bundles, where X is the number of total bundles installed and not just the ones that were explicitly requested? I wonder if we really need this info.

karthikprabhuvinod: I think even the request is to list deps for a single bundle. I could see a way where the user could tally the expected includes and then run the bundle-add to see she/he gets it. But yes, this is super adhoc request.

The user could also infer the same from total installed bundles with what otavio suggested. Also an added incentive in this case is a much cleaner general solution.

karthikprabhuvinod:
@castulo Do you think its worth counting the deps in case of a single bundle?

src/bundle.c
castulo: I wrote this implementation based on my interpretation of this comment from Matthew "I agree that it would be nice to additionally report number of dependencies installed as a separate metric". 

We could just report the total number of bundles installed at the end as @otaviobp suggested, it would be very easy to implement it that way actually, but I'm not convinced it is the best way to do it.

We are very familiar with how swupd works so if we run:
`swupd bundle-add bundleA`

and after running it we see
`Successfully installed 5 bundles`

we could easily assume swupd installed 4 bundles as dependencies. But I'm not sure another less experienced user would make this assumption, maybe they would think something went wrong.

To make this more difficult... what if you add let's say 3 bundles, and each installs 5 dependencies (some of them could be common among them), then if one of the bundles fail to install you would still see something like:
```
swupd bundle-add bundleA bundleB bundleC
Failed to install 1 of 15 bundles
```

otaviobp: @castulo I agree then. Let's go with the solution that prints the dependencies separated from the bundles the user added. just fix the reported problem and it would be nice to have a test covering that too.

castulo: I'll work on that @otaviobp

castulo:
> @castulo Do you think its worth counting the deps in case of a single bundle?

that is not ideal, it may confuse users to show different outputs depending on how many bundles they add to the command

PR #1084, opened 2019-09-11
Show format number

karthikprabhuvinod:
Now, we have global option
and info_verbose logging level.
For this PR, we have verbose for info,
check-update which shows format versions.
This also helps future implementations leverage
the use of a verbose option for their own need.

commands:
swupd check-update --verbose
swupd info --verbose

### DIFF
--- a/src/check_update.c
+++ b/src/check_update.c
@@ -46,6 +46,8 @@ enum swupd_code check_update()
 {
 	int current_version, server_version;
 	enum swupd_code ret = SWUPD_OK;
+	int server_format = -1;
+	int current_format = -1;
 
 	current_version = get_current_version(globals.path_prefix);
 	server_version = version_get_absolute_latest();
@@ -63,11 +65,32 @@ enum swupd_code check_update()
 		ret = SWUPD_SERVER_CONNECTION_ERROR;
 	}
 
-	if (current_version > 0) {
-		info("Current OS version: %d\n", current_version);
+	if (log_get_level() >= LOG_INFO_VERBOSE) {
+		/* Retrieve latest_format */
+		/* Sending latest server_version to get the latest format */
+		server_format = get_server_format(server_version);
+		/* Retrieve current format */
+		current_format = get_current_format();
 	}
+
+	if (current_version > 0) {
+		info("Current OS version: %d", current_version);
+		if (current_format > 0) {
+			info_verbose(" (format %d)", current_format);
+		} else {
+			info_verbose(" (format unknown)");
+		}
+		info("\n");
+	}
+
 	if (server_version > 0) {
-		info("Latest server version: %d\n", server_version);
+		info("Latest server version: %d", server_version);
+		if (server_format > 0) {
+			info_verbose(" (format %d)", server_format);
+		} else {
+			info_verbose(" (format unknown)");
+		}
+		info("\n");
 	}
 
 	if (ret != SWUPD_OK) {
--- a/src/globals.c
+++ b/src/globals.c
@@ -41,6 +41,7 @@
 #define FLAG_QUIET 1002
 #define FLAG_DEBUG 1003
 #define FLAG_ALLOW_INSECURE_HTTP 1004
+#define FLAG_VERBOSE 1005
 
 struct globals globals = {
 	.sigcheck = true,
@@ -60,6 +61,7 @@ static int max_parallel_downloads = -1;
 static int log_level = LOG_INFO;
 static bool quiet = false;
 static bool debug = false;
+static bool verbose = false;
 
 /* Sets the content_url global variable */
 static void set_content_url(char *url)
@@ -492,6 +494,7 @@ static const struct option global_opts[] = {
 	{ "json-output", no_argument, 0, 'j' },
 	{ "allow-insecure-http", no_argument, 0, FLAG_ALLOW_INSECURE_HTTP },
 	{ "wait-for-scripts", no_argument, 0, FLAG_WAIT_FOR_SCRIPTS },
+	{ "verbose", no_argument, 0, FLAG_VERBOSE },
 	{ 0, 0, 0, 0 }
 };
 
@@ -595,6 +598,9 @@ static bool global_parse_opt(int opt, char *optarg)
 	case FLAG_ALLOW_INSECURE_HTTP:
 		globals.allow_insecure_http = optarg_to_bool(optarg);
 		return true;
+	case FLAG_VERBOSE:
+		verbose = true;
+		return true;
 	default:
 		return false;
 	}
@@ -649,6 +655,7 @@ void global_print_help(void)
 	print("   -j, --json-output       Print all output as a JSON stream\n");
 	print("   --allow-insecure-http   Allow updates over insecure connections\n");
 	print("   --quiet                 Quiet output. Print only important information and errors\n");
+	print("   --verbose               Enable verbosity for commands\n");
 	print("   --debug                 Print extra information to help debugging problems\n");
 	print("   --no-progress           Don't print progress report\n");
 	print("   --wait-for-scripts      Wait for the post-update scripts to complete\n");
@@ -742,6 +749,8 @@ int global_parse_options(int argc, char **argv, const struct global_options *opt
 		log_level = LOG_DEBUG;
 	} else if (quiet) {
 		log_level = LOG_ERROR;
+	} else if (verbose) {
+		log_level = LOG_INFO_VERBOSE;
 	}
 	log_set_level(log_level);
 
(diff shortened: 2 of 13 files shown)


src/info.c
```diff
@@ -22,6 +22,8 @@
 
 #include "swupd.h"
 
+static bool cmdline_option_verbose = false;
```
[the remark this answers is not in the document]
karthikprabhuvinod: Done

src/info.c
```diff
@@ -18,14 +18,19 @@
  */
 
 #define _GNU_SOURCE
-#include <stdio.h>
-
 #include "swupd.h"
+#include <stdio.h>
```
otaviobp: Usually it's better to include local headers after includes from the library. Is there a reason you changed that?

otaviobp: And actually it looks like you don't even need that include

src/check_update.c
```diff
@@ -63,13 +65,24 @@ enum swupd_code check_update()
 		ret = SWUPD_SERVER_CONNECTION_ERROR;
 	}
 
+	if (globals.verbose) {
```
otaviobp: Check for log level here so it will work with --debug too

    if (log_get_level() >= LOG_INFO_VERBOSE)

src/swupd.h
```diff
@@ -164,6 +164,8 @@ extern int get_absolute_latest_version(void);
 extern enum swupd_code read_versions(int *current_version, int *server_version, char *path_prefix);
 extern int get_current_version(char *path_prefix);
 extern bool get_distribution_string(char *path_prefix, char *dist);
+extern void get_current_format(char **current_format);
```
otaviobp: Use the same API you use for get_latest_format() for consistency, returning an int instead of saving it in a char.
If it's a negative number you can print "unknown"

src/swupd.h
```diff
@@ -164,6 +164,8 @@ extern int get_absolute_latest_version(void);
...
+extern int get_latest_format(int server_version);
```
otaviobp: Rename this to get_server_format(). You only use this one for the latest, but this function can be used to get the format number of any server version.

src/version.c
```diff
@@ -32,6 +32,48 @@
...
+		format_string
+	};
+
+	string_or_die(&url, "%s/%d/format", globals.version_url, server_version);
+	ret = swupd_curl_get_file_memory(url, &format);
+	free_string(&url);
+
+	if (ret) {
+		return -1;
+	} else {
```
otaviobp: you don't need an else after a return;

src/check_update.c
```diff
@@ -63,13 +65,24 @@ enum swupd_code check_update()
...
 	if (current_version > 0) {
-		info("Current OS version: %d\n", current_version);
+		info("Current OS version: %d", current_version);
+		info_verbose(" (format %s)", (current_format == NULL) ? "unknown" : current_format);
+		info("\n");
 	}
 	if (server_version > 0) {
-		info("Latest server version: %d\n", server_version);
+		info("Latest server version: %d", server_version);
+		info_verbose(" (format %d)", latest_format);
```
otaviobp: do the same you did for current format to print unknown if latest_format < 0

src/info.c
karthikprabhuvinod: Looks it was already there in master

src/check_update.c
karthikprabhuvinod: Good catch! Didnt think about that. Thanks!

src/version.c
karthikprabhuvinod: Done

src/swupd.h
karthikprabhuvinod: Done

src/swupd.h
karthikprabhuvinod: Done

src/check_update.c
karthikprabhuvinod: Done

src/check_update.c
karthikprabhuvinod: Done

karthikprabhuvinod:
@otaviobp : Done with all. I was thinking of a way to implement and print those unknowns in a functional. Not sure how!!

castulo:
You need to sign your commit @karthikprabhuvinod

src/version.c
```diff
@@ -32,6 +32,59 @@
...
+	}
+
+	return ret;
+}
+
+int get_current_format()
+{
+	char *temp_format_buffer = NULL;
+	int ret, err = -1;
+	ret = get_value_from_path(&temp_format_buffer, DEFAULT_FORMAT_PATH, false);
```
castulo: I think we need to support a path prefix set by the user the  using the -p flag

test/functional/checkupdate/chk-update-format-bump.bats
```diff
@@ -29,4 +29,16 @@ test_setup() {
 	)
 	assert_is_output "$expected_output"
 
+	# Also check if verbose options holds
```
castulo: please add this as a new test, tests are ideally atomic unless that cannot happen for some reason. As it is now, if the previous test fails you will never know if this test would pass or fail.

src/version.c
karthikprabhuvinod: Isnt it doing the same ? I thought sending a false as the last argument means it is not an absolute path and would take one by prefixing the state dir path

castulo: ooh, you are right, I missed that 😄

src/check_update.c
```diff
@@ -63,11 +66,30 @@ enum swupd_code check_update()
...
+		int temp_current_format = get_current_format();
+		if (temp_current_format > 0) {
+			sprintf(current_format, "%d", temp_current_format);
+		}
+	}
+
 	if (current_version > 0) {
-		info("Current OS version: %d\n", current_version);
+		info("Current OS version: %d", current_version);
+		info_verbose(" (format %s)", current_format);
```
otaviobp: you can move the logic of the printf here, so you don't need to calculate and store and then print. 

		int temp_current_format = get_current_format();
		if (temp_current_format > 0) {
			info_verbose(" (format %d)", temp_current_format);
		} else {
			info_verbose(" (format unknown)", temp_current_format);
		}

otaviobp:
There's another thing I forgot. Every time we are adding a new flag we need to update the documentation too

src/globals.h
```diff
@@ -49,6 +49,7 @@ extern struct globals {
 	int update_count;
 	int update_server_port;
 	int update_skip;
+	bool verbose;
```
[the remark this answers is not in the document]
karthikprabhuvinod: Done

src/check_update.c
```diff
@@ -63,11 +66,30 @@ enum swupd_code check_update()
 		ret = SWUPD_SERVER_CONNECTION_ERROR;
 	}
 
+	if (log_get_level() >= globals.verbose) {
```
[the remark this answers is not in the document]
karthikprabhuvinod: done

test/functional/checkupdate/chk-update-format-bump.bats
karthikprabhuvinod: Done

src/check_update.c
karthikprabhuvinod: Make sense...remove that extra sprintf

src/globals.c
```diff
@@ -652,6 +658,7 @@ void global_print_help(void)
 	print("   --debug                 Print extra information to help debugging problems\n");
 	print("   --no-progress           Don't print progress report\n");
 	print("   --wait-for-scripts      Wait for the post-update scripts to complete\n");
+	print("   --verbose               Enable verbosity for commands\n");
```
castulo: can you please move this line between --quiet and --debug as you did in the documentation?
we are trying to group the options in the menus based on functionality, so --verbose belong between those two

karthikprabhuvinod: Sure thing Thanks

src/lib/log.c
```diff
@@ -102,6 +102,11 @@ static void log_internal(FILE *out, const char *file, int line, const char *labe
 	}
 }
 
+int log_get_level()
```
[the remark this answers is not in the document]
karthikprabhuvinod: Done

src/version.c
```diff
@@ -32,6 +32,59 @@
...
+	format.data[format.len] = '\0';
+	err = strtoi_err(format.data, &ret);
+	if (err != 0) {
+		return -1;
+	}
+
+	return ret;
+}
+
+int get_current_format()
```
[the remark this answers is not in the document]
karthikprabhuvinod: Done

src/lib/log.c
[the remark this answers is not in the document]
karthikprabhuvinod: Good catch Thanks!

PR #1087, opened 2019-09-16
Testlib fixes

castulo:
This PR includes several fixes and improvements for the swupd test library.
- It adds a `debug_msg` function to print messages for troubleshooting tests.
- It enables printing the debug messages and saving a copy of the test environment separately using two environment variables: DEBUG_TEST and KEEP_ENV.
- It add a `--force` flag to `destroy_test_environment()` to force the removal of an environment even when the `KEEP_ENV` variable is set to true. This is useful to remove leftover environments before starting a new execution.
- It fixes an issue with the logic of the library to call the `global_setup()` and `global_teardown()` functions. The functions were being called correctly when running the tests using "make check" but they were not being called when running the tests using "bats \<directory\>/".
- Adds a missing test ID in a test.

### DIFF
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -109,6 +109,7 @@ generate_certificate() { # swupd_function
 	validate_param "$key_path"
 	validate_param "$cert_path"
 
+	debug_msg "Generating test certificate"
 	# generate self-signed public and private key
 	if [ -z "$config" ]; then
 		sudo openssl req -x509 -sha256 -nodes -days 365 -newkey rsa:2048 \
@@ -121,6 +122,10 @@ generate_certificate() { # swupd_function
 			-subj "/C=US/ST=Oregon/L=Portland/O=Company Name/OU=Org/CN=localhost" \
 			-config "$config"
 	fi
+
+	debug_msg "Generated private key: $key_path"
+	debug_msg "Generated public key: $cert_path"
+	debug_msg "Keys generated succssfully"
 
 }
 
@@ -1577,6 +1582,8 @@ create_test_environment() { # swupd_function
 	validate_param "$env_name"
 	validate_number "$size"
 
+	debug_msg "Creating test environment: $env_name"
+
 	# clean test environment when test interrupted
 	trap 'destroy_test_environment $TEST_NAME' INT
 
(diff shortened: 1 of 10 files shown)


test/functional/testlib.bash
```diff
@@ -214,6 +214,37 @@ print() { # swupd_function
 
 }
 
+# sometimes it is useful to have extra information to debug
+# a test but you don't want to print all this extra info normally,
+# all messages printed with debug_msg will only be shown when the
+# user sets the environment variable DEBUG_TEST=true
+debug_msg() { #swupd_function
```
karthikprabhuvinod: I guess you missed the parameter description in the  comments.

test/functional/testlib.bash
```diff
@@ -1602,20 +1639,30 @@ create_test_environment() { # swupd_function
...
-			      purposes, set DEBUG_TEST=true before running your test.
+			      purposes, set KEEP_ENV=true before running your test.
 			EOM
 		return
 	fi
+	[ "$1" = "--force" ] && { force=true ; shift ; }
+	local env_name=$1
 	validate_param "$env_name"
+	debug_msg "Destroying environment $env_name"
+	debug_msg "The KEEP_ENV environment variable is se to $KEEP_ENV"
```
karthikprabhuvinod: nit pick "is set to"

test/functional/testlib.bash
```diff
@@ -3167,98 +3219,125 @@ list_tests() { # swupd_function
...
+	debug_msg "Test file: $TEST_NAME.bats"
+	debug_msg "BATS test number: $BATS_TEST_NUMBER"
+
+	# run the global_setup only once
+	if [ "$BATS_TEST_NAME" = "${BATS_TEST_NAMES[0]}" ]; then
+
+		# if a global environment exists at this point it means
+		# it was a leftover from a previous execution, remove it
+		if [ -d "$TEST_NAME" ]; then
+			print "\\nAn old test environment was found for this test: $TEST_NAME"
```
karthikprabhuvinod: Should we change this to debug as well? Not really relevant unless we run with DEBUG_TETS

test/functional/testlib.bash
castulo: good eye

test/functional/testlib.bash
castulo: yeah, we can make that a debug message now

castulo:
I forgot to update the certificate tests that only run in Travis, that is why there were failures. I just fixed them, I took the opportunity to add more debug messages to some functions which helped me determine the problem with the failing tests. I think we can add those debug messages on demand when we encounter test failures due struggle to fix.

PR #1088, opened 2019-09-17
test: Change the prefix of real content tests

otaviobp:
Now 3 characters are enforced on test prefix, so renaming the real content
tests to RCT


karthikprabhuvinod:
Th job again failed while running "sudo apt-get install python3-docutils".  So i restarted it

PR #1090, opened 2019-09-18
Progress bar for manifest download

### LINKED ISSUE #1060: Show progress bar when we are downloading manifests
For example if I run:

    $ swupd clean --all
    $ swupd diagnose

Swupd diagnose won't print any output for some time while it's downloading the required manifests to continue the diagnose process. We should print a progress bar in this case

karthikprabhuvinod:
This ensures we have a callback function
for full_file curl download. The function
implements a spinner for progress

Fixes #1060

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -58,10 +58,17 @@ uint64_t total_curl_sz = 0;
 /* alternative CA Path */
 static char *capath = NULL;
 
+static progress_callback_fn_t progress_callback = NULL;
+
 /* Pretty print curl return status */
 static void swupd_curl_strerror(CURLcode curl_ret)
 {
 	error("Curl - Download error - (%d) %s\n", curl_ret, curl_easy_strerror(curl_ret));
+}
+
+void set_progress_callback(progress_callback_fn_t progress_cb)
+{
+	progress_callback = progress_cb;
 }
 
 /*
@@ -779,6 +786,15 @@ CURLcode swupd_curl_set_basic_options(CURL *curl, const char *url, bool fail_on_
 	if (fail_on_error) {
 		/* Avoid downloading HTML files for error responses if the HTTP code is >= 400 */
 		curl_ret = curl_easy_setopt(curl, CURLOPT_FAILONERROR, 1);
+		if (curl_ret != CURLE_OK) {
+			goto exit;
+		}
+	}
+
+	if (progress_callback) {
+		curl_easy_setopt(curl, CURLOPT_NOPROGRESS, 0L);
+		curl_ret = curl_easy_setopt(curl, CURLOPT_PROGRESSFUNCTION,
+					    progress_callback);
 		if (curl_ret != CURLE_OK) {
 			goto exit;
 		}
--- a/src/lib/progress.c
+++ b/src/lib/progress.c
@@ -20,6 +20,7 @@
 #define _GNU_SOURCE
 
 #include <stdlib.h>
+#include <time.h>
 #include <unistd.h>
 
 #include "log.h"
@@ -124,6 +125,40 @@ void progress_report(double count, double max)
 			last_percentage = percentage;
 			last_step = step.current;
 		}
+	}
+}
+
+int progress_spinner_callback()
+{
+	static int index = -1;
+	static time_t begin = 0;
+	const char spinner[4] = { '|', '/', '-', '\\' };
+
+	if (time(NULL) > begin) {
+		begin = time(NULL);
+		index = (index + 1) % 4;
+		info("\r");
+		info("[%c]", spinner[index]);
+		fflush(stdout);
+	}
+
+	return 0;
+}
+
+void progress_set_spinner(bool status_flag)
+{
+	if (log_get_level() == LOG_DEBUG || !isatty(fileno(stdout))) {
+		return;
+	}
+
+	if (status_flag) {
+		if (progress_function) {
+			progress_report(-1, 100);
+			return;
+		}
+		set_progress_callback(progress_spinner_callback);
+	} else {
+		set_progress_callback(NULL);
 	}
 }
 
--- a/src/verify.c
+++ b/src/verify.c
@@ -1024,10 +1024,13 @@ enum swupd_code verify_main(void)
 		}
 	}
 
+	info("Download missing manifests ...\n");
+	progress_set_spinner(true);
 	ret = add_included_manifests(official_manifest, &all_subs);
 	if (cmdline_option_bundles && !(ret & add_sub_ERR)) {
 		ret = add_included_manifests(official_manifest, &bundles_subs);
 	}
+	progress_set_spinner(false);
 	/* if one or more of the installed bundles were not found in the manifest,
 	 * continue only if --force was used since the bundles could be removed */
 	if (ret || invalid_bundle) {
--- a/src/lib/progress.h
+++ b/src/lib/progress.h
@@ -25,6 +25,17 @@ struct step {
 };
 
 /**
+ * @brief progress callback callback type
+ * @see progress.c
+ */
+typedef int (*progress_callback_fn_t)(void);
+
+/**
+ * @brief This function is used to set the callback on the curl for single file download
+ */
+extern void set_progress_callback(progress_callback_fn_t progress_cb);
+
+/**
  * @brief Progress format callback.
  * @see progress_set_format().
  */
--- a/src/swupd_curl.h
+++ b/src/swupd_curl.h
@@ -63,6 +63,11 @@ typedef void (*swupd_curl_free_cb)(void *data);
 typedef int (*swupd_curl_progress_cb)(void *clientp, int64_t dltotal, int64_t dlnow, int64_t ultotal, int64_t ulnow);
 
 /**
+ * @brief function to turn on/off the spinner
+ */
+void progress_set_spinner(bool status_flag);
+
+/**
  * @brief Test a connection with a server.
  *
  * @param url The url to use to test server connectivity
--- a/test/functional/diagnose/diagnose-also-add.bats
+++ b/test/functional/diagnose/diagnose-also-add.bats
@@ -26,6 +26,7 @@ test_setup() {
 	assert_status_is "$SWUPD_NO"
 	expected_output=$(cat <<-EOM
 		Diagnosing version 10
+		Download missing manifests ...
 		Checking for missing files
 		.* Missing file: .*/target-dir/foo
 		.* Missing file: .*/target-dir/foo/test-file1
@@ -62,6 +63,7 @@ test_setup() {
 	assert_status_is "$SWUPD_OK"
 	expected_output=$(cat <<-EOM
 		Diagnosing version 10
+		Download missing manifests ...
 		Checking for missing files
 		Checking for corrupt files
 		Checking for extraneous files
--- a/test/functional/diagnose/diagnose-basics.bats
+++ b/test/functional/diagnose/diagnose-basics.bats
@@ -36,6 +36,7 @@ test_setup() {
 	assert_status_is "$SWUPD_NO"
 	expected_output=$(cat <<-EOM
 		Diagnosing version 20
+		Download missing manifests ...
 		Checking for missing files
 		.* Missing file: .*/target-dir/baz
 		.* Missing file: .*/target-dir/baz/file_3
--- a/test/functional/diagnose/diagnose-boot-file.bats
+++ b/test/functional/diagnose/diagnose-boot-file.bats
@@ -17,6 +17,7 @@ test_setup() {
 	assert_status_is "$SWUPD_NO"
 	expected_output=$(cat <<-EOM
 		Diagnosing version 10
+		Download missing manifests ...
 		Checking for missing files
 		Checking for corrupt files
 		.* Hash mismatch for file: .*/target-dir/usr/lib/kernel/testfile
--- a/test/functional/diagnose/diagnose-bundles.bats
+++ b/test/functional/diagnose/diagnose-bundles.bats
@@ -33,6 +33,7 @@ test_setup() {
 		Limiting diagnose to the following bundles:
 		 - test-bundle1
 		 - test-bundle2
+		Download missing manifests ...
 		Checking for missing files
 		 -> Missing file: $PATH_PREFIX/common
 		 -> Missing file: $PATH_PREFIX/foo/test-file1
@@ -61,6 +62,7 @@ test_setup() {
 		Warning: Bundle "bad-name" is invalid or no longer available
 		Limiting diagnose to the following bundles:
 		 - test-bundle1
+		Download missing manifests ...
 		Error: One or more of the provided bundles are not available at version 10
 		Please make sure the name of the provided bundles are correct, or use --force to override
 		Diagnose did not fully succeed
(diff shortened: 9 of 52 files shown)


src/manifest.c
```diff
@@ -255,7 +256,11 @@ static int retrieve_manifest(int previous_version, int version, char *component,
 	/* Either we're not on mix or it failed, try curl-ing the file if link didn't work */
 	string_or_die(&filename, "%s/%i/Manifest.%s.tar", globals.state_dir, version, component);
 	string_or_die(&url, "%s/%i/Manifest.%s.tar", globals.content_url, version, component);
-
+	if (first_iteration) {
```
castulo: Instead of printing this here, we can print it from the code where this is called, that way we can control where we want to print this. Consider this scenario:

In an operation we need to download a bunch of manifests, so we need to download the MoM first, when we download the MoM this is going to be set and we will print "Downloading manifest", but it was just the MoM so it was pretty quick. Then we may do some more stuff that prints other output in screen then we start downloading many manifests, at that point since we already retrieved a manifest this message won't be printed. And this is the moment where I'd like to see the message printed cause it may take a while downloading manifests.

Besides @otaviobp is going to complaint because of the use of a static variable 🤣

src/curl.c
```diff
@@ -64,6 +64,27 @@ static void swupd_curl_strerror(CURLcode curl_ret)
 	error("Curl - Download error - (%d) %s\n", curl_ret, curl_easy_strerror(curl_ret));
 }
 
+static char spinner[5] = { '|', '/', '-', '\\', '\\' };
```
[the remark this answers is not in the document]
karthikprabhuvinod: That is required to escape the character

src/curl.c
```diff
@@ -64,6 +64,27 @@ static void swupd_curl_strerror(CURLcode curl_ret)
...
+static int progress_callback(void UNUSED_PARAM *clientp, double UNUSED_PARAM dltotal,
+			     double UNUSED_PARAM dlnow, double UNUSED_PARAM ultotal, UNUSED_PARAM double ulnow)
+{
+	static int counter;
+	if (counter % 100000000 == 0) {
+		static int index;
+		info("[%c]", spinner[index % 6]);
+		fflush(stdout);
+		index++;
+		info("\b\b\b");
```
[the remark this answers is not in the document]
karthikprabhuvinod: Good point!

src/curl.c
```diff
@@ -784,6 +805,13 @@ CURLcode swupd_curl_set_basic_options(CURL *curl, const char *url, bool fail_on_
 		}
 	}
 
+	/*Set progress callback*/
+	curl_easy_setopt(curl, CURLOPT_NOPROGRESS, 0L);
```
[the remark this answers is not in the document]
karthikprabhuvinod: Ya that makes sense. @otaviobp gave a similar idea.

src/curl.c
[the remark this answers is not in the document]
castulo: I know, but why '\\\\', '\\\\'? should just be '\\\\''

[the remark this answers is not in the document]
karthikprabhuvinod: ahh yes, I think i added that to test something. Sorry will remove it

src/curl.c
```diff
@@ -64,6 +64,27 @@ static void swupd_curl_strerror(CURLcode curl_ret)
 	error("Curl - Download error - (%d) %s\n", curl_ret, curl_easy_strerror(curl_ret));
 }
 
+static char spinner[5] = { '|', '/', '-', '\\', '\\' };
+
+static int progress_callback(void UNUSED_PARAM *clientp, double UNUSED_PARAM dltotal,
```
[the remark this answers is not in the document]
karthikprabhuvinod: Agreed

src/manifest.c
karthikprabhuvinod: Ya I removed ..will print this elsewhere

src/curl.c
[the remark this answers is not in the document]
karthikprabhuvinod: There is a slight problem with that:

When I do that, i notice a cursor move and flicker which I dont see in case of \b\b\b. Not sure why that happens but. I am checking why

src/curl.c
```diff
@@ -58,12 +59,57 @@ uint64_t total_curl_sz = 0;
...
+{
+#define MAX_THRESHOLD_MULTIPLE_SIX 36000
+
+	if (begin ==0){
+		begin = time(NULL);
+	}
+	
+	if ((time(NULL) - begin) > 0.5) {
+		begin = time(NULL);
+		static int index;
```
otaviobp: I would move this declaration to the top of the function for clarity as it's a static variable.

src/curl.c
```diff
@@ -58,12 +59,57 @@ uint64_t total_curl_sz = 0;
...
+		info("[%c]", spinner[index % 6]);
+		fflush(stdout);
+		index++;
```
otaviobp: index = index + 1 % 6;

So you don't need to have the threshold. And then you don't need to use the % when indexing

src/curl.c
```diff
@@ -58,12 +59,57 @@ uint64_t total_curl_sz = 0;
...
 /* Pretty print curl return status */
 static void swupd_curl_strerror(CURLcode curl_ret)
 {
 	error("Curl - Download error - (%d) %s\n", curl_ret, curl_easy_strerror(curl_ret));
 }
 
+static char spinner[5] = { '|', '/', '-', '\\'};
+static time_t begin = 0;
+
+static int progress_spinner_callback(void UNUSED_PARAM *clientp, double UNUSED_PARAM dltotal,
```
otaviobp: Move this function to progress.c

src/curl.c
```diff
@@ -58,12 +59,57 @@ uint64_t total_curl_sz = 0;
...
+		if (index == MAX_THRESHOLD_MULTIPLE_SIX){
+			index=0;
+		}
+
+	}
+
+	return 0;
+}
+
+void set_spinner_callback(bool onflag) {
```
otaviobp: instead of a boolean here you can get the function that is going to be used for the callback. And instead of calling it set_spinner_callback() you can call it set_progress_callback(). The logic of the function can be the same.
The good thing about this change is that curl will be more generic to support any kind of progress callbacks that we may need in the future. What if we wan't to do more than just printing the spinner for other parts of the code?

otaviobp: Then you can create a 

    void progress_set_spinner(bool onflag);

on progress.c that will call set_progress_callback(progress_spinner_callback) or set_progress_callback(NULL) to turn it on or off.

src/curl.c
```diff
@@ -784,6 +830,15 @@ CURLcode swupd_curl_set_basic_options(CURL *curl, const char *url, bool fail_on_
 		}
 	}
 
+	if (spinner_callback_enable){
```
otaviobp: nice

src/curl.c
karthikprabhuvinod: Done

src/curl.c
karthikprabhuvinod: Done

castulo:
> Some improvements suggested.
> And now another challenge. We need to have a good output in the case the use runs with --debug or outputs that to a file.
> 
> @castulo Any opinions about what we should do? I would just hide the spinner in those cases

yes, kidding the spinner may be the best idea

src/curl.c
karthikprabhuvinod: Good idea

karthikprabhuvinod: Done

src/curl.c
karthikprabhuvinod: Done

src/curl.c
```diff
@@ -58,12 +59,57 @@ uint64_t total_curl_sz = 0;
...
+static int progress_spinner_callback(void UNUSED_PARAM *clientp, double UNUSED_PARAM dltotal,
+			     double UNUSED_PARAM dlnow, double UNUSED_PARAM ultotal, UNUSED_PARAM double ulnow)
+{
+#define MAX_THRESHOLD_MULTIPLE_SIX 36000
+
+	if (begin ==0){
+		begin = time(NULL);
+	}
+	
+	if ((time(NULL) - begin) > 0.5) {
```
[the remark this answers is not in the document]
karthikprabhuvinod: Actually I dont think I need this. Looking to slow down things actually :P

karthikprabhuvinod:
@otaviobp I think now the spinner works only in non-debug mode & wont print to a file. If you like it so far, I can start taking a look at the tests

src/curl.c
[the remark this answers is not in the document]
karthikprabhuvinod: @castulo Not sure why, but the cursor keeps flickering.

src/lib/progress.c
```diff
@@ -127,6 +131,40 @@ void progress_report(double count, double max)
 	}
 }
 
+int progress_spinner_callback()
+{
+	if (log_get_level() == LOG_DEBUG || !isatty(fileno(stdout))) {
+		return 0;
+	}
+
+	static int index;
```
otaviobp: you need to initialize that

src/lib/progress.c
```diff
@@ -127,6 +131,40 @@ void progress_report(double count, double max)
...
+	if (begin == 0) {
+		begin = time(NULL);
+	}
+
+	if (time(NULL) > begin) {
+		begin = time(NULL);
+		index = (index + 1) % 6;
+		info("[%c]", spinner[index]);
+		fflush(stdout);
+		index++;
```
otaviobp: you are increasing index twice

src/lib/progress.c
```diff
@@ -127,6 +131,40 @@ void progress_report(double count, double max)
...
 
+int progress_spinner_callback()
+{
+	if (log_get_level() == LOG_DEBUG || !isatty(fileno(stdout))) {
+		return 0;
+	}
+
+	static int index;
+
+	if (begin == 0) {
```
otaviobp: if you initialize begin here you will start printing the spinner after one second. you can just remove this if because time(NULL) will be always > 0 and begin will be initialized later

src/lib/progress.c
```diff
@@ -31,6 +32,9 @@ static progress_fn_t progress_function = NULL;
 static start_fn_t start_function = NULL;
 static end_fn_t end_function = NULL;
 
+static char spinner[5] = { '|', '/', '-', '\\' };
+static time_t begin = 0;
```
otaviobp: Can't begin be a static inside the spinner function, like index?

src/lib/progress.c
otaviobp: Well, it will be 0 on 19 January 2038, but I think it's fine. hahaha

src/curl.c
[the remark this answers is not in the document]
otaviobp: Visually for me the best is this:

		info("\r");
		info("[%c]", spinner[index]);
 		fflush(stdout); 

Because the cursor will be after the message and not before.

src/lib/progress.c
karthikprabhuvinod: Yes I need to initialize this to -1

src/lib/progress.c
karthikprabhuvinod: Oops..yes no wonder why i was seein only two patterns

src/lib/progress.c
karthikprabhuvinod: I am not sure why did I get an error the first time I tried, but works now :confused:

src/curl.c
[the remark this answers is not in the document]
karthikprabhuvinod: I think that i like that idea..because even if user types something by mistake, the bar does not overwritten.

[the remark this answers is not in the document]
karthikprabhuvinod: Done

karthikprabhuvinod:
`[
{ "type" : "start", "section" : "verify" },
{ "type" : "progress", "currentStep" : 1, "totalSteps" : 7, "stepCompletion" : 100, "stepDescription" : "get_versions" },
{ "type" : "info", "msg" : "Diagnosing version 30940 " },
{ "type" : "progress", "currentStep" : 2, "totalSteps" : 7, "stepCompletion" : 100, "stepDescription" : "cleanup_download_dir" },
{ "type" : "info", "msg" : "Download missing manifests ... " },
{ "type" : "progress", "currentStep" : 3, "totalSteps" : 7, "stepCompletion" : -1, "stepDescription" : "load_manifests" },
{ "type" : "progress", "currentStep" : 3, "totalSteps" : 7, "stepCompletion" : 100, "stepDescription" : "load_manifests" },
{ "type" : "progress", "currentStep" : 4, "totalSteps" : 7, "stepCompletion" : 100, "stepDescription" : "consolidate_files" },
{ "type" : "info", "msg" : " Checking for missing files " },
{ "type" : "progress", "currentStep" : 5, "totalSteps" : 7, "stepCompletion" : 0, "stepDescription" : "add_missing_files" },
{ "type" : "progress", "currentStep" : 5, "totalSteps" : 7, "stepCompletion" : 1, "stepDescription" : "add_missing_files" },
{ "type" : "progress", "currentStep" : 5, "totalSteps" : 7, "stepCompletion" : 2, "stepDescription" : "add_missing_files" },
{ "type" : "progress", "currentStep" : 5, "totalSteps" : 7, "stepCompletion" : 3, "stepDescription" : "add_missing_files" },
{ "type" : "progress", "currentStep" : 5, "totalSteps" : 7, "stepCompletion" : 4, "stepDescription" : "add_missing_files" },`

karthikprabhuvinod:
```console
[
{ "type" : "start", "section" : "verify" },
{ "type" : "progress", "currentStep" : 1, "totalSteps" : 7, "stepCompletion" : 100, "stepDescription" : "get_versions" },
{ "type" : "info", "msg" : "Diagnosing version 30940 " },
{ "type" : "progress", "currentStep" : 2, "totalSteps" : 7, "stepCompletion" : 100, "stepDescription" : "cleanup_download_dir" },
{ "type" : "info", "msg" : "Download missing manifests ... " },
{ "type" : "progress", "currentStep" : 3, "totalSteps" : 7, "stepCompletion" : -1, "stepDescription" : "load_manifests" },
{ "type" : "progress", "currentStep" : 3, "totalSteps" : 7, "stepCompletion" : 100, "stepDescription" : "load_manifests" },
{ "type" : "progress", "currentStep" : 4, "totalSteps" : 7, "stepCompletion" : 100, "stepDescription" : "consolidate_files" },
{ "type" : "info", "msg" : " Checking for missing files " },
{ "type" : "progress", "currentStep" : 5, "totalSteps" : 7, "stepCompletion" : 0, "stepDescription" : "add_missing_files" },
{ "type" : "progress", "currentStep" : 5, "totalSteps" : 7, "stepCompletion" : 1, "stepDescription" : "add_missing_files" },
{ "type" : "progress", "currentStep" : 5, "totalSteps" : 7, "stepCompletion" : 2, "stepDescription" : "add_missing_files" },
{ "type" : "progress", "currentStep" : 5, "totalSteps" : 7, "stepCompletion" : 3, "stepDescription" : "add_missing_files" },
{ "type" : "progress", "currentStep" : 5, "totalSteps" : 7, "stepCompletion" : 4, "stepDescription" : "add_missing_files" },

```

karthikprabhuvinod:
Thats output i get for sudo ./swupd diagnose --json-output

src/verify.c
```diff
@@ -1024,10 +1024,13 @@ enum swupd_code verify_main(void)
 		}
 	}
 
+	info("Download missing manifests ...\n");
```
castulo: I know I'm late to this review, but shouldn't this be "Downloading missing manifests..." instead of "Download..."?

otaviobp: +1 to change to downloading. @karthikprabhuvinod Can you change that?

karthikprabhuvinod: Agreed.. Will do

PR #1091, opened 2019-09-23
List

otaviobp:


src/fullfile.c
```diff
@@ -42,12 +42,7 @@ static int compare_fullfile(const void *a, const void *b)
 	}
 
 	/* they have the same hash, now let's check the version */
-	if (file1->last_change == file2->last_change) {
-		/* we have a match */
-		return 0;
-	}
-
-	return -1;
+	return file1->last_change - file2->last_change;
```
karthikprabhuvinod: We will end up changing the expectation of this function. wont we return a postive no if file1->last_change is bigger?

otaviobp: Yes and that's exactly what I'm suggesting. Before we would return the same value if file1->last_change was bigger or greater than file2->last_change. I'm suggesting to change that so we would return a value that would always put a file that has a last_change lower as before a file that has a last_change that is bigger.

This won't affect the usage of this function so far

karthikprabhuvinod: Okay sounds good.

PR #1094, opened 2019-09-24
list: List deduplicate wasn't deduplicating elements with more than 2…

otaviobp:
… elements

When a duplicated element was found it was removed and then the iterator was
moved to next element. So the original element was never checked again. So if
there were more than 2 identical elements in the list some duplications were left
behind.

Also adding a unit test to validate that the fix is working.

### DIFF
--- a/src/lib/list.c
+++ b/src/lib/list.c
@@ -324,15 +324,17 @@ struct list *list_deduplicate(struct list *list, comparison_fn_t comparison_fn, 
 	struct list *iter = NULL;
 	void *item1, *item2;
 
-	list = list_head(list);
+	iter = list_head(list);
 
-	for (iter = list; iter && iter->next; iter = iter->next) {
+	while (iter && iter->next) {
 
 		item1 = iter->data;
 		item2 = iter->next->data;
 
 		if (comparison_fn(item1, item2) == 0) {
 			list_free_item(iter->next, list_free_data_fn);
+		} else {
+			iter = iter->next;
 		}
 	}
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -341,6 +341,7 @@ BATS = \
 UNIT_TESTS = \
 	test/unit/test_signature.test \
 	test/unit/test_strings.test \
+	test/unit/test_list.test \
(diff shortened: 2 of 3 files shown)


castulo:
Ha, I had already fixed this too in PR #1058

src/lib/list.c
```diff
@@ -324,15 +324,17 @@ struct list *list_deduplicate(struct list *list, comparison_fn_t comparison_fn,
 	struct list *iter = NULL;
 	void *item1, *item2;
 
-	list = list_head(list);
+	iter = list_head(list);
```
castulo: I basically fixed this function the same way here https://github.com/clearlinux/swupd-client/pull/1058/commits/31bf861bb6f36f6d9a60e27a71b6b748a7be5e55, but I think you are missing one thing, I think you need to move the list pointer to the head of the list, the head of the list is the only position we know for sure it won't be removed, otherwise if list is pointing to an element that ends up being removed we will loose the pointer.

karthikprabhuvinod: @castulo aren't we returning the list at the end, so that should be fine! I haven't understood why we do a list_head(list)... doesn't list always need to address of the head element?

karthikprabhuvinod: Sorry realized this was just closed

castulo: > @castulo aren't we returning the list at the end, so that should be fine! I haven't understood why we do a list_head(list)... doesn't list always need to address of the head element?

In this case `list` is a pointer to a list, but that pointer can move across the list, so whenever the function receives the pointer it can be pointing to any element of the list, not necessarily the firs element (head). so let's say at the moment the function was called the pointer was pointing to the 3rd element of the list, if that was a duplicated element and we end up removing it when we return the `list` pointer it will be pointing to NULL, because the data got removed

karthikprabhuvinod: @castulo : Thanks ..so i am guessing it's best to reset it to head of the list before we start deduplicating.

castulo: in this case the first element of the list will never get removed, so we can just return that element

PR #1096, opened 2019-09-24
Fixing missing json messages while file redirect

karthikprabhuvinod:

### DIFF
--- a/src/lib/progress.c
+++ b/src/lib/progress.c
@@ -147,26 +147,22 @@ static int progress_spinner_callback(void)
 
 static void progress_spinner_print_end(void)
 {
-	info("\r[*]\n");
+	info("\r   \r");
 	fflush(stdout);
 }
 
 void progress_set_spinner(bool status_flag)
 {
-	if (log_get_level() == LOG_DEBUG || !isatty(fileno(stdout))) {
-		return;
-	}
-
 	if (status_flag) {
 		if (progress_function) {
 			progress_report(-1, 100);
-		} else {
+		} else if (isatty(fileno(stdout)) && !(log_get_level() == LOG_DEBUG)) {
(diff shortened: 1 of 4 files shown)


src/lib/progress.c
```diff
@@ -153,20 +153,20 @@ static void progress_spinner_print_end(void)
...
-	if (log_get_level() == LOG_DEBUG || !isatty(fileno(stdout))) {
+	if (log_get_level() == LOG_DEBUG) {
 		return;
 	}
 
 	if (status_flag) {
 		if (progress_function) {
 			progress_report(-1, 100);
-		} else {
+		} else if (isatty(fileno(stdout))) {
```
otaviobp: can you do the same on the case of --debug?

if you run

    swupd diagnose --debug --json-output

we should print the -1 too

karthikprabhuvinod: Done

karthikprabhuvinod:
@otaviobp : Fixed it

src/lib/progress.c
```diff
@@ -153,20 +153,20 @@ static void progress_spinner_print_end(void)
...
 {
-	if (log_get_level() == LOG_DEBUG || !isatty(fileno(stdout))) {
-		return;
-	}
-
 	if (status_flag) {
 		if (progress_function) {
 			progress_report(-1, 100);
-		} else {
+		} else if (log_get_level() == LOG_DEBUG) {
```
otaviobp: It works now, but it's clearer if you use the same if for log_debug and isatty. So you don't need a return here.

karthikprabhuvinod: Done

karthikprabhuvinod:
@castulo : Please take a look and let me know if that makes sense
@otaviobp : Please take a look

PR #1098, opened 2019-09-25
Fix os-install progress pts

karthikprabhuvinod:


src/verify.c
```diff
@@ -263,7 +263,9 @@ static int get_required_files(struct manifest *official_manifest, struct list *s
 
 	progress_set_next_step("download_fullfiles");
 	print("\n");
+	progress_set_spinner(true);
```
castulo: why are you setting the spinner here? we do have a progress report for downloading fullfiles

karthikprabhuvinod: I have slightly changed the implementation. Not setting the spinner anymore because it wont use curl sync download feature. so the spinner wont work. Instead I am thinking of implementing this in the form of reporting an indeterminate initially.

src/lib/progress.c
```diff
@@ -169,6 +169,13 @@ void progress_set_spinner(bool status_flag)
 	}
 }
 
+void progress_report_interdeterminate()
```
castulo: "interdeterminate"? is that intended? or did you mean indeterminate?

src/fullfile.c
```diff
@@ -180,7 +180,6 @@ int download_fullfiles(struct list *files, int *num_downloads)
 		if (file->is_deleted || file->do_not_update) {
 			continue;
 		}
-
```
castulo: why did you remove this blank line? 🤔

src/lib/progress.h
```diff
@@ -95,6 +95,8 @@ struct step progress_get_step(void);
  */
 void progress_complete_step(void);
 
+void progress_report_interdeterminate(void);
```
castulo: this declaration is missing its explanation

src/verify.c
```diff
@@ -252,6 +252,7 @@ static int get_required_files(struct manifest *official_manifest, struct list *s
 	if (cmdline_option_install) {
 		progress_set_next_step("download_packs");
 		print("\n");
+		progress_report_interdeterminate();
```
castulo: so this means you are always going to send a -1 followed by actual percentages when downloading packs and/or fullfiles?

src/lib/progress.c
karthikprabhuvinod: Ohh Thats a typo...indeterminate :P

karthikprabhuvinod:
this is the link to the gist

[https://gist.github.com/karthikprabhuvinod/609a5bce99774e87cf025192b3ce4288](url)

src/verify.c
karthikprabhuvinod: Yes, but this does not work really well for downloading packs as that one gets stuck on p_thread_join in last return statement.

So I was initially thinking of not tackling this one.

karthikprabhuvinod: Is it okay for a step to report someting like

{ "type" : "progress", "currentStep" : 5, "totalSteps" : 8, "stepCompletion" : -1, "stepDescription" : "download_packs" }
{ "type" : "progress", "currentStep" : 5, "totalSteps" : 8, "stepCompletion" : 12, "stepDescription" : "download_packs" },
{ "type" : "progress", "currentStep" : 5, "totalSteps" : 8, "stepCompletion" : 25, "stepDescription" : "download_packs" },
{ "type" : "progress", "currentStep" : 5, "totalSteps" : 8, "stepCompletion" : 37, "stepDescription" : "download_packs" },
{ "type" : "progress", "currentStep" : 5, "totalSteps" : 8, "stepCompletion" : 50, "stepDescription" : "download_packs" },
{ "type" : "progress", "currentStep" : 5, "totalSteps" : 8, "stepCompletion" : 62, "stepDescription" : "download_packs" },
 { "type" : "progress", "currentStep" : 5, "totalSteps" : 8, "stepCompletion" : 75, "stepDescription" : "download_packs" },
.
.
.
.

 { "type" : "progress", "currentStep" : 5, "totalSteps" : 8, "stepCompletion" : 100, "stepDescription" : "download_packs" }

src/fullfile.c
karthikprabhuvinod: My bad

src/lib/progress.h
karthikprabhuvinod: Done

src/verify.c
otaviobp: Can't we use the spinner here? So cmdline users would benefit from that too? Let's try to add spinners on any operation that takes a long time.

karthikprabhuvinod:
So you got the pack extraction part right....We have one more place where we need this. When we download fullfiles, we need to find out if the files exist on the same system(check fullfile.c download_fullfile), we have to add an extra step like you did for pack_extraction called check_files_exists. and report a progress bar around it. 

I have written some code already for it..Let me know if sharing helps

karthikprabhuvinod:
https://github.com/clearlinux/swupd-client/blob/master/src/fullfile.c#L177-L188

karthikprabhuvinod:
[https://github.com/clearlinux/swupd-client/blob/master/src/fullfile.c#L177-L188](url)

src/verify.c
castulo: my concern with this approach is that if we send a -1 and installer is going to start a spinner based of this value then they won't take advantage of the real percentage values of the download because a spinner is going to be started anyway.... I think we should not send the -1 unless we really know we won't be able to track progress (unless installer can automatically switch from a spinner to a real progress bar as soon as they get a valid positive value)

src/lib/progress.h
```diff
@@ -95,6 +95,13 @@ struct step progress_get_step(void);
  */
 void progress_complete_step(void);
 
+/**
+ * @brief Marks the current step as indeterminate (progress = -1%).
+ * This is useful in cases json-output needs to scraped to create a
```
castulo: I think you are missing a word(s) somewhere in this sentence

src/verify.c
karthikprabhuvinod: @castulo Good point, but regarding the switching Mark Horn expressed his doubts of effectively switching between the two.

Another approach;
 If we are to do this cleanly, I think we further break a steps like download_packs into sub-steps like pre_precessing packs , download packs, post_processing_packs etc because we are currently doing more than one thing in a given step. download_packs is more a misnomer

otaviobp:
> So you got the pack extraction part right....We have one more place where we need this. When we download fullfiles, we need to find out if the files exist on the same system(check fullfile.c download_fullfile), we have to add an extra step like you did for pack_extraction called check_files_exists. and report a progress bar around it.

I'm unsure if we need that for fullfiles. That's suppose to be way faster than packs extraction. And this isn't suppose to affect installation too.

otaviobp:
I updated the comment in header

src/verify.c
otaviobp: I have this fixed in my commits

karthikprabhuvinod:
@otaviobp : But it will still check for files on the system even for installation. Maybe we can tackle this part later. But I correctly remember Mark pointing this as well. It is marked in this gist
https://gist.github.com/karthikprabhuvinod/609a5bce99774e87cf025192b3ce4288

src/packs.c
```diff
@@ -270,6 +270,8 @@ int download_subscribed_packs(struct list *subs, struct manifest *mom, bool requ
 		/* no packs needs to be downloaded */
 		info("No packs need to be downloaded\n");
 		progress_complete_step();
+		progress_set_next_step("pack_extraction");
+		progress_complete_step();
```
castulo: why are we showing pack_extraction step if there were no packs downloaded? for consistency so we have the same number of steps regardless of if we download packs or not?

src/packs.c
```diff
@@ -325,7 +327,13 @@ int download_subscribed_packs(struct list *subs, struct manifest *mom, bool requ
...
 	}
 	list_free_list(need_download);
+
+	progress_set_next_step("pack_extraction");
 	info("Finishing packs extraction...\n");
+	progress_report_indeterminate();
 
-	return swupd_curl_parallel_download_end(download_handle, NULL);
+	err = swupd_curl_parallel_download_end(download_handle, NULL);
+	progress_report(1, 1);
```
castulo: progress_complete_step should already send a progress_report(1, 1)

src/verify.c
```diff
@@ -253,11 +253,14 @@ static int get_required_files(struct manifest *official_manifest, struct list *s
 		progress_set_next_step("download_packs");
 		print("\n");
 		get_all_files(official_manifest, subs);
+		progress_complete_step();
 	}
 
 	progress_set_next_step("check_files_hash");
 	print("\n");
 	if (check_files_hash(official_manifest->files)) {
+		progress_set_next_step("");
```
castulo: I don't get this, shy are you setting an empty step here?

src/verify.c
```diff
@@ -253,11 +253,14 @@ static int get_required_files(struct manifest *official_manifest, struct list *s
 		progress_set_next_step("download_packs");
 		print("\n");
 		get_all_files(official_manifest, subs);
+		progress_complete_step();
```
castulo: I think you don't need this progress_complete_step(), the download_packs step should have closed automatically once we are done downloading packs, and then you added a packs_extraction step which should already been completed too by now

src/verify.c
```diff
@@ -267,6 +270,7 @@ static int get_required_files(struct manifest *official_manifest, struct list *s
 	if (ret) {
 		error("Unable to download necessary files for this OS release\n");
 	}
+	progress_complete_step();
```
castulo: this step should have been completed automatically too after downloading the files, I don't think you need this one.

src/verify.c
```diff
@@ -1235,8 +1240,12 @@ enum swupd_code verify_main(void)
 		globals.need_update_boot = true;
 		globals.need_update_bootloader = true;
 		timelist_timer_start(globals.global_times, "Run Scripts");
+		progress_set_next_step("run_postupdate_scripts");
+		progress_report_indeterminate();
 		info("\n");
 		scripts_run_post_update(globals.wait_for_scripts);
+		progress_complete_step();
+		progress_report(1, 1);
```
castulo: remember progress_complete_step does a progress_report(1, 1) already

test/functional/bundleadd/add-json.bats
```diff
@@ -27,45 +27,47 @@ test_setup() {
 		[
 		{ "type" : "start", "section" : "bundle-add" },
 		{ "type" : "info", "msg" : "Loading required manifests... " },
-		{ "type" : "progress", "currentStep" : 1, "totalSteps" : 7, "stepCompletion" : 100, "stepDescription" : "load_manifests" },
-		{ "type" : "progress", "currentStep" : 2, "totalSteps" : 7, "stepCompletion" : 100, "stepDescription" : "consolidate_files" },
-		{ "type" : "progress", "currentStep" : 3, "totalSteps" : 7, "stepCompletion" : 100, "stepDescription" : "check_disk_space_availability" },
+		{ "type" : "progress", "currentStep" : 1, "totalSteps" : 8, "stepCompletion" : 100, "stepDescription" : "load_manifests" },
```
castulo: why aren't we showing a -1 at the beginning of loading manifests?

src/packs.c
otaviobp: yes. just for consistency. I'll change that to an empty step

src/packs.c
otaviobp: will fix

src/verify.c
otaviobp: For consistency. To keep the number of steps constant

src/verify.c
otaviobp: No we dont. But I'll move this in another commit

src/packs.c
castulo: good

mdhorn:
Let me know when I can pull this branch and test integration with the installer.

otaviobp:
I'm closing this pull request because I'll bring back this changes in a new branch soon

PR #1099, opened 2019-09-25
Bundle add performance improvements

### LINKED ISSUE #855: software install too slow (swupd bundle-add)
`swupd bundle-add os-clr-on-clr-dev`
Lasts more than 15 minutes.
I noticed that when extracting the process has a processing of up to 2-3% CPU

```
cat /etc/os-release
NAME="Clear Linux OS"
VERSION=1
ID=clear-linux-os
ID_LIKE=clear-linux-os
VERSION_ID=28340
PRETTY_NAME="Clear Linux OS"
ANSI_COLOR="1;35"
HOME_URL="https://clearlinux.org"
SUPPORT_URL="https://clearlinux.org"
BUG_REPORT_URL="mailto:dev@lists.clearlinux.org"
PRIVACY_POLICY_URL="http://www.intel.com/privacy"
5.0.2-717.native #1 SMP Wed Mar 13 23:07:41 UTC 2019 x86_64 GNU/Linux
```

```
swupd bundle-add os-clr-on-clr-dev
Downloading packs...

Extracting os-clr-on-clr-dev pack for version 28330
        ...0%
Extracting editors-dev pack for version 28340

Extracting go-basic-dev pack for version 28230
        ...1%
Extracting kvm-host-dev pack for version 28340

Extracting devpkg-libvirt pack for version 27960

Extracting os-clr-on-clr pack for version 28270
        ...2%
Extracting dev-utils-dev pack for version 28340
        ...3%
Extracting devpkg-metrics-discovery pack for version 28180

Extracting editors pack for version 28030

Extracting emacs pack for version 27210

Extracting joe pack for version 27210

Extracting vim pack for version 28340
        ...4%
Extracting go-basic pack for version 28340

Extracting koji pack for version 28270
        ...5%
Extracting web-server-basic pack for version 27210

Extracting httpd pack for version 28290

Extracting nginx pack for version 28150

Extracting postgresql pack for version 28290
        ...6%
Extracting kvm-host pack for version 28290

Extracting polkit pack for version 27960
...
```

otaviobp:
Fixes #855


src/search_file.c
```diff
@@ -276,25 +279,13 @@ static bool file_matches(const char *filename, const char *search_term)
...
-		return -ENOENT;
-	}
-
-	m = load_manifest(manifest_file->last_change, manifest_file, mom, false, NULL);
-	if (!m) {
-		return -ENOENT;
-	}
-
-	for (l = m->files; l; l = l->next) {
+	for (l = manifest->files; l; l = l->next) {
```
castulo: we've had similar discussions before, but wouldn't it be a good idea to ue list_head to make sure we look at the head of the list? or should we trust the caller?

src/lib/list.c
```diff
@@ -344,6 +344,26 @@ struct list *list_deduplicate(struct list *list, comparison_fn_t comparison_fn,
...
 
+struct list *list_filter_elements(struct list *list, filter_fn_t filter_fn, list_free_data_fn_t list_free_data_fn)
+{
+	struct list *iter, *next;
+
+	iter = list = list_head(list);
+	while (iter) {
+		next = iter->next;
+		if (!filter_fn(iter->data)) {
+			list_free_item(iter, list_free_data_fn);
```
castulo: I think you are going to loose your pointer here because you are freeing the list item

src/update.c
```diff
@@ -255,6 +201,81 @@ static void save_manifest(int version)
 	free_string(&original);
 }
 
+/* Checks to see if the given file is installed under the path_prefix. */
+static bool is_installed_and_verified(struct file *file)
+{
+	/* Not safe to perform the hash check if there was a type change
+	 * involving symlinks. */
+	if (file->is_link != file->peer->is_link) {
+		return false;
```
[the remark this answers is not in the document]
otaviobp: I'm just moving this function around. If it wasn't needed I'll just assume we don't need this for now

src/search_file.c
otaviobp: We need to define who is in charge on keeping the list pointer to head. There are millions of situations where we don't call list_head() and I personally don't think it's a good idea to call it every time just in case. 

My suggestion is:
 - All functions from list.h should work if you pass a pointer to any position of the list. (so we should always call list_head).
 - All functions from list that is suppose to return a list should return the first element of the list. If a function return one item of the list we can return that item (list_append_data, list_remote_item, for example)
 - All functions that receives a list as parameter should expect to receive the first element of the list. 
 - All functions that returns a list should return the first element of the list.

By doing that we are always going to be pointing to the first element of the list. We just need the list_head in the first place because of functions that were returning a list pointing to random positions of the list.

By doing that we are going to be closer (but not there yet) to more used implementations of double linked lists. I would still prefer that functions from list API didn't call list_head, but this would be a big change and not much gain.

If you all agree I can attack this in a new PR.

src/lib/list.c
otaviobp: I save the pointer in next because of that.

otaviobp:
Rebased again on master

src/update.c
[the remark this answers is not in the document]
castulo: ok, sounds good

src/search_file.c
castulo: I like the idea, let's do it

src/lib/list.c
castulo: but in this case this condition is never going to be true `if (item == list)` because you are freeing the pointer before comparing it

castulo: so that means that if you filter out the first element of the list list will be pointing to garbage

otaviobp: When you free a pointer you don't change the value of the pointer. And you are correct that the pointer will be pointing to garbage. So accessing the value pointed by pointer will create a invalid memory access.

But in this case I'm comparing the value of the pointer that hasn't changed after a free. And this is fine. The test test_list_filter_elements() cover that in the test where I add several 'X' in the begging of the list before filtering. Try to debug that and you'll see that the value is fine and it's ok to compare them.

castulo: ahhh I see now, thanks for the explanation

test/functional/bundleadd/add-multiple.bats
```diff
@@ -22,12 +22,10 @@ test_setup() {
...
 		Loading required manifests...
-		Downloading packs for:
-		 - test-bundle1
-		 - test-bundle2
-		Finishing packs extraction...
-		No extra files need to be downloaded
-		Installing bundle(s) files...
+		No packs need to be downloaded
+		Starting download of remaining update content. This may take a while...
+		Staging file content
```
castulo: should we add "..." at the end of the actions written in present participle: "Staging file content" and "Installing files"? We do that with most other phrases that use verbs with that form (with some exceptions like "Calling post-update helper scripts").

test/functional/bundleadd/add-overwrite-files.bats
```diff
@@ -35,7 +36,7 @@ test_setup() {
 	# already exists in the target system, it should replace the file if the
 	# file is different
 
-	run sudo sh -c "$SWUPD bundle-add $SWUPD_OPTS test-bundle2"
+	run sudo sh -c "$SWUPD bundle-add $SWUPD_OPTS test-bundle2 --debug"
```
castulo: is this --debug a leftover or you actually wanted that?

test/functional/bundleadd/add-multiple.bats
castulo: or perhaps we should remove the "..." from the other messages?

castulo: Maybe I'm being too picky haha

src/staging.c
```diff
@@ -359,3 +359,65 @@ int rename_all_files_to_final(struct list *updates)
...
+	/*********** rootfs critical section starts ***************************
+NOTE: the next loop calls do_staging() which can remove files, starting a critical section
+which ends after rename_all_files_to_final() succeeds
+	 */
+
+	/* from here onward we're doing real update work modifying "the disk" */
+
+	/* starting at list_head in the filename alpha-sorted updates list
+	 * means node directories are added before leaf files */
+	info("Staging file content\n");
```
castulo: I like the idea of showing some sort of message here, but I'm uncertain if "Staging file content" is the right message to show to the user. I don't think most users would understand what that means... I even have troubles grasping the concept of staging content.

src/manifest.c
```diff
@@ -1254,3 +1172,110 @@ long get_manifest_list_contentsize(struct list *manifests)
...
+			}
+		}
+		progress_report(complete, total);
+	}
+	progress_report(total, total);
+
+	return last_error;
+}
+
+//struct list *mom_get_installed_manifests(struct manifest *mom)
```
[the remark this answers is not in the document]
otaviobp: it was an old draft. I'll remove those

src/manifest.c
```diff
@@ -1128,21 +1128,14 @@ void deduplicate_files_from_manifest(struct manifest **m1, struct manifest *m2)
 	bmanifest->files = preserver;
 }
 
-struct file *search_bundle_in_manifest(struct manifest *manifest, const char *bundlename)
+int file_bundlename_strcmp(const void *a, const void *b)
```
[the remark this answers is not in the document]
otaviobp: I agree. I go back and forth on the idea to consolidate everything in the same .h or to have them in the .h that makes more sense to them, like file_bundlename_strcmp in a file.h file. I'm inclined to go with a .h consolidating them. What do you think?

test/functional/search/search-content-check-positive.bats
```diff
@@ -57,12 +57,13 @@ global_teardown() {
...
 	assert_status_is 0
 	expected_output=$(cat <<-EOM
+		Downloading all Clear Linux manifests
 		Searching for 'test-bundle1'
 		Bundle test-bundle1 \\(0 MB to install\\)
 		./usr/share/clear/bundles/test-bundle1
 	EOM
 	)
-	assert_regex_is_output "$expected_output"
+	assert_regex_in_output "$expected_output"
```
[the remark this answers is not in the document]
otaviobp: I think that a validation using in_output here is checking better the expected from this search-file than is_output.
By using in_output we are being specific on what we are testing, that in this case is that the bundles were located correctly and in correct order.
We use is_output in most of our tests and this makes tests harder to maintain, because they need to be update if there is any change in output in any other part of the code that is executed by this function. And I'm not convinced that a is_output check in all tests are really testing more than using in_output.
In this specific case the problem is knowing when the cache is or isn't populated to add or not a string "Downloading all Clear Linux manifests" before running the test.
It's possible to do it, but I think it's going to be hard to maintain.

Do you think we are losing something on using in_output here?

src/manifest.h
```diff
@@ -67,6 +67,8 @@ void manifest_free(struct manifest *manifest);
  */
 int mom_get_manifests_list(struct manifest *mom, struct list **manifest_list, filter_fn_t filter_fn);
 
+int recurse_dependencies(struct manifest *mom, const char *bundle, struct list **manifests, filter_fn_t filter_fn);
```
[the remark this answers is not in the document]
otaviobp: oops, this should have a documentation header. I'll add it.

test/functional/bundleadd/add-multiple.bats
otaviobp: no, you're not. I think we should define that. No strong opinion though. Should we add it?

src/staging.c
otaviobp: The major problem here is the progress bar. I was planning to add another one here. I'll just move the installing files to the begging of the function and target the progress in a future PR.

test/functional/bundleadd/add-overwrite-files.bats
otaviobp: leftover. I'll remove it

otaviobp:
Fixed reported problems.
Waiting on feedback for next steps.

src/manifest.c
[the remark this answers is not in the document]
castulo: yeah, I've been contemplating the same two options, and I too think having them in the same file may be easier, so at the moment I'm also inclined towards the first option.

test/functional/search/search-content-check-positive.bats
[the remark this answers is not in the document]
castulo: oh I see, no I think in this case is ok to do it like this.

But, besides special cases like this, in general I do think is better to explicitly use is_output over in_output, I know it makes tests a little bit harder to maintain (not really "harder" to maintain, just more boring to maintain), but that way we make sure the output looks the way we like, if you only tests part of the output in each test you could end up with an output that makes no sense or looks ugly when looking it overall. I'm always very picky on how output looks because the quality of a product is only as good as it is perceived to be by the user, and the output makes a big part of that. Before testlib was in place we were doing it that way, testing only specific parts of the output, and we had many repetitive messages, like multiple warnings for the same thing, etc.

test/functional/bundleadd/add-multiple.bats
castulo: Yeah, let's go with adding the "..." for phrases in present participle for now, if we ended up having too many "..." and that cause the output to look ugly we can reconsider in a future PR.

PR #1102, opened 2019-09-26
Fixes on general usages of stat

otaviobp:
Trying to prevent here to run stat + file operation. Prefer to run the file operation and on errors evaluate the result.

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -311,20 +311,23 @@ bool set_path_prefix(char *path)
 		return false;
 	}
 
-	ret = stat(path, &statbuf);
-	// If path doesn't exit tries to create it
-	if (ret < 0 && errno == ENOENT) {
+	new_path = realpath(path, NULL);
+	if (!new_path) {
+		// If path doesn't exit tries to create it
 		if (mkdir_p(path) < 0) {
 			goto error;
 		}
-		ret = stat(path, &statbuf);
+		new_path = realpath(path, NULL);
 	}
-	if (ret < 0 || !S_ISDIR(statbuf.st_mode)) {
+	if (!new_path) {
 		goto error;
 	}
 
-	new_path = realpath(path, NULL);
-	if (!new_path) {
+	ret = stat(new_path, &statbuf);
+	if (ret < 0 && errno == ENOENT) {
+		goto error;
+	}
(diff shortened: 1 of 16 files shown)


src/lib/sys.c
```diff
@@ -244,6 +255,13 @@ bool file_is_executable(const char *filename)
 	return false;
 }
 
+bool sys_file_is_hardlink(const char *file1, const char *file2)
+{
+	struct stat sb, sb2;
+
+	return !(lstat(file1, &sb) == 0 && lstat(file2, &sb2) == 0 && sb.st_ino == sb2.st_ino);
```
castulo: why are you negating this? if both inodes are the same doesn't that mean it is a hardlink?

src/lib/sys.c
```diff
@@ -348,3 +366,86 @@ bool is_dir(const char *path)
...
+static int sys_rmdir(const char *path)
+{
+	if (rmdir(path) != 0) {
+		return -errno;
+	}
+
+	return 0;
+}
+
+static int sys_rm_dir_recursive(const char *path)
```
castulo: I see the recursiveness will take care of other directories within the directory, but what if any of the directories still have files in it?

src/globals.c
```diff
@@ -311,20 +311,23 @@ bool set_path_prefix(char *path)
...
 	}
-	if (ret < 0 || !S_ISDIR(statbuf.st_mode)) {
+	if (!new_path) {
 		goto error;
 	}
 
-	new_path = realpath(path, NULL);
-	if (!new_path) {
+	ret = stat(new_path, &statbuf);
+	// If path doesn't exit tries to create it
```
castulo: this comment is now in the wrong section of the code

src/lib/sys.c
otaviobp: Wow, this is wrong. Not is wrong here and where I call it, so it's working, but I'll fix that.

src/lib/sys.c
otaviobp: we call sys_rm_recursive() that will remove files too.

src/globals.c
otaviobp: True. I'll move it

otaviobp:
Fixed the reported problems and rebased on master

PR #1111, opened 2019-09-30
Suggestion: Check update

otaviobp:
Suggested change on check-update --verbose. Print extra line with latest version in current format.o

The only 2 cases where this will be different from latest version is in the case of a format bump or
in the case where a format is explicitly stated using -F.

Some samples:

    $ swupd check-update --verbose
    
    current os version: 31130 (format 29) 
    latest server version: 31130 (format 29) 
    latest version in format 29: 31130
    there are no updates available

<BR>

    $ swupd check-update -F staging --verbose
    
    Current OS version: 31140 (format 29) 
    Latest server version: 31130 (format 29) 
    Latest version in format staging: 31140 (format 29) 
    There are no updates available

<BR>

    $ swupd check-update -F 28 --verbose
    
    Current OS version: 31130 (format 29) 
    Latest server version: 31130 (format 29) 
    Latest version in format 28: 30760
    There are no updates available

### DIFF
--- a/src/check_update.c
+++ b/src/check_update.c
@@ -41,6 +41,16 @@ static void print_help(void)
 	global_print_help();
 }
 
+static void print_format(int format)
+{
+	if (format > 0) {
+		info_verbose(" (format %d)", format);
+	} else {
+		info_verbose(" (format unknown)");
+	}
+	info("\n");
+}
+
 /* Return 0 if there is an update available, 1 if not and > 1 on errors. */
 enum swupd_code check_update()
 {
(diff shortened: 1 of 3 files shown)


src/check_update.c
```diff
@@ -93,6 +104,16 @@ enum swupd_code check_update()
 		info("\n");
 	}
 
+	if (latest_version_in_format > 0) {
+		info_verbose("Latest version in format: %d", latest_version_in_format, format_in_format);
```
castulo: You have an extra variable in this interpolation

castulo:
Another option is to show this as:
```
$ sudo ./swupd check-update --verbose
Current OS version: 31090 (format 29)
Latest server version: 31130 (format 29)
Latest version in format 29: 31130
There is a new OS version available: 31130
```

otaviobp:
Implemented it again with your solution. It's slightly different only in the case of the staging.

PR #1113, opened 2019-10-01
Compliant: adding more code style checks

otaviobp:


Makefile.am
```diff
@@ -246,8 +246,8 @@ BATS = \
 	test/functional/os-install/install-multiple.bats \
 	test/functional/os-install/install-no-fullfile-fallback.bats \
 	test/functional/os-install/install-no-packs.bats \
-	test/functional/os-install/install-statedir-cache.bats \
 	test/functional/os-install/install-statedir-cache-offline.bats \
+	test/functional/os-install/install-statedir-cache.bats \
```
castulo: why did you swap these two tests? they were ordered correctly alphabetically

Makefile.am
```diff
@@ -267,10 +267,10 @@ BATS = \
 	test/functional/repair/repair-json.bats \
 	test/functional/repair/repair-missing-file.bats \
 	test/functional/repair/repair-no-disk-space.bats \
-	test/functional/repair/repair-picky.bats \
 	test/functional/repair/repair-picky-downgrade.bats \
 	test/functional/repair/repair-picky-ghosted-missing.bats \
 	test/functional/repair/repair-picky-ghosted.bats \
+	test/functional/repair/repair-picky.bats \
```
castulo: same here, the "ls" command can confirm that

test/code_analysis/compliant.bats
```diff
@@ -0,0 +1,57 @@
...
+	END=$((END - 1))
+	tail -n +$START Makefile.am | head -n $END | sort -c
+}
+
+@test "ANL003: Compliant code style check" {
+	run git diff --quiet --exit-code src
+	# shellcheck disable=SC2154
+	# output and status variable is being assigned and exported by bats
+	if [ "$status" -ne 0 ]; then
+		echo "Error: can only check code style when src/ are clean."
```
castulo: when /src **is** clean?

test/code_analysis/compliant.bats
```diff
@@ -0,0 +1,57 @@
...
+		echo "Stash or commit your changes and try again."
+		return "$status"
+	fi;
```
castulo: you no longer need to keep the ; at the end of the if

test/code_analysis/compliant.bats
```diff
@@ -0,0 +1,57 @@
...
+
+	run clang-format -i -style=file src/*.[ch] src/lib/*.[ch]
+	if [ "$status" -ne 0 ]; then
+		return "$status"
+	fi;
+
+	run git diff --quiet --exit-code src
+	if [ "$status" -ne 0 ]; then \
+		echo "Code style issues found. Run 'git diff' to view issues."; \
```
castulo: same with the ; and \ in this section, no longer needed

Makefile.am
otaviobp: The problem here is that sorting is more complicated than it looks. We need to define how we are going to consider non alphanumeric characters. There are different ways to sort that and I used the one defined by C, where '.' is located after '-'. 

I preferred that because the English sorting order, that is the one you're probably using, was creating an order that didn't look intuitive to me like:

    src/swupd_build_opts.h \
    src/swupd_build_variant.h \
    src/swupd_curl.h \
    src/swupd_curl_internal.h \
    src/swupd_exit_codes.h \
    src/swupd.h \
    src/swupd_internal.h \
    src/swupd_progress.c \
    src/swupd_progress.h \

placing swupd.h in the middle and not on top.

But I'm not picky on this. If you prefer to use English or any other LC_COLLATE I can change that

otaviobp:
Fixed the reported issues except the LC_COLLATE one

test/code_analysis/compliant.bats
otaviobp: fixed

test/code_analysis/compliant.bats
otaviobp: fixed

Makefile.am
castulo: well, is just that I think that when adding tests the easiest way to know the order (without thinking 😄 ) is just run an "ls -l" command in the directory of the new tests...

karthikprabhuvinod: @castulo Lol i had the same problem...I remember I wanted to add a test and didnt know which file to add it ..because didnt know which file had highest test no

castulo: @otaviobp so to me it actually feels more intuitive to keep the alphabetical order because as an example, taking these picky tests (ordered alphabetically):
```
...
repair-picky.bats
repair-picky-downgrade.bats
repair-picky-ghosted.bats
repair-picky-ghosted-missing.bats
```
repair-picky.bats is placed on top of the repair-picky-<VARIANTS>, this makes sense because repair-picky.bats is probably the file with the most basic picky tests, since it has no variant written in the name, while the other files are obviously for more particular cases of picky....

Having said that if you want to keep this order I won't fight it either :P , but you'll have to explain me what files have precedence in this case so I know for future cases.

castulo: > @castulo Lol i had the same problem...I remember I wanted to add a test and didnt know which file to add it ..because didnt know which file had highest test no

btw @karthikprabhuvinod you don't have to add a new test in the file that had the latest test ID, you need to add the new test in the file that makes more sense based on the theme or topic of the file. So in this example again if I was adding a new test that had to do with "--picky" and a ghosted I would probably add it in the repair-picky-ghosted.bats file, regardless of what was the last test ID in that file. You still need to use the next available ID when adding your test, but IDs don't have to be in order within those files. Makes sense?

karthikprabhuvinod: @castulo makes sense. thanks for clarifiying

otaviobp:
Update PR sorting now using en_US.UTF-8

PR #1132, opened 2019-10-03
Fix: To Jsonify helper script output

### LINKED ISSUE #1127: When json format is used we should print the output of post update scripts in a json format
Swupd calls several commands using exec() and we redirect the output of the command to stdout and stderr. Instead we could read it and print using log functions so we would have them printed in a json format.
Another possible approach is redirect every output of exec() to stderr, so it wouldn't clash with json printed in stdout.

karthikprabhuvinod:
Earlier the output of each helper script
run through execve was redirected directly
to the standard-output or standard error.
Hence messages printed like raw strings even if
json-format was selected

This fix ensures that we redirect the output of those
scripts to swupd so that we get a chance to process them
or format them before printing

Fixes #1127

### DIFF
--- a/src/lib/sys.c
+++ b/src/lib/sys.c
@@ -17,6 +17,7 @@
  *
  */
 
+#define _GNU_SOURCE
 #include "sys.h"
 #include "list.h"
 #include "log.h"
@@ -51,6 +52,7 @@ static int replace_fd(int fd, const char *fd_file)
 
 	replaced_fd = dup2(new_fd, fd);
 	close(new_fd);
+
 	return replaced_fd;
 }
 
@@ -59,6 +61,12 @@ int run_command_full_params(const char *stdout_file, const char *stderr_file, ch
 	int pid, ret, child_ret;
 	const char *cmd = params[0];
 	int null_fd;
+	int pipe_fds[2];
+
+	if (pipe2(pipe_fds, O_CLOEXEC) < 0) {
+		error("run_command %s failed to create pipe: \n", cmd);
+		return -1;
+	};
 
 	pid = fork();
 	if (pid < 0) {
@@ -68,6 +76,16 @@ int run_command_full_params(const char *stdout_file, const char *stderr_file, ch
 
 	// Main
 	if (pid > 0) {
+#define MAX_BUFFER_SIZE 1024
+		close(pipe_fds[1]);
+		char buf[MAX_BUFFER_SIZE], *result;
+
+		FILE *fstreamp = fdopen(pipe_fds[0], "r");
+
+		while ((result = fgets(buf, MAX_BUFFER_SIZE, fstreamp)) != NULL) {
+			info("External command: %s", result);
+		}
+
 		while ((ret = waitpid(pid, &child_ret, 0)) < 0) {
 			if (errno != EINTR) {
 				error("run_command %s failed: %i (%s)\n",
@@ -75,11 +93,13 @@ int run_command_full_params(const char *stdout_file, const char *stderr_file, ch
 				return -errno;
 			}
 		}
+		fclose(fstreamp);
 
 		return WEXITSTATUS(child_ret);
 	}
 
 	// Child
+	close(pipe_fds[0]);
 
 	/* replace the stdin with /dev/null. the underlying commands should never be
 	 * run interactively. */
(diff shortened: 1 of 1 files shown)


src/lib/sys.c
```diff
@@ -68,6 +57,14 @@ int run_command_full_params(const char *stdout_file, const char *stderr_file, ch
 
 	// Main
 	if (pid > 0) {
+		close(pipe_fds[1]);
+		char *line = NULL;
+		size_t len = 0;
+
+		while ((getline(&line, &len, pipe_fds[0])) != -1) {
```
otaviobp: getline can't be used with a file descriptor. You need to use read() here.

src/lib/sys.c
```diff
@@ -93,12 +91,16 @@ int run_command_full_params(const char *stdout_file, const char *stderr_file, ch
 	}
 	close(null_fd);
 
-	if (replace_fd(STDOUT_FILENO, stdout_file) < 0) {
```
otaviobp: We still need the replace_fd when stdout_file or stderr_file is set. We need them for:

    run_command_full("file1", "file2", "/usr/bin/ls", "/tmp", NULL);  
    run_command_quiet("/usr/bin/ls", "/tmp", NULL);

src/lib/sys.c
```diff
@@ -93,12 +91,16 @@ int run_command_full_params(const char *stdout_file, const char *stderr_file, ch
...
-		goto error_child;
+	if (stdout_file) {
```
otaviobp: I think you want the opposite. If (!stdout_file) we are printing to stdout, so we want to redirect.

src/lib/sys.c
karthikprabhuvinod: Ya i realized that..got a mighty seg fault 🗡

src/lib/sys.c
karthikprabhuvinod: So if you check the code,

a.) if stdout_file arg is null, replace_fds returns a zero which it will print to STDOUT.
b.) if stdout_file arg is not null, replace_fds returns a replaced fd which points to a file

So I feel it is correct.

src/lib/sys.c
karthikprabhuvinod: done

src/lib/sys.c
karthikprabhuvinod: Done

src/lib/sys.c
karthikprabhuvinod: Done

src/lib/sys.c
```diff
@@ -68,18 +75,27 @@ int run_command_full_params(const char *stdout_file, const char *stderr_file, ch
 
 	// Main
 	if (pid > 0) {
+		close(pipe_fds[1]);
+		char buf[4096];
+
+		while ((read(pipe_fds[0], buf, 4096)) != 0) {
+			info("external command : %s", buf);
```
otaviobp: I'm still unsure on what prefix to use here. @castulo Any idea on this?

src/lib/sys.c
```diff
@@ -101,6 +117,18 @@ int run_command_full_params(const char *stdout_file, const char *stderr_file, ch
 		goto error_child;
 	}
 
+	if (!stdout_file) {
+		if (dup2(pipe_fds[1], STDOUT_FILENO) < 0) {
+			goto error_child;
+		}
+	}
```
otaviobp: you can move the replace_fd into an else clause here to make it clear that we are only replacing the fd if stdout_file != NULL

karthikprabhuvinod: Done

src/lib/sys.c
```diff
@@ -59,6 +60,12 @@ int run_command_full_params(const char *stdout_file, const char *stderr_file, ch
 	int pid, ret, child_ret;
 	const char *cmd = params[0];
 	int null_fd;
+	int pipe_fds[2];
+
+	if (pipe(pipe_fds) < 0) {
```
otaviobp: Use pipe2(pipe_fds, O_CLOEXEC) to prevent leaking the pipe

karthikprabhuvinod: okay

karthikprabhuvinod: Looks like when I build with it, it is just not able to find its declaration

karthikprabhuvinod: 65 |  if (pipe2(pipe_fds, O_CLOEXEC) < 0) {
      |      ^~~~~
      |      pipe
src/lib/sys.c:65:6: warning: nested extern declaration of ‘pipe2’ [-Wnested-externs]
  CCLD     swupd

src/lib/sys.c
```diff
@@ -68,18 +75,28 @@ int run_command_full_params(const char *stdout_file, const char *stderr_file, ch
 
 	// Main
 	if (pid > 0) {
+#define MAX_BUFFER_SIZE 1024
+		close(pipe_fds[1]);
+		char buf[MAX_BUFFER_SIZE];
+
+		while ((read(pipe_fds[0], buf, MAX_BUFFER_SIZE)) != 0) {
```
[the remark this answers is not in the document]
karthikprabhuvinod: Done! Thanks, This is good tip!

src/lib/sys.c
otaviobp: According to the pipe2 manpage you need to

       #define _GNU_SOURCE
       #include <fcntl.h>
       #include <unistd.h>

Adding _GNU_SOURCE should fix this

karthikprabhuvinod: Yes it fixed it..

karthikprabhuvinod: Thanks.. looks like we support both POSIX and GNU linux extensions

src/lib/sys.c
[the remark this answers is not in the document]
karthikprabhuvinod: This is a much better approach but the only downfall I see is somebody earlier trying to parse  one dict has to parse several more. Just makes parsing logic slightly trickier in that case .But the code is more accurate this way.

PR #1136, opened 2019-10-07
Fixing some code errors

castulo:

### DIFF
--- a/src/search_file.c
+++ b/src/search_file.c
@@ -167,6 +167,7 @@ static long get_bundle_size(const char *bundle)
 	}
 
 	b = calloc(1, sizeof(struct bundle_size));
+	ON_NULL_ABORT(b);
 	b->bundle_name = bundle;
 	b->size = compute_bundle_size(bundle);
 	bundle_size_cache = list_prepend_data(bundle_size_cache, b);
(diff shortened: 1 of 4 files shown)


src/search_file.c
```diff
@@ -166,7 +166,10 @@ static long get_bundle_size(const char *bundle)
 		return b->size;
 	}
 
-	b = calloc(1, sizeof(struct bundle_size));
+	b = (struct bundle_size*)calloc(1, sizeof(struct bundle_size));
```
otaviobp: we don't need to cast void*, so it's fine to do:

     	b = calloc(1, sizeof(struct bundle_size));

src/search_file.c
```diff
@@ -166,7 +166,10 @@ static long get_bundle_size(const char *bundle)
...
+	if (!b) {
+		abort();
```
otaviobp: oh, It's better not to call abort here. Use the ON_NULL_ABORT macro

castulo: ahh good call

src/search_file.c
castulo: nice, I was unsure

PR #1138, opened 2019-10-07
Add --path to autoupdate mechanism

### LINKED ISSUE #527: autoupdate does not support --path option
## Problem
`swupd autoupdate` does not accept the `--path=<rootdir>` option like other swupd subcommands: bundle-add, bundle-remove, bundle-list, hashdump, update, verify, check-update, and search. This limits the use of the command to the currently running system.

There is a workaround of using `systemctl` directly, but it would be cleaner to only use the swupd command for consistency and in case systemctl is changed or replaced.

## Use Case
The Clear Linux OS Installer relies on `swupd` for initialization and software installation.

### Enable
`/usr/bin/systemctl --root=<rootdir> unmask --now swupd-update.service swupd-update.timer
`
### Disable
`/usr/bin/systemctl --root=<rootdir> mask --now swupd-update.service swupd-update.timer
`

karthikprabhuvinod:

Fixes #527

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ env:
                 - UPDATE_SUBGROUP2_TOTAL="$((UPDATE_TOTAL - UPDATE_SUBGROUP1_TOTAL))"
                 - UPDATE_SUBGROUP1="$(find test/functional/update -name *.bats | head -n $UPDATE_SUBGROUP1_TOTAL | tr '\n' ' ')"
                 - UPDATE_SUBGROUP2="$(find test/functional/update -name *.bats | tail -n $UPDATE_SUBGROUP2_TOTAL | tr '\n' ' ')"
-                - GROUP1="$UPDATE_SUBGROUP1 $(find test/functional/{bundleinfo,verify-legacy} -name *.bats -printf '%p ')"
+                - GROUP1="$UPDATE_SUBGROUP1 $(find test/functional/{autoupdate,bundleinfo,verify-legacy} -name *.bats -printf '%p ')"
                 - GROUP2="$UPDATE_SUBGROUP2 $(find test/functional/{checkupdate,hashdump,mirror,usability} -name *.bats \( ! -name usa-config-file.bats \) -printf '%p ')"
                 - GROUP3="$(find test/functional/{diagnose,search,os-install,repair} -name *.bats -printf '%p ')"
                 - GROUP4="$(find test/functional/{bundleadd,bundleremove,bundlelist,signature} -name *.bats -printf '%p ')"
@@ -27,7 +27,7 @@ jobs:
                   name: "Static Analysis & Unit Tests"
                   script: make compliant && make shellcheck && sudo sh -c 'umask 0022 && make unit-check' && make docs-coverage
                 - stage: test
-                  name: "Functional Tests - update (group 1), bundle-info, verify-legacy"
+                  name: "Functional Tests - update (group 1), autoupdate, bundle-info, verify-legacy"
                   script: env TESTS="$GROUP1" make -e check
                 - stage: test
                   name: "Functional Tests - update (group 2), checkupdate, hashdump, mirror, usability"
@@ -51,6 +51,7 @@ install:
         - sudo apt-get install doxygen
         - sudo apt-get install check
         - sudo pip install coverxygen
+        - sudo ln -s /bin/systemctl /usr/bin/systemctl
 
         # build bsdiff
         - wget https://github.com/clearlinux/bsdiff/releases/download/v1.0.2/bsdiff-1.0.2.tar.xz
--- a/Makefile.am
+++ b/Makefile.am
@@ -165,6 +165,7 @@ BATS_LOG_DRIVER = $(tap_driver)
 TESTS = $(dist_check_SCRIPTS) $(UNIT_TESTS)
 
 BATS = \
+	test/functional/autoupdate/aup-basic.bats \
 	test/functional/bundleadd/add-alias-basic.bats \
 	test/functional/bundleadd/add-also-add-flag.bats \
 	test/functional/bundleadd/add-bad-hash.bats \
@@ -340,7 +341,6 @@ BATS = \
 	test/functional/usability/usa-external-modules.bats \
 	test/functional/verify-legacy/verify-bundle.bats \
 	test/functional/verify-legacy/verify-flags.bats
-
 
 UNIT_TESTS = \
 	test/unit/test_list.test \
--- a/src/autoupdate.c
+++ b/src/autoupdate.c
@@ -29,56 +29,74 @@
 
 #include "swupd.h"
 
-static void print_help(const char *name)
+static void print_help(void)
 {
 	print("Usage:\n");
-	print("   swupd %s [options]\n\n", basename((char *)name));
-	print("Help Options:\n");
-	print("   -h, --help              Show help options\n");
+	print("   swupd autoupdate [OPTION...]\n\n");
+
+	global_print_help();
+
+	print("Options:\n");
 	print("       --enable            enable autoupdates\n");
 	print("       --disable           disable autoupdates\n");
 	print("\n");
 }
 
-static int enable;
-static int disable;
+#define FLAG_ENABLE 2000
+#define FLAG_DISABLE 2001
+static bool cmdline_option_enable;
+static bool cmdline_option_disable;
 
 static const struct option prog_opts[] = {
 	{ "help", no_argument, 0, 'h' },
-	{ "enable", no_argument, &enable, 1 },
-	{ "disable", no_argument, &disable, 1 },
-	{ 0, 0, 0, 0 }
+	{ "enable", no_argument, 0, FLAG_ENABLE },
+	{ "disable", no_argument, 0, FLAG_DISABLE },
+};
+
+static bool parse_opt(int opt, UNUSED_PARAM char *optarg)
+{
+	switch (opt) {
+	case 'h':
+		print_help();
+		exit(EXIT_SUCCESS);
+	case FLAG_ENABLE:
+		cmdline_option_enable = optarg_to_bool(optarg);
+		return true;
+	case FLAG_DISABLE:
+		cmdline_option_disable = optarg_to_bool(optarg);
+		return true;
+	default:
+		return false;
+	}
+	return false;
+}
+
+static const struct global_options opts = {
+	prog_opts,
+	sizeof(prog_opts) / sizeof(struct option),
+	parse_opt,
+	print_help,
 };
 
 static bool parse_options(int argc, char **argv)
 {
-	int opt;
+	int ind = global_parse_options(argc, argv, &opts);
 
-	while ((opt = getopt_long(argc, argv, "h", prog_opts, NULL)) != -1) {
-		switch (opt) {
-		case '?':
-			print_help(argv[0]);
-			exit(SWUPD_INVALID_OPTION);
-		case 'h':
-			print_help(argv[0]);
-			exit(EXIT_SUCCESS);
-		case 0: /* getopt_long has set the flag */
-			break;
-		default:
-			error("unrecognized option\n\n");
-			goto err;
-		}
+	if (ind < 0) {
+		return false;
 	}
 
-	if (argc > optind) {
+	if (argc > ind) {
 		error("unexpected arguments\n\n");
-		goto err;
+		return false;
+	}
+
+	if (cmdline_option_enable && cmdline_option_disable) {
+		error("Can not use --enable and --disable options at the same time\n\n");
+		return false;
 	}
 
 	return true;
-err:
-	print_help(argv[0]);
-	return false;
 }
 
 static void policy_warn(void)
@@ -89,38 +107,50 @@ static void policy_warn(void)
 
 enum swupd_code autoupdate_main(int argc, char **argv)
 {
+
 	if (!parse_options(argc, argv)) {
+		print_help();
 		return SWUPD_INVALID_OPTION;
 	}
-	if (enable && disable) {
-		error("Can not enable and disable at the same time\n");
-		return SWUPD_INVALID_OPTION;
-	}
-	if (!systemctl_active()) {
+
+	if (!globals.path_prefix && !systemctl_active()) {
 		error("Systemd is inactive - unable to proceed\n");
 		return SWUPD_SUBPROCESS_ERROR;
 	}
 
-	if (enable) {
+	if (cmdline_option_enable) {
 		int rc;
 		check_root();
 		info("Running systemctl to enable updates\n");
-		rc = systemctl_cmd("unmask", "--now", "swupd-update.service", "swupd-update.timer", NULL);
+		rc = systemctl_cmd_path(globals.path_prefix,
+					"unmask", "--now", "swupd-update.service", "swupd-update.timer", NULL);
 		if (rc != 0) {
 			return SWUPD_SUBPROCESS_ERROR;
 		}
-		rc = systemctl_restart("swupd-update.timer");
+
+		/* If globals.path_prefix is given and we have succeeded so far,
+		 * make sense to return SWUPD_OK at this point. In that case, 
+		 * a swupd timer restart cannot be done as there is no
+		 * system daemon to communicate for rootfs.
+		 */
+		if (globals.path_prefix) {
+			warn("Running autoupdate with --path will not restart swupd-update.timer. This will have to be done manually\n");
+			return SWUPD_OK;
+		}
+
+		rc = systemctl_cmd_path(globals.path_prefix, "restart", "swupd-update.timer", NULL);
 		if (rc != 0) {
 			return SWUPD_SUBPROCESS_ERROR;
 		}
 
 		return SWUPD_OK;
-	} else if (disable) {
+	} else if (cmdline_option_disable) {
 		int rc;
 		check_root();
 		policy_warn();
 		info("Running systemctl to disable updates\n");
-		rc = systemctl_cmd("mask", "--now", "swupd-update.service", "swupd-update.timer", NULL);
+		rc = systemctl_cmd_path(globals.path_prefix,
+					"mask", "--now", "swupd-update.service", "swupd-update.timer", NULL);
 		if (rc) {
 			return SWUPD_SUBPROCESS_ERROR;
 		}
@@ -128,14 +158,19 @@ enum swupd_code autoupdate_main(int argc, char **argv)
 	} else {
 		int rc1, rc2;
 		const int STATUS_UNKNOWN = 4;
-		rc1 = systemctl_cmd("is-enabled", "swupd-update.service", NULL);
-		rc2 = systemctl_cmd("is-active", "swupd-update.timer", NULL);
+		rc1 = systemctl_cmd_path(globals.path_prefix, "is-enabled", "swupd-update.service", NULL);
+		rc2 = systemctl_cmd_path(globals.path_prefix, "is-active", "swupd-update.timer", NULL);
 		if (rc1 == SWUPD_OK && rc2 == SWUPD_OK) {
 			print("Enabled\n");
 			return SWUPD_OK;
 		} else if (rc1 >= STATUS_UNKNOWN || rc2 >= STATUS_UNKNOWN) {
 			/* systemctl returns 1,2, or 3 when program dead or not running */
 			error("Unable to determine autoupdate status\n");
+			return SWUPD_SUBPROCESS_ERROR;
+			// Swupd-service is unmasked, static but timer is inactive for --path
+		} else if (rc1 == SWUPD_OK && rc2 != SWUPD_OK && globals.path_prefix) {
+			info("Autoupdate is enabled, but the timer is not running\n");
+			error("Unable to determine autoupdate status with --path option\n");
 			return SWUPD_SUBPROCESS_ERROR;
 		} else {
 			print("Disabled\n");
--- a/src/lib/sys.h
+++ b/src/lib/sys.h
@@ -203,6 +203,12 @@ int sys_rm(const char *filename);
  */
 #define systemctl_cmd(...) run_command_quiet(SYSTEMCTL, __VA_ARGS__)
 
+/**
+ * @brief Run a systemctl command with the informed parameters in a path
+ */
+#define systemctl_cmd_path(path, ...) \
+	path ? run_command_quiet(SYSTEMCTL, "--root", path, __VA_ARGS__) : run_command_quiet(SYSTEMCTL, __VA_ARGS__)
+
 #ifdef __cplusplus
 }
 #endif
--- a/test/functional/README.md
+++ b/test/functional/README.md
@@ -125,6 +125,7 @@ or more of the [assertions](#assertions) provided by the test library.
 
 | Main Business Scenarios                        | Commands      | Group ID |
 | ---------------------------------------------- | ------------- | -------- |
+| Enables/Disables autoupdate                    | auto-update   | AUT      |
 | Install new bundles                            | bundle-add    | ADD      |
 | Uninstall bundles                              | bundle-remove | REM      |
 | List bundles                                   | bundle-list   | LST      |
--- a/test/functional/autoupdate/aup-basic.bats
+++ b/test/functional/autoupdate/aup-basic.bats
@@ -1,0 +1,63 @@
+#!/usr/bin/env bats
+
+# Author: Karthik Prabhu Vinod
+# Email: karthik.prabhu.vinod@intel.com
+
+load "../testlib"
+
+test_setup() {
+
+	if [ -z "$TRAVIS" ]; then
+		skip "This test is intended to run only in Travis (use TRAVIS=true to run it anyway)..."
+	fi
+
+	create_test_environment "$TEST_NAME"
+
+}
+
+test_teardown() {
+
+	if [ -z "$TRAVIS" ]; then
+		return
+	fi
+
+	destroy_test_environment "$TEST_NAME"
+
+}
+
+@test "AUT001: Basic test, Check auto-update enable" {
+
+	# perform autoupdate disable
+	run sudo sh -c "$SWUPD autoupdate --disable --path=$PATH_PREFIX"
+	assert_status_is "$SWUPD_OK"
+	expected_output=$(cat <<-EOM
+		Warning: disabling automatic updates may take you out of compliance with your IT policy
+		Running systemctl to disable updates
+	EOM
+	)
+	assert_is_output "$expected_output"
+
+	# check for files(should exist and point to /dev/null)
+	run sudo sh -c "test -c $PATH_PREFIX/etc/systemd/system/swupd-update.service"
+	assert_status_is "$SWUPD_OK"
+	run sudo sh -c "test -c $PATH_PREFIX/etc/systemd/system/swupd-update.timer"
+	assert_status_is "$SWUPD_OK"
+
+	# perform autoupdate enable
+	run sudo sh -c "$SWUPD autoupdate --enable $SWUPD_OPTS"
+
+	assert_status_is "$SWUPD_OK"
+	expected_output=$(cat <<-EOM
+		Running systemctl to enable updates
+		Warning: Running autoupdate with --path will not restart swupd-update.timer. This will have to be done manually
+	EOM
+	)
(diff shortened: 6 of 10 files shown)


karthikprabhuvinod:
@otaviobp : Please could you provide some early feedback

src/autoupdate.c
```diff
@@ -29,55 +29,75 @@
...
-static int disable;
+static bool enable = false;
+static bool disable = false;
 
 static const struct option prog_opts[] = {
 	{ "help", no_argument, 0, 'h' },
-	{ "enable", no_argument, &enable, 1 },
-	{ "disable", no_argument, &disable, 1 },
-	{ 0, 0, 0, 0 }
+	{ "enable", no_argument, 0, 'e' },
```
otaviobp: We don't need short opts for enable and disable. Take a look at bundle-add for an example

src/autoupdate.c
```diff
@@ -29,55 +29,75 @@
...
+
+	if (argc > ind) {
 		error("unexpected arguments\n\n");
 		goto err;
 	}
 
 	return true;
 err:
-	print_help(argv[0]);
+	print_help();
```
otaviobp: don't call print_help() here.

src/autoupdate.c
```diff
@@ -89,13 +109,18 @@ static void policy_warn(void)
 
 enum swupd_code autoupdate_main(int argc, char **argv)
 {
+	char *root_option;
+
 	if (!parse_options(argc, argv)) {
```
otaviobp: Add a print_help() call here, otherwise `swupd autoupdate --invalid` won't print a help message

src/lib/sys.h
```diff
@@ -203,6 +203,12 @@ int sys_rm(const char *filename);
  */
 #define systemctl_cmd(...) run_command_quiet(SYSTEMCTL, __VA_ARGS__)
 
+/**
+ * @brief Run a systemctl command with the informed parameters in a path
+ */
+#define systemctl_cmd_path(path, ...) \
+	path ? run_command_quiet(SYSTEMCTL, path, __VA_ARGS__) : run_command_quiet(SYSTEMCTL, __VA_ARGS__)
```
otaviobp: Instead of receiving a parameter "--root=my_path", we should receive the path here. And call it

    run_command_quiet(SYSTEMCTL, "--root", path, __VA_ARGS__)

It's a better API to have.

karthikprabhuvinod: Would that work... because systemctl man pages expects its as --root=path

otaviobp: Usually `--param=val` is the same as `--param val`

src/autoupdate.c
karthikprabhuvinod: Done

src/autoupdate.c
karthikprabhuvinod: Done

src/lib/sys.h
karthikprabhuvinod: Tested works! Thanks i didnt know :)

src/autoupdate.c
karthikprabhuvinod: It is good to have short opts isnt it?

otaviobp: We had several problems on conflicts of short opts before. As the number of short opts are limited (we have limited number of letters) we prefer to use it only when needed.

karthikprabhuvinod: ahh I see.

test/functional/autoupdate/aup-basic-tests.bats
```diff
@@ -0,0 +1,41 @@
...
+# Author: Karthik Prabhu Vinod
+# Email: karthik.prabhu.vinod@intel.com
+
+load "../testlib"
+
+test_setup() {
+	create_test_environment "$TEST_NAME"
+
+	# Path to install os-install in
+	run sudo sh -c "sudo mkdir '$TEST_DIRNAME'/testroot"
```
castulo: I don't think you need to create this dir , if you need to install a real bundle into a directory you can install it in "$TARGETDIR" this is supposed to be the target system created by the test environment. However I think this test may be problematic because it is actually installing real bundle so it will need internet connectivity, it is not self contained as the rest are. I understand you need the real bundle so you will have the services there, but that makes me think this test doesn't belong here, maybe belongs to the real content tests https://github.com/clearlinux/swupd-client/tree/master/test/real_content, what do you think @otaviobp

Makefile.am
```diff
@@ -164,6 +164,7 @@ BATS_LOG_DRIVER = $(tap_driver)
 TESTS = $(dist_check_SCRIPTS) $(UNIT_TESTS)
 
 BATS = \
+	test/functional/autoupdate/aup-basic-tests.bats \
```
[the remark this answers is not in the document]
karthikprabhuvinod: make sense cas..will change

src/autoupdate.c
```diff
@@ -29,56 +29,74 @@
...
 	print("\n");
+
+	global_print_help();
 }
 
-static int enable;
-static int disable;
+#define FLAG_ENABLE 1
+#define FLAG_DISABLE 0
+static bool enable;
```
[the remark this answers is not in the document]
karthikprabhuvinod: Done

src/autoupdate.c
```diff
@@ -105,11 +127,20 @@ enum swupd_code autoupdate_main(int argc, char **argv)
...
 		if (rc != 0) {
 			return SWUPD_SUBPROCESS_ERROR;
 		}
-		rc = systemctl_restart("swupd-update.timer");
+
+		/* If root_path is given and we have succeeded so far
+		 * make sense to return SWUPD_OK at this point
+		 */
+		if (root_path) {
+			return SWUPD_OK;
```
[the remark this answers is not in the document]
karthikprabhuvinod: So if we provide root_option, we really cant restart a service because it needs to work with the systemd daemon there. 

So the only thing to be done when we provide root_option is to mask and unmask files.

test/functional/README.md
```diff
@@ -125,6 +125,7 @@ or more of the [assertions](#assertions) provided by the test library.
 
 | Main Business Scenarios                        | Commands      | Group ID |
 | ---------------------------------------------- | ------------- | -------- |
+| Enables/Disables autoupdate                    | auto-update   | AUP      |
```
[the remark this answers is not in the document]
karthikprabhuvinod: Thanks !!

test/functional/autoupdate/aup-basic-tests.bats
```diff
@@ -0,0 +1,41 @@
...
+	run sudo sh -c "$SWUPD os-install --bundles=os-core-update --path='$TEST_DIRNAME'/testroot"
+}
+
+@test "AUP001: Basic test, Check auto-update enable" {
+
+	# Initially the file should be all unmasked as autoupdate is enabled
+	assert_file_not_exists "$TEST_DIRNAME'/testroot/etc/systemd/system/swupd-update.service"
```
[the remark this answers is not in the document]
karthikprabhuvinod: Lol yes!! Sorry my bad!

test/functional/autoupdate/aup-basic-tests.bats
karthikprabhuvinod: Thats a great observation, Thanks for catching that!

src/autoupdate.c
```diff
@@ -29,56 +29,74 @@
 
 #include "swupd.h"
 
-static void print_help(const char *name)
+static void print_help()
 {
-	print("Usage:\n");
-	print("   swupd %s [options]\n\n", basename((char *)name));
 	print("Help Options:\n");
```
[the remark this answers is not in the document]
karthikprabhuvinod: Done ..Thanks!

test/functional/README.md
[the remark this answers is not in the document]
karthikprabhuvinod: Done

test/functional/autoupdate/aup-basic-tests.bats
```diff
@@ -0,0 +1,41 @@
...
+
+	run sudo sh -c "$SWUPD autoupdate --disable --path='$TEST_DIRNAME'/testroot"
+
+	expected_output=$(cat <<-EOM
+		Warning: disabling automatic updates may take you out of compliance with your IT policy
+		Running systemctl to disable updates
+	EOM
+	)
+
+	assert_status_is 0
```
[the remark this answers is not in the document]
karthikprabhuvinod: Done

test/functional/autoupdate/aup-basic-tests.bats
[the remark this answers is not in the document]
karthikprabhuvinod: Done

test/functional/autoupdate/aup-basic-tests.bats
```diff
@@ -0,0 +1,41 @@
...
+	assert_in_output "$expected_output"
```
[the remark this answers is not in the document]
karthikprabhuvinod: Done

karthikprabhuvinod:
@castulo : Thanks for the detailed review.  I have addressed most of the things you asked for and left the others open for @otaviobp to reply on.

test/functional/autoupdate/aup-basic-tests.bats
karthikprabhuvinod: Looks like I cant even get to pass it on my travis without having systemd daemon installed..I getting this failure: https://travis-ci.org/clearlinux/swupd-client/jobs/596251676#L8185

src/autoupdate.c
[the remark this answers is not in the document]
castulo: ok, got it, thanks for the clarification, maybe adding that comment there to explain that in the code for future reference?

test/functional/autoupdate/aup-basic-tests.bats
karthikprabhuvinod: Done with mocking systemctl ! Thanks for the tip

src/autoupdate.c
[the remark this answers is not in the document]
karthikprabhuvinod: Done

karthikprabhuvinod:
@otaviobp @castulo : done please review

src/autoupdate.c
```diff
@@ -29,56 +29,76 @@
...
+
+	print("Options:\n");
 	print("       --enable            enable autoupdates\n");
 	print("       --disable           disable autoupdates\n");
 	print("\n");
 }
 
-static int enable;
-static int disable;
+#define FLAG_ENABLE 1
```
otaviobp: Start at 2000 to avoid clashing with any ascii code or global flags

src/autoupdate.c
```diff
@@ -89,47 +109,63 @@ static void policy_warn(void)
...
 	if (!parse_options(argc, argv)) {
+		print_help();
 		return SWUPD_INVALID_OPTION;
 	}
-	if (enable && disable) {
-		error("Can not enable and disable at the same time\n");
-		return SWUPD_INVALID_OPTION;
+
+	if (globals.path_prefix) {
+		string_or_die(&root_path, "%s", globals.path_prefix);
```
otaviobp: You don't need to duplicate memory for path_prefix. Just use the globals.path_prefix where you used root_path.

test/functional/autoupdate/aup-basic.bats
```diff
@@ -0,0 +1,55 @@
...
+
+	create_test_environment "$TEST_NAME"
+
+	run sudo sh -c "mkdir '$TEST_DIRNAME'/fakepath"
+	# mock systemctl if it doesnt exist to always return true
+	if [ ! -e /usr/bin/systemctl ]; then
+		print "systemctl was not found installed in the system"
+		{
+			printf '#!/bin/bash\n'
+			echo ''
```
otaviobp: echo parameters here so you can check if it has run or not.

test/functional/autoupdate/aup-basic.bats
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bats
+
+# Author: Karthik Prabhu Vinod
+# Email: karthik.prabhu.vinod@intel.com
+
+load "../testlib"
+
+test_setup() {
+
```
otaviobp: Just run this test on if TRAVIS == true because we write files in /usr and we don't want to do it in dev machines.

test/functional/testlib.bash
```diff
@@ -3169,6 +3169,7 @@ get_next_available_id() { # swupd_function
 	id=$((id+1))
 	test_dir=$(basename "$(realpath "$test_dir")")
 	case "$test_dir" in
+		autoupdate) group=AUP;;
```
otaviobp: Why did you change the code?

karthikprabhuvinod: It has to three letter. Thats why. Couldnt think of any other

test/functional/autoupdate/aup-basic.bats
karthikprabhuvinod: Okay

src/autoupdate.c
karthikprabhuvinod: Okay

src/autoupdate.c
karthikprabhuvinod: Okay

test/functional/autoupdate/aup-basic.bats
karthikprabhuvinod: Actually we don't really capture the output of systemctl , so it wont print the parameters.

karthikprabhuvinod:
@otaviobp :So we can turn enabled to disabled, but wont be able to turn disabled to enabled ..for auto-update to be show enabled, swupd-update.timer has to be restarted which we cant for a path

otaviobp:
> @otaviobp :So we can turn enabled to disabled, but wont be able to turn disabled to enabled ..for auto-update to be show enabled, swupd-update.timer has to be restarted which we cant for a path

I think it's safe to ignore the timer if path_prefix != NULL

test/functional/testlib.bash
otaviobp: Just use what we have defined before AUT. I don't know why @castulo picked that but we already have documentation using that name

test/functional/autoupdate/aup-basic.bats
otaviobp: That's true. But we need to find a way to test if the autoupdate is being turned on or not. This test isn't really testing that.

karthikprabhuvinod: For that, we will run this on travis where we install systemd like you said and perform os-install(that will require net access)...but then we can perform real tests.

test/functional/testlib.bash
karthikprabhuvinod: okay

karthikprabhuvinod: Done

test/functional/autoupdate/aup-basic.bats
karthikprabhuvinod: The we can perform real tests.

src/autoupdate.c
karthikprabhuvinod: Done

src/autoupdate.c
karthikprabhuvinod: Done

test/functional/autoupdate/aup-basic.bats
```diff
@@ -0,0 +1,99 @@
...
+	sudo sh -c "ls -l '$PATH_PREFIX'/etc/"
+	sudo sh -c "ls -l '$PATH_PREFIX'/usr/lib"
+
+
+	# perform autoupdate enable
+	run sudo sh -c "$SWUPD autoupdate --enable --path=$PATH_PREFIX"
+	#sudo sh -c "ls -la '$PATH_PREFIX'/etc/systemd/"
+	sudo sh -c "ls -la '$PATH_PREFIX'/usr/lib/systemd/system/"
+	sudo sh -c "ls -la '$PATH_PREFIX'/usr/lib/systemd/system/swupd-update.service"
+	sudo sh -c "ls -la '$PATH_PREFIX'/usr/lib/systemd/system/swupd-update.timer"
```
otaviobp: /usr/lib/systemd/ is not relevant here. Check the /etc.

karthikprabhuvinod: removed that was for test

karthikprabhuvinod:
@otaviobp : Thanks for helping find the issue. Please could you review this?

src/autoupdate.c
```diff
@@ -89,54 +109,71 @@ static void policy_warn(void)
...
+		rc2 = systemctl_cmd_path(globals.path_prefix, "is-active", "swupd-update.timer", NULL);
 		if (rc1 == SWUPD_OK && rc2 == SWUPD_OK) {
 			print("Enabled\n");
 			return SWUPD_OK;
 		} else if (rc1 >= STATUS_UNKNOWN || rc2 >= STATUS_UNKNOWN) {
 			/* systemctl returns 1,2, or 3 when program dead or not running */
 			error("Unable to determine autoupdate status\n");
 			return SWUPD_SUBPROCESS_ERROR;
+			// Swupd-service is unmasked, static but timer is inactive for --path
+		} else if (rc1 == SWUPD_OK && rc2 == (STATUS_UNKNOWN - 1) && globals.path_prefix) {
```
otaviobp: Magic number STATUS_UNKNOWN -1. Is this behavior defined in systemctl documentation?

src/autoupdate.c
```diff
@@ -89,54 +109,71 @@ static void policy_warn(void)
...
+			info("swupd-update.service has been unmasked with --path option, but the timer has not been restarted\n");
```
otaviobp: Change print to: Autoupdate is enabled, but the timer is not running

If you run `swupd autoupdate` before running any other command the service hasn't been unmasked (it was never masked). And no one tried to restart the timer.

src/autoupdate.c
otaviobp: My suggestion on that is just don't run is-active when path_prefix != "/"

test/functional/autoupdate/aup-basic.bats
```diff
@@ -0,0 +1,59 @@
...
+	run sudo sh -c "$SWUPD autoupdate --disable --path=$PATH_PREFIX"
+	assert_status_is "$SWUPD_OK"
+	expected_output=$(cat <<-EOM
+		Warning: disabling automatic updates may take you out of compliance with your IT policy
+		Running systemctl to disable updates
+	EOM
+	)
+	assert_is_output "$expected_output"
+
+	# check for files(should not exist)
```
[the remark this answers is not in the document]
karthikprabhuvinod: Done

test/functional/autoupdate/aup-basic.bats
```diff
@@ -0,0 +1,59 @@
...
+	run sudo sh -c "ls -l '$PATH_PREFIX'/etc/systemd/system/swupd-update.service"
```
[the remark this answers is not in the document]
karthikprabhuvinod: Done

test/functional/autoupdate/aup-basic.bats
```diff
@@ -0,0 +1,59 @@
...
+	assert_status_is "$SWUPD_OK"
+	expected_output=$(cat <<-EOM
+		Running systemctl to enable updates
+		Warning: Running autoupdate with --path will not restart swupd-update.timer. This will have to be done manually
+	EOM
+	)
+	assert_is_output "$expected_output"
+
+	# check for files(should not exist)
+	run sudo sh -c "ls -l '$PATH_PREFIX'/etc/systemd/system/swupd-update.service"
```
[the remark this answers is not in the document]
karthikprabhuvinod: Done

.travis.yml
```diff
@@ -51,6 +51,8 @@ install:
         - sudo apt-get install doxygen
         - sudo apt-get install check
         - sudo pip install coverxygen
+        - sudo apt-get install systemd
```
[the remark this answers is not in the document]
karthikprabhuvinod: Okay will remove this

src/autoupdate.c
karthikprabhuvinod: You are right..changed it

src/autoupdate.c
karthikprabhuvinod: I can do that..problem is the globals.path_prefix has no default value of '/' value in it as we dont perform SWUPD_INIT ad SWUPD_DEINIT in case of autoupdate as it is not really required. Should we?

karthikprabhuvinod: You rightly pointed out ..there is no documentation for isactive for Magic number STATUS_UNKNOWN -1. The only thing they say it returns a number other than zero for in-active.

src/autoupdate.c
```diff
@@ -89,54 +109,71 @@ static void policy_warn(void)
...
+		rc2 = systemctl_cmd_path(globals.path_prefix, "is-active", "swupd-update.timer", NULL);
 		if (rc1 == SWUPD_OK && rc2 == SWUPD_OK) {
 			print("Enabled\n");
 			return SWUPD_OK;
 		} else if (rc1 >= STATUS_UNKNOWN || rc2 >= STATUS_UNKNOWN) {
 			/* systemctl returns 1,2, or 3 when program dead or not running */
 			error("Unable to determine autoupdate status\n");
 			return SWUPD_SUBPROCESS_ERROR;
+			// Swupd-service is unmasked, static but timer is inactive for --path
+		} else if (rc1 == SWUPD_OK && rc2 != SWUPD_OK && globals.path_prefix) {
```
otaviobp: good fix

.travis.yml
```diff
@@ -12,7 +12,7 @@ env:
                 - UPDATE_TOTAL="$(find test/functional/update -name *.bats | wc -l)"
                 - UPDATE_SUBGROUP1_TOTAL=17
                 - UPDATE_SUBGROUP2_TOTAL="$((UPDATE_TOTAL - UPDATE_SUBGROUP1_TOTAL))"
-                - UPDATE_SUBGROUP1="$(find test/functional/update -name *.bats | head -n $UPDATE_SUBGROUP1_TOTAL | tr '\n' ' ')"
+                - UPDATE_SUBGROUP1="$(find test/functional/{autoupdate, update} -name *.bats | head -n $UPDATE_SUBGROUP1_TOTAL | tr '\n' ' ')"
```
castulo: don't add these here, you need to move this in a group not a subgroup

src/autoupdate.c
```diff
@@ -29,56 +29,76 @@
...
-		case 0: /* getopt_long has set the flag */
-			break;
-		default:
-			error("unrecognized option\n\n");
-			goto err;
+	switch (opt) {
+	case 'h':
+		print_help();
+		exit(EXIT_SUCCESS);
+	case FLAG_ENABLE:
```
castulo: I told you this in a previous command but it is better to check flag restrictions within parse_options() as we do in the other commands

test/functional/autoupdate/aup-basic.bats
```diff
@@ -0,0 +1,57 @@
...
+	create_test_environment "$TEST_NAME"
+
+}
+
+test_teardown() {
+
+	if [ -z "$TRAVIS" ]; then
+		return
+	fi
+
```
castulo: you are never cleaning the tea environment , if I run the test with TRAVIS=true the test env will be left behind

test/functional/autoupdate/aup-basic.bats
```diff
@@ -0,0 +1,57 @@
...
+	assert_status_is "$SWUPD_OK"
+	expected_output=$(cat <<-EOM
+		Warning: disabling automatic updates may take you out of compliance with your IT policy
+		Running systemctl to disable updates
+	EOM
+	)
+	assert_is_output "$expected_output"
+
+	# check for files(should exist)
+	assert_file_exists "'$PATH_PREFIX'/etc/systemd/system/swupd-update.service"
```
castulo: why the single quote?

test/functional/autoupdate/aup-basic.bats
```diff
@@ -0,0 +1,57 @@
...
+	if [ -z "$TRAVIS" ]; then
+		return
+	fi
+
+}
+
+@test "AUT001: Basic test, Check auto-update enable" {
+
+	# perform autoupdate disable
+	run sudo sh -c "$SWUPD autoupdate --disable --path=$PATH_PREFIX"
```
castulo: where are you defining PATH_PREFIX?

.travis.yml
castulo: if you check your Travis execution you will notice about 25 tests were not run because of this change

karthikprabhuvinod: Yikes, Yes..I have an extra space. My bad

test/functional/autoupdate/aup-basic.bats
karthikprabhuvinod: Ahh Yes..great catch..neeed to add a destroy

test/functional/autoupdate/aup-basic.bats
karthikprabhuvinod: I thought I  needed it .. doesnt look like. Thanks.

test/functional/autoupdate/aup-basic.bats
karthikprabhuvinod: This is created as part of the test environment ..testlib

.travis.yml
karthikprabhuvinod: Done

test/functional/autoupdate/aup-basic.bats
karthikprabhuvinod: Done

test/functional/autoupdate/aup-basic.bats
karthikprabhuvinod: Done

test/functional/autoupdate/aup-basic.bats
castulo: right I don't know how I missed that, I see it know, thanks

.travis.yml
castulo: please don't add autoupdate to UPDATE_SUBGROUP1 that is ment for update related tests, you want the run as part of worker 1 so add autoupdate to GROUP1

src/autoupdate.c
karthikprabhuvinod: Done

test/functional/autoupdate/aup-basic.bats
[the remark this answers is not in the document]
karthikprabhuvinod: @otaviobp I keep forgetting this, assert_file will not work in this case because the file is symlinked ...I think assert_file does not check for this

.travis.yml
karthikprabhuvinod: Yes corrected it

src/autoupdate.c
karthikprabhuvinod: Done

test/functional/autoupdate/aup-basic.bats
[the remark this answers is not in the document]
karthikprabhuvinod: @castulo and I discussed..looks like there is a test -c function to test for char file..didnt want to change function assert_file_exists

src/autoupdate.c
```diff
@@ -29,56 +29,76 @@
...
+};
+
+static bool parse_opt(int opt, UNUSED_PARAM char *optarg)
+{
+	switch (opt) {
+	case 'h':
+		print_help();
+		exit(EXIT_SUCCESS);
+	case FLAG_ENABLE:
+		cmdline_option_enable = true;
```
castulo: ups, one more thing, instead of directly assigning `true` you need to use the `optarg_to_bool` macro like this `cmdline_option_enable = optarg_to_bool(optarg);` otherwise if those values are added to the config file they won't be parsed correctly.

src/autoupdate.c
```diff
@@ -29,56 +29,76 @@
...
 	}
 
-	if (argc > optind) {
+	if (argc > ind) {
 		error("unexpected arguments\n\n");
-		goto err;
+		return false;
+	}
+
+	if (cmdline_option_enable) {
```
castulo: thanks for making the change. In this case since there are only two conflicting flags I think is better to use just one if statement like:
 `if (cmdline_option_enable && cmdline_option_disable) {`

we can use nested ifs when there are multiple conflicting flags

karthikprabhuvinod: Done

src/autoupdate.c
karthikprabhuvinod: Good catch! Thanks Done..

PR #1144, opened 2019-10-14
Broken tests: Fix broken real_content tests

otaviobp:


src/helpers.c
```diff
@@ -501,7 +501,7 @@ void delete_motd(void)
 
 int get_dirfd_path(const char *fullname)
 {
-	int ret;
+	int ret = -1;
```
castulo: why did you have to revert this?

otaviobp: I reverted the patch because it was introducing one bug in verify that was catch by real_content tests. I reverted all changes in the commit "Fixing some code errors", except things completely unrelated to verify. As code in remove_orphaned_files() was tightly related to get_dirfd_path() I thought it was better to revert both.

The problem can be reproduced by:

    sudo swupd os-install --path ~/path -B os-core -V 31090
    sudo swupd diagnose --path ~/path

There are files that aren't present in the system that are stated as files that should be deleted.

castulo: I see, thanks for the explanation

PR #1146, opened 2019-10-15
Implementing the --recursive flag for bundle-remove

### LINKED ISSUE #455: Implement recursive removal option for bundle-remove
If a user runs `swupd bundle-add` for a bundle that requires new bundles as well (pulled in via bundle includes), it would be helpful if they could `swupd bundle-remove` the same bundle, optionally removing the bundles from the includes list.

This would require a *recursive* bundle remove option, since we would not want to recursively remove the bundles by default.

castulo:
If a bundle has dependencies (**include** or **also-add** bundles) and is removed using the `--recursive` flag, the bundle and all its dependencies should be removed from the system, with the following two exceptions:

- If the dependency is also a dependency of another installed bundle (unless it is also being removed in the same operation). Removing this bundle would break the other installed bundle.
- If the dependency was previously added by the user directly (via bundle-add), because this means the user specifically wanted this bundle installed in the system, to remove this bundle the user would have to specifically remove it via the bundle-remove command.

Tasks:

- [x] Functional tests
- [x] Add `--recursive` flag to bundle-remove (without functionality)
- [x] Add documentation to man pages.
- [x] Add flag to autocompletion scripts swupd.bash and swupd.zsh
- [x] Add dependencies to list of bundles to be removed
- [x] Skip dependency `os-core`
- [x] Skip dependency if tracked (installed by user)
- [x] Skip dependency if required by an installed bundle
- [x] Remove requested bundle and "deletable" dependencies
- [x] Validate integration with other bundle-remove flags
- [x] Check output correctness and consistency with other commands (including output using `--quiet`, `--verbose`, `--debug`)
- [x] Add flag to the config file
- [x] Compare performance of bundle-remove before and after flag implementation
- [x] Test with real content (ad-hoc)
- [x] Update release notes draft.

Closes #455


src/bundle.c
```diff
@@ -575,34 +616,83 @@ enum swupd_code remove_bundles(struct list *bundles)
...
+				list_free_list_and_data(dependencies, manifest_free_data);
+				ret_code = ret;
+			}
+
+			/* sanitize the dependencies and add the user
+			 * provided bundles to the list so they are
+			 * considered when validating the bundle */
+			for (iter = dependencies; iter; iter = iter->next) {
+				struct manifest *m = iter->data;
+				if (is_deletable_dependency(m->component)) {
```
otaviobp: Can't you filter the non deletable dependencies on recursive manifests? Dependencies of non deletable bundles are always non deletable so we don't need to explore them.

src/bundle.c
```diff
@@ -575,34 +616,83 @@ enum swupd_code remove_bundles(struct list *bundles)
...
+					deps = list_prepend_data(deps, m->component);
+				}
+			}
+			deps = list_concat(deps, list_clone(bundles));
+
+			for (iter = deps; iter; iter = iter->next) {
+				char *dep = iter->data;
+				/* if there are dependencies, check if
+				 * they are allowed to be removed */
+				ret = validate_bundle(dep, current_mom, deps, &reqd_by);
```
otaviobp: Some improvements can be made here too:
 - We are checking again if a bundle is not os-core.
 - Is there any chance to get here with a bundle that isn't on MoM?
 - is there any change to get here with a bundle that isn't installed?
 - Can you filter that on recurse_manifests() instead of filtering that later?

src/bundle.c
```diff
@@ -575,34 +616,83 @@ enum swupd_code remove_bundles(struct list *bundles)
...
+			/* sanitize the dependencies and add the user
+			 * provided bundles to the list so they are
+			 * considered when validating the bundle */
+			for (iter = dependencies; iter; iter = iter->next) {
+				struct manifest *m = iter->data;
+				if (is_deletable_dependency(m->component)) {
+					deps = list_prepend_data(deps, m->component);
+				}
+			}
+			deps = list_concat(deps, list_clone(bundles));
```
otaviobp: Just a hit. Usually using list_clone() isn't a good idea. You can always work in the original list if it's a read-only operation. In this case all elements from bundles should be covered by recurse_manifest(), right?

src/bundle.c
otaviobp: We don't need to print the information of all bundles that aren't going to be removed because of --recurse (check output of sudo swupd bundle-remove os-clr-on-clr). This was caused because validate_bundle is calling required_by.

castulo:
> I have made some comments and there is also some problems with the algorithm used to select if a bundle can be removed or not.
> 
> For example, I tested that:
> $ sudo ./swupd os-install -B os-clr-on-clr --path /opt/swupd/root
> $ sudo cp /usr/share/clear/bundles/* /var/lib/swupd/bundles #simulate that all bundles were installed using bundle-add
> $ sudo rm /var/lib/swupd/bundles/vim #except vim
> $ sudo ./swupd bundle-remove os-clr-on-clr --recursive
> ...
> The following bundles are being removed:
> 
> * vim
> * os-clr-on-clr
>   ...
> 
> But vim is also-added by editors, so we can't remove it.
> 
> I also have a strong feeling that is going to be easier to populate a list of bundles to be removed after populating the list of bundles_to_remove and not in the first loop.

when you ran this test, did you also use --path for steps 2, 3 an 4? E.g`$ sudo cp /opt/swupd/root/usr/share/clear/bundles/* /opt/swupd/root/var/lib/swupd/bundles`

otaviobp:
> when you ran this test, did you also use --path for steps 2, 3 an 4? E.g$ sudo cp /opt/swupd/root/usr/share/clear/bundles/* /opt/swupd/root/var/lib/swupd/bundles

Yes, sorry, it wasn't clear in my comments.

castulo:
> > when you ran this test, did you also use --path for steps 2, 3 an 4? E.g$ sudo cp /opt/swupd/root/usr/share/clear/bundles/* /opt/swupd/root/var/lib/swupd/bundles
> 
> Yes, sorry, it wasn't clear in my comments.

Ok, that should not happen, I'll try to recreate the failure you are describing.

castulo:
I just realized I didn't add this flag to the default config file, I'll add it

castulo:
@otaviobp I was able to recreate this behavior but after looking at it closer I think this is working as expected. After following your steps this is the result I got:

```
$ sudo ./swupd bundle-remove --recursive os-clr-on-clr --path /home/castulo/testfs/target --statedir /home/castulo/testfs/target/state
 [|]
The --recursive option was used, the specified bundle and its dependencies will be removed from the system

The following bundles are being removed:
 - vim
 - os-clr-on-clr

Deleting bundle files...
 [100%]

Total deleted files: 1056

Successfully removed 1 bundle
1 bundle that was installed as a dependency was removed
```

Let's analyze what happened. This is what we did:
- we installed os-clr-on-clr (which installs a bunch of dependencies ~311)
- we then marked all installed bundles as "tracked" by copying all files from \<path\>/usr/share/clear/bundles to \<statedir\>/bundles
- then we deleted the tracking file from vim, so this will make swupd think this is the only bundle that was installed as a result of being a dependency
- We ran `bundle-remove --recursive os-clr-on-clr`. This will remove the "os-clr-on-clr" bundle since that was the specified bundle, and "os-clr-on-clr" is not required by any other installed bundle, so we are good for removing it. Then since `--recursive` was used, swupd will check the 311 bundles included by os-clr-on-clr to see if they can be removed as well:
   - All bundles except "vim" were manually installed by the user (tracked files) so swupd skip removing all these
   - vim is not tracked, so swupd proceeds to see if the bundle is required by other bundles. The `editors` bundle also-add this bundle, but it is an optional bundle, so it is ok to remove it.
 - `vim` and `os-clr-on-clr` are removed.

We can confirm this with bundle-info:
```
$ sudo ./swupd bundle-info --path /home/castulo/testfs/target --statedir /home/castulo/testfs/target/state --dependencies editors
__________________________
 Info for bundle: editors
__________________________

Status: Installed

Bundle editors is up to date:
 - Installed bundle last updated in version: 31270
 - Latest available version: 31270

Bundle size:
 - Size of bundle: 692.64 KB
 - Size bundle takes on disk (includes dependencies): 431.51 MB

Direct dependencies (4):
 - os-core
 - emacs (optional, installed)
 - joe (optional, installed)
 - vim (optional, not installed)

Indirect dependencies (6):
 - findutils
 - lib-openssl
 - libstdcpp
 - p11-kit
 - perl-basic
 - python3-basic
```

If I repeat the same steps but using `python3-basic` instead of `vim` I get this:
```
$ sudo ./swupd bundle-remove --recursive os-clr-on-clr --path /home/castulo/testfs/target --statedir /home/castulo/testfs/target/state

The --recursive option was used, the specified bundle and its dependencies will be removed from the system

Bundle "python3-basic" is required by the following bundles:
 - Babel
 - Sphinx
 - <more bundles here>...

The following bundles are being removed:
 - os-clr-on-clr

Deleting bundle files...
 [100%]

Total deleted files: 868

Successfully removed 1 bundle
```

This seems to be working as designed.

otaviobp:
I still think it's a problem. `sudo swupd bundle-remove --recursive os-clr-on-clr` should remove os-clr-on-clr because it "is not required by any other installed bundle", as you stated.
But we shouldn't remove vim, because it is required by an installed bundle, editors. Editors also-add vim so vim shouldn't be removed.

otaviobp:
Not sure if I was clear on the reason where I think this is a problem. I'll describe a set of steps on how this would affect an expected use case:

    $ swupd os-install --path /path
    $ swupd bundle-add --path /path editors
    $ #vim is installed and user is using it
    $ swupd bundle-add --path /path os-clr-on-clr
    $ #user realizes it doesn't need os-clr-on-clr anymore and any other extra dependencies
    $ swupd bundle-remove --path /path os-clr-on-clr
    $ # vim is gone

castulo:
> I still think it's a problem. `sudo swupd bundle-remove --recursive os-clr-on-clr` should remove os-clr-on-clr because it "is not required by any other installed bundle", as you stated.
> But we shouldn't remove vim, because it is required by an installed bundle, editors. Editors also-add vim so vim shouldn't be removed.

I didn't realize "also-add" should be treated as "include", I thought this could be removed. I'll update the PR so also-add bundles are treated as includes as far as `bundle-remove --recursive` is concerned.

otaviobp:
Can you also move all functions that are exclusively needed for bundle removal from bundle.c to bundle_remove.c?

src/bundle.c
```diff
@@ -575,34 +616,83 @@ enum swupd_code remove_bundles(struct list *bundles)
...
+
+		/* move the manifest of the bundle to be removed
+		 * from the list of subscribed bundles to the
+		 * list of bundles to be removed */
 		list_move_item(bundle, &current_mom->submanifests, &bundles_to_remove, find_manifest);
-		info("\nRemoving bundle: %s\n", bundle);
 		remove_tracked(bundle);
+
+		/* if recursive, the dependencies need to be removed too */
+		if (cmdline_option_recursive) {
```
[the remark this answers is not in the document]
castulo: done

src/bundle.c
castulo: done

src/bundle.c
castulo: done

src/bundle.c
castulo: the updated Pr doesn't use list_clone anymore, thanks for the tip.

castulo:
I believe all the reported issues have been addressed and the PR is ready to be reviewed again.

castulo:
I just pushed an update to the release notes draft.
PR is ready for review.

src/bundle.c
```diff
@@ -278,6 +297,85 @@ static enum swupd_code validate_bundle(const char *bundle, struct manifest *mom,
...
+		recurse_dependencies(mom, bundle, &candidates_for_removal, is_deletable_dependency);
+	}
+
+	/* get a list of dependencies still required by bundles
+	 * that won't be removed */
+	struct list *required_bundles = NULL;
+	for (iter = mom->submanifests; iter; iter = iter->next) {
+		m = iter->data;
+		bundle = m->component;
+		if (!list_search(candidates_for_removal, bundle, manifest_bundlename_strcmp)) {
```
otaviobp: Instead of using list_search here can't we use `is_deletable_dependency()`?
And by that we can move this to a function in manifest.h `get_all_tracked_bundles()` (or a better name). This function can then be used to create a command to wipe out every bundle that isn't tracked (or dependency of a tracked bundle)

src/bundle_remove.c
```diff
@@ -89,11 +93,389 @@ static bool parse_options(int argc, char **argv)
 		return false;
 	}
 
+	/* flag restrictions */
+	if (cmdline_option_force && cmdline_option_recursive) {
+		error("--force and --recursive options are mutually exclusive\n\n");
```
otaviobp: With this new implementation of recursive, why they need to be mutually exclusive?
Force will create a list of all bundles including dependencies. And then --recursive will select all dependencies in the system that are leftovers that can be removed.
And your code is already supporting that.

castulo: well it may be supported but using them together makes the command even more dangerous to be used, and more probable something does't work as expected in a corner case since we add more variables to the equation. I don't know if there is a use case for wanting to run both together. Having said that I'm ok to remove the restriction if you think we should support that. Do you think we should support that?

src/bundle.c
castulo: No, that would not work. This loop is not just getting all the tracked bundles, it gets all installed bundles and their dependencies (which includes all tracked bundles and more), except for those we plan to remove. If we only get the tracked bundles and their dependencies it could uninstall bundles that are dependencies of other bundles that are going to remain installed, breaking them.

As an example consider the scenario from test REM026, in this scenario we have 7 bundles installed in the system:
 - test-bundle1 is tracked and includes os-core, test-bundle2, test-bundle4, test-bundle6, and test-bundle3 (via test-bundle2).
 - test-bundle5 is installed (but not tracked) and includes test-bundle4.

If we only had gotten the tracked bundles and their dependencies as you suggested and we run:
`$ swupd bundle-remove --recursive test-bundle1`
Then test-bundle4 will get removed, but it should not sine it is required by test-bundle5.

src/bundle_remove.c
otaviobp: Both --force and --recursive are dangerous. Even isolated. So I'm ok on adding that support.
I wonder if we should add a warning on both --force and --recursive that this should be run with caution.

castulo: I will remove the restriction then.
We already print these messages when using those flags:
```
	/* inform user about the effects of using the options */
	if (cmdline_option_force) {
		info("\nThe --force option was used, the specified bundle%s and all bundles "
		     "that require %s will be removed from the system\n",
		     total > 1 ? "s" : "",
		     total > 1 ? "them" : "it");
	} else if (cmdline_option_recursive) {
		info("\nThe --recursive option was used, the specified bundle%s and %s "
		     "dependencies will be removed from the system\n",
		     total > 1 ? "s" : "",
		     total > 1 ? "their" : "its");
	}
```
I can use `warn` instead of `info`

src/bundle.c
otaviobp: This makes sense. We need to check all installed bundles that aren't going to be removed, not just the tracked ones. So this is correct.

docs/RELEASE_NOTES_DRAFT
```diff
@@ -5,5 +5,10 @@ Release vX.XX.XX
 Bug Fixes:
 
 Features:
+ - Add a bundle-remove --recursive flag to remove a bundle and all its dependencies from
```
otaviobp: Can you try to summarize the feature in just short topic? The idea is to say what the release introduce, not necessary how.

src/bundle_remove.c
otaviobp: No info is fine for that. But when the user sees it it's a bit too late. haha
Let's add a warning in the manual and help messages.

docs/RELEASE_NOTES_DRAFT
castulo: 🤔 I'll try to reword this in a shorter way

src/bundle_remove.c
castulo: done, it will be in the next update

docs/RELEASE_NOTES_DRAFT
castulo: done, it will be in the next update

castulo:
Version 2:

- Removed the flag restriction between --recursive and --force, they are now supported together (suggested by @otaviobp)
- Changed messages about the --recursive and --force flags from type info to warn to attract more attention (suggested by @otaviobp)
- Updated release notes to make the statement shorter (suggested by @otaviobp)
- Changed the bundle-remove command so when --recursive is used, if the tracking directory does not exist or is empty, the command defaults to considering all installed bundles as tracked (suggested by @otaviobp)
- Updated the tests to reflect these changes.

All requested changes have been addressed and the PR is ready for review again.

src/bundle.h
```diff
@@ -47,6 +55,13 @@ void track_bundle_in_statedir(const char *bundle_name, const char *state_dir);
  */
 void track_bundle(const char *bundle_name);
 
+/**
+ * @brief Validates that a tracking directory exists and that it is not
+ * empty, if it is, it creates it and marks all installed bundles as tracked.
+ * @param state_dir The path to the state directory.
+ */
+bool validate_tracking_dir(const char *state_dir);
```
otaviobp: Exporting a validate_tracking_dir here is adds a great possibility of errors in the future. And it doesn't help with the problem you have on os-install that you needed the .init hack.

So when you are tracking a new bundle you are calling the validate_tracking_dir() implicitly and you need to prevent that by creating a file if you don't want to track.  And in the case of remove track files you need to explicitly call the initialize function. This is prone to errors because it's confusing for the users of the library if they need to call or not the initialize function.

There are for sure multiple solutions to this problem. What I would suggest is to call validate_tracking_dir() implicitly on create_state_dir() so we always know that the directory is valid when we are using any tracked function.

 - For bundle_remove and bundle_add we don't need to worry about this anymore 
 - We aren't going to be checking if track directory is valid every time we track a new bundle on os-install and bundle-add
 - On os-install we remove the track of bundles we don't need instead of tracking new bundles. The downside is that we are going to be creating several useless files, but as they are empty and we are doing that on install time, it's totally ok.
 - If we add any other function that uses tracked bundles we don't need to worry about this anymore

test/functional/bundleadd/add-include.bats
```diff
@@ -43,15 +43,18 @@ test_setup() {
 
 @test "ADD048: Adding a bundle that includes another bundle (no tracking state)" {
 
+	# simulate as if test-bundle3 is already installed
+	sudo touch "$TARGETDIR"/usr/share/clear/bundles/test-bundle3
```
karthikprabhuvinod: maybe testlib func: create_dir

castulo:
Version 3:

- Added a warning for `--force` and `--recursive` flags in the swupd documentation (as suggested by @otaviobp).
- The validate_tracking_dir() function is now called on create_state_dir() so we always know that the directory is valid when we are using any tracked function (as suggested by @otaviobp).

All comments have been addressed and the PR is ready to be reviewed again.

test/functional/bundleadd/add-include.bats
castulo: in this case I'm creating a file, not a directory, and even if it was a directory the `create_dir` function in testlib does something more complicated than just creating a directory, it creates a directory, gets its hash and creates a tar for it.

test/functional/bundleremove/remove-recursive.bats
```diff
@@ -0,0 +1,156 @@
...
+	create_bundle -L    -n test-bundle5 -f /test-file5     "$TEST_NAME"
+	create_bundle -L -t -n test-bundle6 -f /test-file6     "$TEST_NAME"
+	# test-bundle1 has 4 direct dependencies: os-core, test-bundle2,
+	# test-bundle4 (also-add), test-bundle6 (tracked)
+	add_dependency_to_manifest -p    "$WEBDIR"/10/Manifest.test-bundle1 os-core
+	add_dependency_to_manifest -p    "$WEBDIR"/10/Manifest.test-bundle1 test-bundle2
+	add_dependency_to_manifest -p -o "$WEBDIR"/10/Manifest.test-bundle1 test-bundle4
+	add_dependency_to_manifest       "$WEBDIR"/10/Manifest.test-bundle1 test-bundle6
+	# and 1 indirect dependency: test-bundle3
+	add_dependency_to_manifest "$WEBDIR"/10/Manifest.test-bundle2 test-bundle3
```
karthikprabhuvinod: Sorry what is indirect dependency?

castulo: bundles that are added to the system because they are included or also-added by another included or also-added bundle. 

So in the specific case of that test, test-bundle1 includes test-bundle2 (direct), and test-bundle2 includes test-bundle3 (direct for test-bundle2 but indirect for test-bundle1)

karthikprabhuvinod: Ahh I see..Thanks for explaining

test/functional/bundleadd/add-include.bats
karthikprabhuvinod: Okay.

lebensterben:
Hey guys,

I came up with an idea that could distinguish bundle added by users from bundle added as dependencies. It also avoids writing anything to `/var` directory, which as pointed out by @otaviobp , it could be removed by the user and that would cause problem.

So currently we store the names of all installed bundles under `/usr/share/clear/bundles/`, and my idea is to add `dependency count` into those files, which are currently empty.

- When the bundle is added, create a file with file name same as bundle name
  - If the bundle is added as a dependency, its initial content is 2
  - If the bundle is added by user specifically, its initial content is 1
- If the bundle is already added, but later on the user add another bundle that depends on it, increment the value by 2
- When removing a bundle, also decrement all its dependencies' contained values by 2
- Therefore,
  - If `/usr/share/clear/bundles/FOO` has an odd value, its added by user
  - If `/usr/share/clear/bundles/FOO` has an even value, its added as a dependency
- A bundle could be safely removed only when its value is 0
- Suppose the user doesn't specify `--recursive` flag when removing a bundle, and some of its dependencies have value 0, then we can also easily find those bundles. Maybe we could have `swupd clean` to remove them.

PR #1147, opened 2019-10-15
Travis problem fixes

otaviobp:


karthikprabhuvinod:
Any reason why you switched to bionic?

.travis.yml
```diff
@@ -46,30 +46,37 @@ jobs:
 # - the Swupd_Root.pem cert must be installed out-of-tree to run the test suite with signature verification enabled
 # - python3-docutils for the rst2man script
 install:
-        - wget https://github.com/libcheck/check/releases/download/0.11.0/check-0.11.0.tar.gz
-        - tar -xvf check-0.11.0.tar.gz
-        - pushd check-0.11.0 && ./configure --prefix=/usr && make -j48 && sudo make install && popd
+        # Install dependencies from ubuntu
+        - sudo apt-get install python3-docutils
+        - sudo apt-get install shellcheck
```
karthikprabhuvinod: Maybe we can combine then by using commas

test/code_analysis/compliant.bats
```diff
@@ -26,6 +26,7 @@ check_sort_makefile()
 
 	run clang-format-9 -i -style=file src/*.[ch] src/lib/*.[ch]
 	if [ "$status" -ne 0 ]; then
+		echo "clang-format-9 failed with status $status. Check if you have clang-format-9 installed"
```
castulo: 🤣 learn from my mistakes

otaviobp:
> Any reason why you switched to bionic?

I added that to the commit message.
 - clang-format was broken for xenial
 - we can use libcheck from ubuntu instead of building from source

And it's better to run that in a more updated system.

PR #1149, opened 2019-10-15
travis: Don't install clearlinux certificates on travis

otaviobp:
We don't need the Clearlinux certificates to run standard tests, so
don't install them anymore

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,6 @@ jobs:
 # Pre-install missing build dependencies:
 # - bsdiff 1.* is the Clear Linux OS fork
 # - libcurl 7.35.0 is too old. Installing a newer version.
-# - the Swupd_Root.pem cert must be installed out-of-tree to run the test suite with signature verification enabled
 # - python3-docutils for the rst2man script
 install:
         # Install dependencies from ubuntu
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -70,10 +69,6 @@ install:
         # Use rst2man from python3
         - sudo ln -s /usr/share/docutils/scripts/python3/rst2man /usr/bin/rst2man.py
 
-        # Install clear linux certificate
-        - wget https://download.clearlinux.org/releases/13010/clear/Swupd_Root.pem
-        - sudo install -D -m0644 Swupd_Root.pem /usr/share/clear/update-ca/Swupd_Root.pem
```
castulo: why we don't need this anymore?

otaviobp: I think we don't need that for a long time. I don't even know what's it doing here.
We don't use the Swupd_Root from clear in any tests we run and we don't want to use. So it's better not to install it. Probably it's a leftover from the tests that were used in the past

karthikprabhuvinod: Looks for all the tests we now use Swupd_Root.pem from test/functiona/Swupd_Root.pem

castulo: thanks for the explanation

PR #1161, opened 2019-10-22
Add third-party skeletal features

karthikprabhuvinod:
This PR starts the third-party repo-add, delete, list feature
for swupd. This PR also add support for configure flag
to build swupd optionally, off by default.

**Repo Add:**
 sudo swupd 3rd-party add <repo_name> \<url\>
**Repo Remove:**
 sudo swupd 3rd-party remove <repo_name>
**Repos List:**
 sudo swupd 3rd-party list

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -32,6 +32,11 @@ bin_PROGRAMS = swupd
 AM_CPPFLAGS = $(AM_CFLAGS) $(libarchive_CFLAGS)
 
 swupd_SOURCES = \
+	src/3rd_party_add.c \
+	src/3rd_party.c \
+	src/3rd_party_internal.h \
+	src/3rd_party_list.c \
+	src/3rd_party_remove.c \
 	src/alias.c \
 	src/alias.h \
 	src/autoupdate.c \
--- a/configure.ac
+++ b/configure.ac
@@ -73,6 +73,25 @@ AS_IF(
 )
 
 AC_ARG_ENABLE(
+  [third-party],
+  [AS_HELP_STRING([--enable-third-party], [Enable 3rd-party (disabled by default)])]
+)
+THIRDPARTY="no"
+AS_IF(
+  [test "x$enable_third_party" = "xyes"],
+  [AC_DEFINE(THIRDPARTY, 1, [Enable third party])
+     THIRDPARTY="yes"],
+  [THIRDPARTY="no"]
+)
+
+AH_TEMPLATE([THIRDPARTY_REPO_PREFIX], [prefix path for third party repos])
+
+AS_IF(
+  [test "$enable_third_party" = "yes"],
+  [AC_DEFINE([THIRDPARTY_REPO_PREFIX], ["/opt/user-repo/"])]
+)
+
+AC_ARG_ENABLE(
   [tests],
   [AS_HELP_STRING([--disable-tests], [Do not enable unit or functional test framework (enabled by default)])]
 )
@@ -383,6 +402,7 @@ Configuration to build swupd-client:
   Version URL:                  ${VERSIONURL}
   Format Identifier:            ${FORMATID}
   Signature verification:       ${SIGVERIFICATION}
+  Third party:                  ${THIRDPARTY}
   Update certificate path:      ${cert_path}
   Configuration file path(s):   ${config_file_path}
   Use bzip compression:         ${BZIP}
--- a/src/3rd_party.c
+++ b/src/3rd_party.c
@@ -1,0 +1,110 @@
+/*
+ *   Software Updater - client side
+ *
+ *      Copyright © 2019 Intel Corporation.
+ *
+ *   This program is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License as published by
+ *   the Free Software Foundation, version 2 or later of the License.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#define _GNU_SOURCE
+#include "config.h"
+#include "swupd.h"
+#include "swupd_internal.h"
+
+#ifdef THIRDPARTY
+
+static struct subcmd third_party_commands[] = {
+	{ "add", "Add third party repo", third_party_add_main },
+	{ "remove", "Remove third party repo", third_party_remove_main },
+	{ "list", "List third party repo", third_party_list_main },
+	{ 0 }
+};
+
+static void print_help(const char *name)
+{
+	print("Usage:\n");
+	print(" or %s SUBCOMMAND [OPTION...]\n\n", basename((char *)name));
+	print("Help Options:\n");
+	print("   -h, --help              Show help options\n");
+	print("Subcommands:\n");
+
+	struct subcmd *entry = third_party_commands;
+
+	while (entry->name != NULL) {
+		print("   %-20s    %-30s\n", entry->name, entry->doc);
+		entry++;
+	}
+	print("\n");
+	print("To view subcommand options, run `%s SUBCOMMAND --help'\n", basename((char *)name));
+}
+
+static const struct option prog_opts[] = {
+	{ "help", no_argument, 0, 'h' },
+	{ 0 }
+};
+
+static int parse_options(int argc, char **argv, int *index)
+{
+	int opt;
+	int ret;
+
+	/* The leading "-" in the optstring is required to preserve option parsing order */
+	while ((opt = getopt_long(argc, argv, "-h", prog_opts, NULL)) != -1) {
+		switch (opt) {
+		case 'h':
+			print_help(argv[0]);
+			exit(EXIT_SUCCESS);
+		case '\01':
+			/* found a subcommand, or a random non-option argument */
+			ret = subcmd_index(optarg, third_party_commands);
+			if (ret < 0) {
+				error("unrecognized subcommand `%s'\n\n",
+				      optarg);
+				goto error;
+			} else {
+				*index = ret;
+				return 0;
+			}
+		case '?':
+			/* for unknown options, an error message is printed automatically */
+			info("\n");
+			goto error;
+		default:
+			/* should be unreachable */
+			break;
+		}
+	}
+
+	/* no subcommands implies -h/--help */
+	print_help(argv[0]);
+	exit(EXIT_SUCCESS);
+error:
+	print_help(argv[0]);
+	return -1;
+}
+
+enum swupd_code third_party_main(int argc, char **argv)
+{
+	int index = 0, ret;
+
+	if (parse_options(argc, argv, &index) < 0) {
+		return SWUPD_INVALID_OPTION;
+	}
+	optind = 0;
+	ret = third_party_commands[index].mainfunc(argc - 1, argv + 1);
+
+	return ret;
+}
+
+#endif
--- a/src/3rd_party_add.c
+++ b/src/3rd_party_add.c
@@ -1,0 +1,94 @@
+/*
+ *   Software Updater - client side
+ *
+ *      Copyright © 2019 Intel Corporation.
+ *
+ *   This program is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License as published by
+ *   the Free Software Foundation, version 2 or later of the License.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#define _GNU_SOURCE
+#include "3rd_party_internal.h"
+#include "config.h"
+
+#ifdef THIRDPARTY
+
+static void print_help(void)
+{
+	print("Usage:\n");
+	print("   swupd 3rd-party add [repo-name] [repo-URL]\n\n");
+
+	global_print_help();
+	print("\n");
+}
+
+static const struct global_options opts = {
+	NULL,
+	0,
+	NULL,
+	print_help,
+};
+
+static bool parse_options(int argc, char **argv)
+{
+#define EXACT_ARG_COUNT 2
+	int ind = global_parse_options(argc, argv, &opts);
+
+	if (ind < 0) {
+		return false;
+	}
+
+	if (argc == 1) {
+		error("The positional args: repo-name and URL are missing\n\n");
+		return false;
+	} else if (argc == 2) {
+		error("The positional args: repo-URL is missing\n\n");
+		return false;
+	}
+
+	/* Ensure that repo add only expects only two args: repo-name, repo-url */
+	if ((argc - ind) != EXACT_ARG_COUNT) {
+		error("Unexpected arguments\n\n");
+		return false;
+	}
+
+	return true;
+}
+
+enum swupd_code third_party_add_main(int argc, char **argv)
+{
+
+	enum swupd_code ret = SWUPD_OK;
+	const int step_in_third_party_add = 1;
+
+	if (!parse_options(argc, argv)) {
+		print_help();
+		return SWUPD_INVALID_OPTION;
+	}
+	progress_init_steps("third-party", step_in_third_party_add);
+
+	ret = swupd_init(SWUPD_NO_ROOT);
+	if (ret != SWUPD_OK) {
+		goto finish;
+	}
+
+	/* TODO implement */
+
+	swupd_deinit();
+
+finish:
+	progress_finish_steps(ret);
+	return ret;
+}
+
+#endif
--- a/src/3rd_party_internal.h
+++ b/src/3rd_party_internal.h
@@ -1,0 +1,26 @@
+#ifndef __THIRD_PARTY_INTERNAL_H
+#define __THIRD_PARTY_INTERNAL_H
+
+#include "swupd.h"
+
+/**
+ * @file
+ * @brief Common functions which can be used by 3rd-party add,
+ * remove, list
+ */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifdef THIRDPARTY
+
+/* TODO Implement */
+
+#endif
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
--- a/src/3rd_party_list.c
+++ b/src/3rd_party_list.c
@@ -1,0 +1,83 @@
+/*
+ *   Software Updater - client side
+ *
+ *      Copyright © 2019 Intel Corporation.
+ *
+ *   This program is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License as published by
+ *   the Free Software Foundation, version 2 or later of the License.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#define _GNU_SOURCE
+#include "3rd_party_internal.h"
+#include "config.h"
+
+#ifdef THIRDPARTY
+
+static void print_help(void)
+{
+	print("Usage:\n");
+	print("   swupd 3rd-party list\n\n");
+
+	global_print_help();
+	print("\n");
+}
+
+static const struct global_options opts = {
+	NULL,
+	0,
+	NULL,
+	print_help,
+};
+
+static bool parse_options(int argc, char **argv)
+{
+	int ind = global_parse_options(argc, argv, &opts);
+
+	if (ind < 0) {
+		return false;
+	}
+
+	if (argc > ind) {
+		error("Unexpected arguments\n\n");
+		return false;
+	}
+
+	return true;
+}
+
+enum swupd_code third_party_list_main(int argc, char **argv)
+{
+	enum swupd_code ret = SWUPD_OK;
+	const int step_in_third_party_list = 1;
+
+	if (!parse_options(argc, argv)) {
+		print_help();
+		return SWUPD_INVALID_OPTION;
+	}
+	progress_init_steps("third-party", step_in_third_party_list);
+
+	ret = swupd_init(SWUPD_NO_ROOT);
+	if (ret != SWUPD_OK) {
+		goto finish;
+	}
+
+	/* TODO Implement */
+
+	swupd_deinit();
+
+finish:
+	progress_finish_steps(ret);
+	return ret;
+}
+
+#endif
--- a/src/3rd_party_remove.c
+++ b/src/3rd_party_remove.c
@@ -1,0 +1,90 @@
+/*
+ *   Software Updater - client side
+ *
+ *      Copyright © 2019 Intel Corporation.
+ *
+ *   This program is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License as published by
+ *   the Free Software Foundation, version 2 or later of the License.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#define _GNU_SOURCE
+#include "3rd_party_internal.h"
+#include "config.h"
+
+#ifdef THIRDPARTY
+
+static void print_help(void)
+{
+	print("Usage:\n");
+	print("   swupd 3rd-party remove [repo-name]\n\n");
+
+	global_print_help();
+	print("\n");
+}
+
+static const struct global_options opts = {
+	NULL,
+	0,
+	NULL,
+	print_help,
+};
+
+static bool parse_options(int argc, char **argv)
+{
+#define EXACT_ARG_COUNT 1
+	int ind = global_parse_options(argc, argv, &opts);
+
+	if (ind < 0) {
+		return false;
+	}
+
+	if (argc == 1) {
+		error("The positional args: repo-name is missing\n\n");
+		return false;
+	}
+
+	/* Ensure that repo remove only expects only one arg: repo-name */
+	if ((argc - ind) != EXACT_ARG_COUNT) {
+		error("Unexpected arguments\n\n");
+		return false;
+	}
+
+	return true;
+}
+
+enum swupd_code third_party_remove_main(int argc, char **argv)
+{
+	enum swupd_code ret = SWUPD_OK;
+	const int step_in_third_party_remove = 1;
+
+	if (!parse_options(argc, argv)) {
+		print_help();
+		return SWUPD_INVALID_OPTION;
+	}
+	progress_init_steps("third-party", step_in_third_party_remove);
+
+	ret = swupd_init(SWUPD_NO_ROOT);
+	if (ret != SWUPD_OK) {
+		goto finish;
+	}
+
+	/* TODO Implement */
+
+	swupd_deinit();
+
+finish:
+	progress_finish_steps(ret);
+	return ret;
+}
+
+#endif
--- a/src/main.c
+++ b/src/main.c
@@ -28,12 +28,6 @@
 #include "swupd_build_opts.h"
 #include "swupd_internal.h"
 
-struct subcmd {
-	char *name;
-	char *doc;
-	enum swupd_code (*mainfunc)(int, char **);
-};
-
 static enum swupd_code external_search_main(int argc, char **argv)
 {
 	int ret;
@@ -49,7 +43,7 @@ static enum swupd_code external_search_main(int argc, char **argv)
 
 // TODO(castulo): remove the superseded command from the help menu by end of November 2019,
 // so it is not visible but the command must remain available in the back so we don't break users
-static struct subcmd commands[] = {
+static struct subcmd main_commands[] = {
 	{ "info", "Show the version and the update URLs", info_main },
 	{ "autoupdate", "Enable/disable automatic system updates", autoupdate_main },
 	{ "check-update", "Check if a new OS version is available", check_update_main },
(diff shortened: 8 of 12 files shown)


karthikprabhuvinod:
@otaviobp : I have a working third_party_repo PR...Tests are still pending and PR needs a bit of finishing. At this moment, could you  take a look from an early feedback & feature standpoint ?

Makefile.am
```diff
@@ -32,6 +32,11 @@ bin_PROGRAMS = swupd
 AM_CPPFLAGS = $(AM_CFLAGS) $(libarchive_CFLAGS)
 
 swupd_SOURCES = \
+	src/3rd_party_add.c \
+	src/3rd_party.c \
+	src/3rd_party_helpers.c \
+        src/3rd_party_list.c \
```
otaviobp: mix tab/spaces?

configure.ac
```diff
@@ -72,6 +72,25 @@ AS_IF(
   [SIGVERIFICATION="no"]
 )
 
+AC_ARG_ENABLE(
+  [thirdparty],
```
otaviobp: I prefer third-party, but I'm ok with all together too.

configure.ac
```diff
@@ -72,6 +72,25 @@ AS_IF(
...
+)
+THIRDPARTY="no"
+AS_IF(
+  [test "x$enable_thirdparty" = "xyes"],
+  [AC_DEFINE(THIRDPARTY, 1, [Enable third party])
+     THIRDPARTY="yes"],
+  [THIRDPARTY="no"]
+)
+
+AH_TEMPLATE([THIRDPARTY_REPO_PREFIX], [prefix path for third party repos])
```
otaviobp: ++ for having a setting for REPO_PREFIX

src/3rd_party.c
```diff
@@ -0,0 +1,173 @@
...
+#define CHOOSE_REMOVE 9003
+static bool cmdline_option_add;
+static bool cmdline_option_list;
+static bool cmdline_option_remove;
+
+#ifdef THIRDPARTY
+
+void get_repo_path(char **repo_path, char *repo_name)
+{
+	string_or_die(repo_path, "%s%s", THIRDPARTY_REPO_PREFIX, repo_name);
```
otaviobp: Use sys_path_join here to make sure we are going to have a '/' separator

src/3rd_party.c
```diff
@@ -0,0 +1,173 @@
...
+
+static void print_help(void)
+{
+	print("Usage:\n");
+	print("   swupd 3rd-party [OPTION...]\n\n");
+
+	global_print_help();
+
+	print("Options:\n");
+	print("   --add <repo_name>		Add third party repo\n");
```
otaviobp: I don't think it's a good idea to use flags for add, list remove. If you use a flag we need to protect against multiple flags used together and this will become hard after adding bundle-add and other commands

src/3rd_party.c
```diff
@@ -0,0 +1,173 @@
...
+	progress_finish_steps(ret);
+	free_string(&repo_path);
+	return ret;
+}
+
+#else
+
+enum swupd_code third_party_main(int UNUSED_PARAM argc, char UNUSED_PARAM **argv)
+{
+	info("This feature is disabled at compile time\n");
```
otaviobp: It's better to avoid a print here so for users this feature won't exist before it's released officially.

src/3rd_party.h
```diff
@@ -0,0 +1,8 @@
+#ifndef __THIRD_PARTY_H
+#define __THIRD_PARTY_H
+
```
otaviobp: I would add the `#ifdef THIRDPARTY` here and never call those functions if it's false.

otaviobp: And follow the standard in other header files to create this one. The define name is different and there's `extern "C" `missing

swupd.bash
```diff
@@ -39,7 +39,7 @@ _swupd()
 	    ("check-update")
 		opts="$global "
 		break;;
-		("update")
+	    ("update")
```
otaviobp: tab/spaces?

src/3rd_party.c
```diff
@@ -0,0 +1,173 @@
...
+	int ind = global_parse_options(argc, argv, &opts);
+	int opt_mask = 0;
+
+	if (ind < 0) {
+		return false;
+	}
+
+	if (!cmdline_option_repo) {
+		if (cmdline_option_add || cmdline_option_remove) {
+			error("The --repo option is a required argument\n\n");
```
otaviobp: same for --repo. I prefer:

`swupd 3rd-party add repo_name url`

src/3rd_party_add.c
```diff
@@ -0,0 +1,59 @@
...
+		error("failed to create %s\n", repo_path);
+		return -1;
+	}
+
+	return SWUPD_OK;
+}
+
+int third_party_add(char *repo_path)
+{
+	if (repo_exists(repo_path)) {
```
otaviobp: We need to have a file in /var/lib/swupd to track the 3rd-party repositories and the corresponding url, so checking for the dir is not enough. And if the dir is completly removed we should be able to restore that.

src/3rd_party.c
karthikprabhuvinod: > I don't think it's a good idea to use flags for add, list remove. If you use a flag we need to protect against multiple flags used together and this will become hard after adding bundle-add and other commands

I was thinking about that....I need to have add, list, remove as sub_commands like add in git remote add

src/3rd_party.c
karthikprabhuvinod: Makes sense will change in next push

Makefile.am
karthikprabhuvinod: Makes sense will change in next push

src/3rd_party.h
karthikprabhuvinod: > I would add the `#ifdef THIRDPARTY` here and never call those functions if it's false.

I think we are using the same pattern using compile time flag for signatures...like in c src/signature.c

karthikprabhuvinod: > And follow the standard in other header files to create this one. The define name is different and there's `extern "C" `missing

Makes sense will change in next push

otaviobp: I mixed 2 comments in same line and I realized now that it's confusing.
 - The #ifdef style is good the way it is, like the signature as you said. But If you just ignore all file it's going to be easier to code and I'm ok with that. But if you want to leave like it is, it's fine too.
 - The style I was talking about is the header file. But I realized that there's not a very well established style. each file does it in a different way. So the only thing really missing is the extern.

src/main.c
```diff
@@ -69,6 +69,7 @@ static struct subcmd commands[] = {
 	{ "clean", "Clean cached files", clean_main },
 	{ "hashdump", "Dump the HMAC hash of a file", hashdump_main },
 	{ "verify", "NOTE: this command has been superseded, please use \"swupd diagnose\" instead", diagnose_main },
+	{ "3rd-party", "Indicate Swupd to use user's third party repo", third_party_main },
```
otaviobp: And use the ifdef here too. I forgot to mention that in my first review

src/3rd_party.c
karthikprabhuvinod: Actually I was thinking this should be an option for sure. Usually command are actions.

It works perfectly in our case as then we can make it required for remove, add and not have it for list.

Let me know what you feel

src/3rd_party.h
karthikprabhuvinod: For point1 I will try to do what you said. I agree with you
For point2, will make sure headers have extern c

src/main.c
karthikprabhuvinod: Ahh yes.. will do

src/3rd_party.c
karthikprabhuvinod: Makes sense will change in next push

src/3rd_party.c
karthikprabhuvinod: Makes sense will change in next push

src/3rd_party.h
karthikprabhuvinod: Makes sense will change in next push

swupd.bash
karthikprabhuvinod: Makes sense will change in next push

src/main.c
karthikprabhuvinod: Makes sense will change in next push

karthikprabhuvinod:
@otaviobp 
I have addressed all your comments with remarks like "Makes sense will change in next push" except two i guess.

With this PR, I made sure the skeleton exists, The add, delete list command lines parsers are created. But they don't do much. 

Please could give it a second quick look? Anything you want to be removed for now?

src/3rd_party.h
```diff
@@ -0,0 +1,44 @@
...
+#include "swupd.h"
+
+/**
+ * @brief Creates a repo folder under THIRDPARTY_REPO_PREFIX 
+ * if it does not exist
+ *
+ * @return 0 on success, SWUPD_THIRD_PARTY_REPO_EXISTS if it already exists
+ * and negative numbers on errors
+ */
+enum swupd_code third_party_add(int argc, char **argv);
```
otaviobp: Rename all functions that are a main to add a _main suffix. This one should be `third_party_add_main`

src/3rd_party.c
```diff
@@ -0,0 +1,136 @@
...
+};
+
+static struct subcmd commands[] = {
+	{ "add", "Add third party repo", third_party_add },
+	{ "remove", "Remove third party repo", third_party_remove },
+	{ "list", "List third party repo", third_party_list },
+	{ 0 }
+};
+
+static int subcmd_index(char *arg)
```
otaviobp: Instead of copying, export this function in swupd_internal.h and use it in 3rd_party

karthikprabhuvinod: @otaviobp This command is slightly diffferent than the one we used in main.c since we dont have to check if it is a verify command etc. Should I still use that same function? Also I will need to pass commands args to it since commands are different from main.c, add.c remove.c

Makefile.am
```diff
@@ -32,6 +32,10 @@ bin_PROGRAMS = swupd
 AM_CPPFLAGS = $(AM_CFLAGS) $(libarchive_CFLAGS)
 
 swupd_SOURCES = \
+	src/3rd_party_add.c \
```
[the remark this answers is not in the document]
karthikprabhuvinod: Makes sense will change in next push

src/3rd_party.h
```diff
@@ -0,0 +1,44 @@
...
+/**
+  * @file
+  * @brief Functions to call for third_party_main
+  */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "swupd.h"
```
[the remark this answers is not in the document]
karthikprabhuvinod: Makes sense will change in next push

src/3rd_party.h
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * @brief Creates a repo folder under THIRDPARTY_REPO_PREFIX
```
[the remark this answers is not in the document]
karthikprabhuvinod: Makes sense will change in next push

src/3rd_party_add.c
```diff
@@ -0,0 +1,120 @@
...
+#include "lib/log.h"
+#include "swupd.h"
+#include "sys/stat.h"
+
+#ifdef THIRDPARTY
+
+static int create_repo_dir(char UNUSED_PARAM *repo_path)
+{
+	//repo_path doesn't exist
+	/*
```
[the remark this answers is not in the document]
karthikprabhuvinod: Makes sense will change in next push

src/3rd_party_internal.h
```diff
@@ -0,0 +1,109 @@
...
+	return SWUPD_OK;
+}
+
+static inline int setup_repo_tracking()
+{
+	char *tracking_dir;
+	char *tracking_file;
+	int ret;
+	tracking_dir = sys_path_join(globals.path_prefix, globals.state_dir);
+	tracking_file = sys_path_join(tracking_dir, "repo_tracker.ini");
```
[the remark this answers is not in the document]
karthikprabhuvinod: Makes sense will change in next push

src/swupd_exit_codes.h
```diff
@@ -47,6 +47,8 @@ enum swupd_code {
 	SWUPD_OUT_OF_MEMORY_ERROR,	     /* 34 swupd ran out of memory */
 	SWUPD_VERIFY_FAILED,		     /* 35 verify could not fix/replace/delete one or more files */
 	SWUPD_INVALID_BINARY,		     /* 36 binary to be executed is missing or invalid */
+	SWUPD_THIRD_PARTY_REPO_EXISTS,	     /* 37 third party repo already exists, error for adding */
```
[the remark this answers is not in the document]
karthikprabhuvinod: Makes sense will change in next push

src/3rd_party.h
karthikprabhuvinod: Makes sense will change in next push

src/3rd_party_internal.h
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * @brief Check if a repo directory exists
+ *
+ * Check if the directory pointed to repo_exists exists
+ *
+ * @return true on success, false on failure
+ */
+static inline bool repo_exists(char *repo_path)
+{
+	struct stat sb;
```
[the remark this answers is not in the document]
karthikprabhuvinod: Makes sense will change in next push

src/3rd_party_internal.h
```diff
@@ -0,0 +1,109 @@
...
+
+	if ((sb.st_uid == 0) &&
+	    (S_ISDIR(sb.st_mode)) &&
+	    ((sb.st_mode & 0777) == 0700)) {
+		return true;
+	}
+	return false;
+}
+
+static int create_tracking_file(char *tracking_file)
```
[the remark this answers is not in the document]
karthikprabhuvinod: @otaviobp : for me, a name like create_repo makes more sense with something like creating a repo directory. Here we are talking about creating a tracking file.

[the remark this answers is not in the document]
otaviobp: So probably I didn't understand the usage of this function. What's a tracking file?

src/3rd_party.c
otaviobp: take a look at  #1180
And you can get that as a parameters. You can rename that if you find a better name and add a comment explaining what the function is doing.

src/3rd_party_internal.h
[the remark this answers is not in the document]
karthikprabhuvinod: Its the repo.ini file where write what repo it is and all the bundles under it

src/3rd_party.c
karthikprabhuvinod: Okay Cool I will move to swupd_internal.h as a inline and include it in main.c and 3rd-party.c

karthikprabhuvinod: Makes sense will change in next push

src/3rd_party_internal.h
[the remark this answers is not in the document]
karthikprabhuvinod: Makes sense will change in next push

src/swupd_internal.h
```diff
@@ -17,5 +17,36 @@ extern enum swupd_code install_main(int argc, char **argv);
...
+	enum swupd_code (*mainfunc)(int, char **);
+};
+
+static inline int subcmd_index(char *arg, struct subcmd *commands)
+{
+	struct subcmd *entry = commands;
+	int i = 0;
+	size_t input_len;
+	size_t cmd_len;
+
```
castulo: why did you remove this from the function?
```
if (strcmp(arg, "verify") == 0) {		
 	verify_set_command_verify(true); // set to true so we know the "verify" command was used		
 	arg = "diagnose";		
 }
```

swupd.zsh
```diff
@@ -175,14 +175,41 @@ if [[ -n "$state" ]]; then
         "mirror:Configure mirror url for swupd content"
         "clean:Clean cached files"
         "hashdump:Dump the HMAC hash of a file"
+	"3rd-party: Indicate Swupd to use user's third party repo"
```
castulo: you seem to be mixing tabs and spaces in these files

src/swupd_internal.h
otaviobp: see #1180

castulo: just did, that explains it, thanks

otaviobp:
> This PR seems to be moving in the right direction but there are still many things that need to be addressed first. I added comments with all my observations in the PR. I also noted there are no tests whatsoever for this PR, it introduces lots of new functionality so I think having tests is crucial.

This PR isn't really adding a new functionality, so there's no way on creating tests to it. It's just adding the skeleton so we can move forward.

otaviobp:
@karthikprabhuvinod Can you remove all functions from this PR that aren't related to the skeleton? And you can add later when you really need them.

src/3rd_party.c
```diff
@@ -0,0 +1,110 @@
...
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#define _GNU_SOURCE
+#include "3rd_party.h"
+#include "config.h"
+#include "lib/log.h"
+#include "swupd.h"
```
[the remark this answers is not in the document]
karthikprabhuvinod: Makes sense will remove it in next push

src/3rd_party_add.c
```diff
@@ -0,0 +1,119 @@
...
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#define _GNU_SOURCE
+#include "3rd_party_internal.h"
+#include "config.h"
+#include "lib/log.h"
+#include "swupd.h"
+#include "sys/stat.h"
```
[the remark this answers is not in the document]
karthikprabhuvinod: There are some others which are redundant as well. Makes sense will remove it in next push

src/3rd_party_add.c
```diff
@@ -0,0 +1,119 @@
...
+		return false;
+	}
+
+	if (argc == 1) {
+		error("The positional args: repo-name and URL are missing\n\n");
+		return false;
+	} else if (argc == 2) {
+		error("The positional args: repo-URL is missing\n\n");
+		return false;
+	}
```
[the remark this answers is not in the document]
karthikprabhuvinod: Good point.  but Actually am not expecting any more args as such for these commands.

src/3rd_party_add.c
```diff
@@ -0,0 +1,119 @@
...
+	if (ret != SWUPD_OK) {
+		goto finish;
+	}
+
+	repo_path = get_repo_path(argv[1]);
+
+	ret = create_repo_dir(repo_path);
+
+	if (!ret) {
+		info("Third party repo successfully created\n");
```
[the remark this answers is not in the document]
karthikprabhuvinod: @castulo Very sharp observation. Yes will change it.

src/3rd_party_add.c
```diff
@@ -0,0 +1,119 @@
...
+		create_repo(argv[1]);
+		goto finish;
```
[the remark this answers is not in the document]
karthikprabhuvinod: Makes sense will remove it in next push

src/3rd_party_add.c
```diff
@@ -0,0 +1,119 @@
...
+	progress_finish_steps(ret);
+	free_string(&repo_path);
+	return ret;
+}
+
+#else
+
+enum swupd_code third_party_add_main(int UNUSED_PARAM argc, char UNUSED_PARAM **argv)
+{
+	return SWUPD_OK;
```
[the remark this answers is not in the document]
karthikprabhuvinod: No this one doesnt need a TODO as it is function definition when we enable-third-party is provided to auto-configure

src/3rd_party_internal.h
```diff
@@ -0,0 +1,98 @@
...
+
+#include "config.h"
+#include "lib/sys.h"
+#include "swupd.h"
+#include <errno.h>
+#include <fcntl.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <sys/stat.h>
+#include <unistd.h>
```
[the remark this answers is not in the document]
karthikprabhuvinod: Thanks . Makes sense will remove it in next push

src/3rd_party_internal.h
```diff
@@ -0,0 +1,98 @@
...
+
+	if (fd < 0) {
+		return -1;
+	}
+	close(fd);
+
+	return SWUPD_OK;
+}
+
+static inline int setup_repo_config()
```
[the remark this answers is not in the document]
karthikprabhuvinod: Thanks. Makes sense will remove it in next push

src/3rd_party_internal.h
```diff
@@ -0,0 +1,98 @@
...
+{
+	char *repo_config_file;
+	int ret;
+	repo_config_file = sys_path_join(globals.state_dir, "/3rd_party/repo.ini");
+
+	sys_file_exists(repo_config_file);
+
+	if (create_repo_config(repo_config_file)) {
+		debug("Issue creating repo config file: %s in %s\n", repo_config_file, globals.state_dir);
+		goto finish;
```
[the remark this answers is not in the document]
karthikprabhuvinod: Makes sense will remove it in next push

src/3rd_party_internal.h
```diff
@@ -0,0 +1,98 @@
...
+	close(fd);
+
+	return SWUPD_OK;
+}
+
+static inline int setup_repo_config()
+{
+	char *repo_config_file;
+	int ret;
+	repo_config_file = sys_path_join(globals.state_dir, "/3rd_party/repo.ini");
```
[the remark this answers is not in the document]
karthikprabhuvinod: Okay. Makes sense will remove it in next push

src/3rd_party_internal.h
```diff
@@ -0,0 +1,98 @@
...
+	return;
+}
+
+static inline void remove_repo(char UNUSED_PARAM *repo_name)
+{
+	/* TODO Implement */
+	return;
+}
+
+static inline char *get_repos()
```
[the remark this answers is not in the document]
karthikprabhuvinod: Makes sense will remove it in next push

src/3rd_party_list.c
```diff
@@ -0,0 +1,106 @@
...
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#define _GNU_SOURCE
+#include "3rd_party_internal.h"
+#include "config.h"
+#include "lib/log.h"
+#include "swupd.h"
```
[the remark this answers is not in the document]
karthikprabhuvinod: Makes sense will change it in next push

src/3rd_party_remove.c
```diff
@@ -0,0 +1,110 @@
...
+#include "3rd_party_internal.h"
+#include "config.h"
+#include "lib/log.h"
+#include "swupd.h"
+
+#ifdef THIRDPARTY
+
+int remove_repo_dir(char UNUSED_PARAM *repo_path)
+{
+	//rm_rf(repo_path);
```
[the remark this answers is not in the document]
karthikprabhuvinod: No, will remove and add a TODO

src/3rd_party_remove.c
```diff
@@ -0,0 +1,110 @@
...
+	return true;
+}
+
+enum swupd_code third_party_remove_main(int argc, char **argv)
+{
+	enum swupd_code ret = SWUPD_OK;
+	const int step_in_third_party_remove = 1;
+	char *repo_path = NULL;
+
+	/* there is no need to report in progress for init at this time */
```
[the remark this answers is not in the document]
karthikprabhuvinod: Makes sense will change it in next push

src/swupd_internal.h
karthikprabhuvinod: Marking this as resolved.

src/3rd_party_add.c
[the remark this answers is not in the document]
otaviobp: You don't need this function here. REmove it and leave everything related to 3rd party inside the #ifdef

src/3rd_party.h
```diff
@@ -0,0 +1,44 @@
...
+#endif
+
+/**
+ * @brief Creates a new third-party repo under THIRDPARTY_REPO_PREFIX
+ * if it does not exist
+ *
+ * @return 0 on success, SWUPD_THIRD_PARTY_REPO_EXISTS if it already exists
+ * and negative numbers on errors
+ */
+enum swupd_code third_party_add_main(int argc, char **argv);
```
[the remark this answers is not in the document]
karthikprabhuvinod: Thanks for that explanation. Makes sense

karthikprabhuvinod:
> @karthikprabhuvinod Can you remove all functions from this PR that aren't related to the skeleton? And you can add later when you really need them.

Okay done

src/3rd_party_add.c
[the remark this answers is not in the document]
karthikprabhuvinod: Done

src/3rd_party_add.c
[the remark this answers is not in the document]
karthikprabhuvinod: Removed and added a TODO

src/3rd_party_list.c
```diff
@@ -0,0 +1,106 @@
...
+
+#ifdef THIRDPARTY
+
+static void print_repo_list(char UNUSED_PARAM *repos)
+{
+	return;
```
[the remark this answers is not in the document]
karthikprabhuvinod: Removed  and replaced with TODO

src/3rd_party_remove.c
```diff
@@ -0,0 +1,110 @@
...
+
+	if (ind < 0) {
+		return false;
+	}
+
+	if (argc == 1) {
+		error("The positional args: repo-name is missing\n\n");
+		return false;
+	}
+
```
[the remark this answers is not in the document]
karthikprabhuvinod: for above: https://github.com/clearlinux/swupd-client/pull/1161#discussion_r340222876

If think I can do this. I can pre define the no of args expected for each operation (**EXACT_ARG_COUNT**).

Fo eg: for add ...that is 2 (repo-url and repo-name) for remove it is 1(repo-name)

Lets take add as an example:

if (argc - ind) == EXACT_ARG_COUNT

I always check if argc - opind == EXACT_ARG_COUNT..if not fail

src/3rd_party_add.c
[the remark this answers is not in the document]
karthikprabhuvinod: Makes sense.. will change it in next push

src/3rd_party_add.c
karthikprabhuvinod: Will address this in a follow-on PR as discussed

src/main.c
```diff
@@ -49,7 +43,7 @@ static enum swupd_code external_search_main(int argc, char **argv)
 
 // TODO(castulo): remove the superseded command from the help menu by end of November 2019,
 // so it is not visible but the command must remain available in the back so we don't break users
-static struct subcmd commands[] = {
+struct subcmd main_commands[] = {
```
otaviobp: I realized that now. This should remain static.

karthikprabhuvinod:
> > ignore
> 
> What should I ignore?

I accidently reviewed my own PR..so had to paste some comment

src/3rd_party.c
```diff
@@ -0,0 +1,110 @@
...
+ */
+
+#define _GNU_SOURCE
+#include "config.h"
+#include "swupd.h"
+#include "swupd_internal.h"
+
+#ifdef THIRDPARTY
+
+struct subcmd third_party_commands[] = {
```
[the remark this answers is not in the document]
karthikprabhuvinod: same as discussed in
https://github.com/clearlinux/swupd-client/pull/1161#discussion_r340778718

src/main.c
otaviobp: No. static means that the variable isn't going to be visible outside of this C source file.Using it as a variable is valid and it's ok. Doing that prevents another source file to do this:

```
extern struct subcmd main_commands[];
main_commands = NEW_VALUE;
```

This would change the value of the original main_commands variable

karthikprabhuvinod: @otaviobp Thanks for explaining that. I know why it works...subcmd_index is inline static fucntion and it isnt considered another file.

karthikprabhuvinod: Fixed this in last push

src/3rd_party.c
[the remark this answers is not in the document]
karthikprabhuvinod: Fixed this in last push

karthikprabhuvinod:
@otaviobp @castulo Looks like all comments are addressed. I think it looks ready now

src/3rd_party_add.c
```diff
@@ -0,0 +1,94 @@
...
+		error("The positional args: repo-name and URL are missing\n\n");
+		return false;
+	} else if (argc == 2) {
+		error("The positional args: repo-URL is missing\n\n");
+		return false;
+	}
+
+	/* Ensure that repo add only expects only two args: repo-name, repo-url */
+	if ((argc - ind) != EXACT_ARG_COUNT) {
+		error("Too many Unexpected args\n\n");
```
castulo: nitpick, "Too many Unexpected args" sounds redundant, use "Too many arguments" or "Unexpected arguments", also you are using a capital letter in Unexpected in the middle of the sentence

castulo: Actually to be consistent with other swupd commands it is better to go with `error("unexpected arguments\n\n");`.

src/3rd_party_list.c
```diff
@@ -0,0 +1,83 @@
...
+static bool parse_options(int argc, char **argv)
+{
+	int ind = global_parse_options(argc, argv, &opts);
+
+	if (ind < 0) {
+		return false;
+	}
+
+	if (argc > ind) {
+		error("Too many Unexpected args\n\n");
```
castulo: same here -> `error("unexpected arguments\n\n");`

src/swupd_internal.h
```diff
@@ -22,6 +25,59 @@ enum swupd_code repair_main(int argc, char **argv);
...
+enum swupd_code third_party_main(int argc, char **argv);
+
+/**
+ * @brief Creates a new third-party repo under THIRDPARTY_REPO_PREFIX
+ * if it does not exist
+ *
+ * @return 0 on success, SWUPD_THIRD_PARTY_REPO_EXISTS if it already exists
+ * and negative numbers on errors
+ */
+extern enum swupd_code third_party_add_main(int argc, char **argv);
```
castulo: I had mentioned this in a previous comment but I could not find it anymore, but you cannot use `enum swupd_code` and return negative numbers on errors. The whole purpose of using `enum soup_code` is to guarantee the function is going to return SWUPD_OK or SWUPD_NO if successful, and any other error within the swupd_exit_codes on errors.

castulo: found the comment I'd made, here it is:

"All these functions are defined as returning a value of enum swupd_code, which at the moment are positive values from 0 to 36.

When defining functions within the swupd code it is possible to define functions that return any kind of value, boolean, positive numbers, negative numbers, etc. However whenever swupd exits, it must always exit with a positive value contained in swupd_code. In other words all functions that return values different than swupd_code can only be used internally, and whatever function that is going to propagate its return code all the way to the user has to return a value defined within swupd_code.

The whole purpose of using the enum swupd_code as return type is to "certify" in a way that a certain function complies with only returning positive values within the range of swupd_code, and this way we know it is safe to propagate its return code all the way to the user. So by this definition it makes no sense to define a function that supposedly complies with only returning values within swupd_code range, but arbitrarily returning negative numbers on errors. If a function returns a swupd_code then it must return SWUPD_OK or SWUPD_NO in a successful operation, and an appropriate error from swupd_code on failures."

karthikprabhuvinod: Sorry about missing that. Will correct it in next push

src/3rd_party_list.c
karthikprabhuvinod: Sure thing Will correct it in next push

src/3rd_party_add.c
karthikprabhuvinod: Sure thing Will correct it in next push

PR #1164, opened 2019-10-24
Track bundles specified in os-install

### LINKED ISSUE #1153: Bundles installed with os-install are not being tracked
**Describe the bug**
When installing a Clear OS using `swupd os-install -B <BUNDLE>`, BUNDLE is not being tracked in <statedir>/bundles/ as it happens when using `swupd bundle-add <BUNDLE>` to install a bundle.

**To Reproduce**
`$ sudo ./swupd os-install -B editors --path <test_dir> --statedir <test_state_dir>`

**Expected behavior**
The file `< test_state_dir>/bundles/editors` should exist.

castulo:
When os-install is used to install an OS and the --bundles option is used
to specify some bundles that should be installed in the system along
with the OS, the bundles specified should be tracked.

Closes #1153

### DIFF
--- a/src/bundle.h
+++ b/src/bundle.h
@@ -27,6 +27,26 @@ bool is_installed_bundle(const char *bundle_name);
  */
 int required_by(struct list **reqd_by, const char *bundle_name, struct manifest *mom, int recursion, struct list *exclusions, char *msg);
 
+/**
+ * @brief Creates a tracking file in the tracking directory within the
+ * specified state directory.
+ * If there are no tracked files in that directory (directory is empty
+ * or does not exist), copy the tracking directory at
+ * path_prefix/usr/share/clear/bundles to the tracking directory to
+ * initiate the tracking files.
+ * This function does not return an error code because weird state in this
+ * directory must be handled gracefully whenever encountered.
+ * @param bundle_name The name of the bundle.
+ * @param state_dir The path to the state directory.
+ */
+void track_bundle_in_statedir(const char *bundle_name, const char *state_dir);
+
+/**
+ * @brief Similar to track_bundle_in_statedir() but uses the global state_dir.
+ * @param bundle_name The name of the bundle.
+ */
+void track_bundle(const char *bundle_name);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/verify.c
+++ b/src/verify.c
@@ -785,6 +785,33 @@ static int find_unsafe_to_delete(const void *a, const void *b)
 	return -1;
 }
 
+static bool init_tracking_dir(const char *state_dir, const char *init_file)
+{
+
+	char *tracking_dir;
+	char *tracking_file;
+
+	/* make sure the state directory exist */
+	if (create_state_dirs(state_dir)) {
+		return false;
+	}
+
+	/* add a temporary control file to the tracking
+	 * directory so it is not empty to avoid tracking
+	 * all installed bundles */
+	tracking_dir = mk_full_filename(state_dir, "bundles");
+	tracking_file = mk_full_filename(tracking_dir, init_file);
+	int fd = open(tracking_file, O_RDWR | O_CREAT, S_IRUSR | S_IWUSR);
+	free_string(&tracking_dir);
+	free_string(&tracking_file);
+	if (fd < 0) {
+		return false;
+	}
+	close(fd);
+
+	return true;
+}
+
 /* This function does a simple verification of files listed in the
  * subscribed bundle manifests.  If the optional "fix" or "install" parameter
  * is specified, the disk will be modified at each point during the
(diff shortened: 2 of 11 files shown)


castulo:
V1 of this PR is ready to be reviewed.

src/bundle.h
```diff
@@ -27,6 +27,26 @@ bool is_installed_bundle(const char *bundle_name);
...
+ * If there are no tracked files in that directory (directory is empty
+ * or does not exist), copy the tracking directory at
+ * path_prefix/usr/share/clear/bundles to the tracking directory to
+ * initiate the tracking files.
+ * This function does not return an error code because weird state in this
+ * directory must be handled gracefully whenever encountered.
+ * @param bundle_name The name of the bundle.
+ * @param tracking_dir The path to the tracking directory.
+ */
+void track_bundle(const char *bundle_name, const char *state_dir);
```
karthikprabhuvinod: Did you mean to tracking_dir in second_arg
@param state_dir in line 40 The path to the state directory.

src/verify.c
```diff
@@ -1153,6 +1180,30 @@ enum swupd_code verify_main(void)
...
+		/* this is a fresh install so the tracking directory
+		 * needs to be created */
+		char *new_os_statedir = mk_full_filename(globals.path_prefix, "/var/lib/swupd");
+		bool tracking = init_tracking_dir(new_os_statedir, ".init");
+		if (tracking) {
+			for (iter = cmdline_option_bundles; iter; iter = iter->next) {
+				char *bundle = iter->data;
+				/* make sure the bundle was in fact valid and will
+				 * be installed before creating the tracking file */
+				if (list_search(bundles_subs, bundle, subscription_bundlename_strcmp)) {
```
karthikprabhuvinod: Is this list_search comparision checking if the bundle was installed manually by user or included as a includes?

If it added manually, then we create a tracking file for it?

src/bundle.h
castulo: ups, no, state_dir is the correct argument, at some point it was the tracking_dir but I changed that, I forgot to update the comment, thanks for catching that, I will update that comment

src/verify.c
castulo: yes, the bundles the user is specifically attempting to install are in the  cmdline_option_bundles list, however there is a chance the user specified an invalid bundle, and we don't want to create a tracking file for an invalid bundle. So by making sure there is a subscription that matches each bundle in the list we make sure the bundle to track was in fact valid.

src/verify.c
```diff
@@ -785,6 +785,33 @@ static int find_unsafe_to_delete(const void *a, const void *b)
...
+	if (create_state_dirs(state_dir)) {
+		return false;
+	}
+
+	/* add a temporary control file to the tracking
+	 * directory so it is not empty to avoid tracking
+	 * all installed bundles */
+	tracking_dir = mk_full_filename(state_dir, "bundles");
+	tracking_file = mk_full_filename(tracking_dir, init_file);
+	int fd = open(tracking_file, O_RDWR | O_CREAT, S_IRUSR | S_IWUSR);
```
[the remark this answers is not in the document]
castulo: I needed to use a temporary file because when installing a new system the state dir will always be empty, so all bundles in /usr/share/clear/bundles are going to be copied as tracked every time, not only the ones requested by --bundles. So it is a hack to make the track_installed() function believe the tracking directory is not empty and only copy those specified.

src/verify.c
```diff
@@ -1153,6 +1180,30 @@ enum swupd_code verify_main(void)
 	 * naming convention: All exit goto labels must follow the "brick_the_system_and_FOO:" pattern
 	 */
 
+	/* track bundles specified by user */
+	if (cmdline_option_install && cmdline_option_bundles) {
+		/* this is a fresh install so the tracking directory
+		 * needs to be created */
+		char *new_os_statedir = mk_full_filename(globals.path_prefix, "/var/lib/swupd");
```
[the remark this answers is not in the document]
castulo: I gave this a lot of thought, and actually the first implementation I wrote used the statedir but decided to change that. 

This is why I think this is better.

When using os-install to install a system, the only required argument is the path where the OS will be installed, as shown here:
`$ swupd os-install /my/installation/path`

Since we are not specifying a statedir, swupd will use the system's default one, which means it will use the one from the system we are using to install, therefore the tracking file will be created in the statedir of that system, which doesn't make sense cause we are installing the OS in a different location. This leads to two problems: 
1) now the system we are using to install will think it has those bundles installed and tracked (and it may not) 
2) the system that does have those bundles installed won't have those bundles as tracked.

Another alternative is to use the statedir specified by the user. This approach has two other problems: 
1) it would mean os-install would really require the --statedir flag to work correctly, this is a problem because the --statedir is optional, we cannot force the user to use an optional flag. 
2) If the user points to the new statedir in the new system using --statedir, this directory will be empty since it is a new system, so the user will never be able to use the content that may already exist in the statedir of the system from which he is installing.

src/verify.c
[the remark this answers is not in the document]
otaviobp: you can touch os-core instead. It's always required to be tracked.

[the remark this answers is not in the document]
otaviobp: Actually not. Leave the way it is

src/verify.c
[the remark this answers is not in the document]
otaviobp: You have a point here. I think that as I said both solutions have problems, but it looks like saving it to path_prefix is better.

src/bundle.h
otaviobp: I am checking the comments again and I think it's a bit hard to understand what is expected from this function by the name.
As we have a `is_tracked_bundle()` I would call the standard function as `track_bundle(bundle_name)`. And the other `track_bundle_in_statedir(bundle_name, state_dir)`.
The function we would expect to be used more often is the track_bundle. The other is for one specific use case.

castulo: sounds good, I'll update that.

castulo:
Updated version 2:
- Renamed function track_bundle() to track_bundle_in_statedir(), and renamed track_installed() to track_bundle() as suggested by @otaviobp.
- Fixed the documentation of  track_bundle_in_statedir() in bundle.h as requested by @karthikprabhuvinod.

PR #1170, opened 2019-10-24
Tabs: Don't mix tabs and spaces

otaviobp:


test/functional/server.py
```diff
@@ -24,279 +24,279 @@
...
-            self.send_header("Content-type", ctype)
-            self.send_header("Content-Length", str(fs[6]))
-            self.send_header("Last-Modified",
-                             self.date_time_string(fs.st_mtime))
-            self.end_headers()
-            return f
-        except Exception:
-            f.close()
-            raise
+	counter = 0
```
castulo: to be honest I prefer using spaces in Python scripts. Using tabs in Python makes formatting the scripts a lot harder, specially sine Python relies in formatting to work properly and since according to PEP8 lines in scripts should not be bigger than 79 characters.

Issue #1064, opened 2019-08-14
Autogenerate Swupd.bash and swupd.zsh


otaviobp:
Maintaining those two files is a lot of work and we could generate them with the information we have on swupd code base.
We have all information we need in "static struct subcmd commands" in main.c and "static const struct option prog_opts" in command files.

phmccarty:
Note that the original implementation of swupd.bash *was* autogenerated, but then we stopped autogenerating it because various commands needed special treatment, like removing the `os-core` autocompletion for `swupd bundle-remove`. So as the number of special cases increased, we hit a threshold of complexity for the autogeneration script.

But the swupd CLI has changed a lot in the meantime, gaining new commands and flags, so the added complexity for the special cases may be worth it, from a maintenance perspective.

PR #1175, opened 2019-10-25
Spaces

otaviobp:


karthikprabhuvinod:
@castulo : Good point ! Maybe then I need to change my PR to do that then. Also look like I need get a different vimrc for python :/

otaviobp:
> I would really to stick to spaces (4 spaces) for indentation in all Python scripts, we can use tabs everywhere else. Using tabs in Python makes formatting the scripts a lot harder, specially sine Python relies in formatting to work properly and since according to PEP8 lines in scripts should not be bigger than 79 characters.

No problem. I'll fix. I like to be consistent, so forcing spaces on all python scripts is fine.
Is tab on shell ok?

lebensterben:
I prefer having spaces, either 2 or 4. This is not for aesthetics but for practicality, because many editors by default use 'soft-tabs', well I think Vim is an exception).

And often developers enables features like 'format on save', which will likely replaces all tabs with spaces.

If a contributor send a patch for review, the git-diff would list many differences unless it's told to ignore white spaces. So either you would see a bloated list of differences, or you might apply a patch which doesn't replace the tabs, but you end up with a script with mixed tabs and spaces.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

castulo:
> No problem. I'll fix. I like to be consistent, so forcing spaces on all python scripts is fine.
> Is tab on shell ok?

Thanks. I have no preference for shell scripts, I believe I always use tabs in my shell scripts, but as @lebensterben mentioned, I can just set my editor to interpret a tab as 4 spaces as well.

otaviobp:
Thank you for the suggestions. In this patch I tried to keep the most used style for each filetype. I'm totally ok on changing the python one to spaces. The source/header files are using tabs for a long time and as clang-forma is checking that it hasn't been a problem so far.
The major problem we had on mixing tabs and spaces were other files.

So for now it looks like the ideal would be:
 - .c .h .bash. .sh
   Always indent with tabs, tabsize is 8 spaces. (same style used for source files from our history)
 - python:
   4 spaces (python standard)
 - zsh
   4 spaces (that's the style used by other zsh autocomplete scripts on clear)
 - Autotools files:
   Autotools is very sensible to this, so I won't change. Just follow the style in the file you are changing.

I'll fix this patch to follow that and send a V2. Is there any other file I should change let me know

otaviobp:
> If a contributor send a patch for review, the git-diff would list many differences unless it's told to ignore white spaces. So either you would see a bloated list of differences, or you might apply a patch which doesn't replace the tabs, but you end up with a script with mixed tabs and spaces.

The ideal is to have an automated check for any style we want to enforce. clang-format is doing that for our sources and make compliant applies some other minor checks. For now let's try to check that manually for other files, but if it's easy to add a script to check I'm ok with that.

otaviobp:
I realized now that there are multiple styles for zsh on clear. So I'll leave it with 2 spaces ident, like the original

otaviobp:
V2: Use 4 spaces instead of a tab in python scripts.

otaviobp:
For some weird reason travis tests are not running on this PR. so please review and I'll run that manually before merging.

lebensterben:
>> If a contributor send a patch for review, the git-diff would list many differences unless it's told to ignore white spaces. So either you would see a bloated list of differences, or you might apply a patch which doesn't replace the tabs, but you end up with a script with mixed tabs and spaces. ...

@otaviobp The only zsh script currently in the project is the auto-completion. Unfortunately shellcheck doesn't support it. And neither does zsh has style checker like bashate.

On top of this, zsh completion system has its own unique syntax, which is too complicated and not well documented. I wrote the script but after 2 months I cannot remember many syntax.

To consistently employ the test-driven paradigm this project has, we may consider do this.

1. The baseline is to use `_gnu_generic` function to automatically generate completion by having zsh call `swupd -h` in the background. And It's also possible to call `swupd -h` manually in background and use `sed`, `awk` or whatever external command to format the result. This approach only requires a consistent help message style.

2. To provide more functionality like bundle name completion, we may let zsh call an external helper script to generate the result. This helper should be coded in a language that automatic tests and code style checks are available, such as bash, Python, or C.

The advantage for this design is:

1. The usage of zsh specific syntax, and especially zsh completion specific syntax, is minimal. So for future maintainers of this project, they don't have to spent hours learning zsh completion system.
2. The completions are up to date with the latest release of swupd-client, as long as the help message is updated.
3. Helper script can be used not only by zsh, but also bash! This would greatly help to improve the bash completion.
4. Helper is could be tested and a consistent format can be enforced.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

otaviobp:
> On top of this, zsh completion system has its own unique syntax, which is too complicated and not well documented. I wrote the script but after 2 months I cannot remember many syntax.

I'm not changing the style of zsh for now then. Just enforcing what we have.

> To provide more functionality like bundle name completion, we may let zsh call an external helper script to generate the result. This helper should be coded in a language that automatic tests and code style checks are available, such as bash, Python, or C.

I like the idea on creating something in common for the bash and zsh autocomplete. My concern is adding an dependency to things we dont need. So a system may have zsh, but not shell or python. I think that autogenerating the autocomplete scripts is easier than trying to maintain another binary that is called to return the autocomplete code. What's is painful on what we have now is that we need to update 2 scripts every time we change any flag in swupd. And all this information is already available in the C code, so it would be way easier to maintain if we had a C code generating the scripts. 

https://github.com/clearlinux/swupd-client/issues/1064

PR #1185, opened 2019-10-31
[DO NOT REVIEW]Adding more concrete features for 3rd-party add,delete, remove

karthikprabhuvinod:
Adding concrete repo list, remove,add  functionality

How to test:
sudo swupd 3rd-party add <repo-name> <repo-url>

sudo swupd 3rd-party remove <repo-name>

sudo swupd 3rd-party list


bryteise:
I'm not sure if adding INI parsing to swupd-client is the right approach here.

karthikprabhuvinod:
@otaviobp Marking Otavio so that he is in loop.

@bryteise : So we choose ini over toml. Is your objection to choice of format?

 We are planning to track all repos which users create & store it in the ini. So parsing is required to perform checks if a repo already exists or can be removed etc.

otaviobp:
@bryteise we decided to store repository information locally in an ini file because we already have a parser that we use for config file so it's the easiest format to store information right now. It's not an exact replace from the toml file and it's not how we are going to read/keep the binary list for example.

@karthikprabhuvinod we already have a parser in src/lib/config_parser.c. You can just use it to read the repository file

bryteise:
Ah alrighty.

bryteise:
I assume that parser has already been fuzzed reasonably well because this is an attack vector.

karthikprabhuvinod:
> I assume that parser has already been fuzzed reasonably well because this is an attack vector.

Right will default to using the same parser

karthikprabhuvinod:
@otaviobp @castulo So I have a working repo-add,remove & list. Wanted to check with you guys before adding tests & cleaning up. Needed an early feedback.

Could you guys give it a shot?

src/3rd_party_config.c
```diff
@@ -0,0 +1,173 @@
...
+	static int repo_field_count = 0;
+	static int global_field_count = 0;
+
+	if (!strncmp(section, "REPO", strlen(section))) {
+		if (key && value) {
+			static struct repo *repo_var;
+			if (repo_field_count == 0) {
+				repo_var = (struct repo *)malloc(sizeof(struct repo));
+			}
+			if (!strncmp(key, "repo-name", strlen(key))) {
```
castulo: I don't see the point of prepending the `repo-` prefix to every key if we are specifically referring to repo properties in the config file

src/3rd_party_config.c
```diff
@@ -0,0 +1,173 @@
...
+				repo_var->repo_name = strdup_or_die(value);
+				repo_field_count++;
+			}
+
+			if (!strncmp(key, "repo-url", strlen(key))) {
```
castulo: I'm not sure about this approach because then you have to hardcode each one of the values of the config file to load one by one, if we want to add more values to the config file this parser would be broken, so it will be harder to maintain. For example, you are missing the `version` value for each repo which should be stored in the config file too.

src/3rd_party_config.c
```diff
@@ -0,0 +1,173 @@
...
+	free_string(&repo_config_file);
+	if (fd < 0) {
+		return -1;
+	}
+	close(fd);
+
+	return SWUPD_OK;
+}
+
+static inline int setup_repo_config()
```
castulo: why did you define this function as "inline" ?

src/3rd_party_config.c
```diff
@@ -0,0 +1,173 @@
...
+				global_var->value2 = strdup_or_die(value);
+				global_field_count++;
+			}
+		}
+		return true;
+	}
+	return false;
+}
+
+static inline char *get_repo_config_file()
```
castulo: missing void argument

src/3rd_party_config.c
```diff
@@ -0,0 +1,173 @@
...
+{
+	return sys_path_join(globals.state_dir, "/3rd_party/repo.ini");
+}
+
+static int create_repo_config(char *repo_config_file)
+{
+	int fd = open(repo_config_file, O_RDWR | O_CREAT, S_IRUSR | S_IWUSR);
+	free_string(&repo_config_file);
```
castulo: don't free repo_config_file here, free it in the same function where it was allocated

src/3rd_party_config.c
castulo: also missing void argument

src/3rd_party_config.c
```diff
@@ -0,0 +1,173 @@
...
+	return SWUPD_OK;
+}
+
+static inline int setup_repo_config()
+{
+	char *repo_config_file;
+	int ret;
+	repo_config_file = get_repo_config_file();
+
+	sys_file_exists(repo_config_file);
```
castulo: you are not checking the returned value from sys_file_exists

Makefile.am
```diff
@@ -350,6 +359,10 @@ BATS = \
 	test/functional/verify-legacy/verify-bundle.bats \
 	test/functional/verify-legacy/verify-flags.bats
 
+BATS = \
```
castulo: isn't this going to crash if THIRD_PARTY_BATS is not defined?

scripts/flag_validator.bash
```diff
@@ -4,9 +4,16 @@ SCRIPTS_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
 
-swupd_commands=(info autoupdate check-update update bundle-add bundle-remove bundle-list search-file diagnose repair os-install mirror clean hashdump)
+swupd_commands=(info autoupdate check-update update bundle-add bundle-remove
+	bundle-list search-file diagnose repair os-install mirror clean hashdump 3rd-party)
 conflict=0
 
+check_command_exist() {
+
+	local command=$1
+	sudo "$SWUPD" "$command" --help 2>&1 > /dev/null | grep "Error: unrecognized subcommand" > /dev/null
```
castulo: this works kind of backwards of what I would expect. It is called check_command_exist but it returns False when it exists, and True when it doesn't.

scripts/flag_validator.bash
```diff
@@ -172,6 +184,7 @@ for val in "${swupd_commands[@]}"; do
 	if [ "$val" == "$arg" ]; then
 		# remove the '-' if it has one
 		flag=$(echo "$flag" | tr --delete '-')
+		echo "here"
```
castulo: leftover?

scripts/flag_validator.bash
```diff
@@ -83,6 +90,11 @@ validate_existing_flags() {
 	find_global_flag_conflict
 
 	for cmd in "${swupd_commands[@]}"; do
+		# Commands which are not compiled dont have to be validated and
+		# can be skipped eg: 3rd-party
+		if check_command_exist "$cmd"; then
+			continue
+		fi
```
castulo: this skips the 3rd-party commands when not installed, but when they are installed, how do you handle the extra argument in the command? I mean "3rd-party" is not really the command you  want to validate, but you want to validate `add`, `remove` and `list`

src/3rd_party_add.c
```diff
@@ -48,21 +62,22 @@ static bool parse_options(int argc, char **argv)
 		return false;
 	}
 
-	if (argc == 1) {
+	int positional_args = argc - ind;
+	/* Ensure that repo add only expects only two args: repo-name, repo-url */
```
castulo: extra "only" word

src/3rd_party_add.c
```diff
@@ -82,7 +97,16 @@ enum swupd_code third_party_add_main(int argc, char **argv)
 		goto finish;
 	}
 
-	/* TODO implement */
+	/* The last two in reverse are the repo-name, repo-url */
+	if (add_repo_config(argv[argc - 2], argv[argc - 1])) {
+		ret = SWUPD_NO;
```
castulo: I don't think SWUPD_NO is the appropriate code to return here, actually SWUPD_NO is considered a successful return code (currently only used in check-update).

src/3rd_party_config.c
castulo: Also what is going to happen if you have multiple repos inside the repo.ini?

src/3rd_party_config.c
```diff
@@ -0,0 +1,377 @@
...
+	ssize_t write_line_len;
+
+	fd = open(repo_config_file_path, O_WRONLY | O_APPEND);
+	if (fd < 0) {
+		return -1;
+	}
+
+	write_line_len = strlen(name);
+
+	if (write_line_len != write(fd, name, write_line_len)) {
```
castulo: where are the `[` `]` in the section name?

src/3rd_party_config.c
```diff
@@ -0,0 +1,377 @@
...
+		error("error writing to file %d", errno);
+		ret = -1;
+	}
+
+	close(fd);
+	free_string(&write_line);
+	return ret;
+}
+
+static int write_empty_line()
```
castulo: (void)

src/3rd_party_config.c
```diff
@@ -0,0 +1,377 @@
...
+				       write_line_length)) {
+		error("error writing to file %d", errno);
+		ret = -1;
+	}
+
+	close(fd);
+	return ret;
+}
+
+static bool repo_config_read()
```
castulo: (void)

src/3rd_party_config.c
```diff
@@ -0,0 +1,377 @@
...
+
+static void generate_path()
+{
+	char *repo_config_dir;
+	repo_config_dir = sys_path_join(globals.state_dir, REPO_CONFIG_DIR);
+	repo_config_file_path = sys_path_join(repo_config_dir, REPO_CONFIG_CONFIG_FILE);
+	free_string(&repo_config_dir);
+}
+
+static int ensure_repo_config_dir()
```
castulo: I'm not gonna write the same in each function, but you are missing adding void in the argument of lots of function definitions

src/3rd_party_config.c
```diff
@@ -0,0 +1,377 @@
...
+		if (ret) {
+			error("Failed to create repository config directory\n");
+		}
+	}
+
+	free_string(&repo_config_dir);
+	return ret;
+}
+
+static int ensure_repo_config()
```
castulo: I'm not sure if "ensure" is the best word for this function, why not "create_repo_config(void)" instead? Ensure sounds more like a validation than a function to create a resource

src/3rd_party_internal.h
```diff
@@ -2,6 +2,9 @@
 #define __THIRD_PARTY_INTERNAL_H
 
 #include "swupd.h"
+#include <config.h>
```
castulo: standard libs first

test/functional/3rd-party/3rd-party-repo-manage-negative.bats
```diff
@@ -0,0 +1,119 @@
...
+
+load "../testlib"
+
+test_setup() {
+
+	create_test_environment "$TEST_NAME"
+
+}
+
+test_teardown() {
```
castulo: these are the default test_setup and test_teardown actions, so there is no need to add them

.travis.yml
```diff
@@ -17,7 +17,7 @@ env:
                 - GROUP1="$UPDATE_SUBGROUP1 $(find test/functional/{autoupdate,bundleinfo,verify-legacy} -name *.bats -printf '%p ')"
                 - GROUP2="$UPDATE_SUBGROUP2 $(find test/functional/{checkupdate,hashdump,mirror,usability} -name *.bats \( ! -name usa-config-file.bats \) -printf '%p ')"
                 - GROUP3="$(find test/functional/{diagnose,search,os-install,repair} -name *.bats -printf '%p ')"
-                - GROUP4="$(find test/functional/{bundleadd,bundleremove,bundlelist,signature} -name *.bats -printf '%p ')"
+                - GROUP4="$(find test/functional/{bundleadd,bundleremove,bundlelist,signature,3rd-party} -name *.bats -printf '%p ')"
```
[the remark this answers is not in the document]
karthikprabhuvinod: > I'm hesitant to have all 3rd-party test under the same 3rd-party directory. I would prefer to start with keeping each 3rd-party command on its own directory for now. So in this case you would need three directories: 3rd_party_add, 3rd_party_remove and 3rd_party_list. Also remember we want to eventually migrate all directory names to use snake case, so use `_` in the directory name instead of `-`.

Ah..I thought about that too, The reason I went with this approach is because all three commands are helpful together. Thats how a user would use it and think of them as some thing called **3rd-party repo manage**. Any specific reason why you wanted a folder for each? Will make to change all to snake case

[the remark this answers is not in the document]
karthikprabhuvinod: > Finally, you forgot to update the description of the job in line 39.

Okay, I will look at #1193 and try to use same approach

karthikprabhuvinod:
@otaviobp @castulo Abandoning this as this is split into #1208 #1204 #1199

PR #1192, opened 2019-11-06
Add ability to install bundles from 3rd-party repositories

### LINKED ISSUE #1177: Create command to add bundles from 3rd-party repository
Create a command `swupd 3rd-party bundle-add` that can install bundles from a 3rd party repository

Command line example:

`swupd 3rd-party bundle-add bundle1`: Look for bundle1 in all 3rd-party repositories and install it. If more than one bundle1 is found, bundle-add should fail with a message saying that the repo need to be selected.

`swupd 3rd-party bundle-add --repo my_repo bundle1`:Install bundle1 from my_repo

Note that we should support installation of multiple bundles with the same name. If there's any name clash the repository name should be required.

Related #1082

castulo:
Implement a command swupd 3rd-party bundle-add that can install bundles from a 3rd party repository.

Tasks:
- [x] Add support for `3rd-party bundle-add` to the `testlib`
- [x] Add functional tests for adding bundles from a 3rd-party repo
- [x] Add new test group to Travis/github actions
- [x] Add the `3rd-party bundle-add` command to swupd (without functionality)
- [ ] Add documentation to man pages.
- [x] Add command to autocompletion script swupd.bash
- [ ] Add command to autocompletion script swupd.zsh
- [x] Add command to the config file
- [x] Search for 3rd-party bundles in a hardcoded 3rd-party repo URL, and install it (as POC)
- [x] Search for the bundle in 3rd-party repos registered in repo.ini
- [x] When getting the URL of the 3rd-party repo, if there are multiple repos, we need to search for the bundle in all of them
- [x] If the bundle is in multiple bundles we need to ask the user to specify a repo
- [x] When using a 3rd-party repo, we need to get the current version of the system from the appropriate section of the repo.ini config file
- [x] When tracking the installed bundle, we need to track it in the state dir that is specific to the 3rd-party repo.
- [x] Check output correctness and consistency with other commands (including output using `--quiet`, `--verbose`, `--debug`)
- [ ] Update release notes draft

Closes #1177 

### DIFF
--- a/src/3rd_party.c
+++ b/src/3rd_party.c
@@ -28,6 +28,7 @@ static struct subcmd third_party_commands[] = {
 	{ "add", "Add third party repository", third_party_add_main },
 	{ "remove", "Remove third party repository", third_party_remove_main },
 	{ "list", "List third party repository", third_party_list_main },
+	{ "bundle-add", "Install a bundle from a third party repository", third_party_bundle_add_main },
 	{ 0 }
 };
 
--- a/src/3rd_party_bundle_add.c
+++ b/src/3rd_party_bundle_add.c
@@ -19,6 +19,7 @@
 
 #define _GNU_SOURCE
 
+#include "3rd_party_repos.h"
 #include "swupd.h"
 
 #ifdef THIRDPARTY
@@ -89,7 +90,16 @@ static bool parse_options(int argc, char **argv)
 enum swupd_code third_party_bundle_add_main(int argc, char **argv)
 {
 	struct list *bundles = NULL;
+	struct list *iter1 = NULL;
+	struct list *iter2 = NULL;
+	struct list *repos = NULL;
+	char *state_dir;
+	char *path_prefix;
 	int ret_code = SWUPD_OK;
+	int ret = SWUPD_OK;
+
+	// TODO(s):
+	// 1) if the bundle is in multiple repos we need to ask the user to choose one
 
 	const int steps_in_bundleadd = 8;
 
@@ -106,6 +116,9 @@ enum swupd_code third_party_bundle_add_main(int argc, char **argv)
 		return ret_code;
 	}
 
+	/* load the existing 3rd-party repos from the repo.ini config file */
+	repos = third_party_get_repos();
+
 	/* move the bundles provided in the command line into a
 	 * list so it is easier to handle them */
 	for (; *cmdline_bundles; ++cmdline_bundles) {
@@ -114,10 +127,86 @@ enum swupd_code third_party_bundle_add_main(int argc, char **argv)
 	}
 	bundles = list_head(bundles);
 
-	//ret_code = execute_bundle_add(bundles);
+	/* backup the original state_dir and path_prefix values */
+	state_dir = strdup_or_die(globals.state_dir);
+	path_prefix = strdup_or_die(globals.path_prefix);
+
+	/* try installing bundles one by one */
+	for (iter1 = bundles; iter1; iter1 = iter1->next) {
+		char *bundle = iter1->data;
+		struct repo *selected_repo = NULL;
+		int count = 0;
+
+		/* search for the bundle in all the 3rd-party repos */
+		info("Searching for bundle %s in the 3rd-party repositories...\n", bundle);
+		for (iter2 = repos; iter2; iter2 = iter2->next) {
+			struct repo *repo = iter2->data;
+			struct manifest *mom = NULL;
+			struct file *file = NULL;
+
+			/* set the appropriate content_dir and state_dir for the selected 3rd-party repo */
+			if (third_party_set_repo(state_dir, path_prefix, repo)) {
+				ret_code = SWUPD_COULDNT_CREATE_DIR;
+				goto clean_and_exit;
+			}
+
+			/* load the repo's MoM*/
+			mom = load_mom(repo->version, false, NULL);
+			if (!mom) {
+				error("Cannot load manifest MoM for 3rd-party repository %s version %i\n", repo->name, repo->version);
+				ret_code = SWUPD_COULDNT_LOAD_MOM;
+				goto clean_and_exit;
+			}
+
+			/* search for the bundle in the MoM */
+			file = mom_search_bundle(mom, bundle);
+			if (file) {
+				/* the bundle was found in this repo, keep a pointer to the repo */
+				info("Bundle %s found in 3rd-party repository %s\n", bundle, repo->name);
+				count++;
+				selected_repo = repo;
+			}
+			manifest_free(mom);
+		}
+
+		/* if the bundle exists in only one repo, we can continue */
+		if (count == 0) {
+			error("bundle %s was not found in any 3rd-party repository\n\n", bundle);
+			ret_code = SWUPD_INVALID_BUNDLE;
+			continue;
+		} else if (count > 1) {
+			error("bundle %s was found in more than one 3rd-party repository, please specify a repository\n\n", bundle);
+			ret_code = SWUPD_INVALID_OPTION;
+			continue;
+		}
+
+		/* set the content_url and state_dir to those from the 3rd-party repo
+		 * where the requested bundle was found */
+		if (third_party_set_repo(state_dir, path_prefix, selected_repo)) {
+			ret_code = SWUPD_COULDNT_CREATE_DIR;
+			goto clean_and_exit;
+		}
+
+		/* execute_bundle_add expects a list so we can send a new list with only
+		 * the one bundle we want to install in this iteration */
+		struct list *bundle_to_install = NULL;
+		bundle_to_install = list_append_data(bundle_to_install, bundle);
+
+		ret = execute_bundle_add(bundle_to_install, selected_repo->version);
+
+		if (ret != SWUPD_OK) {
+			ret_code = ret;
+		}
+		list_free_list(bundle_to_install);
+		info("\n");
+	}
 
 clean_and_exit:
+	free_string(&state_dir);
+	free_string(&path_prefix);
 	list_free_list(bundles);
+	list_free_list_and_data(repos, repo_free_data);
+
 	swupd_deinit();
 	progress_finish_steps(ret_code);
 	return ret_code;
--- a/src/3rd_party_bundle_add.c
+++ b/src/3rd_party_bundle_add.c
@@ -98,10 +98,22 @@ enum swupd_code third_party_bundle_add_main(int argc, char **argv)
 	int ret_code = SWUPD_OK;
 	int ret = SWUPD_OK;
 
+	/*
+	 * Steps for 3rd-party bundle-add:
+	 *
+	 *  1) load_manifests
+	 *  2) download_packs
+	 *  3) extract_packs
+	 *  4) validate_fullfiles
+	 *  5) download_fullfiles
+	 *  6) extract_fullfiles
+	 *  7) install_files
+	 */
+
 	// TODO(s):
 	// 1) if the bundle is in multiple repos we need to ask the user to choose one
 
-	const int steps_in_bundleadd = 8;
+	const int steps_in_bundleadd = 7;
 
 	if (!parse_options(argc, argv)) {
 		print_help();
@@ -191,6 +203,9 @@ enum swupd_code third_party_bundle_add_main(int argc, char **argv)
 		 * the one bundle we want to install in this iteration */
 		struct list *bundle_to_install = NULL;
 		bundle_to_install = list_append_data(bundle_to_install, bundle);
+
+		info("\nBundles added from a 3rd-party repository are forced to run with the --no-scripts flag for security reasons\n\n");
+		globals.no_scripts = true;
 
 		ret = execute_bundle_add(bundle_to_install, selected_repo->version);
 
--- a/src/3rd_party_bundle_add.c
+++ b/src/3rd_party_bundle_add.c
@@ -1,0 +1,126 @@
+/*
+ *   Software Updater - client side
+ *
+ *      Copyright © 2019 Intel Corporation.
+ *
+ *   This program is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License as published by
+ *   the Free Software Foundation, version 2 or later of the License.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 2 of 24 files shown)


src/3rd_party_bundle_add.c
```diff
@@ -0,0 +1,105 @@
...
+	// things:
+	//     a) track it in the same state dir where we track the upstream bundles
+	//        in the target system
+	//     b) track it in the state dir that is specific to the 3rd-party repo
+	//        but then we need to make changes in other functions that look for
+	//        tracked bundles (like bundle-remove) to look for trakced bundles
+	//        in all state dirs
+
+	char *tp_state_dir;
+	string_or_die(&tp_state_dir, "%s/3rd-party/test-repo", globals.state_dir);
```
otaviobp: For now I think we can set all those variables the way you are doing. After we have a clear idea on everything we need to set we can split swupd_init() somehow to create a safe way to set/unset all this variables and make sure we won't forget anything.

castulo: thanks for taking a look at this @otaviobp, I'll continue moving in this direction 😄

src/globals.c
```diff
@@ -411,6 +411,16 @@ bool set_default_urls()
 	return true;
 }
 
+/* Sets the repo_name_3rd_party global variable */
+void set_repo_name_3rd_party(char *repo_name)
```
karthikprabhuvinod: :+1:  This is a good idea, maybe I can also set this in repo_manage and use it to determine path where repos will be created.

src/lib/config_parser.c
```diff
@@ -112,3 +112,101 @@ bool config_parse(const char *filename, parse_config_fn_t parse_config_fn, void
 exit:
 	return ret;
 }
+
```
karthikprabhuvinod: I will sync up with you on this. What I tried to do is create a 3rd-party_config.c where this file is read into a list of struct repo. My initial understanding was bundle-add will build on the same idea of adding bundles to those structs.

keeping @otaviobp loop.

src/3rd_party_bundle_add.c
```diff
@@ -21,69 +21,6 @@
 
 #include "swupd.h"
```
karthikprabhuvinod: Same here, wrap around #ifdef THIRDPARTY

test/functional/testlib.bash
```diff
@@ -1381,6 +1381,56 @@ set_latest_version() { # swupd_function
...
 
+# Creates a new version of the server side content in the specified
+# location to serve as a 3rd-party repository
+# Parameters:
+# - ENVIRONMENT_NAME: the name of the test environment
+# - VERSION: the version of the server side content
+# - FORMAT: the format to use for the version
+# - REPO_NAME: the name of the 3rd-party repo, defaults to "test-repo"
+create_third_party_repo() { #swupd_function
+
```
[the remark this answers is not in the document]
castulo: of course, that will happen when you actually use swupd normally. But this is for testing purposes. When testing we need to mock all dependencies we have to be able to test a specific functionality. This is the same way we are mocking mixer, so we can test all swupd functionality without really having to use upstream content or without having to use mixer to create a local mix to test with it. That way we have full control of our test environment.

src/version.c
```diff
@@ -316,6 +316,31 @@ int get_current_version(char *path_prefix)
 	return v;
 }
 
+int get_current_3rd_party_version(char *state_dir, char *repo_name)
```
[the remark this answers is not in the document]
castulo: yeah, I was planning on doing something like that further down the line. Since this PR is still in very early stages and was more like in a POC state, I didn't want to add more code than necessary just yet. I'll make sure to enclose this in an `ifdef` as you suggest.

castulo:
thanks for the early review @karthikprabhuvinod  👍

src/version.c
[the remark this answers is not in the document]
karthikprabhuvinod: Ahh cool .. 👍

castulo:
There are things that are still not implemented in this PR (noted in the task list in the description), but as suggested by @otaviobp I'm setting this PR as ready for review so it can be merged with the progress it currently has. I can implement the missing tasks in a new PR.

src/lib/config_parser.c
```diff
@@ -112,3 +112,101 @@ bool config_parse(const char *filename, parse_config_fn_t parse_config_fn, void
...
+char *config_get_value(const char *filename, const char *section, const char *key)
```
otaviobp: Instead of re implementing a parser for the config file, use functions in common. There are also inconsistencies between first implementation and the second. Are keys always case insensitive or case sensitive?

test/functional/testlib.bash
```diff
@@ -1456,37 +1515,41 @@ create_version() { # swupd_function
...
+		create_tar "$env_name"/"$content_dir"/"$version"/files/"$hashed_name"
+		FORMAT="$env_name"/"$content_dir"/"$version"/files/"$hashed_name"
 		export FORMAT
 	fi
 	# if the previous version is 0 then create a new MoM, otherwise copy the MoM
 	# from the previous version
 	if [ "$from_version" = 0 ]; then
 
-		mom=$(create_manifest "$env_name"/web-dir/"$version" MoM)
+		debug_msg "This is the fist version in the content directory, creating a new MoM..."
```
karthikprabhuvinod: nit pick: first version

test/functional/testlib.bash
```diff
@@ -1382,6 +1382,56 @@ set_latest_version() { # swupd_function
...
+	sudo chmod -R 0700 "$repo_state_dir"
+	export REPO_STATEDIR="$repo_state_dir"
+
+	# create the basic content for the 3rd-party repo
+	create_version "$env_name" "$version" 0 "$format" "$repo_name"
+	debug_msg "3rd-party repo $repo_name created successfully"
+
+	# add the new repo to the repo.ini file
+	{
+		printf '[%s]\n\n' "$repo_name"
```
karthikprabhuvinod: I think this should be a single \n

src/lib/config_parser.h
```diff
@@ -31,6 +31,16 @@ typedef bool (*parse_config_fn_t)(char *section, char *key, char *value, void *d
...
 
+/**
+ * @brief Searches the INI style configuration file for the requested key within the
+ * specified section
+ * @param filename: the full path to the configuration file where the key will be searched
+ * @param section: the section from the config file we are interested in
+ * @param key: the key we want to get the value from
+ * @returns The value that matches the section and key
+ */
+char *config_get_value(const char *filename, const char *section, const char *key);
```
karthikprabhuvinod: @castulo If you need to query and get specific, may you can query the struct I create in 3rd-party_config.c. We can create a non-static method in 3rd-party_config.c file which queries a  a struct instead of a ini

Let me know what you think

karthikprabhuvinod: https://github.com/clearlinux/swupd-client/pull/1199/files#diff-80aeb8f47ce71855e9a9de63b5be69c6

I will have to change repo_config_deinit & repo_config_init from static to global so that you can call them at the start of 3rd-party each command. You can add the get_value function here and returns   values from the struct. Just a suggestion

test/functional/testlib.bash
castulo: we always have a blank line between sections and key/values -> https://github.com/clearlinux/swupd-client/blob/master/config

src/3rd_party.c
```diff
@@ -96,12 +97,16 @@ static int parse_options(int argc, char **argv, int *index)
...
-	int index = 0, ret;
+	int ret;
+	int index = 0;
 
 	if (parse_options(argc, argv, &index) < 0) {
 		return SWUPD_INVALID_OPTION;
 	}
 	optind = 0;
+
+	globals.use_third_party_repo = true;
```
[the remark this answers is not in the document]
castulo: At this point is only used in two places (in bundle_add.c):
Here is is used to know if we need to print "swupd bundle-add" or "swupd 3rd-party bundle-add":
```
print("   swupd %sbundle-add [OPTION...] [bundle1 bundle2 (...)]\n\n", globals.use_third_party_repo ? "3rd-party " : "");
```
and is also used to know if we want to get the latest version from the os-release file or the repo.ini.

I can make it work without using that state variable, but we probably won't be able to reuse some code (for example the help menu)

src/globals.h
```diff
@@ -82,6 +84,10 @@ bool set_path_prefix(char *path);
 bool set_default_urls(void);
 bool set_state_dir_cache(char *path);
 void set_default_path_prefix(void);
+void set_content_url(char *url);
+bool set_state_dir(char *path);
+void set_default_third_party_config_file(char *path);
```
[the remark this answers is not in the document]
castulo: cool, I'll change that then

src/lib/config_parser.c
castulo: what do you mean I'm reimplementing it? One function is to parse the whole file, the second function is to look for a specific key within a given section of a config file. It is a different use case. 🤔

otaviobp: The second function is almost an exact copy of the first function. This is a big hint that there are code that could be reused. The major problem on doing that is that it's really hard on the long term to keep the behavior on both functions in sync. For example, every bug that we find on parsing the config file will probably be needed to be fixed twice. Any change in the format of the config file, error messages, etc are all duplicated.
The first function is a function that parses the config file and visit all config values found. The second function could use the first to parse the file and filter just the item that you need.

One restriction I would add here is that it's not very efficient to parse files, so if we are planning to get multiple values from the same file it's better to read it only once and to extract everything we need. Something similar to what @karthikprabhuvinod proposed on the other PR.

Anyway, I think that having a config_get_value() function is useful.

castulo: ok, now I get what you mean, I agree there is code to be reused here

src/3rd_party.c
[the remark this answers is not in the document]
castulo: done

src/bundle_add.c
```diff
@@ -417,7 +418,11 @@ enum swupd_code execute_bundle_add(struct list *bundles_list)
 	bool mix_exists;
 
 	timelist_timer_start(globals.global_times, "Load MoM");
-	current_version = get_current_version(globals.path_prefix);
+	if (globals.use_third_party_repo) {
```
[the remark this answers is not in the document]
castulo: done

src/globals.h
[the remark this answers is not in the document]
castulo: done

src/globals.h
```diff
@@ -82,6 +84,10 @@ bool set_path_prefix(char *path);
...
+void set_third_party_repo(char *repo_name);
```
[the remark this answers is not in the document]
castulo: I will remove this 👌

test/functional/testlib.bash
castulo: thanks, fixed

src/lib/config_parser.c
castulo: In the end I decided it was better to just parse all file from the start instead of looking for particular keys in the config file later

castulo:
Version 2 has the following updates:
- Removed global variables related to 3rd-party
- Parse the whole repo config file once and store the data in a list
- Avoid running scripts and heuristics when adding 3rd-party bundles

castulo:
V3

- Added the configuration options from `3rd-part bundle-add` to the default config file
- Added a test to validate we are parsing the config file correctly with sections that include subcommands like [3rd-party-bundle-add]

castulo:
V4:

- Fixing the output of a test that was not deterministic

castulo:
V5:
- Adding `3rd-party bundle-add` options to autocompletion scripts

castulo:
V5:
- Just rebased and updated the PR to use the recently merged code for 3rd-party repos.

castulo:
V6:
- Added the --repo option so the user can specify what 3rd-party repo to use to add a bundle

castulo:
V7:
- make sure the state directories of the 3rd-party repo exist before trying to add a repo

I have tested `3rd-party bundle-add my-bundle` with real content using a mix and it is working.

src/3rd_party_bundle_add.c
otaviobp: It's correct to include that after the ifdef. The config.h is included by swupd.h and it contains the compile flag that the ifdef uses.

src/helpers.c
```diff
@@ -223,17 +223,17 @@ static bool validate_tracking_dir(const char *state_dir)
...
 
-int create_state_dirs(const char *state_dir_path)
+int create_basic_state_dirs(const char *state_dir_path)
 {
 	int ret = 0;
 	unsigned int i;
 	char *dir;
 #define STATE_DIR_COUNT (sizeof(state_dirs) / sizeof(state_dirs[0]))
-	const char *state_dirs[] = { "delta", "staged", "download", "telemetry", "bundles", "3rd_party" };
+	const char *state_dirs[] = { "delta", "staged", "download", "telemetry", "bundles" };
```
otaviobp: Why do we need to create the 3rd_party separate from the rest?

src/3rd_party_repos.c
```diff
@@ -140,11 +142,44 @@ int third_party_add_repo(const char *repo_name, const char *repo_url)
...
 }
 
-void third_party_set_repo(const char *state_dir, const char *path_prefix, struct repo *repo)
+static int third_party_create_state_dirs(const char *state_dir, const char *repo_name)
+{
+	int ret = 0;
+	char *dir;
+
+	/* check for existence */
+	string_or_die(&dir, "%s/3rd_party/%s", state_dir, repo_name);
```
otaviobp: Why don't we just cal create_state_dirs() here?

src/helpers.c
castulo: because the state directories of a 3rd-party repo don't need that directory, so this way we can reuse the same function to create the state directories with these repos

src/3rd_party_repos.c
castulo: well I wasn't just calling create_state_dirs because it was creating an extra directory we don't really need in the state dir of 3rd_party repos. But I guess it really doesn't matter if that extra directory gets created, we just won't use it. I'll update this so we use create_state_dirs instead.

src/helpers.c
castulo: having said that it really doesn't hurt if that extra dir is there... I'll leave this as it was in the next update

castulo:
v8:

- Extracted some code that was really not part of this implementation and submitted it in different PRs
- Rebased the code
- Reusing the create_state_dirs function instead of adding a new one for 3rd-party

castulo:
v9:
- Decoupled the bundle_add.c and 3rd_party_bundle_add.c

castulo:
Code rebased

PR #1193, opened 2019-11-06
Re-balancing functional tests in Travis

castulo:
The number of functional tests that run in Travis have grown
considerably since the last time the jobs that run the tests were
balanced. This commit re-balances these jobs so tests take similar time
to run in each job.


castulo:
Looks like this PR is no longer relevant since the project is moving to using GitHub actions for running Functional tests instead of using Travis (PR #1196).

PR #1200, opened 2019-11-11
test: Don't conflict config tests with other tests

otaviobp:
Always save the config file in another directory so we won't have
problems on running the config tests in parallel. Before this patch
if you run the config tests in parallel with any other tests you could
end up having false-negatives.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -75,10 +75,7 @@ jobs:
       run:  scripts/build_ci.bash
 
     - name: run check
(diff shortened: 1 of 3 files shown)


castulo:
Just as a FYI, remember we have a function in testlib to write to files owned by root:
```
Usage:
    write_to_protected_file [-a] <file> <stream>

Options:
    -a    If used the text will be appended to the file, otherwise it will be overwritten
```

otaviobp:
V2:
 - Rebased on top of master to run tests on actions
 - Remove configuration to run this tests isolated
 - Using write_to_protected_file where appropriate
 - Don't skip test anymore when not in TRAVIS

PR #1203, opened 2019-11-12
actions: Build dependencies in parallel to have a faster build

otaviobp:
Download and build depedencies in parallel to try to maximize usage of CPU,
by building things at the same time we run downloads


otaviobp:
Actually I realized that this is not working. Closing PR

PR #1210, opened 2019-11-19
Repo add: Concrete implementation for third-party repo add

otaviobp:
Opening a new pull request with my proposed changes on top of karthik PR


configure.ac
```diff
@@ -83,12 +83,13 @@ AS_IF(
...
   [THIRDPARTY="no"]
 )
+AM_CONDITIONAL([ENABLE_THIRD_PARTY_TESTS], [test "x$enable_third_party" = "xyes"])
 
 AH_TEMPLATE([THIRDPARTY_REPO_PREFIX], [prefix path for third party repos])
 
 AS_IF(
   [test "$enable_third_party" = "yes"],
-  [AC_DEFINE([THIRDPARTY_REPO_PREFIX], ["/opt/user-repo/"])]
+  [AC_DEFINE([THIRDPARTY_REPO_PREFIX], ["/opt/3rd-party/"])]
```
castulo: I thought this was going to be "3rd_party"

src/3rd_party_repos.c
```diff
@@ -0,0 +1,132 @@
...
+#include "3rd_party_repos.h"
+#include "swupd.h"
+
+#include <errno.h>
+
+#ifdef THIRDPARTY
+
+static char *get_repo_config_path(void)
+{
+	return str_or_die("%s/%s/%s", globals.state_dir, "3rd_party", "repo.ini");
```
castulo: wouldn't it be better to use "sys_path_join" here instead of "str_or_die"?

src/3rd_party_repos.c
```diff
@@ -0,0 +1,132 @@
...
+{
+	int ret = 0;
+	struct list *repos;
+	char *repo_config_file_path = NULL;
+	FILE *fp = NULL;
+
+	repos = third_party_get_repos();
+	if (list_search(repos, repo_name, repo_name_cmp)) {
+		error("The repo: %s already exists\n", repo_name);
+		ret = -ENOENT;
```
castulo: ENOENT doesn't seem like the best error in this case. ENOENT is normally used when an item is not found, not when it already exists... maybe EEXIST instead?

src/3rd_party_repos.c
```diff
@@ -0,0 +1,132 @@
...
+
+	// write name
+	ret = config_write_section(fp, repo_name);
+	if (ret) {
+		goto exit;
+	}
+
+	// write url
+	ret = config_write_config(fp, "url", repo_url);
+
```
castulo: I think we should also add "version = 0" immediately after adding the "url"

src/lib/config_file.c
```diff
@@ -109,3 +110,28 @@ bool config_parse(const char *filename, parse_config_fn_t parse_config_fn, void
...
 }
+
+int config_write_section(FILE *file, const char *section)
+{
+	int ret = 0;
+
+	if (!file) {
+		return -EINVAL;
+	}
+	ret = fprintf(file, "[%s]\n", section);
```
castulo: I think we should leave an empty line before and after a section so it is better formatted:
```

[SECTION1]

url=something
version=0

[SECTION2]

usr=something/else
version=0

File: test/functional/3rd-party-repo-add/3rd-party-repo-add-negative.bats [hunk 6]
```diff
@@ -0,0 +1,55 @@
...
+# Email: karthik.prabhu.vinod@intel.com
+
+load "../testlib"
+
+@test "TRA003: Negative test, repo add usage" {
+
+	run sudo sh -c "$SWUPD 3rd-party add $SWUPD_OPTS"
+	assert_status_is "$SWUPD_INVALID_OPTION"
+	expected_output=$(cat <<-EOM
+			Error: The positional args: repo-name and repo-url are missing
```
castulo: this test have an extra tab in the expected output

src/3rd_party_repos.c
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	// write name
+	ret = config_write_section(fp, repo_name);
+	if (ret) {
+		goto exit;
+	}
+
+	// write url
+	ret = config_write_config(fp, "url", repo_url);
```
castulo: I think we should validate repo_url for two things here before writing it to the config file:
- we should validate the user has specified a protocol (http, https or file), that way we minimize the possible user error of `swupd 4rd-party add my-repo myrepo.com`
- we should warn the user about the risks of using an insecure protocol if he/she decides to use http

configure.ac
otaviobp: Yes, I'll fix that.

src/3rd_party_repos.c
otaviobp: As a general rule yes, but here we have 3 paths to join, so I'm using the str_or_die() to prevent running it twice. We need a sys_path_join() that receives a variable number of parameters.

src/3rd_party_repos.c
otaviobp: haha. true. I'll change

src/3rd_party_repos.c
otaviobp: Why?

src/lib/config_file.c
otaviobp: I personally don't like the idea on using the same separator from one section to the other and the items inside a section. This for example looks weird to me:

```
[section1]

key=value

[section2]

key=value
```

This isn't clear that the `key=value` pair is inside a section. Other problem is that it's hard not to add an empty line break before the beginning of the file, what is also weird to me. As this file isn't going to be necessarily read.

I like the style used by gitconfig, that is very easy to read, but I'm not sure if our parser supports that:

```
[section1]
    key=value
[section2]
    key=value
```

src/3rd_party_repos.c
otaviobp: > we should validate the user has specified a protocol (http, https or file), that way we minimize the possible user error of swupd 4rd-party add my-repo myrepo.com

We should be consistent and we don't validate that in mirror --set, so we sholdn't validate that in 3rd-party repo set. I don't like validating that too. We can trust curl to know how to download from the correct place and print warnings accordingly. One possibility would be to check if the connection is valid here, but I personally prefer to validate that only when needed. (kind of like git remote). If we let curl take care of the protocol all new protocols supported by curl will be automatically supported by swupd for free.

> we should warn the user about the risks of using an insecure protocol if he/she decides to use http

That's true. We need that. I'll add.

otaviobp:
V2:
 - Fixing the incorrect values used and code style
 - Enforcing the http_insecure rule

src/3rd_party_repos.c
castulo: because we read what version our system is relative to the 3rd-party repo from this file, since you are just adding the repo for the first time it means we haven't installed anything yet, that is why we are at version=0. Then when you add a bundle from that repo it try to look for packs to go from 0, which makes sense.

otaviobp:
V3:
 - Adding a line break before new section in config files if not first section
 - Adding a `version=0` line every time a new repository is added.

PR #1212, opened 2019-11-19
Fix memory leak downloading files to a full disk

castulo:
When swupd is downloading fullfiles, if it runs out of free disk space
it is exiting without freeing the curl handle. This commit fixes the
issue by  freeing it when this condition occurrs.

### DIFF
--- a/src/fullfile.c
+++ b/src/fullfile.c
@@ -45,8 +45,9 @@ static int compare_fullfile(const void *a, const void *b)
 	return file1->last_change - file2->last_change;
 }
 
-static void download_mix_file(struct file *file)
+static int download_mix_file(struct file *file)
 {
+	int ret = -1;
 	char *url, *filename;
 
 	string_or_die(&url, "%s/%i/files/%s.tar", MIX_STATE_DIR, file->last_change, file->hash);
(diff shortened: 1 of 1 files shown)


src/curl_async.c
```diff
@@ -453,7 +453,7 @@ static int process_download(struct swupd_curl_parallel_handle *h, struct multi_c
 {
 	CURL *curl = NULL;
 	CURLMcode curlm_ret = CURLM_OK;
-	CURLcode curl_ret = CURLE_OK;
+	CURLcode curl_ret = -1;
```
otaviobp: Try not to mix different types. Use only Curl code for Curl errors.
The ideal too would be to return `CURLcode` in this function too

src/curl_async.c
```diff
@@ -523,7 +524,7 @@ static int process_download(struct swupd_curl_parallel_handle *h, struct multi_c
 	return poll_fewer_than(h, h->max_xfer, h->max_xfer / 2);
 
 out_bad:
-	return -1;
+	return -curl_ret;
```
otaviobp: I don't understant the idea here. Why returning a negative curl_ret? and when you set it to -1 it will return 1. Is this expected?

src/fullfile.c
```diff
@@ -248,7 +251,15 @@ int download_fullfiles(struct list *files, int *num_downloads)
 		if (file->is_mix) {
 			download_mix_file(file);
 		} else {
-			download_file(download_handle, file);
+			ret = download_file(download_handle, file);
+			if (ret == -CURLE_WRITE_ERROR) {
```
otaviobp: What about other curl errors? Are they ok? Why do we need a behavior here different than pack downloads?
Is this problem reproduced with packs?

src/curl_async.c
castulo: I used -1 because the function already returned -1 on any error before, but I agree it is better not to mix types

src/curl_async.c
castulo: ups, my bad, I meant to return -1 here.

I decided to return curl_ret so we could have more information about the error type instead of just returning a -1 for every failure, that way we have the possibility of responding differently depending on the failure

src/fullfile.c
castulo: no, this problem doesn't seem to be happening with packs.

I was only looking for CURL_WRITE_ERROR in this case because at least I know that is a terminal error, and there is no point on keep retrying when that error occurs, with other errors it may be possible to keep trying I guess

src/curl_async.c
castulo: I'll change that

src/fullfile.c
otaviobp: This function is used both here and in packs so I would expect the code that handles it to be similar. If the problem is not happening in packs, try to do the same error handling

castulo:
@otaviobp  All comments have been addressed in this update.

PR #1213, opened 2019-11-20
Fix problem when running config-file tests

castulo:
The commit also fixes a bug with the same tests which was causing the
test environment to be left behind.

### DIFF
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -1780,6 +1780,10 @@ destroy_test_fs() { # swupd_function
 
 create_config_file() { # swupd_function
 
+	if [ "$("$SWUPD" --version | grep "config file path" | awk '{print $4}')" != "./testconfig" ]; then
+		skip "Tests that use a config file require swupd to be built with '--with-config-file-path=./testconfig'"
+	fi
+
 	if [ -e "$SWUPD_CONFIG_FILE" ]; then
 		destroy_config_file
 	fi
@@ -3248,6 +3252,9 @@ list_tests() { # swupd_function
 
 setup() {
 
+	TESTLIB_WD="$(pwd)"
+	export TESTLIB_WD
+
 	print "\\n\\n\\n$sep"
 	print "$alt_sep"
 	print "$sep\\n"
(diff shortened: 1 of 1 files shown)


test/functional/testlib.bash
```diff
@@ -1780,22 +1778,47 @@ destroy_test_fs() { # swupd_function
 
 create_config_file() { # swupd_function
 
+	SWUPD_CONFIG_DIR="$("$SWUPD" --version | grep "config file path" | awk '{print $4}')"
```
otaviobp: This isn't going to work for paths separated with ':'. like the default path.

test/functional/testlib.bash
```diff
@@ -1780,22 +1778,47 @@ destroy_test_fs() { # swupd_function
...
+
+	# when the config dir is set to a relative path, lets redirect it to the
+	# directory of the test environment so every test has its own config file
+	# this is useful when running the config tests in parallel with other tests,
+	# if the path is absolute lets leave it as it is
+	if [ "$SWUPD_CONFIG_DIR" == "./testconfig" ]; then
```
otaviobp: Why do we need to have this case supported? And why it should be specific for ./testconfig/

test/functional/testlib.bash
otaviobp: And I'm a bit concerned on overwriting any possible config file that the user may have.

castulo: yes, that is right, this won't work if the user has more than one path defined.
And about the possibility of overwriting an config file a user may have,  that should not happen , that is why I'm backing it up first if config file already exists.

test/functional/testlib.bash
castulo: well, currently if we want to run the tests in parallel, we need to create one config file per test, which forces us to change the directory where swupd reads the config file from. So if the user wants to run tests in parallel he/she will need to configure the build with ` --with-config-file-path=./testconfig` like we are doing right now. That is why I left this particular condition here, I don't see how else we could support multiple config files without having to add a flag for setting up the config file. I'm open to suggestions if you have any.

castulo:
V2:

- Removed the code that looked for an absolute path in the config file
- Skip the config tests if the env is not configured correctly to run the tests
- Make sure the working directory is returned to the original so the test environment gets removed correctly

PR #1214, opened 2019-11-21
Repo list: Rework of 3rd-party list

otaviobp:


.github/workflows/ccpp.yml
```diff
@@ -129,7 +129,7 @@ jobs:
       run:  test/functional/generate-cert.prereq
 
     - name: run check
-      run: env TESTS="$(find test/functional/os-install -name *.bats -printf '%p ')" make -e -j2 check
+      run: env TESTS="$(find test/functional/{os-install,3rd-party-repo} -name *.bats -printf '%p ')" make -e -j2 check
```
castulo: it may be a good idea to rename the job too, from test_os_install to tes_os_install_3rd_party or something like that

PR #1215, opened 2019-11-22
Repo remove: Rework of 3rd-party remove

otaviobp:


test/functional/3rd-party-repo/3rd-party-repo-remove-negative.bats
```diff
@@ -0,0 +1,88 @@
...
+@test "TPR009: Negative test, repo remove usage" {
+
+	run sudo sh -c "$SWUPD 3rd-party remove $SWUPD_OPTS"
+	assert_status_is "$SWUPD_INVALID_OPTION"
+	expected_output=$(cat <<-EOM
+			Error: The positional args: repo-name is missing
+			*.*
+	EOM
+	)
+	assert_regex_in_output "$expected_output"
```
castulo: wouldn't it look cleaner that instead of using "assert_regex_in_output" you use "assert_in_output" i these tests and you can remove the *.* part?
I don't see the point of using a regex to check the output when you are anyway only using _in_ to check for only one deterministic line within the output.

test/functional/3rd-party-repo/3rd-party-repo-remove-negative.bats
```diff
@@ -0,0 +1,88 @@
...
+
+	run sudo sh -c "$SWUPD 3rd-party remove test-repo1 junk_positional $SWUPD_OPTS"
+	assert_status_is "$SWUPD_INVALID_OPTION"
+	expected_output=$(cat <<-EOM
+			Error: Unexpected arguments
```
castulo: also, again these tests are incorrectly tabbed (with double tab instead of a single one)

test/functional/3rd-party-repo/3rd-party-repo-remove-negative.bats
```diff
@@ -0,0 +1,88 @@
...
+		[test5]
+		url=www.lmn.com
+		\n
+	EOM
+	)
+
+	repo_config_file="$STATEDIR"/3rd_party/repo.ini
+	run sudo sh -c "mkdir -p $STATEDIR/3rd_party/"
+	run sudo sh -c "mkdir -p $PATH_PREFIX/opt/3rd-party/{test1,test2,test3,test4,test5}"
+	write_to_protected_file -a "$repo_config_file" "$contents"
```
castulo: what is the point of adding all this to the test setup if it is not being used/validated anywhere on the tests. The tests are only checking the return code and the output, so what's the point of all these?
This gets proven as not useful since the directories created with this command `run sudo sh -c "mkdir -p $PATH_PREFIX/opt/3rd-party/{test1,test2,test3,test4,test5}"`are not correct since you are using the wrong directory (3rd-party instead of 3rd_party) and yet the test doesn't complain at all.

test/functional/3rd-party-repo/3rd-party-repo-remove.bats
```diff
@@ -0,0 +1,99 @@
...
+
+}
+
+test_teardown(){
+
+	destroy_test_environment "$TEST_NAME"
+
+}
+
+@test "TPR008: Remove a multiple repos" {
```
castulo: extra word "a"

test/functional/3rd-party-repo/3rd-party-repo-remove.bats
```diff
@@ -0,0 +1,99 @@
...
+
+	#remove at start of file
+	run sudo sh -c "$SWUPD 3rd-party remove test1 $SWUPD_OPTS"
+	assert_status_is "$SWUPD_OK"
+	expected_output=$(cat <<-EOM
+		Repository test1 and its contents removed successfully
+	EOM
+	)
+	assert_is_output "$expected_output"
+	assert_dir_not_exists "$PATH_PREFIX/opt/3Nrd-party/test1"
```
castulo: wrong path in this test and all below: 3Nrd-party -> 3rd_party,

src/3rd_party_remove.c
```diff
@@ -74,14 +98,29 @@ enum swupd_code third_party_remove_main(int argc, char **argv)
...
+		goto exit;
+	}
+
+	if (remove_repo_directory(argv[argc - 1]) < 0) {
+		ret = SWUPD_COULDNT_REMOVE_FILE;
+		goto exit;
 	}
 
-	/* TODO Implement */
+	info("Repository %s and its contents removed successfully\n", argv[argc - 1]);
```
castulo: I think this statement is grammatically incorrect,  the word "contents" in plural has a different meaning that what is intended here, I believe the right work is just "content" https://dictionary.cambridge.org/grammar/british-grammar/content-or-contents

src/3rd_party_repos.c
```diff
@@ -142,4 +144,88 @@ int third_party_add_repo(const char *repo_name, const char *repo_url)
...
+{
+	struct repo *repo;
+	struct list *repos;
+	int ret = 0;
+
+	repos = third_party_get_repos();
+	repo = list_remove(repo_name, &repos, repo_name_cmp);
+
+	if (!repo) {
+		info("Repository not found\n");
```
castulo: shouldn't this be an error too instead of info?

src/3rd_party_remove.c
otaviobp: wow, I didn't know that. thanks

otaviobp:
V2: Fixed the reported problems

PR #1216, opened 2019-11-25
Implement 3rd-party bundle-list

castulo:
Implementation of the `swupd 3rd-party bundle-list` command.

Signed-off-by: Castulo Martinez <castulo.martinez@intel.com>

Closes #1179

PR #1226, opened 2019-12-02
3rd party os core

castulo:
When using a 3rd-party repository we need to know the current version of the system related to the 3rd-party repository. This version may be different than the version related to the upstream content. The original idea of 3rd-party repositories was that this version would be tracked inside the repo.ini config file. This was not ideal cause it differ from the way swupd tracks version for upstream content. This PR makes a few updates so 3rd-party bundles are handled similarly to the way we handle upstream content.

The PR includes these updates:
 - Removes all the code related to handling the version key/value in the
    repo.ini file.
 - When doing 3rd-party operations, instead of getting the current version of a 3rd-party repo from the repo.ini file, it will now be gotten from the os-release file in the appropriate path_prefix for the repo.
 - Installs the os-core bundle when a 3rd-party repository is added so swupd can track the version of the 3rd-party repo.
 - Updates functions in testlib that facilitate the creation of tests.
 - Updates 3rd-party tests.

**Note**: This PR is built on top of #1216  and #1225, these need to be reviewed and merged first.

### DIFF
--- a/src/3rd_party_add.c
+++ b/src/3rd_party_add.c
@@ -72,6 +72,9 @@ enum swupd_code third_party_add_main(int argc, char **argv)
 	enum swupd_code ret = SWUPD_OK;
 	const int step_in_third_party_add = 1;
 	const char *name, *url;
+	struct list *repos = NULL;
+	struct repo *repo = NULL;
+	int repo_version;
 
 	if (!parse_options(argc, argv)) {
 		print_help();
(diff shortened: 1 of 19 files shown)


src/3rd_party_add.c
```diff
@@ -93,13 +96,44 @@ enum swupd_code third_party_add_main(int argc, char **argv)
...
+	/* get repo's latest version */
+	repo_version = get_latest_version(repo->url);
+	if (repo_version < 0) {
+		error("Unable to determine the latest version for repository %s\n\n", repo->name);
+		ret = SWUPD_INVALID_REPOSITORY;
+		goto finish;
+	}
+
+	/* the repo's "os-core" bundle needs to be installed at this moment
+	 * so we can track the version of the repo */
```
bryteise: I'm sadly a little late to this. Is the os-core bundle from a 3rd-party repository essentially the /usr/lib/os-release file?

otaviobp: We have been going back and forth on adding an os-core with only the os-release file or tracking the version number ourselves. Last conclusion was that it would be better to have an os-core with only os-release file, but I'm open to discuss that.
It's not late.

bryteise: Seems reasonable to me, thanks =).

PR #1232, opened 2019-12-11
Implementing the "3rd-party diagnose" command

castulo:
This PR adds the ability to diagnose bundles from 3rd-party repos.

### DIFF
--- a/src/3rd_party_repos.c
+++ b/src/3rd_party_repos.c
@@ -410,6 +410,76 @@ clean_and_exit:
 	return ret_code;
 }
 
+enum swupd_code third_party_run_operation_multirepo(const char *repo, run_operation_fn_t run_operation_fn, enum swupd_code expected_ret_code)
+{
+	enum swupd_code ret_code = SWUPD_OK;
+	enum swupd_code ret;
+	struct list *repos = NULL;
+	struct list *iter = NULL;
+	struct repo *selected_repo = NULL;
+	char *state_dir;
+	char *path_prefix;
+
+	/* load the existing 3rd-party repos from the repo.ini config file */
+	repos = third_party_get_repos();
+
+	/* backup the original state_dir and path_prefix values */
+	state_dir = strdup_or_die(globals.state_dir);
+	path_prefix = strdup_or_die(globals.path_prefix);
+
+	/* if the repo to be used was specified, use it,
+	 * otherwise perform operation in all 3rd-party repos */
+	if (repo) {
+
+		selected_repo = list_search(repos, repo, repo_name_cmp);
+		if (!selected_repo) {
+			error("3rd-party repository %s was not found\n\n", repo);
+			ret_code = SWUPD_INVALID_REPOSITORY;
+			goto clean_and_exit;
+		}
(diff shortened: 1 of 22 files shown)


src/3rd_party_repos.c
```diff
@@ -410,6 +410,76 @@ enum swupd_code third_party_run_operation(struct list *bundles, const char *repo
...
+			goto clean_and_exit;
+		}
+
+		/* set the appropriate variables for the selected 3rd-party repo */
+		ret_code = third_party_set_repo(state_dir, path_prefix, selected_repo);
+		if (ret_code) {
+			goto clean_and_exit;
+		}
+
+		ret_code = run_operation_fn(NULL);
```
karthikprabhuvinod: Why are we not checking the expected_ret_code in case of a repo passed as a argument?

src/3rd_party_update.c
```diff
@@ -170,66 +159,23 @@ enum swupd_code thir_party_update_main(int argc, char **argv)
...
 	update_set_option_keepcache(cmdline_option_keepcache);
 
-	/* load the existing 3rd-party repos from the repo.ini config file */
-	repos = third_party_get_repos();
-
-	/* backup the original state_dir and path_prefix values */
-	state_dir = strdup_or_die(globals.state_dir);
-	path_prefix = strdup_or_die(globals.path_prefix);
+	/* update 3rd-party bundles */
+	ret_code = third_party_run_operation_multirepo(cmdline_option_repo, update_repos, SWUPD_NO);
```
karthikprabhuvinod: expected_ret_Code = SWUPD_NO??

src/3rd_party_repos.c
castulo: because when we are just running the operation in one repo, we can just cascade the return code that we got from the operation, we don't need to handle multiple return codes.

src/3rd_party_update.c
castulo: if you run "update" on a system that is up to date you will get SWUPD_NO as return code. If an update is needed and performed correctly then SWUPD_OK is returned. In this case if any repo does an update it will return SWUPD_OK. Both of the codes are expected by the way.

karthikprabhuvinod: Make sense. Thanks for explaining.

castulo: you know what? I think I may be mistaken, I just tried running update in my system and it returned SWUPD_OK in both cases, when the system updates and when the system is up to date.... I'll double check, and if I'm wrong I'll update this PR with SWUPD_OK as expected_ret_code

castulo: confirmed, update returns SWUPD_OK when it updates and when it is already up to date. I'll update the PR. Thanks for noticing it @karthikprabhuvinod

castulo: when I was updating this in the code I realized this was setup like this so the "update --status" command works properly, that command expects SWUPD_NO if there are no updates available. When the --status flag is not being used we are overwriting the return code so it will work properly too. So it is fine as it is. 😆

PR #1235, opened 2019-12-12
3rd party housekeeping

castulo:
Housekeeping on 3rd-party commands
This commit does some housekeeping in the 3rd-party commands:
 - Initialize the function "progress" only after swupd has initialized successfully.
 - Make functions that are not used outside of a file static.
 - Avoid using regex or partial verifications on tests when possible.
 - Add 3rd-party command flags to the default config file.
 - Don't use "repo" in help menus, use the whole "repository" word.
 - Add 3rd-party sub-commands to the flag_validator script.
 - Add missing values from the autocompletion scripts.

This PR is built on top of #1232, #1233 and #1234 and have to be reviewed first.

### DIFF
--- a/config
+++ b/config
@@ -198,6 +198,22 @@
 #has_dep=[BUNDLE]
 
 
+[bundle-info]
+
+#
+# Options for the "swupd bundle-info" command
+#
+
+# Show the bundle info for the specified version (integer value)
+#version=[VER]
+
+# Show the bundle dependencies as part of the information displayed (boolean value)
+#dependencies=<true/false>
+
+# Show the bundle files as part of the information displayed (boolean value)
+#files=<true/false>
+
+
 [search-file]
 
 #
@@ -237,6 +253,9 @@
 # Attempt to proceed even if non-critical errors found (boolean value)
 #force=<true/false>
 
+# Don't check for corrupt files, only find missing files (boolean value)
+#quick=<true/false>
+
 # List files which should not exist (boolean value)
 #picky=<true/false>
 
@@ -254,7 +273,7 @@
 # Example: --bundles=os-core,vi
 #bundles=[BUNDLES]
 
-# Like --picky, but it only searches for extra files
+# Like --picky, but it only searches for extra files (boolean value)
 #extra_files_only=<true/false>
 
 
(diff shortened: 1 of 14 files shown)


karthikprabhuvinod:
@castulo: nit pick: Since you have this open, Please could you spell correct thir_party_update_main to third_party_update_main in these files src/3rd_party.c, src/3rd_party_update.c and src/swupd_internal.h

castulo:
> @castulo: nit pick: Since you have this open, Please could you spell correct thir_party_update_main to third_party_update_main in these files src/3rd_party.c, src/3rd_party_update.c and src/swupd_internal.h

Hey @karthikprabhuvinod nice catch, I'll add that correction to this PR

karthikprabhuvinod:
@castulo : The PR title is a little misleading. We are saying housekeeping but implement new 3rd-party functions

config
```diff
@@ -198,6 +198,22 @@
 #has_dep=[BUNDLE]
 
 
+[bundle-info]
+
+#
+# Options for the "swupd bundle-info" command
```
karthikprabhuvinod: Are we missing the 3rd-party repo add, delete and list here?

castulo: well, this PR sits on top of other PRs (that still have not merged), so you are looking at more that what this PR brings to the table. I forgot to add the disclaimer in the PR details. I'll add it to avoid confusions.

scripts/flag_validator.bash
```diff
@@ -5,7 +5,7 @@ SWUPD_DIR="$SCRIPTS_DIR"/..
 SWUPD="$SWUPD_DIR"/swupd
 
 swupd_commands=(info autoupdate check-update update bundle-add bundle-remove bundle-list bundle-info search-file diagnose repair os-install mirror clean hashdump 3rd-party)
```
karthikprabhuvinod: Suggestion: Maybe we have more comments for this file for certain parts to be easy to understand?

config
castulo: I just added this disclaimer "This PR is built on top of #1232, #1233 and #1234 and have to be reviewed first".

castulo: yes, I didn't included "add, remove and list" to the config file because they have no flags.

scripts/flag_validator.bash
castulo: The script does have some comments, I'll review it and see what other comments can be added to it to make it more straightforward, I'll do that in another PR if that is ok though.

castulo: The script does have some comments, I'll review it and see what other comments can be added to it to make it more straightforward, I'll do that in another PR if that is ok though.

PR #1236, opened 2019-12-16
Revert operation when theres an error adding repos

castulo:
When adding 3rd-party repositories, errors may occur, in these cases the
operation has to be reverted completely because we don't want to be left
with an invalid repo.


karthikprabhuvinod:
@castulo LGTM.

PR #1238, opened 2019-12-18
Consistency with signature verification messages

castulo:
This commit enhance the consistency and readability, of the messages
shown when the signature verification is skipped or fails.

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -95,19 +95,19 @@ bool signature_init(const char *certificate_path, const char *crl)
 			info("Certificate validity is:\n");
 			b = BIO_new_fp(stdout, BIO_NOCLOSE);
 			if (b == NULL) {
-				error("Failed to create BIO wrapping stream\n");
+				debug("Failed to create BIO wrapping stream\n");
 				goto fail;
 			}
 			/* The ASN1_TIME_print function does not include a newline... */
 			if (!ASN1_TIME_print(b, X509_get_notBefore(cert))) {
 				info("\n");
-				error("Failed to get certificate begin date\n");
+				debug("Failed to get certificate begin date\n");
 				goto fail;
 			}
 			info("\n");
 			if (!ASN1_TIME_print(b, X509_get_notAfter(cert))) {
 				info("\n");
-				error("Failed to get certificate expiration date\n");
+				debug("Failed to get certificate expiration date\n");
 				goto fail;
 			}
 			info("\n");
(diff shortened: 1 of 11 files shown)


src/signature.c
```diff
@@ -241,43 +242,45 @@ bool signature_verify(const char *file, const char *sig_file, bool print_errors)
 	/* get the signature */
 	sig_fd = open(sig_file, O_RDONLY);
 	if (sig_fd == -1) {
-		string_or_die(&errorstr, "Failed open %s: %s\n", sig_file, strerror(errno));
+		string_or_die(&errorstr, "Failed open %s: %s", sig_file, strerror(errno));
```
karthikprabhuvinod: Failed to open

src/signature.c
```diff
@@ -241,43 +242,45 @@ bool signature_verify(const char *file, const char *sig_file, bool print_errors)
...
-		string_or_die(&errorstr, "Failed to mmap %s signature\n", sig_file);
+		string_or_die(&errorstr, "Failed to mmap %s signature", sig_file);
 		goto error;
 	}
 	/* get the data to be verified */
 
 	data_fd = open(file, O_RDONLY);
 	if (data_fd == -1) {
-		string_or_die(&errorstr, "Failed open %s\n", file);
+		string_or_die(&errorstr, "Failed open %s", file);
```
karthikprabhuvinod: Failed to open

src/signature.c
```diff
@@ -444,12 +446,12 @@ static int validate_certificate(X509 *cert, const char *certificate_path, const
 
 	/* create a verification context and initialize it */
 	if (!(verify_ctx = X509_STORE_CTX_new())) {
-		error("Failed X509_STORE_CTX_new() for %s\n", certificate_path);
+		debug("Failed X509_STORE_CTX_new() for %s\n", certificate_path);
```
karthikprabhuvinod: why are we changing these from error to debug

src/version.c
```diff
@@ -176,10 +176,9 @@ static int get_version_from_url(char *url)
 	if (globals.sigcheck) {
 		sig_verified = verify_signature(url, &tmp_version);
 	} else {
-		error("FAILED TO VERIFY SIGNATURE OF %s. Operation proceeding due to \n"
-		      " --nosigcheck, but system security may be compromised\n",
-		      url);
-		journal_log_error("swupd security notice:  --nosigcheck used to bypass version signature");
+		warn("The --nosigcheck flag was used, THE SIGNATURE OF %s WILL NOT BE VERIFIED\n", url);
```
karthikprabhuvinod: Maybe we should stop using Capitalization

src/version.c
```diff
@@ -176,10 +176,9 @@ static int get_version_from_url(char *url)
...
+		info("Be aware that this compromises the system security\n\n");
```
karthikprabhuvinod: I think we need this: info("Be aware that this compromises the system security\n"); You have two line feeds

src/signature.c
castulo: Well that typo in the sentence was not part of my commit, but sure, I can fix it.

src/signature.c
castulo: Same here

src/signature.c
castulo: I explained that in the commit message:

"When performing signature verification to the MoM or to the version
file, if there is any error during the process, a series of errors are
printed to stderr. Many of these error messages are duplicated and most
of them are too detailed.

This commit filters those detailed messages so they are only seen if the
user is using the --debug flag, showing only a few more general
error messages about the signature failure."

src/version.c
castulo: yeah, I don't like the capitalization either, I didn't want to change it to lower case just so I was more conservative in my change, I can certainly change it to lowercase y @otaviobp is ok with that.

src/version.c
castulo: those two lines "\n" are on purpose, I usually add two line breaks when I print important errors to highlight them. I think it looks nicer in the output when the error message is separated from the normal workflow.

castulo:
> looks good

@otaviobp are you ok with converting those messages that right now are with UPPER CASE to be normal lower case? @karthikprabhuvinod and I agree it would look better, but wanted to double check with you

src/version.c
castulo: I'll change that to normal lower case

otaviobp:
> > looks good
> 
> @otaviobp are you ok with converting those messages that right now are with UPPER CASE to be normal lower case? @karthikprabhuvinod and I agree it would look better, but wanted to double check with you

Yes, I liked that. upper case looks like it's screaming.

src/version.c
castulo: after some discussion, we decided to leave these messages as upper case

castulo:
v2:
 - I just change "Failed open" with "Failed to open"

We decided to leave the certificate messages as uppercase.

PR #1240, opened 2019-12-19
3rd party certificate

castulo:
When running an operation, swupd uses a certificate to validate things like the MoM and the version file. The certificate that needs to be used for 3rd-party repos is different than the one that has to be used for upstream.

 This commit enables the use of the 3rd-party repo certificate that will be installed in the repo's directory with all swupd operations, except when adding the repository, in those cases, since the os-core bundle hasn't been installed yet, we need to provide the cert path through other means, like using the -C flag.

note: this PR is built on top of other PRs that have to be reviewed first.


test/functional/testlib.bash
```diff
@@ -1409,7 +1420,15 @@ create_third_party_repo() { #swupd_function
 
 	debug_msg "Creating 3rd-party repo $repo_name..."
 	create_version -r "$env_name" "$version" 0 "$format" "$repo_name"
-	create_bundle -n os-core -v "$version" -f /usr/lib/os-release:"$OS_RELEASE",/usr/share/defaults/swupd/format:"$FORMAT" -u "$repo_name" "$env_name"
+
+	# we need to create os-core which should include the os-release and Swupd_Root.pem
+	debug_msg "Creating os-core with Swupd_Root.pem and os-release..."
+	hashed_name=$(sudo "$SWUPD" hashdump --quiet "$TEST_ROOT_DIR"/Swupd_Root.pem)
```
karthikprabhuvinod: Could we a openssl bash utility to use this? Technically we shouldn't be using a tool in a test that is trying to test itself.

test/functional/testlib.bash
```diff
@@ -1470,9 +1489,13 @@ add_third_party_repo() { #swupd_function
 	debug_msg "3rd-party repo content dir: $TPWEBDIR"
 
 	# every 3rd-party repo needs to have at least the os-core bundle so this should be
-	# added by default
+	# added by default which should include the os-release and Swupd_Root.pem
 	debug_msg "Adding bundle os-core to the 3rd-party repo"
-	create_bundle -L -n os-core -v "$version" -f /usr/lib/os-release:"$OS_RELEASE",/usr/share/defaults/swupd/format:"$FORMAT" -u "$repo_name" "$env_name"
+	hashed_name=$(sudo "$SWUPD" hashdump --quiet "$TEST_ROOT_DIR"/Swupd_Root.pem)
```
karthikprabhuvinod: same here

test/functional/testlib.bash
castulo: Yeah, this is known limitation we have in our test library. When I initially wrote the test library I asked about this and I was told to just use hashdump cause it was very hard to implement something to do the same thing. I agreed because the risk seems to be small and really the only tests that get somehow affected by this are those for the hashdump itself, it was definitely not worth the effort (I still think it is not).

Anyway the testilb already uses hashdump all over the place for generating hashes for files, so even if we decided to try an implement this ourselves (which I don't think we ever will) it would be part of a different PR since it is unrelated work.

karthikprabhuvinod: Sounds good. Ya looks like we are doing something adhoc there.

PR #1242, opened 2019-12-20
Remove the state dir when removing 3rd-party repo

castulo:
When we remove a 3rd-party repo from the system, we remove the directory
where all its contents got installed, but we don't remove its state
directory.

This commit removes it since we no longer need that state directory.


karthikprabhuvinod:
Should we add a funcitonal test?

castulo:
> Should we add a funcitonal test?

Good idea, I think I can add a validation of this into an existing test

castulo:
V2:
 - Adding validation of the state dir being removed in a functional test

PR #1243, opened 2019-12-20
Verifying the URL of a repo doesn't exist already

castulo:
When adding a 3rd-party repository we need to make sure that the URL
provided by the user is not already assigned to another 3rd-arty
repository, otherwise we can end up having the same repository multiple
times, just with different names.

This commit checks the URL doesn't already exists.

### DIFF
--- a/src/3rd_party_add.c
+++ b/src/3rd_party_add.c
@@ -127,7 +127,7 @@ enum swupd_code third_party_add_main(int argc, char **argv)
 	ret = third_party_add_repo(name, url);
 	if (ret) {
 		if (ret != -EEXIST) {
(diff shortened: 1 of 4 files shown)


src/3rd_party_add.c
```diff
@@ -102,7 +102,7 @@ enum swupd_code third_party_add_main(int argc, char **argv)
 	ret = third_party_add_repo(name, url);
 	if (ret) {
 		if (ret != -EEXIST) {
-			error("Failed to add repo: %s to config\n", name);
+			error("Failed to add repository: %s to config\n", name);
```
otaviobp: +1 for using repository instead of repo. Let's try to use that in docs and messages.

PR #1245, opened 2019-12-23
Formating 3rd-party add,list,remove command output

castulo:
This commit makes a few small changes to the output of the commands to
make it look cleaner and to be consistent with the output of other swupd
commands.

### DIFF
--- a/src/3rd_party_add.c
+++ b/src/3rd_party_add.c
@@ -128,17 +128,17 @@ enum swupd_code third_party_add_main(int argc, char **argv)
 
 	/* The last two in reverse are the repo-name, repo-url */
 	progress_next_step("add_repo", PROGRESS_UNDEFINED);
+	info("Adding 3rd-party repository %s...\n\n", name);
 	ret = third_party_add_repo(name, url);
 	if (ret) {
 		if (ret != -EEXIST) {
-			error("Failed to add repository: %s to config\n", name);
+			error("Failed to add repository %s to config\n\n", name);
 			ret_code = SWUPD_COULDNT_WRITE_FILE;
 		} else {
 			ret_code = SWUPD_INVALID_OPTION;
 		}
 		goto finish;
 	}
-	info("Repository %s added successfully\n", name);
 
 	/* at this point the repo has been added to the repo.ini file */
 	repos = third_party_get_repos();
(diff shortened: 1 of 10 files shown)


src/3rd_party_add.c
```diff
@@ -180,16 +180,20 @@ enum swupd_code third_party_add_main(int argc, char **argv)
...
-		info("There was an error adding 3rd-party repository %s, reverting...\n", name);
 		set_path_prefix(path_prefix);
 		set_state_dir(state_dir);
 		ret = remove_repo(name);
-		if (!ret) {
-			info("Operation reverted\n");
-		} else {
-			error("There was an error removing the repository (errno: %d)\n", ret);
+		if (ret) {
+			error("There was an error removing the corrupt repository (errno: %d)\n\n", ret);
```
karthikprabhuvinod: Getting a little verbose .. how about "error removing the corrupt repository"

src/3rd_party_repos.c
```diff
@@ -183,15 +183,15 @@ static int overwrite_repo_config_file(struct list *repos)
 	// open it in trunc mode as we re-write all the contents again
 	fp = fopen(repo_config_file_path, "w");
 	if (!fp) {
-		error("Impossible to write to file %s", repo_config_file_path);
+		error("Impossible to write to file %s\n", repo_config_file_path);
```
karthikprabhuvinod: Unable to write to the file?

src/3rd_party_repos.c
```diff
@@ -183,15 +183,15 @@ static int overwrite_repo_config_file(struct list *repos)
...
 		goto exit;
 	}
 
 	repo = repos;
 	while (repo) {
 		ret = write_repo(fp, repo->data);
 		if (ret < 0) {
-			error("Write error when trying to write to %s", repo_config_file_path);
+			error("Write error when trying to write to %s\n", repo_config_file_path);
```
karthikprabhuvinod: Error when trying to write to

karthikprabhuvinod:
@castulo : left you with some suggestions for wording the errors.

src/3rd_party_add.c
castulo: Hmm, that would look strange cause the error() function is already going to add "Error: ", so to would read " Error: error removing the corrupt repository".
I could change it to something like "The corrupt repository failed to be removed", what do you think?

src/3rd_party_repos.c
castulo: Well, my commit didn't really modified that message I only added a missing line break, I'm not too concerned with this error message cause I don't expected it to be shown a lot. Having said that, sure, I could change it to what you suggests, the current error message seems a little odd.

src/3rd_party_repos.c
castulo: Same problem as before, starting the sentence with "error" will result in "Error": error when..."

src/3rd_party_add.c
karthikprabhuvinod: I would put the event first... but upto you.

how abt you?
Failed to remove corrupt repository

src/3rd_party_repos.c
karthikprabhuvinod: "Failed to write to %s" in that case

PR #1248, opened 2019-12-26
Incraese the total steps if inaccurate

castulo:
When performing a swupd operation, if the progress steps are defined incorrectly, a buffer overflow may be triggered. This commit makes sure the steps are still within the specified range by redefining the total number of steps to avoid the memory problem.

PR built on top of #1249 and needs to be reviewed first


src/progress.c
```diff
@@ -213,9 +218,12 @@ void progress_next_step(const char *step_title, enum progress_type type)
 		debug("Please initialize the steps\n");
 		return;
 	}
+	debug("Current step name: %s\n", step_title);
+	debug("Current step number: %d (out of %d)\n", current_step + 1, total_steps);
 	if (current_step >= total_steps) {
-		debug("The current step (%d) is >= total steps (%d)\n", current_step, total_steps);
-		return;
+		debug("Warning: The current steps exceed the number of steps defined for the current operation\n");
```
karthikprabhuvinod: the current step exceeds the number of steps defined for the current operation

castulo: I actually meant the "current steps", because what we define for each operation is the Total number of steps, so here it is saying that the current number of steps that have run exceed the amount defined

PR #1249, opened 2019-12-27
Fix the number of steps in swupd functions

castulo:
In order to report progress accuratelly, currently swupd requires a
hardcoded number of steps per operation so we can report how far in the
operation we are. Most of these step totals are wrong.

This commit fixes the number of steps in many swupd functions.


src/os_install.c
```diff
@@ -199,8 +185,26 @@ enum swupd_code install_main(int argc, char **argv)
...
+	 *  9) run_postupdate_scripts
+	 */
+	if (cmdline_option_download) {
+		steps_in_os_install = 7;
+	} else {
+		steps_in_os_install = 9;
+	}
 	progress_init_steps("os-install", steps_in_os_install);
+
+	/* install */
```
karthikprabhuvinod: verify?

src/3rd_party_bundle_add.c
```diff
@@ -114,22 +114,10 @@ enum swupd_code third_party_bundle_add_main(int argc, char **argv)
...
-	 *  2) download_packs
-	 *  3) extract_packs
-	 *  4) validate_fullfiles
-	 *  5) download_fullfiles
-	 *  6) extract_fullfiles
-	 *  7) install_files
-	 */
-
-	const int steps_in_bundleadd = 7;
+	const int steps_in_bundleadd = 8;
```
karthikprabhuvinod: bundle_add

karthikprabhuvinod: did we decide to use them as a single_word?

src/os_install.c
castulo: this is actually an install, but we use the verify function for it

src/3rd_party_bundle_add.c
castulo: here it does not matter, is just a variable name

src/os_install.c
karthikprabhuvinod: okay

PR #1250, opened 2020-01-02
Show what indirect dependencies are optional

castulo:
When using "bundle-info BUNDLE --dependencies", information regarding
BUNDLE is presented to the user, this information includes the list of
direct and indirect dependencies BUNDLE has. All direct includes that
are optional (also-add) are marked as such, while there is no way to
distiguish what indirect dependencies are optional with the provided
information.

This commit expands the information shown by the command so indirect
dependencies that are optional are shown as such along with info about
the status of that dependency.

### DIFF
--- a/src/bundle_info.c
+++ b/src/bundle_info.c
@@ -129,6 +129,7 @@ static bool parse_options(int argc, char **argv)
 static void print_bundle_dependencies(struct manifest *manifest, struct list *indirect_includes)
 {
 	struct list *iter;
+	struct sub *sub;
 	char *dep;
 
 	/* print direct includes/also-add first */
@@ -144,14 +145,14 @@ static void print_bundle_dependencies(struct manifest *manifest, struct list *in
 			/* optional dependencies can be installed or not, so we
 			 * can also provide this info to the user, this only make
 			 * sense if the requested bundle is also installed */
+			info(" - ");
+			print("%s", dep);
 			if (is_installed_bundle(manifest->component)) {
-				info(" - ");
-				print("%s", dep);
 				info(" (optional, %s)", is_installed_bundle(dep) ? "installed" : "not installed");
-				print("\n");
 			} else {
-				print(" - %s (optional)\n", dep);
+				info(" (optional)");
 			}
+			print("\n");
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


castulo:
There is a similar problematic in this command to that discussed in PR #1251.

Currently when using the `--dependencies` option with `bundle-info`, swupd shows the list of dependencies needed by the specified bundle, using "editors" as an example:
```
$ sudo swupd bundle-info editors --dependencies
__________________________
 Info for bundle: editors
__________________________

Status: Installed

Bundle editors is up to date:
 - Installed bundle last updated in version: 31980
 - Latest available version: 31980

Bundle size:
 - Size of bundle: 921.13 KB
 - Size bundle takes on disk (includes dependencies): 427.76 MB

Direct dependencies (4):
 - os-core
 - emacs (optional, installed)
 - joe (optional, installed)
 - vim (optional, installed)

Indirect dependencies (6):
 - findutils
 - lib-openssl
 - libstdcpp
 - p11-kit
 - perl-basic
 - python3-basic
```
As it is shown in this example, the direct dependencies show which of these are optional and which are not. However the Indirect dependencies don't make this distinction. This can be misleading because it looks like all indirect dependencies are mandatory, when in fact, they are not. In this example, all of the indirect dependencies are **mandatory** dependencies of "emacs", "joe" or "vim", which are **optional**, and because of this, they may or may not be installed when the "editors" bundle is installed.

To avoid the problem mentioned above, with this PR, I decided to show all dependencies (mandatory or optional) as optional if the bundle that is including them is optional. Using the same example it would show like this"
```
$ sudo ./swupd bundle-info editors --dependencies
__________________________
 Info for bundle: editors
__________________________

Status: Installed

Bundle editors is up to date:
 - Installed bundle last updated in version: 31980
 - Latest available version: 31980

Bundle size:
 - Size of bundle: 921.13 KB
 - Size bundle takes on disk (includes dependencies): 427.76 MB

Direct dependencies (4):
 - os-core
 - emacs (optional, installed)
 - joe (optional, installed)
 - vim (optional, installed)

Indirect dependencies (6):
 - findutils (optional, installed)
 - lib-openssl (optional, installed)
 - libstdcpp (optional, installed)
 - p11-kit (optional, installed)
 - perl-basic (optional, installed)
 - python3-basic (optional, installed)
```
This example is a little radical because all its bundles are optional.

If direct dependencies are mandatory, then their indirect bundles show either as optional or mandatory depending on their actual status. For example:
```
$ sudo ./swupd bundle-info editors-dev --dependencies
Password:
______________________________
 Info for bundle: editors-dev
______________________________

Status: Installed

Bundle editors-dev is up to date:
 - Installed bundle last updated in version: 31930
 - Latest available version: 31930

Bundle size:
 - Size of bundle: 149.30 KB
 - Size bundle takes on disk (includes dependencies): 1.32 GB

Direct dependencies (18):
 - binutils
 - dev-utils
 - devpkg-ncurses
 - devpkg-systemd
 - diffutils
 - editors
 - git
 - glibc-locale
 - gzip
 - make
 - os-core
 - os-core-dev
 - patch
 - plzip
 - strace
 - unzip
 - which
 - xz

Indirect dependencies (92):
 - bc
 - bison
 - c-basic
 - ... lots more bundles...
 - python3-basic
 - rsync
 - texinfo
 - the_silver_searcher (optional, not installed)
 - valgrind
 - vim (optional, installed) 
```
 
But I'm unsure if this is the way to go or if I should just be printing the actual status of an indirect bundle regardless of the status of the direct bundle that includes it. I'd like to hear your opinions @otaviobp @karthikprabhuvinod .

PR #1251, opened 2020-01-03
Show optional includes with bundle-list --has-dep

### LINKED ISSUE #1159: swupd bundle-list --has-dep is not showing optional includes (also-add)
The `swupd bundle-list --has-dep BUNDLE` should list all bundles that have BUNDLE as an include or as an also-add. Currently it is only listing bundles that include BUNDLE.

**To Reproduce**
```
$ swupd bundle-list --has-dep vim
No bundles have vim as a dependency


$ sudo swupd bundle-info --dependencies editors
__________________________
 Info for bundle: editors
__________________________

Status: Installed

Bundle editors is up to date:
 - Installed bundle last updated in version: 31270
 - Latest available version: 31270

Bundle size:
 - Size of bundle: 692.64 KB
 - Size bundle takes on disk (includes dependencies): 440.29 MB

Direct dependencies (4):
 - os-core
 - emacs (optional, not installed)
 - joe (optional, installed)
 - vim (optional, installed)

Indirect dependencies (6):
 - findutils
 - lib-openssl
 - libstdcpp
 - p11-kit
 - perl-basic
 - python3-basic
```

castulo:
Using the "bundle-list --has-dep BUNDLE" command shows a list of bundles
that have BUNDLE as a dependency. However this list is not considering
the bundles that have BUNDLE as an optional dependency.

This commit completes the list by including these bundles.

Closes #1159 

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -62,8 +62,9 @@ bool is_tracked_bundle(const char *bundle_name)
 
 	return ret;
 }
+
 /* Return list of bundles that include bundle_name */
-int required_by(struct list **reqd_by, const char *bundle_name, struct manifest *mom, int recursion, struct list *exclusions, char *msg)
+int required_by(struct list **reqd_by, const char *bundle_name, struct manifest *mom, int recursion, struct list *exclusions, char *msg, bool include_optional)
 {
 	struct list *b, *i;
 	char *name;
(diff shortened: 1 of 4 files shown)


castulo:
Here is an example of how this implementation works. If you run the command to see what bundles have "vim" as a dependency it show this:
```
$ sudo ./swupd bundle-list --has-dep vim
Installed bundles that have vim as a dependency:
 - editors (as optional)

Total: 1
```
So swupd will tell you that the "editors" bundle has "vim" as an optional dependency.

However there are four other bundles that have "editors" as a mandatory dependency. So if you run the same command for the editors bundle, you will get this result:
```
$ sudo ./swupd bundle-list --has-dep editors
Installed bundles that have editors as a dependency:
 - editors-dev
 - os-clr-on-clr
 - os-clr-on-clr-dev
 - user-basic

Total: 4
```
In this case, since vim is only an optional dependency for editors, it may not necessarily be a dependency for the other four bundles, so I decided to not recurse bundles when looking for bundles that optionally depend on another bundle. If I had recurse bundles, the result from the first command would have looked like this:
```
$ sudo ./swupd bundle-list --has-dep vim
Installed bundles that have editors as a dependency:
 - editors (as optional)
 - editors-dev
 - os-clr-on-clr
 - os-clr-on-clr-dev
 - user-basic

Total: 5
```
I think this just makes the result confusing because it would look like vim is a mandatory dependency of the other bundles, so I prefer not to show those bundles....

I guess another option would be to just show it as an optional dependency of the other bundles as well, something like this:
```
$ sudo ./swupd bundle-list --has-dep vim
Installed bundles that have editors as a dependency:
 - editors (as optional)
 - editors-dev (as optional)
 - os-clr-on-clr (as optional)
 - os-clr-on-clr-dev (as optional)
 - user-basic (as optional)

Total: 5
```
 
I would like to hear opinions? @otaviobp @karthikprabhuvinod

karthikprabhuvinod:
@castulo : I feel that term "optional dependency" is kind of misnomer. Generally, as a user, I would expect a dependency to be something mandatory(like a blocker for other bundles) and not print any optional ones.


```
sudo ./swupd bundle-list --has-dep vim
No bundles have vim as a dependency
Total: 0
```

I think isnt this the current behavior?

**OR**

Maybe we can enable what you said by means of another option like --optional. Just an idea

castulo:
> @castulo : I feel that term "optional dependency" is kind of misnomer. Generally, as a user, I would expect a dependency to be something mandatory(like a blocker for other bundles) and not print any optional ones.
> 
> ```
> sudo ./swupd bundle-list --has-dep vim
> No bundles have vim as a dependency
> Total: 0
> ```
> 
> I think isnt this the current behavior?
> 

Yes, the current behavior is not showing those, but in a previous discussion we determined that to be a bug.

> 
> Maybe we can enable what you said by means of another option like --optional. Just an idea

Even if we added an --optional flag it would not solve the problem I asked above about how to print the information in a way that is clear to the user.

PR #1255, opened 2020-01-03
Adding command descriptions

### LINKED ISSUE #1108: Add --help description of commands
Some commands don't have a description on --help, check-update, bundle-add, bundle-list, bundle-remove, mirror, update.

castulo:
This commit adds a shoort description for the help menu of all swupd
commands.

Closes #1108 

### DIFF
--- a/src/check_update.c
+++ b/src/check_update.c
@@ -34,6 +34,7 @@
 
 static void print_help(void)
 {
+	print("Checks whether an update is available and prints out the information if so\n\n");
 	print("Usage:\n");
 	print("   swupd check-update [OPTION...]\n\n");
 	//TODO: Add documentation explaining this command
--- a/src/3rd_party.c
+++ b/src/3rd_party.c
@@ -42,6 +42,7 @@ static struct subcmd third_party_commands[] = {
 
 static void print_help(const char *name)
 {
+	print("Gives users the ability for handling custom content not provided upstream\n\n");
 	print("Usage:\n");
 	print("    swupd %s SUBCOMMAND [OPTION...]\n\n", sys_basename(name));
 	print("Help Options:\n");
--- a/src/3rd_party_add.c
+++ b/src/3rd_party_add.c
@@ -28,6 +28,9 @@
 
 static void print_help(void)
 {
+	/* TODO(castulo): we need to change this description to match that of the
+	 * documentation once the documentation for this content is added */
+	print("Adds a repository from which 3rd-party content can be installed\n\n");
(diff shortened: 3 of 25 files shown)


src/check_update.c
```diff
@@ -34,6 +34,7 @@
 
 static void print_help(void)
 {
+	print("Checks whether an update is available and prints out the information if so\n\n");
```
karthikprabhuvinod: I would quote this as

Checks whether an update is available along with current & server version

karthikprabhuvinod:
@castulo : Hey Castulo , this is a good addition. I was thinking we should have something like example section for all swupd commands which lists a few popular ways of using it.... helps a beginner ramp up fast if he/she comes from another linux distro world?

for eg for update: 
**Usage:**

**Description:**

**Example:**
swupd update -V 10  -- update to a specific version
swupd update            -- updates to latest version 

Maybe we can discuss this once @otaviobp is back. Let me know what you feel?

castulo:
> @castulo : Hey Castulo , this is a good addition. I was thinking we should also have something like example section for all swupd commands which lists a few popular ways of using it.... helps a beginner ramp up fast if he/she comes from another linux distro world?
> 
> for eg for update:
> **Usage:**
> 
> **Description:**
> 
> **Example:**
> swupd update -V 10 -- update to a specific version
> swupd update -- updates to latest version
> 
> Maybe we can discuss this once @otaviobp is back. Let me know what you feel?

👍 yeah we can discuss it with @otaviobp, I like the idea of having examples but I'm not sure the "--help" menus are the best place for them or if we should just have them in the manual. In either case that is separate from this PR.

src/check_update.c
castulo: Well, to check if there is an update available it needs to check the current version and the server version. So it kind of seems redundant to me that way.

karthikprabhuvinod: hmm thinking of it makes sense

karthikprabhuvinod:
@castulo : Suggest to add a print inside print_help for third_party command like

print("3rd-party provides users a mechanism to build, add custom packages which are not present upstream \n\n");

Because if I am new to swupd and type sudo ./swupd 3rd-party ...there is no way for me to really understand what the term actually means

castulo:
> @castulo : Suggest to add a print inside print_help for third_party command like
> 

Good catch @karthikprabhuvinod , I forgot that general command. I'll add a short description to it.

castulo:
Moved this commit to PR #1260 so I'm closing this PR.

Issue #1120, opened 2019-10-01
Create autoupdate tests


otaviobp:

karthikprabhuvinod:
We started with adding one with https://github.com/clearlinux/swupd-client/pull/1138. What other tests were you thinking of @otaviobp ??

PR #1256, opened 2020-01-04
autoupdate: adding auto-update test back

karthikprabhuvinod:
Signed-off-by: Karthik Prabhu Vinod karthik.prabhu.vinod@intel.com

### DIFF
--- a/scripts/build_ci.bash
+++ b/scripts/build_ci.bash
@@ -40,3 +40,11 @@ sudo apt-get install -y clang-format-9
 autoreconf --verbose --warnings=none --install --force
 ./configure CFLAGS="$CFLAGS -fsanitize=address -Werror" --prefix=/usr --enable-third-party --with-fallback-capaths=./swupd_test_certificates --with-systemdsystemunitdir=/usr/lib/systemd/system --with-config-file-path=./testconfig
 make -j$CORES
+
+# Needed to initialize the host for auto-update. Without these steps auto-update
+# within a --path would just not work
+sudo mkdir -p /usr/lib/systemd/system/
+sudo cp data/swupd-update.service /usr/lib/systemd/system/
+sudo cp data/swupd-update.timer /usr/lib/systemd/system/
+sudo systemctl enable swupd-update.timer
+sudo systemctl restart swupd-update.timer
--- a/test/functional/autoupdate/aup-basic.bats
+++ b/test/functional/autoupdate/aup-basic.bats
@@ -7,26 +7,20 @@ load "../testlib"
 
 test_setup() {
 
-	skip "This test is broken on github actions"
-
 	create_test_environment "$TEST_NAME"
 
 }
 
 test_teardown() {
 
-	if [ -z "$TRAVIS" ]; then
-		return
-	fi
-
 	destroy_test_environment "$TEST_NAME"
 
 }
 
-@test "AUT001: Basic test, Check auto-update enable" {
+@test "AUT001: auto-update disable-enable & status" {
 
 	# perform autoupdate disable
-	run sudo sh -c "$SWUPD autoupdate --disable --path=$PATH_PREFIX"
+	run sudo sh -c "$SWUPD autoupdate --disable $SWUPD_OPTS"
 	assert_status_is "$SWUPD_OK"
 	expected_output=$(cat <<-EOM
 		Warning: disabling automatic updates may take you out of compliance with your IT policy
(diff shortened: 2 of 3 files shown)


karthikprabhuvinod:
https://github.com/clearlinux/swupd-client/issues/1120

test/functional/autoupdate/aup-basic.bats
```diff
@@ -59,3 +52,45 @@ test_teardown() {
 	assert_status_is_not "$SWUPD_OK"
 
 }
+
+@test "AUT002: Check auto-update status in different root" {
```
castulo: aren't you doing basically the same thing as the previous test in this one?

test/functional/autoupdate/aup-basic.bats
```diff
@@ -7,26 +7,19 @@ load "../testlib"
...
-	if [ -z "$TRAVIS" ]; then
-		return
-	fi
-
 	destroy_test_environment "$TEST_NAME"
 
 }
 
-@test "AUT001: Basic test, Check auto-update enable" {
+@test "AUT001: auto-update disable-enable in different root" {
```
castulo: why did you add "in different root"? because we are using a `--path`?

test/functional/autoupdate/aup-basic.bats
karthikprabhuvinod: So the second one is exclusively for checking the status of auto-update. I created a different test because the previous test was getting too big

test/functional/autoupdate/aup-basic.bats
karthikprabhuvinod: Yes...should I use a different name?

castulo: all our tests (or almost all) use --path, so I would just not specify that at all.

test/functional/autoupdate/aup-basic.bats
castulo: but you are still doing the same stuff from the first test plus some more stuff, so I don't see how that is going to help with that....

Ideally every test should be atomic, so AUT001 should ideally be split into two tests, one which enables autoupdate (from an initially disabled system) and another one which disables it (from an initially enabled system). The same thing for the autoupdate tests you are writing (for checking the status), there should be two tests, one for checking the status when the system has autoupdate disabled and one when is enabled. Those conditions, the ones I'm referring to when saying "initially" are supposed to be set up in the test_setup() function, so the tests would be split into two or more test files.

Having said that, if you still just want to add that other verification I would just do it in the same test (AUT001), the only thing you are not doing from AUT001 in AUT002 is to check the existence of files:
```
# check for files(should exist and point to /dev/null)
run sudo sh -c "test -c $PATH_PREFIX/etc/systemd/system/swupd-update.service"
assert_status_is "$SWUPD_OK"
run sudo sh -c "test -c $PATH_PREFIX/etc/systemd/system/swupd-update.timer"
assert_status_is "$SWUPD_OK"
```
So you could just add your verifications to this test and maybe change the code above with something like this"
```
assert_file_exists $PATH_PREFIX/etc/systemd/system/swupd-update.service
assert_file_exists $PATH_PREFIX/etc/systemd/system/swupd-update.timer
```
At least that way you are saving a couple of lines of code, plus you would be using the assertions from the testlib which is supposed to be the correct way of verifying results.

test/functional/autoupdate/aup-basic.bats
karthikprabhuvinod: Okay will change that

test/functional/autoupdate/aup-basic.bats
karthikprabhuvinod: Makes sense. Okay I will add the verifications to the same test.

for the assert_file_exists i remember you asking of this change before..The problem these files are charcters files as they point to /dev/null and assert_file_exists. I will change the assert_file_exist in that case.

castulo: > for the assert_file_exists i remember you asking of this change before..The problem these files are charcters files as they point to /dev/null and assert_file_exists. I will change the assert_file_exist in that case.

Ahh I didn't remember that, but it makes sense that I complaint about that before haha 😝

test/functional/autoupdate/aup-basic.bats
```diff
@@ -53,9 +53,7 @@ test_teardown() {
 	assert_is_output "$expected_output"
 
 	# check for files(should not exist)
-	run sudo sh -c "test -c $PATH_PREFIX/etc/systemd/system/swupd-update.service"
-	assert_status_is_not "$SWUPD_OK"
-	run sudo sh -c "test -c $PATH_PREFIX/etc/systemd/system/swupd-update.timer"
-	assert_status_is_not "$SWUPD_OK"
+	assert_file_not_exists "$PATH_PREFIX/etc/systemd/system/swupd-update.service"
+	assert_file_not_exists "$PATH_PREFIX/etc/systemd/system/swupd-update.timer"
```
castulo: don't you also need to run "autoupdate" again to check for the correct status after enabling it?

karthikprabhuvinod: I didn't add it initially as it would have created confusion. But I see why not. Please let me know if that explanation looks good for it

castulo: oh so there is not way to use autoupdate to get the status when using a --path? I'm confused now 😛

scripts/build_ci.bash
```diff
@@ -10,6 +10,8 @@ sudo apt-get install check
 sudo apt-get install bats
 sudo pip install coverxygen
 sudo ln -s /bin/systemctl /usr/bin/systemctl
+sudo systemctl mask swupd-update.service
+sudo systemctl mask swupd-update.timer
```
otaviobp: If you mask the timer and service you aren't testing if the disabled is really working.
I checked the current error and systemctl is falling with message:

```
External command: Created symlink /home/runner/work/swupd-client/swupd-client/aup-basic_AUT001/testfs/target-dir/etc/systemd/system/swupd-update.service → /dev/null.
External command: Created symlink /home/runner/work/swupd-client/swupd-client/aup-basic_AUT001/testfs/target-dir/etc/systemd/system/swupd-update.timer → /dev/null.
External command: Failed to stop swupd-update.service: Unit swupd-update.service not loaded.
External command: Failed to stop swupd-update.timer: Unit swupd-update.timer not loaded.
```

Did you try to initialize the service instead of masking it?

karthikprabhuvinod: @otaviobp : You were right...placing starting the service unit files inside and starting the swupd-update.timer helped it. I did need the mask.

PR #1262, opened 2020-01-08
Add functional test for info

karthikprabhuvinod:
Add info tests

### DIFF
--- a/test/functional/info/info-basic.bats
+++ b/test/functional/info/info-basic.bats
@@ -1,0 +1,51 @@
+#!/usr/bin/env bats
+
+# Author: Karthik Prabhu Vinod
+# Email: karthik.prabhu.vinod@intel.com
+
+load "../testlib"
+
+test_setup() {
+
+	create_test_environment "$TEST_NAME" 10 1
+
+}
+
+test_teardown() {
+
+	destroy_test_environment "$TEST_NAME"
+
+}
+
+
+@test "INF001: Basic test, Check info verbose" {
+
+	run sudo sh -c "$SWUPD info $SWUPD_OPTS --verbose"
+	assert_status_is "$SWUPD_OK"
+	expected_output=$(cat <<-EOM
+		Distribution:      Swupd Test Distro
+		Installed version: 10 (format 1)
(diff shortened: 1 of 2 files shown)


test/functional/info/info-basic.bats
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bats
+
+# Author: Karthik Prabhu Vinod
+# Email: karthik.prabhu.vinod@intel.com
+
+load "../testlib"
+
+test_setup() {
```
castulo: you don't need to change anything in this PR, but for future reference, if you don't define any test_setup() a default one is run by testlib, and the default one looks like the one you created, it just creates a simple test environment. Same with the test_teardown(), so you could have gotten away with not adding them to this test.

test/functional/info/info-basic.bats
```diff
@@ -0,0 +1,51 @@
...
+	assert_status_is "$SWUPD_OK"
+	expected_output=$(cat <<-EOM
+		Distribution:      Swupd Test Distro
+		Installed version: 10 \(format 1\)
+		Version URL:       .*web-dir
+		Content URL:       .*web-dir
+	EOM
+	)
+
+	assert_regex_is_output "$expected_output"
```
castulo: avoid using regex to check output when possible, in this case you can just do:
```
	expected_output=$(cat <<-EOM
		Distribution:      Swupd Test Distro
		Installed version: 10 (format 1)
		Version URL:       file://$TEST_DIRNAME/web-dir
		Content URL:       file://$TEST_DIRNAME/web-dir
	EOM
	)

```
I think it looks much cleaner and provides better output when there are failures.

karthikprabhuvinod: Ya agreed. Thanks will change

test/functional/info/info-basic.bats
karthikprabhuvinod: I need a to create an environment with a format other than staging. Looks like the teardown wasn't needed.  But will keep it for clarity.

castulo: ahh I see, then yes, you needed that 😄

PR #1263, opened 2020-01-09
Showing how much space was freed when cleaning

### LINKED ISSUE #1107: On Swupd clean print the number of bytes removed after running function

castulo:
Closes #1107

### DIFF
--- a/src/clean.c
+++ b/src/clean.c
@@ -30,6 +30,8 @@
 #define FLAG_ALL 2000
 #define FLAG_DRY_RUN 2001
 
+static const int BLOCK_SIZE = 512;
+
 static void print_help(void)
 {
 	print("Remove cached content used for updates from state directory\n\n");
@@ -52,6 +54,7 @@ static struct {
 
 static struct {
 	int files_removed;
+	long bytes_removed;
 } stats;
 
 int clean_get_stats(void)
@@ -114,6 +117,8 @@ static enum swupd_code remove_if(const char *path, bool dry_run, remove_predicat
 {
 	int ret = SWUPD_OK;
 	DIR *dir;
+	struct stat st;
+	long size;
 
 	dir = opendir(path);
 	if (!dir) {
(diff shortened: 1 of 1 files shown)


src/clean.c
```diff
@@ -142,12 +147,25 @@ static enum swupd_code remove_if(const char *path, bool dry_run, remove_predicat
...
 		}
 
+		if (lstat(file, &st) == 0) {
+			/* a file being removed (unlinked) may have many hardlinks which may
+			 * stay in the system, so the only files being trully removed are
+			 * those which only have one inode left in the system.
+			 * This also means that when doing dry-run we can only guess how
+			 * much space we would free since we cannot know what inodes will
+			 * have all their hardlinks removed */
+			size = st.st_nlink == 1 ? (st.st_blocks * BLOCK_SIZE) : 0;
```
otaviobp: Isn't there a way to get the size of the file using stat?
We can't assume the block size is 512

phmccarty: Are you referring to `st.size`?

castulo: @otaviobp yeah, I tried using stat at first, explicitly `st.size` as @phmccarty mentioned, but the calculation was off by a few mega bytes when I did that.

I don't know how disk allocation in the file system works for sure, I looked everywhere to try to find information about this and I could only find pieces of information here and there, so I can only theorize about how this works. Having said that, my theory is that `st.size` reports the actual size of the data in the file, but the file may be a little larger because space for it is reserved in blocks of 512 bytes, so there is reserved space not accounted for if just looking at `st.size`, it is a marginal size but it adds up if there are many files. Another thing that throws off the calculation is that empty files have a `st.size` of 0, but I think they are actually not 0 since they should have at least a few bytes of space reserved for metadata. One more thing to factor in is that lots of the files we have in the statedir are hardlinked somewhere else in the system, so the actual data for the file is really not going away, only one of the hard links (and I guess these have also a few bytes of metadata).

Bottom line, when I was using `st.size` to calculate how much space was being freed it was always calculating a few mega bytes short (about 20 Mb in my tests) and when I started using the blocks to calculate that, I started getting a lot closer to the actual bytes freed (about 2 Mb in my tests if I recall correctly). For reference, I was using the `du -bs STATEDIR` command to manually calculate how much space was actually freed and to compare that to what swupd reported.

src/clean.c
```diff
@@ -30,6 +30,8 @@
 #define FLAG_ALL 2000
 #define FLAG_DRY_RUN 2001
 
+static const int BLOCK_SIZE = 512;
```
otaviobp: Use a define here.

src/clean.c
otaviobp: The block size of a filesystem can be different, but stat is always reporting the number of blocks used in 512 bytes, so it's correct the way it is.

PR #1265, opened 2020-01-10
Fix name of 3rd-party directory

castulo:
The name of the 3rd-party repository directory is currently `3rd_party`, this name is incorrect and needs to be changed to `3rd-party`. This PR fixes that.

### DIFF
--- a/src/3rd_party_repos.c
+++ b/src/3rd_party_repos.c
@@ -28,7 +28,7 @@
 
 static char *get_repo_config_path(void)
 {
-	return str_or_die("%s/%s/%s", globals.state_dir, "3rd_party", "repo.ini");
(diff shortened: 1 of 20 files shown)


src/3rd_party_repos.c
```diff
@@ -28,7 +28,7 @@
 
 static char *get_repo_config_path(void)
 {
-	return str_or_die("%s/%s/%s", globals.state_dir, "3rd_party", "repo.ini");
+	return str_or_die("%s/%s/%s", globals.state_dir, "3rd-party", "repo.ini");
```
otaviobp: Can we have a constant defining that?

castulo: sounds good, I'll change that

castulo:
V2:
 - Use a constant to define the name of the 3rd-party directory to use instead of hard-coding it.

PR #1266, opened 2020-01-13
Use the user specified path for 3rd-party cert

castulo:
When using 3rd-party repos, swupd looks for the certificate to use
inside the repository directory, if the user specifies another location
it is just ignored.

If the user specifies a different path for the certificate, it should be
used. This commit fixes that.

### DIFF
--- a/src/3rd_party_repos.c
+++ b/src/3rd_party_repos.c
@@ -276,9 +276,14 @@ enum swupd_code third_party_set_repo(struct repo *repo, bool sigcheck)
 	set_content_url(repo->url);
 	set_version_url(repo->url);
 
-	/* set up swupd to use the certificate from the 3rd-party repository */
+	/* set up swupd to use the certificate from the 3rd-party repository,
+	 * unless the user is specifying a path for the certificate to use */
 	string_or_die(&repo_cert_path, "%s%s/%s%s", globals_bkp.path_prefix, SWUPD_3RD_PARTY_BUNDLES_DIR, repo->name, CERT_PATH);
-	set_cert_path(repo_cert_path);
+	if (!globals.user_defined_cert_path) {
+		/* the user did not specify a cert, use repo's default */
+		set_cert_path(repo_cert_path);
+	}
+
 	/* if --nosigcheck was used, we do not attempt any signature checking */
 	if (sigcheck) {
 		signature_deinit();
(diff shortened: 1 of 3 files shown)


karthikprabhuvinod:
LGTM to me @castulo

src/3rd_party_repos.c
```diff
@@ -263,12 +263,19 @@ enum swupd_code third_party_set_repo(const char *state_dir, const char *path_pre
 	char *repo_path_prefix;
 	char *repo_cert_path;
 
+	/* TODO(castulo): we may want to inform the user that some global
```
otaviobp: is this still valid?

castulo: no, per the discussion we had over IRC the other day we no longer need this TODO... I'll remove it.

castulo:
V2:

- Removed a TODO that was not applicable anymore.

src/3rd_party_repos.c
```diff
@@ -266,9 +266,14 @@ enum swupd_code third_party_set_repo(const char *state_dir, const char *path_pre
 	set_content_url(repo->url);
 	set_version_url(repo->url);
 
-	/* set up swupd to use the certificate from the 3rd-party repository */
+	/* set up swupd to use the certificate from the 3rd-party repository,
+	 * unless the user is specifying a path for the certificate to use */
 	string_or_die(&repo_cert_path, "%s/%s/%s/%s", path_prefix, SWUPD_3RD_PARTY_BUNDLES_DIR, repo->name, CERT_PATH);
-	set_cert_path(repo_cert_path);
+	if (strcmp(globals.cert_path, CERT_PATH) == 0) {
```
otaviobp: What if you run:
```swupd 3rd-party bundle-add test -C /usr/share/clear/update-ca/Swupd_Root.pem```

castulo: then that value would be overwritten. The problem is that we don't have a way of telling if the user set up a global variable or not... so this approach works fine with the exception of the one corner case you just mentioned. Do you have another suggestion that would not have this limitation?

otaviobp: It's not a minor corner case. If you have a 3rd party repo that is maintained by the same maintainer as your main distro you may want to use the same public key to validate that, so I think that this is a very possible use case.
I'll take a look at the code and see what we can do about that.

castulo:
V3:
- added a new global variable that would indicate if a user specified a cert path.

PR #1267, opened 2020-01-13
Moving the content from 3rd-party bundles

castulo:
This commit moves the path where the content from 3rd-party bundles is
going to be installed from opt/3rd-party/<bundle_name> to
/opt/3rd-party/bundles/<bundle_name> so bundle directories are separated
from other top level directories like bin.


bryteise:
LGTM

PR #1268, opened 2020-01-14
Do not run scripts after updating from 3rd-party

castulo:
Swupd should not run post-update scripts after updating content from a
3rd-party repo since it may threaten the security of the system.

### DIFF
--- a/src/3rd_party_update.c
+++ b/src/3rd_party_update.c
@@ -127,10 +127,12 @@ static enum swupd_code update_repos(UNUSED_PARAM char *unused)
 {
 	/* Update should always ignore optional bundles */
 	globals.skip_optional_bundles = true;
+	globals.no_scripts = true;
 
 	if (cmdline_option_status) {
 		return check_update();
 	} else {
+		info("Updates from a 3rd-party repository are forced to run with the --no-scripts flag for security reasons\n\n");
 		return execute_update();
 	}
 }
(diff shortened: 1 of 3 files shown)


karthikprabhuvinod:
whats the security threat here?

src/3rd_party_update.c
```diff
@@ -127,10 +127,12 @@ static enum swupd_code update_repos(UNUSED_PARAM char *unused)
 {
 	/* Update should always ignore optional bundles */
 	globals.skip_optional_bundles = true;
+	globals.no_scripts = true;
 
 	if (cmdline_option_status) {
 		return check_update();
 	} else {
+		info("Updates from a 3rd-party repository are forced to run with the --no-scripts flag for security reasons\n\n");
```
otaviobp: we don't need to print that. We just don't have scripts to run on 3rd party.
No need to print any information. From the user perspective we aren't forcing any flag.

castulo: The thing is that if we don't print that message then this warning may cause confusions:

'''
...
Update was applied
Warning: post-update helper scripts skipped due to --no-scripts argument
Update successful - System updated from version 10 to version 20
'''

Because it mentions that the `--no-script` flag was used.

otaviobp: We shouldn't print the warning about --no-scripts then

castulo: > We shouldn't print the warning about --no-scripts then

not ever? not even when running updates from upstream content? Because from the update function there is no way to distinguish when the update is being run for upstream content, a mix or  3rd-party content. If we remove the warning `Warning: post-update helper scripts skipped due to --no-scripts argument` we will remove it for all scenarios. Is that ok?

karthikprabhuvinod: IMO, it is good to keep the warnings as it is transparent to the user how a 3rd-party update is different from a regular update

PR #1269, opened 2020-01-16
Organizing compare-like functions

### LINKED ISSUE #1167: Organize all compare, and filter functions throughout the code
We have a couple of type definitions that we use often throughout the code: 
```
typedef int (*comparison_fn_t)(const void *a, const void *b);
typedef bool (*filter_fn_t)(const void *a);
```
Functions that use these type definitions should be organized somehow so it is easier to find them so they can be reused.

At the moment we have some duplication on that kind of functions.
For example here:

https://github.com/clearlinux/swupd-client/blob/master/src/lib/sys.c#L215
```
static int lex_sort(const void *a, const void *b)
{
	const char *name1 = (char *)a;
	const char *name2 = (char *)b;
	return strcmp(name1, name2);
}
```

https://github.com/clearlinux/swupd-client/blob/master/src/lib/list.c#L319
```
int list_strcmp(const void *a, const void *b)
{
	return strcmp((const char *)a, (const char *)b);
}
```

castulo:
We used compare-like functions as arguments to other functions that
filter, sort, compare data. These functions are scattered accros the
code and they are sometimes difficult to identify so there is a lot of
code duplication.

This commit organizes all the compare like functions throughout the code
in the same place and with standard naming so it is easier to find them
and reuse them.

Closes #1167

### DIFF
--- a/src/swupd.h
+++ b/src/swupd.h
@@ -15,6 +15,7 @@
 #include "config_loader.h"
 #include "globals.h"
 #include "lib/archives.h"
+#include "lib/comp_functions.h"
 #include "lib/config_file.h"
 #include "lib/formatter_json.h"
 #include "lib/list.h"
@@ -26,6 +27,7 @@
 #include "manifest.h"
 #include "progress.h"
 #include "scripts.h"
+#include "swupd_comp_functions.h"
 #include "swupd_curl.h"
 #include "swupd_exit_codes.h"
 #include "swupd_progress.h"
@@ -175,9 +177,6 @@ extern int get_server_format(int server_version);
 extern bool ignore(struct file *file);
 extern void apply_heuristics(struct file *file);
 
-extern int file_sort_filename(const void *a, const void *b);
-extern int file_sort_filename_reverse(const void *a, const void *b);
-extern int file_sort_hash(const void *a, const void *b);
 extern struct manifest *load_mom(int version, bool mix_exists, int *err);
 extern struct manifest *load_manifest(int version, struct file *file, struct manifest *mom, bool header_only, int *err);
 extern struct manifest *load_manifest_full(int version, bool mix);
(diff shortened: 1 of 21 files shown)


src/swupd.h
```diff
@@ -409,6 +405,27 @@ enum swupd_code check_update();
 
 /****************************************************************/
 
+/* swupd_comp_funtions.c */
+
+int find_manifest(const void *a, const void *b);
```
otaviobp: It makes more sense to group all this functions on a new header file instead of just adding it here.
Grouping all functions is a single header is not a good approach, so we should try to move functions from swupd.h and not to group them here.

src/swupd.h
```diff
@@ -409,6 +405,27 @@ enum swupd_code check_update();
...
+int find_subscription(const void *a, const void *b);
+
+int cmp_file_filename(const void *a, const void *b);
+int cmp_file_filename_ptr(const void *a, const void *b);
+int cmp_file_hash(const void *a, const void *b);
+int cmp_filerecord_filename(const void *a, const void *b);
+int cmp_manifest_component(const void *a, const void *b);
+int cmp_subscription_component(const void *a, const void *b);
+
+int sort_fullfile(const void *a, const void *b);
```
otaviobp: The name of this function is not good. The function doesn't sort anything, it's a compare function.
We could have a better naming standard for cmp functions.
Maybe something like: `cmp_ + type of first parameter + type of second parameter + some other sufix.`
So in this case `cmp_file_file`.

otaviobp: The same applies to all find_, sort_, filter_ functions here

src/swupd.h
castulo: I did consider having all these  in `src/swupd_comp_functions.h`, then just having `swupd.h` include `src/swupd_comp_functions.h`, that way all files that needed a comparison function will work as they were. But the problem I encountered was that the comparison functions needed things that are defined in `swupd.h`, like the file or manifest structs, and I didn't want to cause a circular reference in the header files. So I decided to go with this approach so we didn't have to specifically include `src/swupd_comp_functions.h` in all files that needed a function from there. But if you prefer it that way @otaviobp I can change that back to have their own `src/swupd_comp_functions.h` header file.

otaviobp: There's no struct used on any function definition, so you can create a new header file for them without including swupd.h

castulo:
V2:

- Removed the categories and renamed all the functions following the same name convention
- Returned functions that were too specific (and unlikely to reuse) as static functions in the file where they are used
- Moved all functions in their own header file instead of using `swupd.h`

PR #1271, opened 2020-01-16
3rd party manual

### LINKED ISSUE #1270: Document the swupd 3rd-party related features
The `swupd 3rd-party` features need to be documented in the swupd man pages before the functionality is released.

castulo:
This PR makes a few updates to the `swupd` man page, and adds the `3rd-party` related documentation.

Closes #1270

### DIFF
--- a/docs/swupd.1.rst
+++ b/docs/swupd.1.rst
@@ -42,6 +42,8 @@ the actual content is provided to clients in the form of archive files.
 format the installed version of ``swupd`` understands. For more information
 about how these artifacts are generated see ``mixer``\(1) and ``os-format``\(7).
 
+The ``swupd`` tool can also manage 3rd-party content, this allows users to
+install, remove, and update bundles from 3rd-party repositories.
 
 OPTIONS
 =======
@@ -99,50 +101,50 @@ used to modify the core behavior and resources that swupd uses.
    Specify an alternate swupd state directory. Normally ``swupd`` uses
    ``/var/lib/swupd``.
 
-- ``-C, --certpath``
+-  ``-C, --certpath``
 
    Specify alternate path to swupd certificate store (pem file).
    Default is /usr/share/clear/update-ca/Swupd_Root.pem
 
-- ``-W, --max-parallel-downloads``
+-  ``-W, --max-parallel-downloads``
 
    Set the maximum number of parallel downloads.
 
-- ``-r, --max-retries``
+-  ``-r, --max-retries``
 
    Maximum number of retries for download failures.
 
-- ``-d, --retry-delay``
+-  ``-d, --retry-delay``
 
    Initial delay in seconds between download retries, this will be
    doubled for each retry until the download succeeds or the maximum
    number of retries has been reached.
 
-- ``-n, --nosigcheck``
+-  ``-n, --nosigcheck``
 
    Do not attempt to enforce certificate or signature checking.
 
-- ``-I, --ignore-time``
+-  ``-I, --ignore-time``
 
    Ignore system/certificate time when validating signature.
 
-- ``-t, --time``
+-  ``-t, --time``
 
    Show verbose time output for swupd operations.
 
-- ``-N, --no-scripts``
+-  ``-N, --no-scripts``
 
    Do not run the post-update scripts and boot update tool.
 
-- ``-b, --no-boot-update``
+-  ``-b, --no-boot-update``
 
    Do not update the boot files using clr-boot-manager.
 
-- ``-j, --json-output``
+-  ``-j, --json-output``
 
    Prints the swupd output as a machine readable JSON stream.
 
-- ``--allow-insecure-http``
+-  ``--allow-insecure-http``
 
    For security reasons, swupd only allows system updates using
    secure https connections by default. This option forces swupd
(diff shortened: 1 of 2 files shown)


bryteise:
LGTM

karthikprabhuvinod:
@castulo The content changes its formatting right after first occurrence of

 -C, --certpath

I am not sure why!! checking

castulo:
> @castulo The content changes its formatting right after first occurrence of
> 
> -C, --certpath
> 
> I am not sure why!! checking

What do you mean? Did you use `rst2man` to convert the file to man format and opened it using the `man` command and you saw problems with the formatting? I did that and it looks good on my side.

Plus there is only one occurrence of `--certpath`, so I'm not sure what you mean. 😅

karthikprabhuvinod:
> > @castulo The content changes its formatting right after first occurrence of
> > -C, --certpath
> > I am not sure why!! checking
> 
> What do you mean? Did you use `rst2man` to convert the file to man format and opened it using the `man` command and you saw problems with the formatting? I did that and it looks good on my side.
> 
> Plus there is only one occurrence of `--certpath`, so I'm not sure what you mean. 😅

if you open the file in github and check the first occurence you will see the explanation of --certpath is greyed out

castulo:
> if you open the file in github and check the first occurence you will see the explanation of --certpath is greyed out

I still don't know what you mean, I don't see anything wrong/weird, and as I said `--certpath` exists only once in the document, so I don't know why you see more occurrences.

phmccarty:
I see the rendering oddity as well:

- Click `Files changed` tab
- Click the `...` dropdown at the top-right of the diff view of `docs/swupd.1.rst`, and select `View file`
- Scroll down the where the `-S, --statedir={path}` options are documented. You will see that the documentation for that option is rendered normally. But compare with the rendering of the doc for `-C, --certpath` that follows. There is a grey vertical bar that appears to indicate a blockquote, and it's indented further. The remainder of the man page renders in a similar, unexpected way, where lines are blockquoted.

castulo:
> I see the rendering oddity as well:
> 
> * Click `Files changed` tab
> * Click the `...` dropdown at the top-right of the diff view of `docs/swupd.1.rst`, and select `View file`
> * Scroll down the where the `-S, --statedir={path}` options are documented. You will see that the documentation for that option is rendered normally. But compare with the rendering of the doc for `-C, --certpath` that follows. There is a grey vertical bar that appears to indicate a blockquote, and it's indented further. The remainder of the man page renders in a similar, unexpected way, where lines are blockquoted.

Ahh I see it now, thanks for the detailed explanation  @phmccarty.

It is strange in deed, as I mentioned, if I convert that file to a man page locally and open it with man it renders correctly. I'll investigate further. 👍

castulo:
I see now that the document looks like that in master too. But I think I know what is causing it, I'll submit an update to fix it.

karthikprabhuvinod:
@castulo: Its corrected now but still appears in Subcommands section now of the same document

castulo:
Ok, I see now what is going on @karthikprabhuvinod @phmccarty. A paragraph with different indentation in reStructuredText (.rst) is considered a blockquote as @phmccarty mentioned. 

_"A text block that is indented relative to the preceding text, without preceding markup indicating it to be a literal block or other content, is a block quote. "_

So when GitHub is rendering the file, it is rendering it as an rst file. However we really don't care about the correct rendering of the file in rst mode since we are further converting that file from .rst to man format using `rst2man`. If I remove the indentation in the document so it renders properly in rst form, we loose the indentation we want in the man page, which is the one we care about.

This is an excerpt of the man page when removing the indentation from the description of the `info` command:

```
SUBCOMMANDS
       info

       Shows the current OS version and the URLs used for updates.

       autoupdate [--enable|--disable]
          Enables or disables automatic updates, or reports current status. Enabling updates does not cause an immediate update -
          use swupd update to force one if desired.
```

You can see how the description from `info` has different indentation than that of `autoupdate`, so we cannot really remove the indentation so it renders fine on .rst unless we are ok with loosing the indentation in the man page as shown above.

In that example doesn't look that bad since it was one line, but if removing the indentation from more lines it looks like this:

```
SUBCOMMANDS
       info

       Shows the current OS version and the URLs used for updates.

       autoupdate [--enable|--disable]

       Enables  or  disables  automatic updates, or reports current status. Enabling updates does not cause an immediate update -
       use swupd update to force one if desired.

       check-update

       Checks whether an update is available and prints out the information if so. Does not download update content.

       update

       Performs a system software update.

       The program will contact the version server at the version url, and check to see if a system software update is available.
       If  an  update  is  available, the update content will be downloaded from the content url and stored in the /var/lib/swupd
       state path. Once all content is downloaded and verified, the update is applied to the system.

       In case any problem arises during a software update, the program attempts to correct the issue, possibly by  performing  a
       swupd repair operation, which corrects broken or missing files and other issues.

       After the update is applied, the system performs an array of post-update actions. These actions are triggered through sys‐
       temd(1) and reside in the update-triggers.target(4) system target.

       • -V, --version

         Update to a specific version, also accepts 'latest' (default).
```

IMO the swupd commands get lost very easily. I like it better when they are indented.
Opinions @otaviobp @karthikprabhuvinod @phmccarty ?

phmccarty:
@castulo Any reason to not use a bulleted list for the subcommands as well, similar to the option list?

castulo:
> @castulo Any reason to not use a bulleted list for the subcommands as well, similar to the option list?

Not really, at least not from my side, whoever created the manual initially decided to go with this format, but it could be changed I guess.

otaviobp:
> @castulo Any reason to not use a bulleted list for the subcommands as well, similar to the option list?

Actually for me it's very weird to use a bullet list for options. I never saw a man page like that. So I would remove that. Options list from globals is also rendered differently from the list in the sub-commands. But as this wasn't introduced by this patch, I won't consider this a problem.

otaviobp:
merged.

otaviobp:
I created an issue to cover the reported problems: #1285

PR #1284, opened 2020-01-22
Fix return value overwriting during bundle-list

castulo:

### DIFF
--- a/src/bundle_list.c
+++ b/src/bundle_list.c
@@ -208,6 +208,7 @@ static enum swupd_code list_local_bundles(int version)
 /* Return recursive list of included bundles */
 static enum swupd_code show_included_bundles(char *bundle_name, int version)
 {
+	enum swupd_code ret_code = SWUPD_OK;
 	int ret = 0;
 	struct list *bundles = NULL;
 	struct list *subs = NULL;
@@ -221,7 +222,7 @@ static enum swupd_code show_included_bundles(char *bundle_name, int version)
 	mom = load_mom(version, false, NULL);
 	if (!mom) {
 		error("Cannot load official manifest MoM for version %i\n", version);
-		ret = SWUPD_COULDNT_LOAD_MOM;
+		ret_code = SWUPD_COULDNT_LOAD_MOM;
 		goto out;
 	}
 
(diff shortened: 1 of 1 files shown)


src/bundle_list.c
```diff
@@ -257,7 +257,11 @@ static enum swupd_code show_included_bundles(char *bundle_name, int version)
 
 	if (verbose) {
 		ret = subscription_get_tree(bundles, &subs, mom, true, 0);
+		if (ret == add_sub_NEW) {
```
otaviobp: This looks incorrect. What about other values of subscription_get_tree() return?

castulo: the possible values are these:
- `0` : nothing valid was found, so nothing will be listed. 0 is already the same as `SWUPD_OK`.
- `add_sub_BADNAME` : an invalid bundle was found (a bundle that is not in the MoM), this condition will never happen during `bundle-list --deps` because at this point in the code the bundle was already validated.
- `add_sub_ERR` : the manifest of a bundle could not be downloaded. This is an important error we want to avoid since it will mean the list will be incomplete.
- `add_sub_NEW` : we have gone through at least one valid bundle.

So the only two valid conditions for us at this point is either `0` or `add_sub_NEW`, we should fail if we find `add_sub_ERR`, and `add_sub_BADNAME ` will never happen.

otaviobp: show_included_bundles() is suppose to return a ```enum swupd_code```. In the case you described you are possibly returning the ```add_sub_``` values you don't want.
This can lead to swupd returning the error codes with meanings that aren't listed on the manual.

One suggestion to avoid this kind of problems is never using the same variable for return codes with different meanings, even if they have the same type. The ideal here would be to have one exit_code with type ```enum swupd_code``` and another with type int for the int return codes.

castulo: ahh you are right, we don't want to propagate those error codes up, I overlooked that, I'll correct that. Thanks for pointing that out.

castulo:
V2:
- fixed the error codes returned by the function.

src/bundle_list.c
```diff
@@ -249,15 +250,21 @@ static enum swupd_code show_included_bundles(char *bundle_name, int version)
 	const int MINIMAL_SUBSCRIPTIONS = 2;
 	if (strcmp(bundle_name, "os-core") == 0 || list_len(subs) <= MINIMAL_SUBSCRIPTIONS) {
 		info("\nNo included bundles\n");
-		ret = SWUPD_OK;
+		ret_code = SWUPD_OK;
```
otaviobp: ret_code is already SWUPD_OK here.

otaviobp: same for other lines setting it to SWUPD_OK after this line

castulo: no, at this point the ret_code will be `add_sub_NEW` 🤔

otaviobp: what line sets ret_code to add_sub_NEW?

castulo: line 231: `ret = add_subscriptions(bundles, &subs, mom, true, 0);`

otaviobp: that's ret and not ret_code

castulo: 🤣 that's true. Man, I've got to pay more attention. I'll make the change.

castulo:
V3:
- removed unnecessary ret_code assignments

PR #1286, opened 2020-01-23
docs: Explicitly adding a list of files in the doc check exclude list

otaviobp:
We have a list of header files that weren't documented yet. Add them explicitly
to the check to prevent us to lower the quality of the documentation of headers
that are well documented.

Also fixing some documentation that weren't included because of incorrect style.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -266,7 +266,7 @@ functional-check:
 
 docs-coverage:
 	doxygen docs/Doxyfile || die; \
-	python -m coverxygen --xml-dir xml/ --src-dir ./ --output doc-coverage.info --exclude ".*/src/.*" --include ".*/src/lib/.*" --format json || die; \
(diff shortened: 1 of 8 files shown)


Makefile.am
```diff
@@ -266,7 +266,7 @@ functional-check:
 
 docs-coverage:
 	doxygen docs/Doxyfile || die; \
-	python -m coverxygen --xml-dir xml/ --src-dir ./ --output doc-coverage.info --exclude ".*/src/.*" --include ".*/src/lib/.*" --format json || die; \
+	python -m coverxygen --xml-dir xml/ --src-dir ./ --output doc-coverage.info --exclude ".*/src/globals.h" --exclude ".*/src/swupd_build_opts.h" --exclude ".*/src/swupd_build_variant.h" --exclude ".*/src/swupd_exit_codes.h" --exclude ".*/src/swupd.h" --exclude ".*/src/swupd_internal.h" --exclude ".*/src/manifest.h" --exclude ".*/src/swupd_curl_internal.h" --exclude ".*/src/swupd_curl.h" --format json || die; \
```
castulo: I'm curious, how did you come up with the list of files to exclude?

otaviobp: I excluded all files that aren't properly documented yet.

castulo: ahh ok, so this patch is to fix some of the files but not all yet. Got it.

PR #1287, opened 2020-01-23
Implementation: create links to binaries exported by a 3rd-party repository on bundle-add

### LINKED ISSUE #1277: 3rd_party: Create links to binaries exported by a 3rd party repository on bundle-add
On a 3rd party repository some files can be exported according to one whitelist. Exported files are going to be marked on Manifest setting the last flag to x like the example:

```F..x    4163f4a329da5276291ffeee367268dfbefda37e927ff7d51dc230359031f652        32000   /usr/bin/binary```

Swupd should create a link to all binaries set as export on /opt/3rd-party/bin/ on a bundle-add.

### PR DESCRIPTION
**Pull request by castulo**
This PR implements the workflow described in issue #1277.

When installing a bundle from a 3rd-party repository, if the bundle has files that are marked with 'x' in the manifest (as exported), and the files have one of the qualifying paths for binaries ('/bin', '/usr/bin', '/usr/local/bin'), these files will be made available to the user via a script file that will reside in '/opt/3rd-party/bin'.

These scripts will export the appropriate PATH and libraries and will call the correct binary from the 3rd-party content.

Closes #1277

### CODE REVIEW

File: src/3rd_party_bundle_add.c [hunk 1]
```diff
@@ -94,6 +96,136 @@ static bool parse_options(int argc, char **argv)
...
+	bin_directory = sys_path_join(tmp2, "/bin");
+	free_string(&tmp1);
+	free_string(&tmp2);
+
+	script = sys_path_join(bin_directory, sys_basename(filename));
+	binary = sys_path_join(globals.path_prefix, filename);
+
+	/* if a file already exists, report the problem to the user */
+	if (sys_file_exists(script)) {
+		warn("There is already a binary called %s in %s, it will be skipped\n", sys_basename(filename), bin_directory);
```
bryteise: Having a conflict notification at this point seems a little late. Couldn't we catch this earlier in the bundle-add instead and not actually add the bundle in that case?

castulo: sure it can be done @bryteise . So the desired behavior would be that if there is a conflict inform the user and skip adding the bundle, correct?

bryteise: Yes, I'd expect we'd not add the bundle in that case since it would only be partially installed without the symlink.

castulo: gotcha, I'll make the change, thanks

castulo:
V2:
- Validating that the binaries won't have conflicts before installing the bundle

castulo:
V3:
- added a test to validate that bundles with conflicting binaries won't get installed

castulo:
V4:
- Rebased against master.

src/3rd_party_bundle_add.c
```diff
@@ -94,6 +98,174 @@ static bool parse_options(int argc, char **argv)
...
+	struct list *iter = NULL;
+	struct file *file = NULL;
+	char *filename = NULL;
+	int count = 0;
+	int number_of_files = list_len(installed_files);
+
+	info("%s", msg);
+
+	progress_next_step(step, PROGRESS_BAR);
+	for (iter = list_head(installed_files); iter; iter = iter->next) {
```
otaviobp: why do you need a list_head() here?

src/3rd_party_bundle_add.c
```diff
@@ -94,6 +98,174 @@ static bool parse_options(int argc, char **argv)
...
+		file = iter->data;
+		filename = file->filename;
+		if (is_binary(filename) && file->is_exported) {
```
otaviobp: Check if is_exported() before checking if is_binary() because it's way faster.

src/3rd_party_bundle_add.c
```diff
@@ -94,6 +98,174 @@ static bool parse_options(int argc, char **argv)
...
+	bin_directory = sys_path_join(tmp2, "/bin");
+	free_string(&tmp1);
+	free_string(&tmp2);
+
+	script = sys_path_join(bin_directory, sys_basename(filename));
+
+	/* if a file already exists, report the problem to the user */
+	if (sys_file_exists(script)) {
+		error("There is already a binary called %s in %s\n", sys_basename(filename), bin_directory);
+		ret_code = SWUPD_INVALID_BINARY;
```
otaviobp: Error code here is different from what's documented.
I think SWUPD_COULDNT_CREATE_FILE  is more appropriate

src/3rd_party_bundle_add.c
```diff
@@ -94,6 +98,174 @@ static bool parse_options(int argc, char **argv)
...
+
+	fprintf(fp, "#!/bin/bash\n");
+	fprintf(fp, "\n");
+	fprintf(fp, "export PATH=%s%s%s\n", third_party_bin_path, env_path ? ":" : "", env_path ? env_path : "");
+	fprintf(fp, "export LD_LIBRARY_PATH=%s%s%s\n", third_party_ld_path, env_ld_path ? ":" : "", env_ld_path ? env_ld_path : "");
+	fprintf(fp, "\n");
+	fprintf(fp, "%s \"$@\"\n", binary);
+
+	/* set the file mode to 0755 */
+	if (chmod(script, mode) < 0) {
```
otaviobp: Instead of setting the mode later you can use `open()` to create the file with correct permissions and have one less location to check for errors.

otaviobp: `fdopen()` can help too.

src/3rd_party_bundle_add.c
castulo: I don't really since this function should run immediately after `consolidate_files()` which is going to already return list_head(). I like to do that sometimes just to make sure the head of the list is what we've got, also I think the code is easier to follow when you explicitly see you are working with the head of the list. The list_head() function should return right away if you already are at the head of the list, so there is really no penalty from doing it when you are already at the head.

src/3rd_party_bundle_add.c
castulo: good point, actually I had it like that initially but switched the positions so I could comment out the `file->is_exported` condition easier when testing locally. I will change that.

src/3rd_party_bundle_add.c
castulo: ok, I will change that.

src/3rd_party_bundle_add.c
castulo: I didn't know I could use `fopen` setting the permissions at once, I will change that.

src/3rd_party_bundle_add.c
otaviobp: > I don't really since this function should run immediately after consolidate_files() which is going to already return list_head().

Using list_head() in locations like can mask problems instead of solving it. All functions should return the head of the list.

> I like to do that sometimes just to make sure the head of the list is what we've got, also I think the code is easier to follow when you explicitly see you are working with the head of the list.

That's an unusual implementation of a list. Usually lists should point to the element you consider the begging of the list, not any element of the list. One great advantage of this implementation is that you can call functions that will iterate partially at the list. I don't see code using list_head() or something similar in python, go, glib or any other library I have worked before.

> The list_head() function should return right away if you already are at the head of the list, so there is really no penalty from doing it when you are already at the head.

It's not about performance it's about not masking problems in other areas of the code.

src/3rd_party_bundle_add.c
otaviobp: I think you cant. You need to use `open()`. Then `fdopen()` to get the FILE*

src/3rd_party_bundle_add.c
castulo: I see your point, I'll remove the list_head from there then, thanks for the explanation

castulo:
@otaviobp I addressed all your comments in this new update

castulo:

PR #1290, opened 2020-01-27
Implementation: remove links to binaries exported by a 3rd-party repository on bundle-remove

### LINKED ISSUE #1278: 3rd_party: Remove links to binaries exported by a 3rd party repository on bundle-remove
Swupd should create links to binaries as stated on #1277
On bundle-remove these links should be removed.

castulo:
When removing a 3rd-party bundle, all scripts that export the binaries of the bundle should be removed from the target system unless they are still required by another installed 3rd-party bundle.

Closes #1278 

Note: This PR was created o top of PR #1287, that PR needs to be reviewed first.


src/3rd_party_bundle_add.c
```diff
@@ -98,21 +96,12 @@ static bool parse_options(int argc, char **argv)
...
-	    strncmp(filename, "/usr/bin", strlen("/usr/bin")) == 0 ||
-	    strncmp(filename, "/usr/local/bin", strlen("/usr/local/bin")) == 0) {
-		return true;
-	}
-
-	return false;
-}
-
-static enum swupd_code create_wrapper_script(const char *filename)
+static enum swupd_code create_wrapper_script(void *data)
```
otaviobp: I couldn't understand the change you made here that needed to change the type from filename to void*. Can you point me what triggered that?

castulo: I had two type of functions to process data:
one that received a `const char *`:
`typedef enum swupd_code (*process_data_fn_t)(const char *name);`
and one that received a `struct file *`:
`typedef enum swupd_code (*process_data_fn_t)(struct file *file);`
So instead of having those two very similar type definitions I decided to make the parameter a void pointer so we could reuse the same type definition. And since the places in the code where those are used receive a function for extra processing, I thought it was more flexible if they would receive a void pointer as data so we could cast it to whatever we needed and have more flexibility.

otaviobp: The problem here is that `process_bundle_binaries()` is always calling the callback with a char*. So if you use a callback that receives a struct file it will crash. Creating a type definition as you had before enables the compiler to warn about those problems.
It's a good idea to use a void* for that if the data type is unknown by the function that will call the callback. That's the case of list functions, for example. The list module doesn't know the type of its elements. Or any function that receives both the callback and the data, like the progress_cb related functions. Otherwise it's better to set the type and let the compiler handle that.
Another case where we need void* data is when we are using the callback on functions that we don't have control. In this case we really need to follow the specifications.

Unless there's any possible code reusable or a generic function that works for both cases, I would rowback this to how it was before

castulo:

PR #1293, opened 2020-01-28
Rebalancing tests for github actions

castulo:
This PR brings back down the test execution time to about ~15 minutes.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -30,7 +30,7 @@ jobs:
       if: failure()
       run: cat test-suite.log
 
-  test_update1:
+  test_3rd_party1:
     runs-on: ubuntu-latest
     timeout-minutes: 60
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ccpp.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       if: failure()
       run: cat test-suite.log
 
-  test_update1:
+  test_3rd_party1:
```
otaviobp: It's too much manual work to keep balancing tests often. Can't we do better and have something auto balanced?

otaviobp: We could for example to list all tests divide them equally into 10 jobs, instead of having explicit jobs for different cases.
It's easier to increase the number of jobs and we wouldn't need to worry about re-balancing anymore. One downside is that it would be harder to know what is broken without looking at the specific job. But I think it's a price worth paying

castulo: We can certainly do that, but there are two things to consider with that approach:
- Lest's say we have 100 test files, if divided into 10 jobs that means each job is going to have 10 test files, but 1 test file may have only 1 test while other test file may have 5.
- Depending on the test being run 1 test may take 2 minutes to run while other test may take 2 seconds. For example, we only have 19 `3rd-party` test files with 55 tests among all of them, and these are the slowest tests of all. If we used 10 jobs to run the tests, most likely 1 job is going to run all 19 of the 3rd-party tests, which will cause the execution to be as slow as before.

The upside of this approach is that we don't have to be manually balancing the tests across the jobs, adding more jobs is pretty easy.

The downside is that because of the reasons given above, the control we have over the balancing is limited, some jobs may still take considerably more than others.

I tested this approach and divided the tests among 10 jobs, the job that took the most time to run ran during ~20 minutes, the shortest one took ~5 minutes, so you can see the big disparity. 

One thing we could do to mitigate this would be to sort the list of test files randomly first so each job has tests from all commands, that way they may be more balanced, but it would be even more difficult to pinpoint what area is broken when tests are failing. Testing out this approach I got: the fastest job took ~6 minutes, and the slowest took ~14 minutes. So we got basically about the same result as when we hand picked the jobs, but it will be difficult to predict the times since the list would be random every time.

Test -> https://github.com/castulo/swupd-client/actions/runs/32716095

So bottom line is we are sacrificing time and clarity for not having to manually re-balance the tests once in a while.

PR #1294, opened 2020-01-28
verify: Quick and extra files are not mutually exclusive

otaviobp:
According to the man page, "Omit checking hash values. Instead only looks
for missing files and directories and/or symlinks.".

When --quick and --picky or --extra-files-only is used that's exactly what
swupd does, so there's no reason to block that.

For --picky there's a real use case that is repair your system without hash
checks. For --extra-files-only in practice the --quick will be ignored, but
it's not inconsistent, so there's no reason to block it.

### DIFF
--- a/src/3rd_party_diagnose.c
+++ b/src/3rd_party_diagnose.c
@@ -163,16 +163,6 @@ static bool parse_options(int argc, char **argv)
 			return false;
 		}
 	}
-	if (cmdline_option_quick) {
-		if (cmdline_option_picky) {
-			error("--quick and --picky options are mutually exclusive\n");
-			return false;
-		}
-		if (cmdline_option_extra_files_only) {
-			error("--quick and --extra-files-only options are mutually exclusive\n");
-			return false;
-		}
-	}
 	if (cmdline_option_extra_files_only) {
 		if (cmdline_option_picky) {
 			error("--extra-files-only and --picky options are mutually exclusive\n");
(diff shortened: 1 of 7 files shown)


castulo:
- When `--quick` and `--picky` are used, swupd will only add missing files and exit, so using `--picky` will have no effect, it will just get ignored.

- When `--quick` and `--extra-files-only` are used, swupd will only check for extra files and will ignore adding missing files, so `--quick` gets ignored.

I think that if we remove the flag restrictions is just going to confuse users, specially since the flags are really not that clear. `--extra-files-only` is very self explanatory, but `--picky` and `--quick` are not. `--quick` should really be called something like `--missing-files-only` or something like that so it would make sense. My point is that we have flags that isolate the execution of only one of the 4 tasks verify does (like --picky or --extra-files-only), while we have flags that do not isolate the execution of the task but run it in addition to the other tasks, IMO this behavior is very confusing as a user.

otaviobp:
I agree that `--picky` is confusing, but --quick is self explanatory and documentation is very clear.
And I really miss not having `swupd repair --picky --quick`. Right now there's no way to run that and the only reason is because we artificially block that.

castulo:
> I agree that `--picky` is confusing, but --quick is self explanatory and documentation is very clear.
> And I really miss not having `swupd repair --picky --quick`. Right now there's no way to run that and the only reason is because we artificially block that.

But I'm still not getting why you are missing that, if you just run `--quick` now you will get the same outcome as if we didn't block `--picky` + `--quick` and you would run that, since the code enabled by`--picky` gets skipped because of `--quick`.

otaviobp:
If the code is skipped, it shouldn't. So I need to stop using this in this PR

castulo:
> So I need to stop using this in this PR

What do you mean by that?

PR #1295, opened 2020-01-29
Remove all exported binaries when removing a 3rd-party repo

### LINKED ISSUE #1291: 3rd_party: Remove links to all binaries exported by a 3rd party repository on `swupd 3rd-party remove <repo>`
When removing a 3rd-party repo, all the content installed from that repo is deleted along with the repo. The scripts that export the binaries of all the bundles of that repo should also be deleted when deleting the repo.

castulo:
When a 3rd-party repository is removed, all the content installed from it must be removed from the system. But also those exported binaries from the repo being deleted should also be deleted.

Closes #1291 

Note: This PR is built on top of #1290 and has to be reviewed and merged first.


castulo:

PR #1303, opened 2020-02-05
Export binaries of 3rd-party bundles after installing them

### LINKED ISSUE #1277: 3rd_party: Create links to binaries exported by a 3rd party repository on bundle-add

castulo:
When a 3rd-party bundle is installed, in order for the user to be able to use the binaries the bundle provides, it needs to have them exported.

This PR exports those binaries after they have been successfully installed.

Closes #1277


src/3rd_party_bundle_add.c
```diff
@@ -94,18 +96,134 @@ static bool parse_options(int argc, char **argv)
...
+	fp = fdopen(fd, "w");
+	if (!fp) {
+		error("The file %s failed to be created\n", script);
+		ret_code = SWUPD_COULDNT_CREATE_FILE;
+		goto close_and_exit;
+	}
+
+	/* get the current values of the environment */
+	env_path = getenv("PATH");
+	env_ld_path = getenv("LD_LIBRARY_PATH");
```
bryteise: Didn't we decide to not use the values of the environment at script creation time but instead using the symbols directly in the script so they'll be evaluated at runtime?

src/swupd_comp_functions.c
```diff
@@ -136,6 +136,8 @@ int cmp_file_filename_is_deleted(const void *a, const void *b)
 	return 0;
 }
 
+#ifdef THIRDPARTY
```
otaviobp: This is in the wrong commit.

src/3rd_party_bundle_add.c
castulo: yes we did, I'm just using `env_path` and `env_ld_path` to see if the variables are already defined so I can add the ":" to the PATH. There is no option on appending a ":" to the end of the PATH if there are no paths to append at the end.

src/3rd_party_bundle_add.c
```diff
@@ -94,18 +96,134 @@ static bool parse_options(int argc, char **argv)
...
+	if (sys_file_exists(script)) {
+		/* the binary already exists, this condition should never happen
+		 * since we checked before installing */
+		ret_code = SWUPD_UNEXPECTED_CONDITION;
+		error("There is already a binary called %s in %s, it will be skipped\n", sys_basename(filename), bin_directory);
+		goto close_and_exit;
+	}
+
+	/* if the SWUPD_3RD_PARTY_BIN_DIR does not exist, attempt to create it */
+	if (!sys_file_exists(bin_directory)) {
```
otaviobp: to avoid a TOCTOU just call the mkdir_p() and check if the return is EEXIST

src/3rd_party_bundle_add.c
```diff
@@ -94,18 +96,134 @@ static bool parse_options(int argc, char **argv)
...
+	/* get the path for the 3rd-party content */
+	third_party_bin_path = str_or_die("%sbin:%susr/bin:%susr/local/bin", globals.path_prefix, globals.path_prefix, globals.path_prefix);
+	third_party_ld_path = str_or_die("%susr/lib64:%susr/local/lib64", globals.path_prefix, globals.path_prefix);
+
+	fprintf(fp, "#!/bin/bash\n");
+	fprintf(fp, "\n");
+	fprintf(fp, "export PATH=%s%s%s\n", third_party_bin_path, env_path ? ":" : "", env_path ? "$PATH" : "");
+	fprintf(fp, "export LD_LIBRARY_PATH=%s%s%s\n", third_party_ld_path, env_ld_path ? ":" : "", env_ld_path ? "$LD_LIBRARY_PATH" : "");
+	fprintf(fp, "\n");
+	fprintf(fp, "%s \"$@\"\n", binary);
```
otaviobp: This is a bit hard to read. I would change that to:
```
#define WRAPPER_TEMPLATE "#!/bin/bash\n" \
                         "export PATH=%s..."

fprintf(fp, TEMPLATE, parameters...);

File: src/3rd_party_bundle_add.c (hunk 1 above)
Comment (otaviobp) [reply src/3rd_party_bundle_add.c]: It's not a problem to add PATH even if it's not defined. And we should add that because it could be defined later.

Comment (bryteise) [reply src/3rd_party_bundle_add.c]: Isn't that something you'd only know for sure at runtime? Doing the check at script creation time you may have an empty path (well doubtful since I think some swupd commands like tar would fail anyway) causing it to not add path and then at runtime path would be defined but not be in the script.

File: src/3rd_party_repos.c [hunk 5]
```diff
@@ -526,4 +526,52 @@ void third_party_repo_header(const char *repo_name)
 	free_string(&header);
 }
 
+static bool is_binary(const char *filename)
+{
+	if (strncmp(filename, "/bin", strlen("/bin")) == 0 ||
```
otaviobp: Keep an array of PATHs and check each one in the array instead of hardcoding that.

src/3rd_party_repos.c
```diff
@@ -526,4 +526,52 @@ void third_party_repo_header(const char *repo_name)
...
+	    strncmp(filename, "/usr/local/bin", strlen("/usr/local/bin")) == 0) {
+		return true;
+	}
+
+	return false;
+}
+
+bool third_party_file_is_binary(struct file *file)
+{
+	if (is_binary(file->filename) && file->is_exported) {
```
otaviobp: Change the order of the arguments

src/3rd_party_add.c
```diff
@@ -174,7 +174,7 @@ enum swupd_code third_party_add_main(int argc, char **argv)
 	globals.no_scripts = true;
 	struct list *bundle_to_install = NULL;
 	bundle_to_install = list_append_data(bundle_to_install, "os-core");
-	ret_code = bundle_add(bundle_to_install, repo_version);
+	ret_code = bundle_add(bundle_to_install, repo_version, NULL, NULL);
```
otaviobp: Multiple NULL or boolean parameters are hard to read in a function. In cases like that it's preferable to have 2 functions with different number of arguments (sad that C doesn't support that with the same name).
So I would create
```

 enum swupd_code bundle_add_full(struct list *bundles_list, int version, extra_proc_fn_t pre_add_fn, extra_proc_fn_t post_add_fn); //or bundle_add_extra
 enum swupd_code bundle_add(struct list *bundles_list, int version)
{
    return bundle_add_full(bundles_list, version, NULL, NULL);
}
```

src/3rd_party_bundle_add.c
castulo: that is a very good point.... I'll just add it regardless of if the PATH is empty or not. Thanks for catching that!

src/swupd_comp_functions.c
castulo: ups, that's true, I'll move it to the right commit

src/3rd_party_bundle_add.c
castulo: ok will do that

src/3rd_party_bundle_add.c
castulo: done

src/3rd_party_repos.c
castulo: done

src/3rd_party_repos.c
castulo: done

PR #1304, opened 2020-02-05
Remove exported 3rd-party binaries after bundle-remove

### LINKED ISSUE #1278: 3rd_party: Remove links to binaries exported by a 3rd party repository on bundle-remove

castulo:
When removing a 3rd-party bundle, all scripts that export the binaries of the bundle should be removed from the target system unless they are still required by another installed 3rd-party bundle.

This PR is built on top of #1303 and it has to be reviewed and merged first.

Closes #1278

### DIFF
--- a/src/3rd_party_repos.c
+++ b/src/3rd_party_repos.c
@@ -583,4 +583,28 @@ enum swupd_code third_party_process_binaries(struct list *files, const char *msg
 	return ret_code;
 }
 
+enum swupd_code third_party_remove_binary(char *filename)
+{
+	enum swupd_code ret_code = SWUPD_OK;
+	int ret;
+	char *script = NULL;
+	char *binary = NULL;
+
+	script = third_party_get_binary_path(sys_basename(filename));
+	binary = sys_path_join(globals.path_prefix, filename);
+
+	if (!sys_file_exists(binary)) {
+		ret = sys_rm(script);
+		if (ret != 0 && ret != -ENOENT) {
+			error("File %s could not be removed\n\n", script);
+			ret_code = SWUPD_COULDNT_REMOVE_FILE;
+		}
+	}
+
+	free_string(&script);
+	free_string(&binary);
+
+	return ret_code;
+}
+
 #endif
--- a/src/bundle_remove.c
+++ b/src/bundle_remove.c
@@ -290,10 +290,10 @@ static void print_remove_summary(unsigned int requested, unsigned int bad, unsig
 /*  The function removes one or more bundles
  *  passed in the bundles list.
  */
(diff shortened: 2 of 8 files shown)


src/3rd_party_bundle_add.c
```diff
@@ -94,18 +96,134 @@ static bool parse_options(int argc, char **argv)
...
+	char *third_party_bin_path = NULL;
+	char *third_party_ld_path = NULL;
+	mode_t mode = 0755;
+
+	bin_directory = third_party_get_bin_dir();
+	script = third_party_get_binary_path(sys_basename(filename));
+	binary = sys_path_join(globals.path_prefix, filename);
+
+	if (!sys_filelink_is_executable(binary)) {
+		goto close_and_exit;
```
bryteise: It might be worthwhile to have some way to notify the user that the mix bundle they installed isn't configured properly so they can know something about the quality of the mix they are using.

src/bundle_remove.c
```diff
@@ -490,7 +494,7 @@ enum swupd_code bundle_remove_main(int argc, char **argv)
 	 */
 	progress_init_steps("bundle-remove", steps_in_bundle_remove);
 
-	ret = execute_remove_bundles(bundles_list);
+	ret = execute_remove_bundles(bundles_list, NULL);
```
otaviobp: same comment from last PR here

src/3rd_party_bundle_add.c
castulo: Hmm this comment applied to part of the code was really part of PR #1303, which has already merged. I'll address this comment in a separate PR.

src/3rd_party_repos.c
```diff
@@ -574,4 +574,28 @@ enum swupd_code third_party_process_binaries(struct list *files, const char *msg
...
+{
+	enum swupd_code ret_code = SWUPD_OK;
+	int ret;
+	char *script = NULL;
+	char *binary = NULL;
+
+	script = third_party_get_binary_path(sys_basename(filename));
+	binary = sys_path_join(globals.path_prefix, filename);
+
+	if (sys_file_exists(script) && !sys_file_exists(binary)) {
```
otaviobp: To avoid a TOCTOU error don't check if the script exists. If the binary doesn't exist just remove it.

src/bundle_remove.c
castulo: done

src/3rd_party_repos.c
castulo: good idea, done

castulo:
V2:
- Rebased and addressed comments from @otaviobp

PR #1305, opened 2020-02-05
Remove all exported binaries from 3rd-party bundles when removing a repo

### LINKED ISSUE #1291: 3rd_party: Remove links to all binaries exported by a 3rd party repository on `swupd 3rd-party remove <repo>`

castulo:
When a 3rd-party repository is removed, all the content installed from it must be removed from the system. But also those exported binaries from the repo being deleted should also be deleted.

This PR is built on top of #1303 and #1304 and they have to be reviewed first in that order.

Closes #1291

### DIFF
--- a/src/3rd_party_bundle_remove.c
+++ b/src/3rd_party_bundle_remove.c
@@ -93,15 +93,20 @@ static bool parse_options(int argc, char **argv)
 	return true;
 }
 
-static enum swupd_code remove_bundle(char *bundle)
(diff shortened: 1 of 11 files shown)


castulo:
This PR has already been rebased

PR #1307, opened 2020-02-06
Warn user when a 3rd-party binary isn't executable

castulo:
This commit notifies the user that the 3rd-party bundle they installed
isn't configured properly so they can know something about the quality
of the 3rd-party repository they are using.


bryteise:
LGTM

PR #1310, opened 2020-02-10
Update exported binaries during a "swupd 3rd-party update"

### LINKED ISSUE #1279: 3rd_party: Add and Remove links to binaries exported by a 3rd party repository on update
Related to #1277 and #1278 
On update whitelisted binaries can be added and removed. Create and remove links accordingly

castulo:
When a 3rd-party repository is updated, if one of the updated bundles
has updated binaries, the scripts that export those binaries should also
be updated. This commit implements that update.

Closes #1279

### DIFF
--- a/src/3rd_party_bundle_add.c
+++ b/src/3rd_party_bundle_add.c
@@ -19,8 +19,6 @@
 
 #define _GNU_SOURCE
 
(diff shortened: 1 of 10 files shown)


bryteise:
This looks fine to me.

PR #1312, opened 2020-02-12
3rd-party: Import certificate from 3rd-party repository on repo add

otaviobp:
When trying to add the repository for the first time, download the public
certificate published by the 3rd party repository and with a user confirmation,
use it to install the 3rd-party os-core. After that the official certificate will
be installed for future commands.

### DIFF
--- a/src/3rd_party_add.c
+++ b/src/3rd_party_add.c
@@ -19,9 +19,12 @@
 
 #define _GNU_SOURCE
 
+#include <ctype.h>
 #include <errno.h>
+#include <unistd.h>
 
 #include "3rd_party_repos.h"
+#include "signature.h"
 #include "swupd.h"
 
 #ifdef THIRDPARTY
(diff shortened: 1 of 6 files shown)


src/signature.c
```diff
@@ -492,6 +492,59 @@ static int verify_callback(int ok, X509_STORE_CTX *stor)
...
+	fclose(f);
+}
+
+void signature_print_info(const char *path)
+{
+	X509 *cert;
+	char *subj, *issuer;
+
+	if (!path) {
+		error("Invalid swupd certificate - Empty");
```
castulo: Missing "\n"

src/signature.c
```diff
@@ -492,6 +492,59 @@ static int verify_callback(int ok, X509_STORE_CTX *stor)
...
+		return;
+	}
+
+	cert = get_cert_from_path(path);
+	if (!cert) {
+		error("Invalid swupd certificate - Error parsing the certificate");
```
castulo: same here ☝️

src/signature.h
```diff
@@ -57,6 +57,16 @@ bool signature_verify(const char *file, const char *sig_file, bool print_errors)
  */
 bool signature_verify_data(const unsigned char *data, size_t data_len, const unsigned char *sig_data, size_t sig_data_len, bool print_errors);
 
+/**
+ * Print extra information about the certificate pointed by PATH.
+ *
+ * Currenttly printing Issuer, Subject and the dump of the certificate data.
```
castulo: Typo: "Currenttly"

src/version.c
```diff
@@ -34,36 +34,43 @@
 
 #define MAX_VERSION_STR_SIZE 11
 
-int get_server_format(int server_version)
+int get_int_from_url(const char *url)
```
castulo: since this is a public function it would be better to follow the standard we adopted of starting the function name with something common for all functions of this file ("version").

For example maybe something like: `version_from_url`, I think the "int" part is implied by the return type so no need to reference it.

src/version.c
```diff
@@ -34,36 +34,43 @@
...
+	err = strtoi_err(tmp_data.data, &value);
 	if (err != 0) {
 		return -1;
 	}
 
-	return ret;
+	return value;
+}
+
+int get_server_format(int server_version)
```
castulo: same thing here with the name, although in this case a name that starts with "version" doesn't make sense: `version_get_serveer_format`. This brings the question, does this function belongs in this file? Maybe it should go in the helpers.c file where we have more general functions. It could keep the current name in that file.

test/functional/3rd-party/3rd-party-repo-add-certificate.bats
```diff
@@ -0,0 +1,100 @@
...
+load "../testlib"
+
+export repo1
+export repo2
+
+global_setup() {
+	# Skip this test for local development because we write the certificate on /
+    # This is necessary because the default certificate location is hardcoded on build time and we
+	# need to run swupd without -C parameter to test this feature.
+	if [ -z "${TRAVIS}" ]; then
```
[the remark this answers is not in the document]
otaviobp: I'm ok on changing that. Github actions have one env that is set too, but I think I prefer something more generic. `RUNNING_IN_CI`?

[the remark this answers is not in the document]
castulo: that would work

bryteise:
Overall LGTM

src/3rd_party_add.c
```diff
@@ -89,6 +92,138 @@ static int remove_repo(const char *repo_name)
...
+}
+
+static int import_third_party_certificate()
+{
+	char *url;
+	int tmp_version;
+
+	url = str_or_die("%s/version/format%s/latest", globals.content_url, globals.format_string);
+	tmp_version = get_int_from_url(url);
+	free(url);
```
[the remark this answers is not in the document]
otaviobp: Why free_string() instead of free()?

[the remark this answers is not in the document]
castulo: haha you tell me, I don't really know why we have free_string when it is basically the same as free

src/version.c
otaviobp: In general functions exported in swupd.h are undocumented, not following a good name system and spread in .c that not necessary make sense. We need to try to organize them, as we improve the quality, but I usually don't concern that much with them, unless it's something significant that is being changed.

src/version.c
otaviobp: I don't like the name version_from_url() because that's not what this function does. This function is really reading an int from an url. It doesn't matter if it's a version or not. The ideal would be to reorganize several things in version.c to improve that, but I don't think it's worth to do it right now. I would leave it like this for now.

otaviobp:
V2: Fixed the proposed changes, other than the ones I commented.

src/3rd_party_add.c
```diff
@@ -89,6 +92,138 @@ static int remove_repo(const char *repo_name)
...
+	manifest_free(manifest);
+	unlink(os_core_tar);
+	unlink(os_core);
+	free(os_core);
+	free(os_core_tar);
+	free(url);
+	return ret;
+}
+
+static int import_third_party_certificate()
```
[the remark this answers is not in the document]
castulo: this comment was not addressed

src/signature.c
castulo: This comment was not addressed either

otaviobp:
v3: Sorry, now addressing them

PR #1313, opened 2020-02-12
test: Auto rebalance test CI based on the number of tests

otaviobp:
Count the number of tests and distribute tests in jobs to try to have around
the same number of tests for each job.

### DIFF
--- a/scripts/filter_bats_list.sh
+++ b/scripts/filter_bats_list.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+# Filter a list of jobs
+# Parameter 1: Job number
+# Parameter 2: Number of jobs
+#
+# To test if any test is missing:
+# for i in $(seq $NUM_JOBS); do
+#    ./scripts/filter_bats_list2.sh $i $NUM_JOBS >> list;
+# done
+
+JOB_NUM=$(($1 - 1))
+NUM_JOBS=$2
+
+penalty() {
+    i=$1
+    TESTS=$2
+
+    if [[ $i == *'slow'* || $i == *'3rd-party'* || $i == *'certificate'* || $i == *'repair'* ]]; then
+        TESTS=$((TESTS * 3))
+    fi
+
+    echo $TESTS
+}
+
+TOTAL_TESTS=0
+for i in $(find test/functional/ -name *.bats); do
+    TESTS=$(grep "^@test" $i | wc -l)
+    TOTAL_TESTS=$((TOTAL_TESTS + TESTS))
+done
+
+NUM_TESTS="$((TOTAL_TESTS/(NUM_JOBS - 1)))"
+START="$((JOB_NUM * NUM_TESTS + 1))"
+END="$(((JOB_NUM + 1) * NUM_TESTS))"
+
+COUNT=0
+for i in $(find test/functional/ -name *.bats); do
+    TESTS=$(grep "^@test" $i | wc -l)
+    TESTS=$(penalty $i $TESTS)
+
+    COUNT=$((COUNT + TESTS))
+
+    if [ $COUNT -gt $END ]; then
+        break
+    fi
+
+    if [ $COUNT -ge $START ]; then
+        echo -n "$i "
+    fi
+
+done
--- a/scripts/filter_bats_list_by_file.sh
+++ b/scripts/filter_bats_list_by_file.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+# Filter a list of jobs
+# Parameter 1: Job number
+# Parameter 2: Number of jobs
+#
+# To test if any test is missing:
+# for i in $(seq $NUM_JOBS); do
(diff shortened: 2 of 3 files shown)


scripts/filter_bats_list_by_file.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
castulo: This script is not being used at the moment correct?

otaviobp: no, this file isn't being used.

otaviobp:
I added repair to the list of penalty tests and it got better. Now we are faster than the human-balanced one even with less jobs (10).

scripts/filter_bats_list.sh
```diff
@@ -0,0 +1,50 @@
...
+# done
+
+JOB_NUM=$(($1 - 1))
+NUM_JOBS=$2
+
+penalty() {
+    i=$1
+    TESTS=$2
+
+    if [[ $i == *'slow'* || $i == *'3rd-party'* || $i == *'certificate'* || %i == *'repair'* ]]; then
```
castulo: I think you made a mistake here, you added `%i` instead of `$i`

scripts/filter_bats_list.sh
```diff
@@ -0,0 +1,50 @@
...
+    if [[ $i == *'slow'* || $i == *'3rd-party'* || $i == *'certificate'* || $i == *'repair'* ]]; then
+        TESTS=$((TESTS * 3))
+    fi
+
+    echo $TESTS
+}
+
+TOTAL_TESTS=0
+for i in $(find test/functional/ -name *.bats); do
+    TESTS=$(grep "^@test" $i | wc -l)
```
castulo: looks like you need to account for the penalty here too:
`TESTS=$(penalty $i $TESTS)`

castulo:
I played around with this code in my system, I fixed the issue I reported earlier and made a few tweaks and I ran this in GitHub actions -> https://github.com/castulo/swupd-client/actions/runs/39460437, I still see a 1 minute improvement in execution time vs the manually balanced tests.
- Auto balanced: 21 minutes, 8 seconds (with 10 jobs)
- Manually balanced: 22 minutes, 9 seconds (with 13 jobs)

castulo:
Found another problem (the last job was not processing anything) and fixed it locally, and here I was able to bring this down to 18 minutes and 37 seconds -> https://github.com/castulo/swupd-client/actions/runs/39502659. That makes a 3.5 minutes improvement.

castulo:
As I mentioned before, I like the "weighting system" you are implementing for balancing the tests across the jobs, but I think we can do better when weighting those tests. 

For example, these are the average times each test group take to run (on my system):
( total seconds / number of test )
- 3rd-party = 261 / 58 = 4.5  ~ 4 s/t
- repair = 148 / 44 = 3.4 ~ 3 s/t
- diagnose = 82 / 27 = 3 s/t
- signature = 51 / 20 = 2.5 ~ 2 s/t
- update = 140 / 60 = 2.3	~ 2 s/t
- bundleremove = 62 / 29 = 2.1 ~ 2 s/t
- verify-legacy = 12 / 6 = 2 s/t
- checkupdate = 21 / 11 = 1.9 ~ 2 s/t
- bundlelist = 47 / 26 = 1.8 ~ 2 s/t
- bundleadd = 81 / 53 = 1.5 ~ 1 s/t
- bundleinfo = 20 / 14 = 1.4 ~ 1 s/t
- os-install = 35 / 25 = 1.4 ~ 1 s/t
- autoupdate = 1 s/t
- usability = 17 / 19 = 0.9 ~ 1 s/t
- search = 23 / 30 = 0.8 ~ 1 s/t
- info = 1.6 / 2 = 0.8 ~ 1s/t
- mirror = 7 / 18 = 0.3s/t
- hashdump = 1 / 3 = 0.3 ~ 1 s/t

So if we round up or down to the closes integer for each group we can create 4 different "weight groups" that go from 1 to 4 (which represent average seconds a test in a group takes to complete). Using those 4 values we can average like this:
```
average() {
    i=$1
    TESTS=$2

	# start with the special cases
	if [[ $i == *'UPD025'* ]]; then
		TESTS=$((TESTS * 120))
	elif [[ $i == *'/3rd-party/'* ]]; then
		TESTS=$((TESTS * 4))
	elif [[ $i == *'/repair/'* || $i == *'/diagnose/'* ]]; then
		TESTS=$((TESTS * 3))
	elif [[ $i == *'/signature/'* || $i == *'/update/'* || $i == *'/bundleremove/'* || $i == *'/verify-legacy/'* || $i == *'/checkupdate/'* || $i == *'/bundlelist/'* ]]; then
		TESTS=$((TESTS * 2))
	fi
    echo $TESTS
}
```

I have made some other small tweaks that you can see here -> https://github.com/clearlinux/swupd-client/compare/master...castulo:auto_balancing

otaviobp:
Closed in favor of #1321

PR #1314, opened 2020-02-13
Do not install dangerous files from 3rd-party repos unless user accepts the risk

### LINKED ISSUE #1281: 3rd-party: dangerous flags handling step1 - Don't install files with dangerous flags
Some file flags are too permissive for a 3rd party repo, so they need shouldn't be installed unless whitelisted.

For this issue we should block installation of files the following dangerous flags:
 - setuid
 - set group id
 - sticky

### LINKED ISSUE #1282: 3rd-party: dangerous flags handling step2 - Only install files with dangerous flag when whitelisted
Depends on #1281 
Only install files with dangerous flags when whitelisted. Description on how the whitelist should behave is still pending.

castulo:
If a 3rd-party bundle include files that have either the setuid, setgid
or sticky bits set, then it should not be installed since they could
compromise the security of the system. In this case the user should be
asked to see if they want to assume the risk and continue, or abort the
process.

Closes #1281
Closes #1282


src/3rd_party_bundle_add.c
```diff
@@ -120,14 +123,71 @@ static enum swupd_code validate_binary(struct file *file)
...
 }
 
+static enum swupd_code validate_permissions(struct file *file)
+{
+	enum swupd_code ret_code = SWUPD_OK;
+	struct stat file_stats;
+	char *staged_file = NULL;
+
+	if (!file) {
+		return ret_code;
```
bryteise: I'm not sure on the usage but if the file doesn't exist this return SWUPD_OK, is that intended?

bryteise:
One question but LGTM.

src/3rd_party_bundle_add.c
castulo: yeah, that was intended.
This is a condition that really will never happen since this function is being called from a loop that is going through existing files... but I decided to go with SWUPD_OK because this function is supposed to perform some validation on a "file", but if "file" is NULL is really because we passed NULL to the function, not because the file doesn't meet the validation we are doing, so it felt more appropriate to just return SWUPD_OK.

PR #1319, opened 2020-02-19
Don't allow 3rd-party updates with dangerous flags

### LINKED ISSUE #1283: 3rd-party: Create an auto-update mechanism for 3rd-party repositories

castulo:
This PR implements the following things:

- If an update of 3rd-party content includes files with dangerous flags, the user should be notified and the update should only continue if the user interactively accepts the risk, otherwise the update should be aborted.
- It introduces the `--non-interactive=[Y|N]` flag that can be used to set an automatic response to prompts to run commands that otherwise would require interactive interaction, non-interactively.
- Uses the `--non-interactive=no" flag to run "3rd-party update" during autoupdate.

**Note:** This PR is built on top of #1314 and it needs to be reviewed/merged first.

Closes #1283


castulo:
@bryteise I added you as a reviewer, but it is optional, feel free to ignore my request to review 😅

src/globals.c
```diff
@@ -412,6 +414,36 @@ bool set_default_urls()
...
+	}
+
+	option_lower = str_tolower(option);
+
+	if (strcmp(option_lower, "y") == 0 || strcmp(option_lower, "yes") == 0) {
+		globals.interactive = false;
+		globals.non_interactive_yes = true;
+		ret = true;
+	} else if (strcmp(option_lower, "n") == 0 || strcmp(option_lower, "no") == 0) {
+		globals.interactive = false;
```
bryteise: So if you pass y or n interactive is set to false. Seems like a typo? Also having the option being a negative seems a little awkward. Maybe have the option be interactive instead?

bryteise:
LGTM other than my one comment.

src/globals.c
castulo: interactive = true is the default value, so if the user doesn't use `--non-interactive` swupd will interactively ask for either 'n' or 'y'. But when the user chooses non interactive mode by using the flag they can choose `--non-interactive=yes` or `--non-interactive=no`, in either case it is non interactive, the first case is just saying "respond Yes to all prompts", the second case is saying "respond No to all prompts".

bryteise: Ah that's interesting, I didn't understand the man page change very well clearly X(. Seems fine to me.

castulo: do you have suggestions on how to make the wording for the --help menu and the man pages more clear? I have the concern of people getting confused with this flag now 🤔 

This is what I'm showing now in the help menu:
`--non-interactive=[yes/no] Sets an automatic response to prompts to run non-interactively`

And the same in the Man page:
```
 • --non-interactive={yes|no}

         Sets an automatic response to prompts to run non-interactively.
```

src/helpers.c
```diff
@@ -1199,3 +1200,17 @@ char *get_tracking_dir(void)
 {
 	return sys_path_join(globals.state_dir, "bundles");
 }
+
+bool confirm_action(const char *warning_msg, const char *action)
+{
+	int response;
+
+	if (warning_msg) {
```
otaviobp: This doesn't need to be here. It's more flexible to remove this argument and the caller can print anything before calling confirm_action()

src/helpers.c
```diff
@@ -1199,3 +1200,17 @@ char *get_tracking_dir(void)
...
+		warn("%s\n", warning_msg);
+	}
+	info("Do you want to continue%s? (y/N): ", action ? action : "");
```
otaviobp: I would rephrase this in a way that you don't need to have an argument for action.
For example:
```
The 3rd-party bundle you are about to install contains files with dangerous permission
Do you want to continue? (y/N): 
```

src/helpers.c
```diff
@@ -1199,3 +1200,17 @@ char *get_tracking_dir(void)
...
+	response = tolower(getchar());
+	info("%s\n", response == 'y' ? "y" : "N");
```
otaviobp: Why do we need to print that? It seams redundant because getchar() already echoes the character.

src/globals.h
```diff
@@ -33,6 +33,8 @@ extern struct globals {
 	bool sigcheck;
 	bool timecheck;
 	bool wait_for_scripts;
+	bool is_interactive;
```
otaviobp: Instead of 2 boolens, can't we have an enum with values interactive, assume_yes, assume_no? So we won't end up in cases where is_interactive == false and assume_yes == true.

src/helpers.c
castulo: hmm I didn't see that printed in the tests output I thought it wasn't being printed.

src/helpers.c

src/helpers.c

src/helpers.c
castulo: but you are right, I just tested that and it is echoed. I'll remove this.

src/globals.h
castulo: good idea, I'll make the change

PR #1326, opened 2020-02-26
Handle conflicting 3rd-party binaries

### LINKED ISSUE #1322: Handle 3rd-party files that are exported in one bundle but not exported in another bundle
When in a 3rd-party repository one bundle has one binary file that is exported, and the same file is included in another bundle, only it is not exported in this bundle, swupd will load  a file reference from both manifests and treat the file as being the same one (since it has the same hash) so it will use one reference and will discard the other one based on an algorithm that doesn't take the "exported" flag into consideration. This means that swupd could discard the file reference that has the exported flag, this will lead to swupd not exporting the file.

A similar conflict could happen when removing one of those two bundles from the system, if we are removing the bundle that has the exported flag, swupd would notice that the file is still needed in the system hence it would not remove the script that exports the binary. But it should remove it since the bundle that remains in the system does not have the file as exported.

castulo:
When in a 3rd-party repository one bundle has one binary file that
is exported, and the same file is included in another bundle, only
it is not exported in this bundle, when we remove the bundle that
exports the file, the file should be unexported regardless of the file
still being installed since the bundle that is still in the system is
not exporting it.

Closes #1322

### DIFF
--- a/src/3rd_party_bundle_remove.c
+++ b/src/3rd_party_bundle_remove.c
@@ -22,6 +22,8 @@
 #include "3rd_party_repos.h"
 #include "swupd.h"
 
+#include <errno.h>
+
 #ifdef THIRDPARTY
 
 static char **cmdline_bundles;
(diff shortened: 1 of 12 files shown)


src/3rd_party_bundle_remove.c
```diff
@@ -93,9 +93,37 @@ static bool parse_options(int argc, char **argv)
...
+	/* some binary files from the removed bundle may not have been removed due to
+	 * being needed by bundles still installed in the system, however, if
+	 * the bundles that are still installed didn't export those binaries then
+	 * the scripts from those binaries must be deleted */
+	for (iter = common_files; iter; iter = iter->next) {
+		file = iter->data;
+		if (!file->is_file || !is_binary(file->filename) || file->is_exported) {
+			continue;
+		}
+		/* if the file is exported, we can unexport it now */
```
bryteise: I think the comment should read "if the file isn't exported".

bryteise:
LGTM with maybe one fix to a comment.

src/3rd_party_bundle_remove.c
castulo: Hmm, I see now how that message could be confusing, the file is really exported because at that point of time a script exists that exports that file (and that is what I meant), but it should no longer be exported because the remaining bundles don't have the is_exported flag set (which I believe is what you mean't).

I will rephrase the message so it is not confusing.

bryteise: Ah okay, your explanation makes perfect sense with that context, thanks!

PR #1328, opened 2020-02-26
Move 3rd-party repo.ini out of the state directory

castulo:
The repo.ini file was being created in the swupd state directory, this
was not ideal since it prevented the state directory from being used for
multiple target paths.

This commit moves the repo.ini file out of the state directory and into
the 3rd-party content directory.

### DIFF
--- a/src/3rd_party_add.c
+++ b/src/3rd_party_add.c
@@ -254,6 +254,7 @@ enum swupd_code third_party_add_main(int argc, char **argv)
 	 */
 	progress_init_steps("third-party-add", step_in_third_party_add);
 
+	/* The last two in reverse are the repo-name, repo-url */
 	name = argv[argc - 2];
 	url = argv[argc - 1];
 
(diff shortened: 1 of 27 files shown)


bryteise:
My thinking this through means that if you give `--path=/foo` for third party operations then the structure looks like `/foo/opt/3rd-party/{bin,bundles,repo.ini}` with the statedir still under `/var/lib/swupd` (I forget how third party state is namespaced). Before you would get `repo.ini` under `/var/lib/swupd` still. This enables multiple roots each with their own 3rd-party configurations with passing only `--path=` where as before you'd need `--statedir=` as well and for them to stay in sync. Seems like a usability win and if there is oddly statedir type conflicts that can still be passed separately as well.

otaviobp:
@bryteise My plan was to keep the statedir only for cache files and save everything else relative to the path. Right now the only other thing that is not cache that lives in the statedir is the list of tracked bundles and it's very hard to reuse the statedir for multiple paths because of that. Do you have any restriction on moving that from statedir to a path relative dir?

bryteise:
As long as swupd knows where to find tracking files, it is fine with me =).

PR #1333, opened 2020-02-27
3rd-party: Support adding symlinks to 3rd party repositories

otaviobp:
When a symlink is added to a 3rd-party repository it will be broken
on the statedir and stat command will fail. We need to use lstat, to
get information about the symlink and not the file that it's pointing
to.

### DIFF
--- a/src/3rd_party_bundle_add.c
+++ b/src/3rd_party_bundle_add.c
@@ -133,7 +133,7 @@ static enum swupd_code validate_permissions(struct file *file)
 	}
 
 	string_or_die(&staged_file, "%s/staged/%s", globals.state_dir, file->hash);
-	if (stat(staged_file, &file_stats) == 0) {
+	if (lstat(staged_file, &file_stats) == 0) {
 		if ((file_stats.st_mode & S_ISUID) || (file_stats.st_mode & S_ISGID) || (file_stats.st_mode & S_ISVTX)) {
 			warn("File %s has dangerous permissions\n", file->filename);
 			ret_code = SWUPD_NO;
--- a/test/functional/3rd-party/3rd-party-bundle-add-symlink.bats
+++ b/test/functional/3rd-party/3rd-party-bundle-add-symlink.bats
@@ -1,0 +1,47 @@
+#!/usr/bin/env bats
+
+# Author: Otavio Pontes
+# Email: otavio.pontes@intel.com
+
+load "../testlib"
+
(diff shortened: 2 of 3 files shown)


test/functional/3rd-party/3rd-party-bundle-add-symlinkc.bats
```diff
@@ -0,0 +1,55 @@
...
+	create_third_party_repo -a "$TEST_NAME" 10 1 test-repo1
+
+	files_dir="$TEST_NAME"/3rd-party/test-repo1/10/files/
+	sudo ln -s broken "$files_dir"/symlink
+	hashed_name=$("$SWUPD" hashdump --quiet "$files_dir"/symlink)
+	sudo mv "$files_dir"/symlink "$files_dir"/"$hashed_name"
+
+	create_tar "$files_dir"/"$hashed_name"
+
+	create_bundle -n test-bundle1 -f /foo/symlink:"$files_dir"/"$hashed_name" -u test-repo1 "$TEST_NAME"
```
castulo: Just a tip for the future, testlib supports the creation of bundles with dangling links with the use of the `-b` flag, it would have saved you some effort:
```
Usage:
    create_bundle [-L] [-t] [-e] [-n] <bundle_name> [-v] <version> [-u] <3rd-party repo> [-d] <list of dirs> [-f] <list of files> [-l] <list of links> ENV_NAME

Options:
    -L    When the flag is selected the bundle will be 'installed' in the target-dir, otherwise it will only be created in content dir
    -t    The bundle will be 'tracked' in the state-dir regardless of if its content is installed on the system
    -e    The bundle will be marked as experimental
    -n    The name of the bundle to be created, if not specified a name will be autogenerated
    -v    The version for the bundle, if non selected version 10 will be used
    -u    If the bundle is to be created inside a 3rd-party repo, the repo's name should be specified here
    -d    Comma-separated list of directories to be included in the bundle
    -f    Comma-separated list of files to be created and included in the bundle
    -l    Comma-separated list of symlinks to files to be created and included in the bundle
    -c    Comma-separated list of symlinks to directories to be created and included in the bundle
    -b    Comma-separated list of dangling (broken) symlinks to be created and included in the bundle
```

otaviobp: Nice, I didn't find that when I was looking for it. I'll use that in this PR

PR #1335, opened 2020-02-27
scripts: Always run ld_config before running any update scripts

otaviobp:
It's possible to create a post_update script to run ld_config, but we
had problems with this in the past, so it was decided to hardcode that
on swupd instead of relying on scripts.

### DIFF
--- a/src/scripts.c
+++ b/src/scripts.c
@@ -94,6 +94,21 @@ static void exec_post_update_script(bool reexec, bool block)
 	free(params[0]);
 }
 
+static void run_ldconfig(void)
+{
+	int err;
+
+	if (strcmp("/", globals.path_prefix) == 0) {
+		err = run_command_quiet("/usr/bin/ldconfig", NULL);
+	} else {
+		err = run_command_quiet("/bin/chroot", globals.path_prefix, "/usr/bin/ldconfig", NULL);
+	}
+
+	if (err) {
+		debug("Running ldconfig failed with error code = %d\n", err);
+	}
+}
+
 static void update_triggers(bool block)
 {
 	if (strlen(POST_UPDATE) == 0) {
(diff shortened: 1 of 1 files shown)


bryteise:
I'm curious if there will be improvements to https://github.com/clearlinux/distribution/issues/1765 with this change.

src/scripts.c
```diff
@@ -94,6 +94,21 @@ static void exec_post_update_script(bool reexec, bool block)
...
+	int err;
+
+	if (strcmp("/", globals.path_prefix) == 0) {
+		err = run_command_quiet("/usr/bin/ldconfig", NULL);
+	} else {
+		err = run_command_quiet("/bin/chroot", globals.path_prefix, "/usr/bin/ldconfig", NULL);
+	}
+
+	if (err) {
+		debug("Running ldconfig failed with error code = %d", err);
```
castulo: you forgot the '\n' again 😄

src/scripts.c
```diff
@@ -94,6 +94,21 @@ static void exec_post_update_script(bool reexec, bool block)
 	free(params[0]);
 }
 
+static void run_ld_config()
```
castulo: you forgot the 'void' argument here, I learnt that from you 😝

otaviobp:
V2: Fixed the minor problems

src/scripts.c
castulo: I think you forgot this ☝️

otaviobp:
Sorry, I didn't add that to the commit

mythi:
> I'm curious if there will be improvements to clearlinux/distribution#1765 with this change.

I tried adding `RUN ldconfig` next to `swupd` commands in my `Dockerfile` and the image still fails.

PR #1336, opened 2020-02-28
PoC: Make tests 10x faster by saving a cache

otaviobp:
I wrote a PoC to show how it's possible to make the swupd tests 10x faster by saving all generated files between one execution to the other. There's still room for improvements and I made some hacks to implement the PoC faster so I can have some feedback to make sure it's worth pursuing this changes.

The basic idea of this PR is to run everything on test_setup() and save the results in tarballs to be committed to the repository. So when we need to run the tests we don't need to execute test_setup() again. We just extract the tarballs to the correct location and uses it to execute the tests.

For comparison reasons I created 2 jobs on github actions to run in serial all repair tests using the new approach and the old approach. Tests using the tarball cache are around 10 times faster than tests that generate all the data. I have tested that only with repair tests.

Some improvements necessary:
 - Instead of having specific generate scripts we could keep the generation code in bats file
 - Create a script to automatically clean and regenerate the cache
 - Check if we can reuse the web-dir for multiple tests (there's one  tests that breaks if we do that)
 - Check if there's a better place to put the tarballs other than the git repository


bryteise:
What is the process for generating the test tarballs? Generally I don't like checking in binaries to the repo itself especially if they change and this is a lot of them.

otaviobp:
> What is the process for generating the test tarballs? Generally I don't like checking in binaries to the repo itself especially if they change and this is a lot of them.

I made a hack script to generate them. We need a better script. The only reason it's a tarball and not a text file is root permissions. We could change the permissions and fix them before running the test if we prefer to keep them as text files.

We also use a random file generator to create the text files. If we want to commit them as files it would be ideal to have a fixed seed so they wouldn't change every execution.

Another possible option would be to save this tarballs somewhere else and download them to run the tests. The major reason I created this PoC is to show that it's possible to have way faster tests by saving the metadata.

bryteise:
I think saving the generated content for reuse is definitely a win (as this PoC demonstrates). Since the how is open, I would be interested in testing storing the text and seeing how much of a performance difference that makes. Storing the tarballs elsewhere is an interesting option but I think the logistics are a bit more difficult.

castulo:
I think this idea could work

otaviobp:
Closing in favor of #1385

PR #1337, opened 2020-02-28
Use of --assume and --yes flags to confirm a cert

castulo:
When users add a 3rd-party repository they are asked if they want to
trust the certificate, this commit makes possible to use the
--assume flag to answer that question.

### DIFF
--- a/src/3rd_party_add.c
+++ b/src/3rd_party_add.c
@@ -94,18 +94,12 @@ static int remove_repo(const char *repo_name)
 
 static bool confirm_certificate(const char *cert)
 {
-	int c;
-
 	info("Importing 3rd-party repository public certificate:\n\n");
 	signature_print_info(cert);
 
 	info("\n");
-	info("Do you want to accept this certificate? (y/N): ");
-
-	c = tolower(getchar());
-	info("\n");
-
-	return c == 'y';
+	info("To add the 3rd-party repository you need to accept this certificate\n");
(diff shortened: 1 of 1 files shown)


src/3rd_party_add.c
```diff
@@ -94,18 +94,11 @@ static int remove_repo(const char *repo_name)
 
 static bool confirm_certificate(const char *cert)
 {
-	int c;
-
 	info("Importing 3rd-party repository public certificate:\n\n");
 	signature_print_info(cert);
 
 	info("\n");
-	info("Do you want to accept this certificate? (y/N): ");
```
otaviobp: I think it's important to print a message asking if we want to accept the certificate. It's not clear to me just to say "Do you want to continue?"

castulo: how about now?

src/3rd_party_add.c
```diff
@@ -94,18 +94,12 @@ static int remove_repo(const char *repo_name)
...
-
-	c = tolower(getchar());
-	info("\n");
-
-	return c == 'y';
+	info("To add the 3rd-party repository you need to accept this certificate. ");
```
otaviobp: I would add a \n

castulo: haha that is just what I was thinking... the line looked to long without it

castulo:
V3: Addressed review comments

PR #1344, opened 2020-03-03
Do not apply heuristics to 3rd-party files

### LINKED ISSUE #1342: 3rd-party bundle-add should be allowed install state files
When swupd is installing a bundle, it skips installing files considered to be state files, these include files in:
```
/data
/dev
/home
/lost+found
/proc
/root
/run
/sys
/tmp
/usr/src
/var
```

For swupd 3rd-party, a user should be able to install files in those directories.

castulo:
When adding an upstream bundle, we apply some heuristics to the bundle
files and based on that some files are skipped from being installed.
These heuristics don't apply to 3rd-party bundles.

This commit skips running heuristics when adding 3rd-party bundles.

Closes #1342 

### DIFF
--- a/src/bundle_add.c
+++ b/src/bundle_add.c
@@ -215,23 +215,29 @@ static int compute_bundle_dependecies(struct manifest *mom, struct list *bundles
 	return 0;
 }
 
+static enum swupd_code apply_heuristics_for_new_files(struct list *files)
+{
+	struct list *iter;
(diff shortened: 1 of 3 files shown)


bryteise:
LGTM, I like the test change quite a bit too. We may come to a point where those pre/post/val functions would be better as a function list that are applied in order but not something to consider at this point.

PR #1348, opened 2020-03-04
Forcing the removal of corrupt 3rd-party repos

### LINKED ISSUE #1343: Make swupd 3rd-party remove more error prone
Right now swupd 3rd-party remove depens on the installed 3rd-part content to be sane in order to proceed. This is problematic because if there's a corruption (or destructive changes) in the installed directory there's no way to remove that permanently.

For example the following would fail:
```
# sudo ./swupd 3rd-party add repo http://my_mix
# rm /opt/3rd-party/bundles/repo/usr/lib/os-release
# sudo ./swupd 3rd-party remove repo
```

We need a better approach on when there are errors removing the wrappers. Maybe a --force to remove the repo anyway? And for sure better warning messages.

castulo:
This commit allow users to force the deletion of corrupt 3rd-party repos
which will end up deleting most of the repo's files, except for the
scripts that exported files from that repo.

Closes #1343

### DIFF
--- a/src/3rd_party_remove.c
+++ b/src/3rd_party_remove.c
@@ -25,22 +25,43 @@
 
 #ifdef THIRDPARTY
 
+static bool cmdline_option_force = false;
+
+static const struct option prog_opts[] = {
+	{ "force", no_argument, 0, 'x' },
+};
+
 static void print_help(void)
 {
 	/* TODO(castulo): we need to change this description to match that of the
 	 * documentation once the documentation for this content is added */
 	print("Removes a 3rd-party repository from the list of available repositories in the system\n\n");
 	print("Usage:\n");
-	print("   swupd 3rd-party remove [repo-name]\n\n");
+	print("   swupd 3rd-party remove [OPTION...] [repo-name]\n\n");
 
 	global_print_help();
+
+	print("Options:\n");
+	print("   -x, --force             Attempt to proceed even if non-critical errors found\n");
 	print("\n");
 }
 
+static bool parse_opt(int opt, char *optarg)
+{
+	switch (opt) {
+	case 'x':
+		cmdline_option_force = optarg_to_bool(optarg);
+		return true;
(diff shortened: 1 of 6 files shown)


castulo:
@otaviobp please check the messages I'm printing and let me know if we need to provide more information to the user.

src/3rd_party_remove.c
```diff
@@ -146,10 +167,29 @@ enum swupd_code third_party_remove_main(int argc, char **argv)
...
 
+	/* if there are errors at this point the repo may be corrupt,
+	 * continue only if forced by the user */
+remove_repo:
+	if (ret) {
+		if (cmdline_option_force) {
+			char *bin_dir;
+			bin_dir = third_party_get_bin_dir();
+			warn("The --force option is specified; forcing the removal of the repository\n");
+			info("Please be aware that this may leave dangling files in %s\n", bin_dir);
```
otaviobp: s/dangling/exported/ ?

src/3rd_party_remove.c
```diff
@@ -146,10 +167,29 @@ enum swupd_code third_party_remove_main(int argc, char **argv)
...
+			free_and_clear_pointer(&bin_dir);
+			ret_partial = ret;
+			ret = SWUPD_OK;
+		} else {
+			error("The 3rd-party repository %s is corrupt and cannot be removed cleanly\n", name);
```
otaviobp: I think it's better to write something here about the possible dangling files too. Something like:
```
The 3rd-party repository %s is corrupt and cannot be removed cleanly
Some exported files can be left behind at /opt/3rd-party/bin/
```

castulo: sounds good, I'll make the change

src/3rd_party_remove.c
castulo: sure, I'll go with 'exported' then

PR #1352, opened 2020-03-05
Staging: Don't create hardlinks when installing symlinks

otaviobp:
Staging: Don't create hardlinks when installing symlinks
    
    Alternative solution to solution proposed on #1318
    
    Instead of trying to create a link just don't try to link symlinks.
    Tested patch using bsdtar and implemented automated testing to
    validate solution.
    
### DIFF
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -2303,7 +2303,7 @@ create_bundle() { # swupd_function
 	cb_usage() {
 		cat <<-EOM
 		Usage:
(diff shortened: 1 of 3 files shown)


otaviobp:
PR reworked and ready to be reviewed now

test/functional/testlib.bash
```diff
@@ -2395,10 +2399,12 @@ create_bundle() { # swupd_function
 		return
```
castulo: you need to change this call to the help menu here (line 2398), at the moment is using `create_bundle -h` to display the help menu, since there was no `h` parameter it was just defaulting to showing the help menu, but since you are defining that parameter then we need to call the help menu by just calling `cb_usage`

PR #1366, opened 2020-03-10
scripts: Add a script to run swupd basic tests

otaviobp:
Besides having a script that is used to run tests on github
actions CI, adding another script that can be used in any
system to build swupd with different configurations and
run all needed tests.

### DIFF
--- a/scripts/build_and_run_tests.bash
+++ b/scripts/build_and_run_tests.bash
@@ -1,0 +1,43 @@
+#!/bin/bash
+# Simple script to build swupd with multiple configurations and run all needed tests with different configurations
+
+set -e
+
+#configurable parameters
+if [ -z "$JOB_COUNT" ]; then
+	JOB_COUNT=2
+fi
+
+CONFIGURE_ARGS="--with-fallback-capaths=./swupd_test_certificates -with-systemdsystemunitdir=/usr/lib/systemd/system --with-config-file-path=./testconfig"
+
+export RUNNING_IN_CI=true
+run_build() {
+	make distclean || echo "No distclean - System already cleaned"
+	autoreconf -fi
+	# shellcheck disable=SC2068
+	# We want to expand the variable, so ignore shellcheck warning
+	./configure $@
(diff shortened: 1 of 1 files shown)


scripts/build_and_run_tests.bash
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+# Simple script to build swupd with multiple configurations and run all needed tests with different configurations
+
+set -e
+
+#configurable parameters
+JOB_COUNT=8
+CONFIGURE_ARGS="--with-fallback-capaths=$PWD/swupd_test_certificates --with-config-file-path=./testconfig --enable-third-party"
```
castulo: `--enable-third-party` is the default now, so you can remove this.

scripts/build_and_run_tests.bash
```diff
@@ -0,0 +1,36 @@
...
+	make distcheck
+}
+
+# Run tests without any security sanitizer
+CFLAGS="" run_build --disable-optimizations "$CONFIGURE_ARGS"
+run_checks
+
+# Run build with basic configurations
+CFLAGS="$CFLAGS -fsanitize=address -fno-omit-frame-pointer -Werror" run_build "$CONFIGURE_ARGS"
+run_checks
```
castulo: What is the purpose of running the tests twice once without security sanitizer and once with?

scripts/build_and_run_tests.bash
otaviobp: done

scripts/build_and_run_tests.bash
otaviobp: This helps us to detect code problems that sanitizers and optimizations may hide. I remember we had a case in the past, but I don't remember what it is specifically.

castulo: I would have thought that this combination would be enough:
- optimization **off**
- sanitizers **on**

But if you think there is value in running the tests twice changing those values I won't oppose. ☮️

Issue #1369, opened 2020-03-11
swupd os-install: Error: Missing --bundle argument


mythi:
Our container builds have:

```
RUN mkdir /install_root \
    && swupd os-install \
    ${CLEAR_LINUX_VERSION} \
    --path /install_root \
    --statedir /swupd-state \
    --bundles=os-core \
    --no-boot-update \
    && rm -rf /install_root/var/lib/swupd/*
```

but is has started failing with an error: `Error: Missing --bundle argument`

There's no documentation about `--bundle`. 

However, if I change  to `--bundles=os-core,vi` (as shown by the help text), this starts working*. It looks `os-core` is somehow ignored/skipped and if the list becomes empty the error is given.

(* However, `vi` gives `Warning: Bundle "vi" is invalid or no longer available`)

otaviobp:
This is a bug introduced on last release. I'm reproduced it here and I'll move it to swupd repository.
The problem is that os-core is always added on os-install, so it's filtered and we print a warning as if we didn't have any bundles in the parameter.
And `--bundle` should be `--bundles`. It's a typo

otaviobp:
As a workaround for now, don't use the --bundles if you are only installing os-core. And if you are adding more than one bundle, this should work.
Also vi was rename to vim. We need to update that in the documentation.

mythi:
> As a workaround for now, don't use the --bundles if you are only installing os-core

I'll wait for the fix because I don't want to rework the Dockerfiles after the fixes unless it's guaranteed that `swupd os-install` without `--bundles` defaults to `os-core` always.

otaviobp:
It's guaranteed that os-install will always install os-core when no arguments are provided to --bundles and there are no plans on changing that.
But we are working in a fix that should be ready and released soon too, so it's up to you.

mythi:
> It's guaranteed that os-install will always install os-core when no arguments are provided to --bundles

OK. I think that should be documented too and be visible to users who run `swupd os-install --help`.

PR #1370, opened 2020-03-11
docs: Replace vi with vim

otaviobp:
Vi isn't a bundle anymore. Replacing all examples to vim.

Related to #1369

### DIFF
--- a/docs/swupd.1.rst
+++ b/docs/swupd.1.rst
@@ -441,9 +441,9 @@ SUBCOMMANDS
 
        Examples:
 
-         -  ``--bundles os-core,vi``
+         -  ``--bundles xterm,vim``
 
-            Diagnoses only bundles `os-core` and `vi`.
+            Diagnoses only bundles `xterm` and `vim`.
 
     -  ``-Y, --picky``
 
@@ -521,9 +521,9 @@ SUBCOMMANDS
 
        Examples:
 
-         -  ``--bundles os-core,vi``
+         -  ``--bundles xterm,vim``
 
-            Repairs only bundles `os-core` and `vi`.
+            Repairs only bundles `xterm` and `vim`.
 
     -  ``-Y, --picky``
 
@@ -596,9 +596,9 @@ SUBCOMMANDS
 
        Examples:
 
-         -  ``--bundles xterm,vi``
+         -  ``--bundles xterm,vim``
 
-            Installs bundles `xterm` and `vi`, along with `os-core` (installed by default).
+            Installs bundles `xterm` and `vim`, along with `os-core` (installed by default).
 
     -  ``-s, --statedir-cache={PATH}``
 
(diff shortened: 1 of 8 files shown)


docs/swupd.1.rst
```diff
@@ -441,7 +441,7 @@ SUBCOMMANDS
 
        Examples:
 
-         -  ``--bundles os-core,vi``
+         -  ``--bundles os-core,vim``
```
castulo: you also need to replace `vi` then in line 446

docs/swupd.1.rst
```diff
@@ -521,7 +521,7 @@ SUBCOMMANDS
...
```
castulo: same here...

docs/swupd.1.rst
```diff
@@ -596,7 +596,7 @@ SUBCOMMANDS
 
        Examples:
 
-         -  ``--bundles xterm,vi``
+         -  ``--bundles xterm,vim``
```
castulo: ... and here

otaviobp:
V2: Reported problems fixed.

swupd.zsh
```diff
@@ -351,7 +351,7 @@ if [[ -n "$state" ]]; then
           local -a osinstall; osinstall=(
             $global_opts
             '(help -x --force)'{-x,--force}'[Attempt to proceed even if non-critical errors found]'
-            '(help -B --bundles)'{-B,--bundles=}'[Include the specified BUNDLES in the OS installation. Example: --bundles=os-core,vi]:diagnose: _swupd_all_bundles'
+            '(help -B --bundles)'{-B,--bundles=}'[Include the specified BUNDLES in the OS installation. Example: --bundles=os-core,vim]:diagnose: _swupd_all_bundles'
```
castulo: same here, change os-core with something else

src/os_install.c
```diff
@@ -53,7 +53,7 @@ static void print_help(void)
 	print("Options:\n");
 	print("   -V, --version=[VER]         If the version to install is not the latest, it can be specified with this option\n");
 	print("   -x, --force                 Attempt to proceed even if non-critical errors found\n");
-	print("   -B, --bundles=[BUNDLES]     Include the specified BUNDLES in the OS installation. Example: --bundles=os-core,vi\n");
+	print("   -B, --bundles=[BUNDLES]     Include the specified BUNDLES in the OS installation. Example: --bundles=os-core,vim\n");
```
castulo: I just thought one more thing, many we should also remove os-core from that list, change it with `openssl` or something like that since `os-core` is installed by default, this may confuse people.

otaviobp: yes. it's a good idea. I am going to add joe, because it's small. It's a good bundle for testing

otaviobp:
v3: Replacing os-core to xterm

PR #1371, opened 2020-03-12
Bug fix: os-install --bundles with os-core fails

### LINKED ISSUE #1369: swupd os-install: Error: Missing --bundle argument

castulo:
When doing a system install with "os-install --bundles" if the list of
bundles to install includes only "os-core" the install fails.

This commit fixes the issue.

Closes #1369 

### DIFF
--- a/src/os_install.c
+++ b/src/os_install.c
@@ -88,14 +88,12 @@ static bool parse_opt(int opt, char *optarg)
 		char *arg_copy = strdup_or_die(optarg);
 		char *token = strtok(arg_copy, ",");
 		while (token) {
-			if (strcmp(token, "os-core") != 0) {
-				cmdline_bundles = list_prepend_data(cmdline_bundles, strdup_or_die(token));
-			}
+			cmdline_bundles = list_prepend_data(cmdline_bundles, strdup_or_die(token));
 			token = strtok(NULL, ",");
 		}
 		free(arg_copy);
 		if (!cmdline_bundles) {
-			error("Missing --bundle argument\n\n");
+			error("Missing --bundles argument\n\n");
 			return false;
 		}
 		return true;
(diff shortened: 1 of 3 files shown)


src/os_install.c
```diff
@@ -87,15 +87,18 @@ static bool parse_opt(int opt, char *optarg)
 	case 'B': {
 		char *arg_copy = strdup_or_die(optarg);
 		char *token = strtok(arg_copy, ",");
+		bool oscore_requested = false;
 		while (token) {
-			if (strcmp(token, "os-core") != 0) {
+			if (strcmp(token, "os-core") == 0) {
```
otaviobp: Why we just don't add os-core to the list?

castulo: To be honest, IDK 🤣 , I assumed we had a reason no to, so I kept the same code logic. But now that I'm looking at the code I cannot find a reason why not to. I'll take a closer look and will run a few tests, if everything looks good I'll just add it as you mentioned.

castulo:
V2: I just tested the fix as @otaviobp suggested just by allowing os-core to be added to the bundles list when requested, and it works fine. So this update includes that change.

PR #1374, opened 2020-03-13
Shellcheck-all: Run shellcheck all on new commits on master

otaviobp:


otaviobp:
As job doesn't run on pull requests there's no run status for this PR. If you want to check the action performed you can look at my repo: https://github.com/otaviobp/swupd-client/actions/runs/55271759

test/code_analysis/shellcheck-all.bats
```diff
@@ -4,11 +4,24 @@
 # Email: otavio.pontes@intel.com
 
 @test "ANL002: Shellcheck on all files" {
-	find . -regex ".*\.bats\|.*\.bash" -print0 | while IFS= read -d '' -r i; do
-		echo checking "$i"
+	local error=0
+	status=0
+	output=""
+	TESTS=$(find . -regex ".*\\.bats\\|.*\\.bash" | sort)
```
castulo: do we want to include files that are .sh?

otaviobp:
> Nice one, the only comment I have is that it may be worth adding .sh files to be checked by shellcheck, also in case at some point someone adds one. Anyway I'm approving this PR since we don't really have .sh files so it is just a nice to have.

I'll address that in a new PR

PR #1380, opened 2020-03-17
Testlib: add support to create and update bundles based on content from a directory

castulo:
This PR adds support to create and update bundles based on the content of a directory. The idea of these improvements is to facilitate the creation of more complex test scenarios.

The PR also includes a few bug fixes in the test library.


test/functional/update/update-type-changes-symlink-to-file.bats
```diff
@@ -0,0 +1,87 @@
...
+
+test_setup() {
+
+	create_test_environment "$TEST_NAME"
+
+	# create a directory with some content that will
+	# be used to create our test bundle
+	sudo mkdir -p "$TEST_NAME"/tmp/dir1/dirA
+	write_to_protected_file "$TEST_NAME"/tmp/fileA "$(generate_random_content)"
+	sudo ln -s --relative "$TEST_NAME"/tmp/fileA "$TEST_NAME"/tmp/file1
```
otaviobp: I wonder if we need to test absolute symlinks too

test/functional/update/update-type-changes-symlink-to-file.bats
```diff
@@ -0,0 +1,87 @@
...
+		Update successful - System updated from version 10 to version 20
+	EOM
+	)
+	assert_is_output "$expected_output"
+
+	assert_file_exists "$TARGETDIR"/file1
+	assert_dir_exists "$TARGETDIR"/dir1/dir2
+	assert_file_exists "$TARGETDIR"/dir1/dir2/file2
+	assert_file_exists "$TARGETDIR"/file3
+	assert_file_exists "$TARGETDIR"/file4
```
otaviobp: We need to check here if file type is correct.

test/functional/update/update-type-changes-file-to-symlink.bats
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env bats
```
otaviobp: Test looks good, but it's a good idea to add a comment on all file type changes being tested so we can check if there's nothing left.

test/functional/testlib.bash
```diff
@@ -2659,6 +2694,429 @@ create_bundle() { # swupd_function
...
+	validate_param "$bundle_name"
+
+	echo "Updating bundle $bundle_name using content from $content_dir..."
+
+	# default values
+	# shellcheck disable=SC2010
+	# SC2010: Don't use ls | grep. Use a glob.
+	# Exception: ls has sorting options that are tricky
+	# to get right with other commands.
+	version=${version:-$(ls "$env_name"/web-dir | grep -E '^[0-9]+$' | sort -rn | head -n1)}
```
otaviobp: This looks like something we should have a function

test/functional/testlib.bash
```diff
@@ -2659,6 +2694,429 @@ create_bundle() { # swupd_function
...
+	old_version=$(sudo cat "$env_name"/web-dir/"$version"/Manifest.MoM | grep -m1 "$bundle_name" | cut -f 3)
```
otaviobp: same here

test/functional/testlib.bash
```diff
@@ -2659,6 +2694,429 @@ create_bundle() { # swupd_function
...
+	web_dir="$env_name"/web-dir
+	target_dir="$env_name"/testfs/target-dir
+	state_dir="$env_name"/testfs/state
+	version_dir="$web_dir"/"$version"
+	old_version_dir="$web_dir"/"$old_version"
+
+	# to create the bundle update we will start treating the update as
+	# if it was a new bundle, then we will just adjust the generated artifacts
+	# to look like an update, so we will create a "new bundle" based on the new content first
+	create_bundle_from_content -n "$bundle_name" -v "$version" -c "$content_dir" "$env_name"
```
otaviobp: The approach on using another function and undoing things that aren't necessary is not ideal.
The ideal is to find the things in common and create a function to perform that. And things that are different should be performed by each individual functions.

test/functional/testlib.bash
```diff
@@ -2659,6 +2694,429 @@ create_bundle() { # swupd_function
...
+
+	debug_msg "Creating bundle $bundle_name using content from $content_dir..."
+
+	# default values
+	bundle_name=${bundle_name:-$(generate_random_name test-bundle-)}
+	# shellcheck disable=SC2010
+	# SC2010: Don't use ls | grep. Use a glob.
+	# Exception: ls has sorting options that are tricky
+	# to get right with other commands.
+	version=${version:-$(ls "$env_name"/web-dir | grep -E '^[0-9]+$' | sort -rn | head -n1)}
```
otaviobp: this is repeated 4 times. should be a function

test/functional/testlib.bash
```diff
@@ -2659,6 +2694,429 @@ create_bundle() { # swupd_function
...
+		recursive_path="$content_dir"
+	fi
+
+	# shortcut to useful directories
+	version_dir=$(dirname "$manifest")
+	files_dir="$version_dir"/files
+
+	# get an array with all the content we want to add to the manifest
+	# from the current recursive path
+	mapfile -t content < <(ls "$recursive_path")
```
otaviobp: Can you use find here? So we don't need to make this function recursive

test/functional/testlib.bash
```diff
@@ -861,6 +862,8 @@ add_to_manifest() { # swupd_function
 		          updates are to be done in the manifest to avoid extra processing
 		    -e    If set (experimental), the bundle will be added to the manifest as
 		          experimental. This should only be used when <manifest> is a MoM
+		    -l    If set (do not follow links), if a link is being added, the file
```
otaviobp: Instead of adding this parameter, I think that function add_to_manifest() should never include the file that is being liked in the manifest. If the user wants to include it, the user can do that explicitly. That's how most tool works and how swupd usually handle files. For example:
```
touch a
ln -s a b
tar cjf file.tar b # a is not automatically included.
```

otaviobp: Another win on that is that we don't need a recursion in add_to_manifest(). I have the feeling that this function has an infinite loop if you have a link pointing to itself.

test/functional/testlib.bash
castulo: turns out we already have a function called `get_latest_version` which does exactly that, I had forgotten about it

test/functional/testlib.bash
castulo: it may work, it never occurred to me I could use find for that, it is worth trying

test/functional/testlib.bash
castulo: looks like we were already explicitly adding those files anyway, so making this change shouldn't be a problem. Good eye.

castulo:
V2 addresses all requested changes:
- stop adding files pointed by links when adding symlinks to a manifest, they will have to be added explicitly
- using `find` when collecting the content of a directory during `add_content_to_bundle()` instead of traversing the directories
- using `get_latest_version()` to retrieve version numbers
- Refactored `update_bundle_from_content()` to remove steps it did not need

test/functional/testlib.bash
```diff
@@ -2791,6 +2834,399 @@ create_bundle() { # swupd_function
...
+	local filename
+	local content
+
+	# shortcut to useful directories
+	version_dir=$(dirname "$manifest")
+	files_dir="$version_dir"/files
+
+	# get an array with all the content we want to add to the manifest
+	# from the current recursive path
+	mapfile -t content < <(find "$content_dir" -mindepth 1 -printf "%P\n")
```
otaviobp, line 2866: I didn't know this mapfile command. Pretty cool

test/functional/testlib.bash
```diff
@@ -1188,6 +1201,42 @@ update_manifest() { # swupd_function
 
 }
 
+get_entry_from_manifest() { # swupd_function
```
otaviobp, line 1204: I liked this function too and I can see several usages for that. But there are some problems

test/functional/testlib.bash
```diff
@@ -1188,6 +1201,42 @@ update_manifest() { # swupd_function
...
+			    get_entry_from_manifest <manifest> [file_hash or file_name]
+			EOM
+		return
+	fi
+	validate_item "$manifest"
+
+	# if no key is provided, return a list with all
+	# files/entries from the manifest
+	if [ -z "$entry_key" ]; then
+		sudo cat "$manifest" | grep -E -x "(D|F|L)..."$'\t'".*"$'\t'".."$'\t'"/.*" | awk '{ print $4 }'
```
otaviobp, line 1223: For me, this doesn't work when version number of file has more than 2 digits.

test/functional/testlib.bash
```diff
@@ -1188,6 +1201,42 @@ update_manifest() { # swupd_function
...
+		return
+	fi
+
+	# a key was provided, try with the key as a file
+	# path first which should be unique in a Manifest
+	value=$(sudo cat "$manifest" | grep "$entry_key"$)
+
+	# if an entry was not found using the file path, try with the hash next
+	if [ -z "$value" ]; then
+		value=$(sudo cat "$manifest" | grep ^"...."$'\t'"$entry_key")
```
otaviobp, line 1233: It looks like the output format is different here than when no parameter is used. I think we need to be consistent in the format of the output

test/functional/testlib.bash
castulo: ahh, good catch my friend, I'll fix that

test/functional/testlib.bash
castulo: you are right, this is because I first I had two separate functions, and when I merged them I forgot about consistency. I'll fix that too.

PR #1382, opened 2020-03-18
curl: Don't resume download in case of HTTP 416 error

chenming1986:
When tens of thousands of files are transferred (eg. update/repair to a
recent build from a very old one), there are chances that curl returns
CURLE_RECV_ERROR when downloading some of the files even if those files
are already successfully downloaded. In such cases resuming download
will always fail with HTTP 416 error - requested range not satisfiable.

This change adds a new download status for HTTP 416 and in case of that
status, start the download from the beginning all over again for the
affected files.

Signed-off-by: Ming Chen <ming.a.chen@intel.com>


chenming1986:
Attaching the log when I was attempting to repair to Clear Linux 32400 from 32610. Private mirror url was hidden in the log. Updating to the latest version from an very old one also experienced the same issue.
[swupd.log](https://github.com/clearlinux/swupd-client/files/4348802/swupd.log)

chenming1986:
Thanks for merging!

PR #1385, opened 2020-03-19
Test cache: First batch of changes to support test caching

otaviobp:
This PR isn't adding the caching capabilities yet, but it is introducing one step used to create metadata in a way that can be cached.


test/functional/testlib.bash
```diff
@@ -1,6 +1,3 @@
-#!/usr/bin/bash
```
castulo: if you remove this, shellcheck won't work

test/functional/testlib.bash
```diff
@@ -1979,8 +1983,6 @@ create_config_file() { # swupd_function
 
 	# these values have to be exported in this function because the
 	# TEST_NAME changes during the setup
```
castulo: I'm confused about moving the setup of the config dir to the top of the file, it was being called from `create_config_file()` because the name of the tests change for each setup, there is even a comment about this here:
```
# these values have to be exported in this function because the
# TEST_NAME changes during the setup
```

test/functional/testlib.bash
```diff
@@ -488,6 +488,12 @@ set_env_variables() { # swupd_function
 	debug_msg "STATEDIR: $STATEDIR"
 	export PATH_PREFIX="$TEST_DIRNAME"/testfs/target-dir
 	debug_msg "PATH_PREFIX: $PATH_PREFIX"
+	export TP_BASE_DIR="$TEST_DIRNAME"/3rd-party
+	debug_msg "TP_BASE_DIR: $TP_BASE_DIR"
+	export TP_BASE_STATEDIR="$STATEDIR"/3rd-party
+	debug_msg "TP_BASE_STATEDIR: $TP_BASE_STATEDIR"
+	export TP_TARGETDIR="$env_name"/testfs/target-dir/"$THIRD_PARTY_BUNDLES_DIR"
+	debug_msg "3rd-party repo target dir: $TP_TARGETDIR"
```
castulo: Again I'm confused about how moving these variables here make them relative to the current test, they may be relative to the current test file, but a test file may have many tests, so the only way of having variables relative to a test is if the variable is set up from within the setup() once we know the name of the test

test/functional/testlib.bash
```diff
@@ -1537,14 +1537,6 @@ create_third_party_repo() { #swupd_function
 		create_bundle -n os-core -v "$version" -f /usr/lib/os-release:"$OS_RELEASE",/usr/share/clear/update-ca/Swupd_Root.pem:"$cert",/usr/share/defaults/swupd/format:"$FORMAT" -u "$repo_name" "$env_name"
 	fi
 
-	if [ "$TEST_ENV_ONLY" = true ]; then
```
castulo: why are you removing this part? I find very useful to use the `create_test_environment_only` function from my terminal to create an environment and then debug a scenario manually within that environment. The info printed here is helpful to set up the debugger.

otaviobp: This variables doesn't exist anymore. I can review again if there's anything left behind, but I am not opposed to export all variables needed to debug that when create_test_environment_only is used.

test/functional/testlib.bash
otaviobp: my bad, it was a mistake

test/functional/testlib.bash
```diff
@@ -500,8 +500,8 @@ set_env_variables() { # swupd_function
 
 	export CLIENT_CERT_DIR="$testfs_path/target-dir/etc/swupd"
 	export CLIENT_CERT="$CLIENT_CERT_DIR/client.pem"
-	export PORT_FILE="$path/$env_name/port_file.txt" # stores web server port
-	export SERVER_PID_FILE="$path/$env_name/pid_file.txt" # stores web server pid
+	export PORT_FILE="$path/$TEST_NAME_SHORT/port_file.txt" # stores web server port
```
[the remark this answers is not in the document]
otaviobp: I can review this again and fix the problem. I'm ok on trying to support that function running directly from the terminal too.

test/functional/testlib.bash
```diff
@@ -1696,6 +1697,25 @@ create_version() { # swupd_function
 
 }
 
+
+# Creates the directory to save test files
+# If it already exists, destroy it.
+# Parameters:
+# - DIRECTORY_NAME: the path of the directory
+create_test_env_directory() { # swupd_function
```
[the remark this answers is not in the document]
otaviobp: ok, I'll review the usage of swupd_functions too

test/functional/testlib.bash
otaviobp: I'm moving everything that is relative to TEST_NAME to set_env_variables to enforce that everything will be relative to the test. Any file that is generated and needed by the test should be in it's own directory.

otaviobp:
Let me try to explain a bit here the overall idea of the major changes in this PR. btw, if when it's good to merge, we need to document that better on testlib doc.

Today files are created for tests in 2 different moments. On global_setup() or test_setup(). And we support just one of those cases. If global_setup() is used all tests are running with the same environment (what's bad) and the result of one test impact the other. If test_setup() is used the environment is recreated for every test. global_setup() and test_setup() are also used to start/stop services that may be needed by the test (like the webserver).

The major proposal of this PR is to create another step that is used to generate any file that would be reused by all tests on the test file in a way that it could be cached. I called this step metadata_setup(). Metadata_setup is called only once for each test file and all generated files are copied to the specific test directory to be reused by the test.

Changes made to accomplish that:
 - Separate the creation of a test directory from the creation of a mix. We may want to use the testlib for tests that doesn't require any mix to be created.
 - Explicit mix creation. Instead of creating an env on test_setup with default parameters, it's up to the user to define how to create the environment. This prevents us to need override test_setup() with an empty body. Directory for the test data is still created and mandatory. It's cheap and it can be used by testlib to store any file needed.
 - Create keys to sign the release on testlib and not in another test. Now we are creating one key for each test file because it's easier to cache.
 - Create a step to be used to create metadata and copy it to every test.

So now, if we have 3 tests in a example.bats file we will have the following workflow:
 - global_setup() #TEST_NAME=example
 - metadata_setup() #TEST_NAME=example_META
 - test_setup() #TEST_NAME=example_001 and everything saved to example_META is copied to example_001 before starting
 - test_teardown() #TEST_NAME=example_001
 - test_setup() #TEST_NAME=example_002 and everything saved to example_META is copied to example_002 before starting
 - test_teardown() #TEST_NAME=example_002
 - test_setup() #TEST_NAME=example_003 and everything saved to example_META is copied to example_003 before starting
 - test_teardown() #TEST_NAME=example_003
 - global_teardown() # TEST_NAME=example

castulo:
> Today files are created for tests in 2 different moments. On global_setup() or test_setup().

Yes, and this the way all other test frameworks work

> And we support just one of those cases. 

What do you mean we only support one of those cases? You can use both if you want in a test at the same time. This would be the workflow of a test file if all setups and teardown are used (assuming we have 2 tests in the file):
```
global setup

test setup
test 1 runs
test teardown

test setup
test 2 runs
test teardown

global teardown
```
And again, this is pretty standard of any test framework.

> If global_setup() is used all tests are running with the same environment (what's bad) and the result of one test impact the other. If test_setup() is used the environment is recreated for every test. 

I disagree with this statement, whatever is done in the global_setup() impacts all the tests, but that is definitely not bad, you only have to choose the right actions to run there (if any). For example in an application that needs the user to authenticate you may want to use the global setup to login to the application that is going to be tested so you don't have to do it for every test. In our case, global setups are really not that useful because of the nature of our application and our tests, most of the times we don't want to use it precisely because we use global_setup and test_setup to create environments that we modify with tests, and we cannot really reuse an environment if it was already modified by another test or we would have the conflict you talk about where one test could impact the result of the next. I still think global_setup is useful for a few of our tests, for example tests that only list things without changing anything, we can create only one environment and reuse it in all tests, that makes the test faster.

> global_setup() and test_setup() are also used to start/stop services that may be needed by the test (like the webserver).

> The major proposal of this PR is to create another step that is used to generate any file that would be reused by all tests on the test file in a way that it could be cached. I called this step metadata_setup(). Metadata_setup is called only once for each test file and all generated files are copied to the specific test directory to be reused by the test.

I'm all about saving execution time by caching stuff that we create over and over, specially since the creation of that test data consumes most of the time of the test. I think it is a good idea. I just want us to be careful about it so we don't impact the experience of using the test library when run locally and interactively. I'm not sure how you use the test library, but I use it a lot interactively for debugging. Also I wouldn't want the library to divert from what is expected from a test library, for example as I said a "global setup" is pretty standard in any test framework, so I would expect it to behave similarly to a global setup in other frameworks.

> Changes made to accomplish that:
> 
> * Separate the creation of a test directory from the creation of a mix. We may want to use the testlib for tests that doesn't require any mix to be created.
> * Explicit mix creation. Instead of creating an env on test_setup with default parameters, it's up to the user to define how to create the environment. This prevents us to need override test_setup() with an empty body. Directory for the test data is still created and mandatory. It's cheap and it can be used by testlib to store any file needed.

I'm ok with this, in fact, I also prefer it this way. I ended up adding the default actions in `test_setup` and `test_teardown` as a result of a direct request when folks were reviewing the implementation of the test library.

> * Create keys to sign the release on testlib and not in another test. Now we are creating one key for each test file because it's easier to cache.
> * Create a step to be used to create metadata and copy it to every test.



> So now, if we have 3 tests in a example.bats file we will have the following workflow:
> 
> * global_setup() #TEST_NAME=example
> * metadata_setup() #TEST_NAME=example_META
> * test_setup() #TEST_NAME=example_001 and everything saved to example_META is copied to example_001 before starting
> * test_teardown() #TEST_NAME=example_001
> * test_setup() #TEST_NAME=example_002 and everything saved to example_META is copied to example_002 before starting
> * test_teardown() #TEST_NAME=example_002
> * test_setup() #TEST_NAME=example_003 and everything saved to example_META is copied to example_003 before starting
> * test_teardown() #TEST_NAME=example_003
> * global_teardown() # TEST_NAME=example

I'm a little confused in this workflow. If you are creating metadata_setup between global setup and test setup... does that mean that metadata will only contain stuff that otherwise would have been created during the global setup?

otaviobp:
As there are many topics here that we are not on the same page, I'll open issues and pull request for the points that it looks like there is an agreement and we can discuss them on a different PR. Most of the changes here are bug fixes or enhancements on the workflow that were needed by my proposed change to cache metadata files. I think most of those changes are possible to be made on different PRs, so I'll also open tickets for them.

I'll try to answer here though some of your questions.

> And we support just one of those cases.

I read again what I wrote and it's hard to understand. I'll try to explain my point here with examples. Maybe it's not really a missing feature, it's just something I didn't know how to do.

For example, if I use global_setup() to create my environment I have changes persisted from test to test (https://gist.github.com/otaviobp/18c7ccc01720864780c01ed6e0f4c28b), as I would expect. If I use test_setup (https://gist.github.com/otaviobp/8f1c0a344701a2859adc3542ad3ee9d4), changes are not persisted, as I would expect. What I don't know how to do and I think it is a testlib limitation is to do both in the same test. I tried it here (https://gist.github.com/otaviobp/3d8fe8d5bf75c57b55e954b063389f72). If there's a way to do it, let me know. 

When I said that the user needs to choose one, I meant that the user needs to choose to create the mix for global or for local.

> I'm a little confused in this workflow. If you are creating metadata_setup between global setup and test setup... does that mean that metadata will only contain stuff that otherwise would have been created during the global setup?

metadata will contain only things generated in metadata_setup(). nothing from global_setup() or test_setup(). metadata_setup() is a new step to create/consume any data that you want to cache/restore in multiple executions. Maybe cache_setup() would be a better name

PR #1387, opened 2020-03-19
Enforcing signature verification even when signature is missing

otaviobp:
Also adding more tests and other fixes

### DIFF
--- a/docs/swupd.1.rst
+++ b/docs/swupd.1.rst
@@ -124,6 +124,10 @@ used to modify the core behavior and resources that swupd uses.
 
    Do not attempt to enforce certificate or signature checking.
 
+-  ``-n, --nosigcheck-latest``
+
+   Do not attempt to enforce signature checking when retrieving the latest version number.
+
 -  ``-I, --ignore-time``
 
    Ignore system/certificate time when validating signature.
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -1240,3 +1240,23 @@ void free_and_clear_pointer(char **s)
 		*s = NULL;
 	}
 }
+
+void warn_nosigcheck(const char *file)
+{
+	static bool first_time = true;
+
+	if (first_time) {
+		const char *flag = globals.sigcheck ? "--nosigcheck-latest" : "--nosigcheck";
+		char journal_msg[LINE_MAX] = { 0 };
+		warn("The %s flag was used and this compromises the system security\n", flag);
+
+		snprintf(journal_msg, sizeof(journal_msg), "swupd security notice: %s used to bypass MoM signature verification", flag);
+
+		journal_log_error(journal_msg);
+
+		first_time = false;
+	}
+
+	warn("\n");
+	warn("THE SIGNATURE OF %s WILL NOT BE VERIFIED\n\n", file);
+}
--- a/test/functional/signature/version-sig-check.bats
+++ b/test/functional/signature/version-sig-check.bats
@@ -16,16 +16,13 @@ test_setup() {
 
 @test "SIG019: Verify signature for absolute latest version with swupd check-update" {
 
-	#TODO: Remove on FB 30
-	skip "Skipping until FB 30 - feature disabled"
 	# absolute latest version get checked during check-update
 	sudo sh -c "mv $WEBDIR/version/latest_version.sig $WEBDIR/version/latest_version_bad_name.sig"
 	run sudo sh -c "$SWUPD check-update $SWUPD_OPTS"
 	assert_status_is "$SWUPD_SIGNATURE_VERIFICATION_FAILED"
 
 	expected_output=$(cat <<-EOM
-		Error: Failed to retrieve size for signature file: file://$TEST_DIRNAME/web-dir/version/latest_version.sig
-		Error: Signature Verification failed for URL: file://$TEST_DIRNAME/web-dir/version/latest_version
+		Error: Signature for latest file (file://$TEST_DIRNAME/web-dir/version/latest_version) is missing
 		Error: Unable to determine the server version as signature verification failed
 		Current OS version: 10
 	EOM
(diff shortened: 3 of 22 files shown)


otaviobp:
@bryteise After this PR signature verification for latest file is going to be completely enforced. If signature is missing or incorrect, any swupd operation will be aborted. If user wants to proceed there's a new flag --nosigcheck-latest that will ignore latest signature, but continue to check Manifests, so it's safer than --nosigcheck.

On an update, if there is a mirror set we always check if the mirror is out of sync with upstream. In the case where signature for mirror is invalid or missing, we consider the mirror out of sync and update will proceed using the default url (upstream).

Does this approach looks good to you? This code is still aborting in the case where signature is expired. This fix will come next.

docs/swupd.1.rst
```diff
@@ -124,6 +124,10 @@ used to modify the core behavior and resources that swupd uses.
 
    Do not attempt to enforce certificate or signature checking.
 
+-  ``-n, --nosigcheck-latest``
+
+   Do not attempt to enforce certificate or signature checking on latest file.
```
castulo: don't you think we need to add more context here to what the "latest" file is? I'm thinking this will make no sense to the common user.

test/functional/signature/version-sig-check.bats
```diff
@@ -16,16 +16,14 @@ test_setup() {
...
 	# absolute latest version get checked during check-update
 	sudo sh -c "mv $WEBDIR/version/latest_version.sig $WEBDIR/version/latest_version_bad_name.sig"
 	run sudo sh -c "$SWUPD check-update $SWUPD_OPTS"
 	assert_status_is "$SWUPD_SIGNATURE_VERIFICATION_FAILED"
 
 	expected_output=$(cat <<-EOM
-		Error: Failed to retrieve size for signature file: file://$TEST_DIRNAME/web-dir/version/latest_version.sig
-		Error: Signature Verification failed for URL: file://$TEST_DIRNAME/web-dir/version/latest_version
+		Error: Signature for latest file (file://$TEST_DIRNAME/web-dir/version/latest_version) is missing
+		Error: Signature verification failed for URL: file://$TEST_DIRNAME/web-dir/version/latest_version
```
castulo: I don't love this output, it seems very redundant, we basically have the same error 3 times with different wording.

test/functional/signature/version-sig-check.bats
```diff
@@ -65,4 +63,104 @@ test_setup() {
...
+
+	expected_output=$(cat <<-EOM
+		Update started
+		Warning: The --nosigcheck flag was used, THE SIGNATURE OF file://$TEST_DIRNAME/web-dir/version/formatstaging/latest WILL NOT BE VERIFIED
+		Be aware that this compromises the system security
+		Preparing to update from 10 to 20
+		Warning: The --nosigcheck flag was used, THE MANIFEST SIGNATURE WILL NOT BE VERIFIED
+		Be aware that this compromises the system security
+		Warning: The --nosigcheck flag was used, THE MANIFEST SIGNATURE WILL NOT BE VERIFIED
+		Be aware that this compromises the system security
```
castulo: same here, we have the same warning 2 times 😩

test/functional/testlib.bash
```diff
@@ -501,6 +501,8 @@ set_env_variables() { # swupd_function
 	export CACERT_DIR="$SWUPD_DIR/swupd_test_certificates" # trusted key store path
 	export PORT_FILE="$path/$env_name/port_file.txt" # stores web server port
 	export SERVER_PID_FILE="$path/$env_name/pid_file.txt" # stores web server pid
+	SWUPD_DIR_ABS="$(realpath "$SWUPD_DIR")"
+	export SWUPD_DIR_ABS
```
castulo: if `SWUPD_DIR_ABS` is not dependent of a specific test environment I think it should be exported with the other independent variables at the top of the file
```
# global variables
FUNC_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
export FUNC_DIR
TEST_ROOT_DIR="$(pwd)"
export TEST_ROOT_DIR
TEST_FILENAME=$(basename "$BATS_TEST_FILENAME")
export TEST_FILENAME
export TEST_NAME=${TEST_FILENAME%.bats}
export THEME_DIRNAME="$BATS_TEST_DIRNAME"
export TEST_NAME_SHORT="$TEST_NAME"
export SWUPD_DIR="$FUNC_DIR/../.."
SWUPD_DIR_ABS="$(realpath "$SWUPD_DIR")"
export SWUPD_DIR_ABS
```

otaviobp:
V2
 - Fixing all reported problems
 - Adding another commit changing phrasing of --nosigcheck

src/helpers.c
```diff
@@ -1240,3 +1240,23 @@ void free_and_clear_pointer(char **s)
...
 }
+
+void warn_nosigcheck(const char *file)
+{
+	static bool first_time = true;
+
+	if (first_time) {
+		const char *flag = globals.sigcheck ? "--nosigcheck-latest" : "--nosigcheck";
+		char journal_msg[LINE_MAX] = { 0 };
+		warn("The %s flag was used and this will compromises the system security\n", flag);
```
castulo: Small typo here `compromises` -> `compromise`

src/helpers.c
```diff
@@ -1240,3 +1240,23 @@ void free_and_clear_pointer(char **s)
...
+
+		snprintf(journal_msg, sizeof(journal_msg), "swupd security notice: %s used to bypass MoM signature verification", flag);
```
castulo: this message doesn't have a `\n` at the end, is that expected?

otaviobp: Yes, this is correct

otaviobp:
V3: Fixing comment - s/will compromises/compromises/

PR #1389, opened 2020-03-20
Remove extra slash from path in 3rd-party funcs

castulo:

### DIFF
--- a/src/3rd_party_repos.c
+++ b/src/3rd_party_repos.c
@@ -34,27 +34,27 @@
 
 char *third_party_get_bin_dir(void)
 {
-	return sys_path_join(globals_bkp.path_prefix, SWUPD_3RD_PARTY_BIN_DIR);
+	return sys_path_join("%s/%s", globals_bkp.path_prefix, SWUPD_3RD_PARTY_BIN_DIR);
 }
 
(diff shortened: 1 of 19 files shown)


castulo:
@otaviobp take a look at this PR and let me know what you think. I'm unsure about two things:
1) the name of the function `sys_path`
2) if the function should live in sys.c or in helpers.c

src/3rd_party_repos.c
```diff
@@ -34,27 +34,27 @@
 
 char *third_party_get_bin_dir(void)
 {
-	return sys_path_join(globals_bkp.path_prefix, SWUPD_3RD_PARTY_BIN_DIR);
+	return sys_path("%s%s", globals_bkp.path_prefix, SWUPD_3RD_PARTY_BIN_DIR);
```
otaviobp: I would prefer to always use "%s/%s" to make sure we will never end up merging 2 paths without separators. sys_path is going to remove that anyway, so it's fine.

castulo: good point

castulo:
V2, addressed all the comments from review:
- replaced the current `sys_path_join` function with the new implementation that can join many paths
- fixed the arguments of the function everywhere where the function was already being used to match the new arguments

PR #1404, opened 2020-03-27
actions: cache dependencies for swupd

otaviobp:
Some packages we need to build swupd are not available on ubunutu-latest
so we need to download and build from source. This process is slow and
caching can save us some significant time on build.

### DIFF
--- a/scripts/github_actions/build_ci_dependencies.bash
+++ b/scripts/github_actions/build_ci_dependencies.bash
@@ -1,0 +1,46 @@
+#!/bin/bash
+# Build all custom dependencies needed by swupd:
+# - Curl
+# - Bsdiff
+# - libarchive
+#
+# As process is slow, it can be cached. If directory dependencies exists,
+# just install already build dependencies.
+
+set -e
+CORES=2
+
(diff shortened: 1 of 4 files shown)


scripts/github_actions/build_ci_dependencies.bash
```diff
@@ -0,0 +1,46 @@
...
+	exit
+fi
+
+mkdir dependencies
+pushd dependencies
+
+# build bsdiff
+wget https://github.com/clearlinux/bsdiff/releases/download/v1.0.2/bsdiff-1.0.2.tar.xz
+tar -xvf bsdiff-1.0.2.tar.xz
+pushd bsdiff-1.0.2 && ./configure --prefix=/usr --disable-tests && make -j$CORES && sudo make install && popd
```
castulo: missing "" around j$CORES, are you running shellcheck against this script too?

otaviobp: shellcheck doesn't  complain. I think it knows the value is hardcoded

otaviobp:
v2: Added \" around variable

PR #1406, opened 2020-03-27
Remove extra slash from paths

castulo:
This commit removes the extra slash '/' from the path from:
- the update_boot function
- the verifytime script

### DIFF
--- a/src/verifytime.c
+++ b/src/verifytime.c
@@ -24,6 +24,7 @@
 
 #include "lib/log.h"
 #include <errno.h>
+#include <stdarg.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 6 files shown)


src/verifytime.c
```diff
@@ -37,9 +37,27 @@ static unsigned long int get_versionstamp(char *path_prefix)
 	FILE *fp = NULL;
 	char data[11];
 	char *filename;
+	char *prefix = NULL;
 	unsigned long int version_num;
+	int len;
 
-	if (asprintf(&filename, "%s/usr/share/clear/versionstamp", path_prefix ? path_prefix : "") < 0) {
+	/* if the path_prefix has a trailing slash on it, remove it */
+	if (path_prefix) {
```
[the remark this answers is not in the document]
castulo: is there a reason no do it? it annoys me to see multiple separators in the output, IMO it makes the tool look low quality.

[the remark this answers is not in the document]
otaviobp: ok, but we need to have same behavior here than we use in sys_path_join(). Right now we are keeping duplicated slashes that are present on path_prefix. As we are limiting the linking on verify_time, we can just copy-and-paste sys_path_join() with appropriate changes to this file.

castulo:
V2:
- copied `sys_path_join()` to verifytime.c to fix the path of he versionstamp file.

src/verifytime.h
```diff
@@ -12,6 +12,11 @@
 extern "C" {
 #endif
 
+/**
+ * @brief Character used as PATH separator.
+ */
+#define PATH_SEPARATOR '/'
```
otaviobp: this shouldn't be exported. define in the .c file

src/verifytime.c
```diff
@@ -32,17 +33,50 @@
 
 #define DAY_SECONDS 86400
 
+static char *path_join(const char *fmt, ...)
```
otaviobp: add a comment here saying that this is a copy.

castulo: ok

src/verifytime.h
castulo: I can do that. But I'd like to understand this better, can you please explain me why this should not be exported?

castulo:
V3:
- defining `PATH_SEPARATOR` in verifytime.c
- added comment showing the `path_join()` is a copy of `sys_path_join()`

src/verifytime.h
otaviobp: On header files we export everything that we want others to access and use. Things that we don't want others to use, we should keep only in the C file. As this is a header for a library to verify time, I don't see a reason on exporting a path separator.

Besides that we already export a constant with the same name in other header file. In this case the value is also identical, but if for some reason it is changed we would have files that import both headers with different values for the same constant.

castulo: I see. Thanks for explaining that :)

PR #1408, opened 2020-03-30
Bug_fix: Do not follow symlinks when removing files during staging or update can end up removing incorrect files

### LINKED ISSUE #890: Update can remove files if a directory is replaced by a symlink
**Describe the bug**
If a directory is replaced by a symlink in an update, i.e. it was a directory in v10 and it's now a symlink in v20, staging process will remove all files from that directory after the symlink is created. This could lead to the removal of files under the directory that the symlink is pointing too.

**To Reproduce**

    $ sudo swupd verify --install -B c-basic --path /custom_path -m 28760 -F 27
    $ sudo swupd update --path /custom_path -m 28780
    $ ls -la  /custom_path/usr/include/c++/x86_64-generic-linux/bits #directory not found
    $ sudo swupd verify --path /custom_path #several files are missing

As /usr/include/c++/x86_64-generic-linux/32/ was replaced by a symlink to /usr/include/c++/x86_64-generic-linux/, files under /usr/include/c++/x86_64-generic-linux/ are deleted in this update.

**Expected behavior**

We should see files on /usr/include/c++/x86_64-generic-linux/ and no file should go missing on a swupd verify

 - Clear Linux OS Version: update from 28760 to 28780
 - Platform: any

castulo:
When during an update a directory was removed and replaced by a symlink to another directory that contains one or more files with the same name  as files in the directory that was removed, swupd will remove the directory from the target system and then it will try to remove each one     of the files that were contained in the directory. Since the directory is now a symlink to another existing directory, it may end up deleting the files from the other directory. swupd should not follow links when removing files during the staging process.

Closes #890

Note: for the tests in this PR to work #1380 needs to be reviewed and merged first.

### DIFF
--- a/src/lib/sys.c
+++ b/src/lib/sys.c
@@ -286,6 +286,29 @@ bool sys_file_exists(const char *filename)
 	return false;
 }
 
+bool sys_path_is_absolute(const char *filename)
+{
+	char *abspath = NULL;
+	bool ret = false;
+
+	abspath = realpath(filename, NULL);
+	if (!abspath) {
+		if (errno != ENOENT) {
+			error("Failed to get absolute path of %s (%s)\n", filename, strerror(errno));
+		}
+		ret = false;
+		goto exit;
+	}
+
+	if (strcmp(abspath, filename) == 0) {
+		ret = true;
+	}
+
+exit:
+	free(abspath);
+	return ret;
+}
+
 bool sys_filelink_exists(const char *filename)
 {
 	struct stat sb;
--- a/src/lib/sys.h
+++ b/src/lib/sys.h
@@ -105,6 +105,11 @@ bool sys_file_exists(const char *filename);
 bool sys_filelink_exists(const char *filename);
 
 /**
+ * @brief Checks if all elements of a path are absolute and don't contain symlinks
+ */
+bool sys_path_is_absolute(const char *filename);
+
+/**
  * @brief Checks if a file exists in the filesystem and is executable.
  * Follows symlinks
  */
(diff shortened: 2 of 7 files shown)


src/lib/sys.c
```diff
@@ -286,6 +286,22 @@ bool sys_file_exists(const char *filename)
 	return false;
 }
 
+bool sys_absolute_path_exists(const char *filename)
```
otaviobp: I think that the most appropriate name for this function would be `sys_path_is_absolute()` because it's not checking only if the path exists.

src/lib/sys.c
```diff
@@ -286,6 +286,22 @@ bool sys_file_exists(const char *filename)
...
+	char *abspath = NULL;
+	bool ret = false;
+
+	abspath = realpath(filename, NULL);
+	if (strcmp(abspath, filename) == 0) {
+		ret = true;
+	}
+
+	if (abspath) {
+		free(abspath);
```
otaviobp: free(NULL) works fine. You don't need to check for NULL here.

src/lib/sys.c
castulo: nice, I was unsure about the name of the function and I was going to ak you for your opinion. `sys_path_is_absolute()` sounds good to me.

src/lib/sys.c
castulo: cool, thanks for the tip.

castulo:
> Also this fix is independent from PR #1380, so you don't need to carry all it's commits

The problem is that the tests for this fix do need PR #1380

castulo:
V2:
- Changed the name of the function to `sys_path_is_absolute()`
- Fixed an error in the code caused by having an extra "/" in the path

otaviobp:
Can you rebase this?

castulo:
Rebased

castulo:
Rebased again

PR #1417, opened 2020-04-01
Strings: some improvements in string usage on swupd

otaviobp:


src/lib/strings.h
```diff
@@ -6,12 +6,19 @@
...
+#include <stdint.h>
+
 #include "list.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+/**
+ * helper to use strnlen() instead of string_len.
```
castulo, line 18: I think you meant `strlen` in the comment, not `string_len`

test/code_analysis/warning_functions.bats
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bats
+
+# Author: Otavio Pontes
+# Email: otavio.pontes@intel.com
+@test "ANL004: Functions to avoid" {
+
+	# There are some C functions that are tricky to use and it's prefered to
+	# avoid using them in swupd. If any function in this list is really needed
+	# we need to review the usage and add an exception here.
```
castulo, line 9: Good idea 👌

otaviobp:
V2: Removing 'a' and fixing strlen comment

PR #1426, opened 2020-04-03
3rd-party: Regenerate wrapper on template change

### LINKED ISSUE #1413: 3rd-party: Regenerate binary wrapper on updates and swupd repair
We need to have a way to regenerate the wrappers on swupd 3rd-party. First we need to regenerate all wrappers on any update where the binary is updated.
Regenerating on update won't apply any changes we made on swupd SCRIPT_TEMPLATE though. We need to have a way to force swupd to regenerate all wrappers.

 - [x] swupd update should regenerate the wrapper if binary has changed
 - [x] swupd repair should regenerate all wrappers of binaries that it is handling, even if binary is fine.
 - [ ] Wrappers are generated using the variable SCRIPT_TEMPLATE on src/3rd_party_repos.c. This variable can change every time swupd is updated and when this happens, we need to regenerate the wrappers.

castulo:
When the template used for creating the wrapper scripts changes, all
exported binaries should have the scripts regenerated during updates
regardless of if the binary changed.

Closes #1413

### DIFF
--- a/src/3rd_party_repos.c
+++ b/src/3rd_party_repos.c
@@ -27,14 +27,6 @@
 
 #ifdef THIRDPARTY
 
-#define SCRIPT_TEMPLATE "#!/bin/bash\n\n"                              \
-			"export PATH=%s:$PATH\n"                       \
-			"export LD_LIBRARY_PATH=%s:$LD_LIBRARY_PATH\n" \
-			"export XDG_DATA_DIRS=%s:$XDG_DATA_DIRS\n"     \
-			"export XDG_CONF_DIRS=%s:$XDG_CONF_DIRS\n"     \
-			"\n"                                           \
-			"%s \"$@\"\n"
-
 char *third_party_get_bin_dir(void)
 {
 	return sys_path_join("%s/%s", globals_bkp.path_prefix, SWUPD_3RD_PARTY_BIN_DIR);
@@ -294,7 +286,7 @@ enum swupd_code third_party_set_repo(struct repo *repo, bool sigcheck)
 
 	/* set up swupd to use the certificate from the 3rd-party repository,
 	 * unless the user is specifying a path for the certificate to use */
-	string_or_die(&repo_cert_path, "%s%s/%s%s", globals_bkp.path_prefix, SWUPD_3RD_PARTY_BUNDLES_DIR, repo->name, CERT_PATH);
+	repo_cert_path = sys_path_join("%s/%s/%s/%s", globals_bkp.path_prefix, SWUPD_3RD_PARTY_BUNDLES_DIR, repo->name, CERT_PATH);
 	if (!globals.user_defined_cert_path) {
 		/* the user did not specify a cert, use repo's default */
 		set_cert_path(repo_cert_path);
@@ -615,9 +607,33 @@ enum swupd_code third_party_remove_binary(struct file *file)
 	return ret_code;
 }
 
+static enum swupd_code third_party_bin_directory_exist(void)
+{
+	enum swupd_code ret_code = SWUPD_OK;
+	char *bin_directory = NULL;
+
+	bin_directory = third_party_get_bin_dir();
+
+	/* if the SWUPD_3RD_PARTY_BIN_DIR does not exist, attempt to create it */
+	if (mkdir_p(bin_directory)) {
+		error("The directory %s for 3rd-party content failed to be created\n", bin_directory);
+		ret_code = SWUPD_COULDNT_CREATE_DIR;
+		goto exit;
+	}
+
+	if (!sys_filelink_is_dir(bin_directory)) {
+		error("The path %s for 3rd-party content exists but is not a directory\n", bin_directory);
+		ret_code = SWUPD_COULDNT_CREATE_DIR;
+	}
+
+exit:
+	free_and_clear_pointer(&bin_directory);
+	return ret_code;
+}
+
 enum swupd_code third_party_create_wrapper_script(struct file *file)
 {
-	enum swupd_code ret_code = 0;
+	enum swupd_code ret_code = SWUPD_OK;
 	int fd;
 	FILE *fp = NULL;
 	char *bin_directory = NULL;
@@ -652,15 +668,9 @@ enum swupd_code third_party_create_wrapper_script(struct file *file)
 		goto close_and_exit;
 	}
 
-	/* if the SWUPD_3RD_PARTY_BIN_DIR does not exist, attempt to create it */
-	if (mkdir_p(bin_directory)) {
-		ret_code = SWUPD_COULDNT_CREATE_DIR;
-		goto close_and_exit;
-	}
-
-	if (!sys_filelink_is_dir(bin_directory)) {
-		error("The path %s for 3rd-party content exists but is not a directory\n", bin_directory);
-		ret_code = SWUPD_COULDNT_CREATE_DIR;
+	/* make sure the SWUPD_3RD_PARTY_BIN_DIR exist */
+	ret_code = third_party_bin_directory_exist();
+	if (ret_code) {
 		goto close_and_exit;
 	}
 
--- a/src/3rd_party_update.c
+++ b/src/3rd_party_update.c
@@ -195,6 +195,43 @@ static enum swupd_code validate_file_permissions(struct list *files_to_be_update
 	return ret_code;
 }
 
+static enum swupd_code regenerate_all_wrapper_scripts(UNUSED_PARAM char *unused)
+{
+	enum swupd_code ret_code;
+	struct list *current_subs = NULL;
+	struct manifest *current_mom = NULL;
+	int version;
+
+	/* get currently installed 3rd-party bundles */
+	read_subscriptions(&current_subs);
+
+	/* load the MoM */
+	version = get_current_version(globals.path_prefix);
+	current_mom = load_mom(version, NULL);
+	if (!current_mom) {
+		ret_code = SWUPD_COULDNT_LOAD_MOM;
+		goto exit;
+	}
+
+	/* get a list of all 3rd-party files installed */
+	current_mom->submanifests = recurse_manifest(current_mom, current_subs, NULL, false, NULL);
+	if (!current_mom->submanifests) {
+		ret_code = SWUPD_RECURSE_MANIFEST;
+		goto exit;
+	}
+	current_mom->files = consolidate_files_from_bundles(current_mom->submanifests);
+
+	ret_code = third_party_process_files(current_mom->files, "Regenerating scripts...\n", "update_binaries", third_party_update_wrapper_script);
+	if (ret_code == SWUPD_OK) {
+		info("Scripts regenerated successfully\n");
+	}
+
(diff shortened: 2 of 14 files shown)


src/helpers.c
```diff
@@ -1289,3 +1289,66 @@ void warn_mixin_deprecation(void)
...
+	ret = fprintf(fp, "%s", content);
+	if (ret < 0) {
+		error("There was an error writing to file %s\n", file);
+	}
+
+	fclose(fp);
+	return ret;
+}
+
+char *read_file_into_memory(char *file)
```
otaviobp, line 1312: Use mmap for that, like on signature.c

src/helpers.c
```diff
@@ -1289,3 +1289,66 @@ void warn_mixin_deprecation(void)
 		show = false;
 	}
 }
+
+int write_into_file(char *file, char *content)
```
otaviobp, line 1293: use fwrite() for that.

src/3rd_party_update.c
```diff
@@ -260,7 +300,43 @@ enum swupd_code third_party_update_main(int argc, char **argv)
...
+	template = read_file_into_memory(template_file);
+
+	if (!template || (template && strcmp(template, SCRIPT_TEMPLATE) != 0)) {
+		/* there is no template file, or the template changed,
+		 * all scripts need to be recreated */
+		info("The scripts that export binaries from 3rd-party repositories need to be regenerated\n\n");
+		ret_code = third_party_run_operation_multirepo(NULL, regenerate_all_wrapper_scripts, SWUPD_OK, "regenerate_scripts", steps_in_update);
+
+		/* update the template */
+		ret = sys_rm(template_file);
```
otaviobp, line 318: you don't need to remove. Just update it.

src/3rd_party_update.c
```diff
@@ -260,7 +300,43 @@ enum swupd_code third_party_update_main(int argc, char **argv)
 
 	/* update 3rd-party bundles */
 	ret_code = third_party_run_operation_multirepo(cmdline_option_repo, update_repos, SWUPD_NO, "update", steps_in_update);
+	if (ret_code) {
+		goto exit;
+	}
+
+	/* read the current template copy */
+	template_file = third_party_get_binary_path(SWUPD_3RD_PARTY_TEMPLATE_FILE);
```
bryteise, line 308: I don't think the template should be in the binary path since that poses name collision issues.

Also given the template is distro content and shouldn't need to be written to (outside of doing a system update) couldn't the file just be in /usr/share/swupd? I think we had talked about this a little before so excuse my memory if you had explained why that kind of thing would be problematic X(.

castulo: I'm ok with storing it in `/usr/share/swupd` if you think it is more appropriate for it to be there. There is currently no `/usr/share/swupd` directory, so if we store it there it would be the only file there.

bryteise: Ah sorry, /usr/share/defaults/swupd is the folder in use.

castulo: 👌

otaviobp: That's true, we shouldn't save it in /opt/3rd-party/bin.
But I think there's a confusion on what this file is. This is not the template that will be used to generate the wrappers. That is still a const in code. This is a file to realize if the template has changed or not. So it shouldn't be installed by clear and we can't install that on /usr/.
I think the best place to put that is in /opt/3rd-party.

Now regarding having the template to be used to create wrappers on /usr, @bryteise do you think it's important to have that? If so, it's independent from this PR.

bryteise: @otaviobp Yea you are right, the file is being used for state at the moment. It isn't content but state data. In that case I'd probably keep it in /var/lib/swupd but I don't remember at the moment if you decided to keep state there or in /opt/3rd-party. That said, I think my confusion stems from seeing it as content and moving it to /usr and taking mtime as the mechanism for deciding when to regenerate the wrappers seems like a reasonable thing to do.

castulo:
V2: comments from review were addressed

bryteise:
LGTM

src/helpers.c
```diff
@@ -1250,3 +1251,63 @@ void warn_nosigcheck(const char *file)
 	warn("\n");
 	warn("THE SIGNATURE OF %s WILL NOT BE VERIFIED\n\n", file);
 }
+
+int write_into_file(char *file, char *content, char *mode)
```
otaviobp, line 1255: We have several file operations under sys. I think you can put this there

src/helpers.c
```diff
@@ -1250,3 +1251,63 @@ void warn_nosigcheck(const char *file)
...
+	if (fwrite(content, sizeof(char), sizeof(content), fp) <= 0) {
+		ret = -1;
+		error("There was an error writing to file %s\n", file);
+	}
+
+	fclose(fp);
+	return ret;
+}
+
+int map_file_into_memory(const char *file, unsigned char **buffer, size_t *file_length)
```
otaviobp, line 1274: I would move the map functions to sys too.

src/3rd_party_update.c
```diff
@@ -260,7 +301,40 @@ enum swupd_code third_party_update_main(int argc, char **argv)
...
+		info("The scripts that export binaries from 3rd-party repositories need to be regenerated\n\n");
+		ret_code = third_party_run_operation_multirepo(NULL, regenerate_all_wrapper_scripts, SWUPD_OK, "regenerate_scripts", steps_in_update);
+
+		/* update the template */
+		ret_code = third_party_bin_directory_exist();
+		if (ret_code) {
+			goto exit;
+		}
+
+		ret = write_into_file(template_file, SCRIPT_TEMPLATE, "w");
```
otaviobp, line 328: If you write the file here the first update will regenerate wrappers even if not needed
```
$ Swupd 3rd-party add ...
$ swupd 3rd-party bundle-add ...
# there's no template
$ swupd 3rd-party update # First time you run update you always regenerate.
```

src/helpers.c
otaviobp: Why unsigned char?

otaviobp:
It's worth adding a test to check if the template was update correctly too. There's one test generating an incorrect template/

src/helpers.c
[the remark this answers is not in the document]
castulo: > What are the use cases for mode here? I think you can hardcode it to "w". Not sure if we are ever going to need "a".

If you don't think we will need the "append" I can hardcode to "w", and if we need the append at some point then we can add it then.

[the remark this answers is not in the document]
castulo: > And I think you need the content size here. I think it's better than using strlen because not necessary the file is a text file.

I don't understand this comment, I'm not using strlen

src/helpers.c
```diff
@@ -1250,3 +1251,63 @@ void warn_nosigcheck(const char *file)
...
+{
+	FILE *fp = NULL;
+	int ret = 0;
+
+	fp = fopen(file, mode);
+	if (!fp) {
+		return -errno;
+	}
+
+	if (fwrite(content, sizeof(char), sizeof(content), fp) <= 0) {
```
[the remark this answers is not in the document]
castulo: I'll take a look

src/helpers.c
castulo: ok

src/helpers.c
castulo: > Why unsigned char?

that is a good question, it was like that in https://github.com/clearlinux/swupd-client/blob/master/src/signature.c#L270, so I just copied it like that, I didn't understand the reason but decided to let it that way

src/3rd_party_update.c
castulo: if there is no template the first update will regenerate the wrappers, I think this behavior is correct, the wrapper don't need to be regenerated at that point but I don't see the harm on it either, it happens only once. What are you suggesting?

src/helpers.c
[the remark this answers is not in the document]
otaviobp: you are using sizeof() what's not working. I'm suggesting you not to fix it by using strlen()

src/3rd_party_update.c
otaviobp: that's true. it's just in the first update. it's fine then

src/helpers.c
otaviobp: You can use char here then

castulo:
V3: addressed all new comments in the review.

src/3rd_party_repos.c
```diff
@@ -615,9 +607,33 @@ enum swupd_code third_party_remove_binary(struct file *file)
 	return ret_code;
 }
 
+enum swupd_code third_party_bin_directory_exist(void)
```
otaviobp, line 610: I would keep this function static

src/3rd_party_update.c
```diff
@@ -260,7 +301,32 @@ enum swupd_code third_party_update_main(int argc, char **argv)
...
 	ret_code = third_party_run_operation_multirepo(cmdline_option_repo, update_repos, SWUPD_NO, "update", steps_in_update);
+	if (ret_code) {
+		goto exit;
+	}
+
+	/* read the current template copy */
+	template_file = sys_path_join("%s/%s/%s", globals_bkp.path_prefix, SWUPD_3RD_PARTY_DIR, SWUPD_3RD_PARTY_TEMPLATE_FILE);
+	template = sys_mmap_file(template_file, &template_len);
+
+	if (!template || strcmp(template, SCRIPT_TEMPLATE) != 0) {
```
otaviobp, line 312: mmap return is not guaranteed to be NULL terminated. Change to: `strncmp(template, SCRIPT_TEMPLATE, template_len);`

otaviobp: With current implementation of sys_mmap_file this if will be true even on errors because template won't be NULL on errors.

src/lib/sys.c
```diff
@@ -613,3 +614,60 @@ int sys_dir_is_empty(const char *path)
...
+
+	if (fstat(fd, &st) != 0) {
+		debug("Failed to stat %s file", file);
+		goto error;
+	}
+	*file_length = st.st_size;
+
+	buffer = mmap(NULL, *file_length, PROT_READ, MAP_PRIVATE, fd, 0);
+	if (buffer == MAP_FAILED) {
+		debug("Failed to mmap %s content", file);
```
otaviobp, line 657: In this case your function will return MAP_FAILED and !template will be false, in the code bellow. I would change this function to return NULL on errors.

src/3rd_party_repos.c
castulo: good idea, this function is no longer being used anywhere outside that file

src/lib/sys.c
castulo: ahh, I missed that, true!

castulo:
V4, rebased and addressed new issues:
- used `strncmp` to compare the template
- returning NULL when `sys_mmap_file` encounter errors

castulo:
> looks good now :)

thanks for the reviews, you caught many things I overlooked 👏

PR #1429, opened 2020-04-06
Changing the way we create and destroy environments

otaviobp:


test/functional/testlib.bash
```diff
@@ -4242,7 +4242,6 @@ global_teardown() {
 test_setup() {
 
 	debug_msg "No test_setup was defined, using the default one..."
```
castulo, line 4250: This message has to be changed now, we are not using a "default" environment any more.

test/functional/testlib.bash
```diff
@@ -4218,6 +4218,12 @@ teardown() {
 		global_teardown
 		debug_msg "Global teardown finished\\n"
 
+		destroy_test_environment "$TEST_NAME"
```
castulo, line 4221: If there is no test environment, the `destroy_test_environment()` function is going to return 1, isn't that going to cause the script to exit there? Maybe we need to add something like:
`destroy_test_environment "$TEST_NAME" || true` to avoid that?

otaviobp: no. If the test doesn't exist destroy_test_environment returns 0.
If the directory exists and it is not a valid test env, it returns 1 and it fails the test. I think that this is expected and it's like it was before.

otaviobp:
V2: Rebasing and fixing comment.

test/functional/testlib.bash
castulo: The function returns 1 when the env doesn't exist:
https://github.com/clearlinux/swupd-client/blob/master/test/functional/testlib.bash#L1925

The comment in that part of the code specifies we should not terminate the script if the env doesn't exist but looks like the code is wrong cause it is returning 1, we should just return with 0. Looks like this wasn't giving us problems before because we were checking the env existed everywhere where we used the function. Like this:
```
if [ -d "$TEST_NAME" ]; then
	debug_msg "\\nAn old test environment was found for this test: $TEST_NAME"
	debug_msg "Deleting it..."
	destroy_test_environment --force "$TEST_NAME"
fi
```
I just tested this and when the test environment doesn't exist it crashes the test.

Maybe the right way to fix this is not to add `destroy_test_environment "$TEST_NAME" || true` like I suggested, but just 'return 0' here instead -> https://github.com/clearlinux/swupd-client/blob/master/test/functional/testlib.bash#L1925

otaviobp: You are correct about returning 1, but for some reason it's not failing the test for me (test/functional/usability/usa-completion-basic.bats for example). I'll change to 0 because it makes sense, but how did you reproduce that? 
The line you are pointing now is never going to be executed if there's no test environment because of the if.

One thing I noticed though is that we are trying to destroy the environment twice for the last test. I am going to fix that.

castulo: I reproduced it by adding `destroy_test_environment $TEST_NAME` in the setup before creating the environment. I know this is something we would not do, but the point is that maybe we should not crash the test if there was no test environment. I guess it is possible we could have run a test without adding a test environment, although we are not doing it now. If that happens the teardown should not crash because the env didn't exist.

castulo: For some reason this doesn't seem to be a problem during `teardown()` is like bats just disregards that RC=1 during teardown. We can leave like that then until we find a reason to change it.

PR #1430, opened 2020-04-06
actions: Validate if all tests are going to be executed

otaviobp:
Validate if we are really calling make check for all tests in
test/functional directory.


otaviobp:
V2: Rebased and removing tarball added by accident

PR #1434, opened 2020-04-07
Showing status after enabling/disabling autoupdate

castulo:

### DIFF
--- a/src/autoupdate.c
+++ b/src/autoupdate.c
@@ -135,6 +135,7 @@ enum swupd_code autoupdate_main(int argc, char **argv)
 		 */
 		if (globals.path_prefix) {
 			warn("Running autoupdate with --path will not restart swupd-update.timer. This will have to be done manually\n");
+			info("Autoupdate enabled\n");
 			return SWUPD_OK;
 		}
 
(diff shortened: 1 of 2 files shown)


src/autoupdate.c
```diff
@@ -135,6 +135,7 @@ enum swupd_code autoupdate_main(int argc, char **argv)
 		 */
 		if (globals.path_prefix) {
 			warn("Running autoupdate with --path will not restart swupd-update.timer. This will have to be done manually\n");
+			info("Autoupdate enabled\n");
```
otaviobp, line 138: 2 things here. First I would use the same wording used when no parameter is used. And also I think we should to run the same systemctl commands to make sure that the autoupdate is really enabled.

otaviobp: The only problem is that if a restart is needed we would print something like
```
Running systemctl to enable updates
Disabled
```

Not sure if it's clear enough

castulo: > 2 things here. First I would use the same wording used when no parameter is used. And also I think we should to run the same systemctl commands to make sure that the autoupdate is really enabled.

Sure, I can change this

castulo: > The only problem is that if a restart is needed we would print something like
> 
> ```
> Running systemctl to enable updates
> Disabled
> ```
> 
> Not sure if it's clear enough

I do think this is very confusing

otaviobp: So maybe we should let the way it is or find a better way to report this

castulo:
Canceling this PR until we come up with a better solution

PR #1438, opened 2020-04-08
API: output for [3rd-party] check-update

castulo:
- swupd check-update
- swupd 3rd-party check-update

### DIFF
--- a/src/3rd_party_repos.c
+++ b/src/3rd_party_repos.c
@@ -543,11 +543,10 @@ clean_and_exit:
 void third_party_repo_header(const char *repo_name)
 {
 	char *header = NULL;
-	bool quiet = (log_get_level() == LOG_ERROR);
 
 	string_or_die(&header, " 3rd-Party Repo: %s", repo_name);
 	print_header(header);
-	if (quiet) {
+	if (log_is_quiet()) {
 		print("[%s]\n", repo_name);
 	}
 	free_and_clear_pointer(&header);
--- a/src/check_update.c
+++ b/src/check_update.c
@@ -120,11 +120,12 @@ enum swupd_code check_update(void)
 	}
 
 	if (current_version < server_version) {
-		print("There is a new OS version available: %d\n", server_version);
+		info("There is a new OS version available: ");
+		print("%d\n", server_version);
 		update_motd(server_version);
 		return SWUPD_OK; /* update available */
 	} else if (current_version >= server_version) {
-		print("There are no updates available\n");
+		info("There are no updates available\n");
 	}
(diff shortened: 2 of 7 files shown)


src/3rd_party_repos.c
```diff
@@ -543,9 +543,13 @@ enum swupd_code third_party_run_operation_multirepo(const char *repo, process_bu
 void third_party_repo_header(const char *repo_name)
 {
 	char *header = NULL;
+	bool quiet = (log_get_level() == LOG_ERROR);
```
otaviobp, line 546: This is hard to maintain.
I would create a `#define LOG_QUIET LOG_ERROR` in log.h or some kind of function log_is_debug() and log_is_quiet().
And the ideal would be `log_get_level() < LOG_ERROR` to include LOG_ALWAYS too.

src/check_update.c
```diff
@@ -120,11 +120,12 @@ enum swupd_code check_update(void)
 	}
 
 	if (current_version < server_version) {
-		print("There is a new OS version available: %d\n", server_version);
+		info("There is a new OS version available: ", server_version);
```
otaviobp, line 123: extra parameter

test/functional/api/api-3rd-party-checkupdate.bats
```diff
@@ -0,0 +1,57 @@
...
+
+@test "API008: 3rd-party checkupdate" {
+
+	run sudo sh -c "$SWUPD 3rd-party check-update $SWUPD_OPTS --quiet"
+
+	assert_status_is "$SWUPD_OK"
+	expected_output=$(cat <<-EOM
+		[repo1]
+		20
+		[repo2]
```
otaviobp, line 30: I am unsure about this. Should we print the header [repo2] if there's no update?

src/3rd_party_repos.c
castulo: good idea

src/check_update.c
castulo: oops!

test/functional/api/api-3rd-party-checkupdate.bats
castulo: It would look nicer if we didn't, but I think is a limitation we have. I think it would be very difficult to not print it, I couldn't think of a way of doing it without doing a massive refactoring.

castulo:
V2: addressed all the comments except the one regarding removing the [repo_name] from the repo that did not have output. As I said, I think this is a limitation, it would look better if we didn't print the repos that don't have output, but I don't think it can be done without having to do a big refactoring, and IMO is not worth it.

PR #1439, opened 2020-04-08
Staging: Some improvements and reorganization of the staging module

otaviobp:
File installation and removal from system was spread in different parts of swupd. Grouping everything one module to make it easier to maintain.

I renamed this module from staging to system_root because all operations are relative to installing/removing files from the clear system installed. Let me know if you find this name confusing or if you have any other suggestion.


src/staging.c
```diff
@@ -370,6 +370,11 @@ enum swupd_code staging_install_all_files(struct list *files, struct manifest *m
 	int complete = 0;
 	unsigned int list_length = list_len(files);
 
+	if (!list_is_sorted(files, cmp_file_filename_is_deleted)) {
+		warn("List of files to install is not sorted - fixing\n");
+		files = list_sort(files, cmp_file_filename_is_deleted);
+	}
```
bryteise: This warning seems not useful to end users, more of a developer warning that somehow we got here with an unsorted list. I think getting to the point where we don't sort all over is a good goal so the logic is a step in the right direction.

otaviobp: You are correct. I didn't get concerned with the warning because as this is never suppose to happen, users are not suppose to see that. But I see your point.
The reason why I didn't use debug() is because I wanted this to break our tests if it ever happen.

I was wondering other approaches we could take here. I know some projects that uses assert() for those cases. There are other areas of the code I think we could start using this, but I remember that some people in the past didn't like the idea. Do you have any opinion about that?

bryteise: assert in swupd for things that are not fatal isn't something I want to add. If it is only an assert when built for running our tests and disabled in production that could work.

castulo: if this is a message that is really not supposed to be printed as you mentioned @otaviobp, you could just rephrase it in a way that has so meaning to the user just in the case it does get printed for some reason. Then we can use that message to catch errors in the code with tests.

An example of a message would be:
"Preparing files to be installed..."

src/staging.c
```diff
@@ -88,7 +88,7 @@ enum swupd_code do_staging(struct file *file, struct manifest *MoM)
 
 	/* make sure the directory where the file should be copied to exists
 	 * and is in deed a directory */
-	targetpath = sys_path_join("%s/%s", globals.path_prefix, rel_dir);
+	targetpath = sys_path_join("%s/%s", globals.path_prefix, dir);
```
castulo, line 92: can't you just use `targetpath` in line 175 so you get rid or rel_dir entirely?
Like
```
string_or_die(&tarcommand, TAR_COMMAND " -C '%s' " TAR_PERM_ATTR_ARGS " -cf - './%s' 2> /dev/null | " TAR_COMMAND " -C '%s' " TAR_PERM_ATTR_ARGS " -xf - 2> /dev/null", rename_tmpdir, base, targetpath);
```

src/staging.c
otaviobp: I am replacing for debug in this PR. Let's try to find a good solution on how to fix this kind of problems. There are multiple places in code where we do something similar, so we could have a nice approach to test that.

otaviobp:
V3:
 - Using target_root instead of system_root as suggested.
 - Don't print warnings to user that doesn't make sense
 - Use sys_path_join() in more places
 - Handle return code on function

src/staging.c
otaviobp: good catch here. I haven't noticed that.

otaviobp:
V4: REmove rel_dir completly

PR #1448, opened 2020-04-10
Suggestion WIP: Create UNEXPECTED() macro

otaviobp:
There are several points in swupd that we find conditions that should never
    happen, but we try to correct them anyway to be more reliable. Because of that
    we may hide programming mistakes that are corrected later.
    
    So, to make sure we are going to catch those problems in debug mode in our development
    machines and test environment, I am suggesting a UNEXPECTED() macro that aborts swupd
    if --enable-debug was used to build swupd.
    
    I am adding some examples on what we could do with that.

What are your opinions about that?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -28,6 +28,16 @@ AC_CHECK_LIB([pthread], [pthread_create])
 # Program checks
 AC_CHECK_PROGS(TAR, tar)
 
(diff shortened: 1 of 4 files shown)


bryteise:
Looks good to me =).

otaviobp:
I am going to fix comments then before merging

Issue #671, opened 2018-11-02
swupd-client should support filenames with unusual characters...


bktan8:
See https://github.com/clearlinux/mixer-tools/issues/494

Currently, mixer is blacklisting filenames with unusual characters.  But once mixer is able to handle them, I wonder how swupd-client will react to them.  This issue is a reminder to deal with this in the future.

phmccarty:
Recently, mixer started to ban the control character `\a` in filenames, so that should be factored in to this feature.

See the commit added this PR: https://github.com/clearlinux/mixer-tools/pull/596

otaviobp:
Fixed

PR #1457, opened 2020-04-15
Target root: Change the way we install directories and files

otaviobp:
Currently we install directories and files using a tar command piped to another tar command.
There are multiple problems related to that, like:
 - performance
 - dependency on tar binary
 - lack support on unusual characters on filenames

This pull request changes the way we install directories by creating them with mkdir and updating all attributes that we control on swupd, update directories by update all attributes and install files by using the cp binary, when link is not supported.

In some preliminary tests we got a nice improvement in installation speed, around 60~70% faster on big bundles, like desktop and os-clr-on-clr.

Related to #671 and  #640. Almost fixed and shouldn't be a problem anymore, but there are still changes required on install_dir_using_tar() and install_file_using_tar()

### DIFF
--- a/test/functional/update/update-unusual-file-names.bats
+++ b/test/functional/update/update-unusual-file-names.bats
@@ -1,0 +1,84 @@
+#!/usr/bin/env bats
+
+# Author: Otavio Pontes
+# Email: otavio.pontes@intel.com
+
+load "../testlib"
+
+test_setup() {
+
+	create_test_environment "$TEST_NAME"
+	create_version -p "$TEST_NAME" 20 10
+
+	update_bundle "$TEST_NAME" os-core --add "/usr/simple_file"
+	file=$(get_hash_from_manifest "$TEST_NAME"/web-dir/20/Manifest.os-core /usr/simple_file)
+
+	# Unusual directory names
+	update_bundle "$TEST_NAME" os-core --add "/usr/unusual_'dirname'/normal_file1"
+	update_bundle "$TEST_NAME" os-core --add '/usr/unusual_"dirname"/normal_file2'
+	update_bundle "$TEST_NAME" os-core --add '/usr/very_unusual_,¹²³!@#$%^&*()_;[]{}\`_name/normal_file3'
+	update_bundle "$TEST_NAME" os-core --add "/usr/unusual:dirname/normal_file4:""$TEST_NAME"/web-dir/20/files/"$file"
+
+	# Unusual file names
+	update_bundle "$TEST_NAME" os-core --add "/usr/unusual_files/'file1'"
+	update_bundle "$TEST_NAME" os-core --add '/usr/unusual_files/"file2"'
+	update_bundle "$TEST_NAME" os-core --add '/usr/unusual_files/very_unusual_,¹²³!@#$%^&*()_;[]{}\_file3'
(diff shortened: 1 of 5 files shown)


bryteise:
I really like the direction of these changes!

otaviobp:
V2: Adding a test to validate if files with unusual filenames are successfully installed

test/functional/update/update-unusual-file-names.bats
```diff
@@ -0,0 +1,84 @@
...
+	# Unusual file names
+	update_bundle "$TEST_NAME" os-core --add "/usr/unusual_files/'file1'"
+	update_bundle "$TEST_NAME" os-core --add '/usr/unusual_files/"file2"'
+	update_bundle "$TEST_NAME" os-core --add '/usr/unusual_files/very_unusual_,¹²³!@#$%^&*()_;[]{}\_file3'
+	update_bundle "$TEST_NAME" os-core --add "/usr/unusual_files/file:4:""$TEST_NAME"/web-dir/20/files/"$file"
+
+	# Unusual link names
+	create_bundle -n test-bundle1 -l "/usr/links/'link1'",'/usr/links/"link2"','/usr/links/very_unusual_¹²³!@#$%^&*()_;[]{}\_link3','/usr/links/link:4' "$TEST_NAME"
+	add_dependency_to_manifest "$WEBDIR"/20/Manifest.os-core test-bundle1
+
```
castulo, line 31: wow, you are really pushing it with this test, nice!
😲

test/functional/update/update-unusual-file-names.bats
```diff
@@ -0,0 +1,84 @@
...
+		4 files were not in a pack
+		Update successful - System updated from version 10 to version 20
+	EOM
+	)
+	assert_is_output "$expected_output"
+
+	assert_file_exists "$TARGETDIR""/usr/simple_file"
+
+	# Unusual dir names
+	assert_file_exists "$TARGETDIR""/usr/unusual_'dirname'/normal_file1"
```
castulo, line 68: you could even use `assert_dir_exists` here to make sure it is a directory

test/functional/update/update-unusual-file-names.bats
```diff
@@ -0,0 +1,84 @@
...
+	assert_file_exists "$TARGETDIR"'/usr/unusual:dirname/normal_file4'
+
+	# Unusual file names
+	assert_file_exists "$TARGETDIR""/usr/unusual_files/'file1'"
+	assert_file_exists "$TARGETDIR"'/usr/unusual_files/"file2"'
+	assert_file_exists "$TARGETDIR"'/usr/unusual_files/very_unusual_,¹²³!@#$%^&*()_;[]{}\_file3'
+	assert_file_exists "$TARGETDIR""/usr/unusual_files/file:4"
+
+	# Unusual link names
+	assert_file_exists "$TARGETDIR""/usr/links/'link1'"
```
castulo, line 80: same here with `assert_symlink_exists`

test/functional/update/update-unusual-file-names.bats
otaviobp: I am actually testing the file that is inside the directory. I want to check if swupd can create dirs and can install files into them. So we are also checking if it's a dir by free

PR #1461, opened 2020-04-16
API: output for [3rd-party] repair

castulo:

 - swupd repair
 - swupd 3rd-party repair


test/functional/api/api-repair.bats
```diff
@@ -0,0 +1,99 @@
...
+
+@test "API059: repair (failure to repair)" {
+
+	# force some repairs in the target system
+	set_current_version "$TEST_NAME" 20
+	sudo touch "$TARGETDIR"/usr/untracked_file
+
+	# force failures while repairing
+	sudo touch "$TARGETDIR"/baz
+	sudo chattr +i "$TARGETDIR"/usr/untracked_file
```
otaviobp, line 78: great idea!

PR #1465, opened 2020-04-16
API: output for [3rd-party] clean

castulo:

- swupd clean
- swupd 3rd-party clean


castulo:
@otaviobp this PR should be good to be reviewed, again, the error shown is in the Tests IDs, they should be good once the other PRs opened with tests are merged and the code base is rebased.

PR #1467, opened 2020-04-17
API: output for 3rd-party add, remove, list

castulo:

- swupd 3rd-party add
- swupd 3rd-party remove
- swupd 3rd-party list

### DIFF
--- a/test/functional/api/api-3rd-party-add.bats
+++ b/test/functional/api/api-3rd-party-add.bats
@@ -1,0 +1,23 @@
+#!/usr/bin/env bats
+
+# Author: Castulo Martinez
+# Email: castulo.martinez@intel.com
+
+load "../testlib"
+
+test_setup() {
+
+	create_test_environment "$TEST_NAME"
+	create_third_party_repo "$TEST_NAME" 10 staging repo1
+	export repo1="$TPURL"
+
+}
+
+@test "API071: 3rd-party add --certpath CERTIFICATE" {
+
+	run sudo sh -c "$SWUPD 3rd-party add $SWUPD_OPTS my-repo file://$repo1 --quiet"
+
+	assert_status_is "$SWUPD_OK"
+	assert_output_is_empty
+
+}
--- a/src/3rd_party_add.c
+++ b/src/3rd_party_add.c
@@ -102,7 +102,7 @@ static bool confirm_certificate(const char *cert)
 	signature_print_info(cert);
 
 	info("\n");
-	info("To add the 3rd-party repository you need to accept this certificate\n");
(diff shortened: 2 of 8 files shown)


test/functional/api/api-3rd-party-add.bats
```diff
@@ -0,0 +1,55 @@
...
+
+}
+
+@test "API069: 3rd-party add" {
+
+	run sudo sh -c "echo y | $SWUPD 3rd-party add $SWUPD_OPTS_NO_CERT my-repo file://$repo1 --quiet"
+
+	assert_status_is "$SWUPD_OK"
+	expected_output=$(cat <<-EOM
+		Issuer: /C=US/ST=Oregon/L=Portland/O=Company Name/OU=Org/CN=localhost
```
otaviobp, line 22: I am unsure if we need to show that if --assume=yes is used, what do you think?

castulo: hmm you are right, it may be a good idea 🤔

castulo:
V2:
- The certificate was already being printed when `--assume=yes` was being used (as requested by @otaviobp )
- Corrected a test that was wrong

castulo:
V3:
- Copying the certificate to the default location so we can use `$SWUPD_OPTS_NO_CERT`

test/functional/api/api-3rd-party-add.bats
otaviobp: Actually my idea was to hide it, but I really don't know. Let's leave how it is.

test/functional/api/api-3rd-party-add.bats
```diff
@@ -0,0 +1,94 @@
...
+	# Skip this test for local development because we write the certificate on /
+	# This is necessary because the default certificate location is hardcoded on build time and we
+	# need to run swupd without -C parameter to test this feature.
+	if [ -z "${RUNNING_IN_CI}" ]; then
+		skip "Skipping test because it will make changes to your system"
+	fi
+
+	create_test_environment "$TEST_NAME"
+
+	if [ ! -f /usr/share/clear/update-ca/Swupd_Root.pem ]; then
```
otaviobp, line 19: You can't guarantee that a test like this will run fine when running in parallel with other tests. Move that to test/functional/only_in_ci_system/

test/functional/api/api-3rd-party-add.bats
castulo: ooh I see... well, let me know what you think

castulo:
V4:
- Moved tests that should only run in CI to the `functional/only_in_ci_system` directory

PR #1468, opened 2020-04-17
API: output for mirror

castulo:

- swupd mirror

### DIFF
--- a/src/mirror.c
+++ b/src/mirror.c
@@ -365,6 +365,7 @@ enum swupd_code mirror_main(int argc, char **argv)
 		print_help();
 		return SWUPD_INVALID_OPTION;
 	}
+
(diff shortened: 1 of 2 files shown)


src/mirror.c
```diff
@@ -402,7 +405,16 @@ enum swupd_code mirror_main(int argc, char **argv)
 	}
 
 	/* print new configuration */
-	print_update_conf_info();
+	if (quiet) {
```
otaviobp, line 408: why not just use `if (log_is_quiet())` here?

castulo: 🤔 good point 😆

castulo:
V2:
- using the `log_is_quiet()` function directly without storing the value in a variable

PR #1471, opened 2020-04-17
Telemetry: Some changes in the telemetry reports we create

otaviobp:


otaviobp:
@mbelluzzo: following your suggestions and some other improvements.
@jurobystricky and @alexjch: This may interest you too. I am changing the severity of some telemetry records created by swupd, so if you are looking at some swupd records you may be affected by this PR.

src/telemetry.h
```diff
@@ -0,0 +1,36 @@
...
+extern "C" {
+#endif
+
+/**
+ * Severity of a telemetry record
+ */
+enum telemetry_severity {
+	/** @brief Temetry severity DEBUG */
+	TELEMETRY_DEBG = 1,
+	/** @brief Temetry severity DEBUG */
```
mbelluzzo, line 19: Just missing to update the @brief lines for each entry.

alexjch:
@otaviobp this is transparent to telemetry, the change is fine with me.

src/telemetry.c
```diff
@@ -32,6 +32,16 @@
 
 #define RECORD_VERSION 2
 
+#ifdef DEBUG_MODE
```
castulo, line 35: nice, good idea

PR #1490, opened 2020-04-21
docs: Changes on man pages to make is similar to system man pages

### LINKED ISSUE #1285: Make swupd global and local options consistent in the manual
Options on manpage is rendered differently when they are global or local options. We could use the same style for both. 

On #1271 there were some discussions on how we put bullet points before options and subcommands. In all man pages I have reviewed recently I didn't see any bullet point used for that, so I would take them out too.

otaviobp:
Using a man page formating similar to what other man pages in the system
uses.

Fixes #1285

### DIFF
--- a/docs/swupd.1.rst
+++ b/docs/swupd.1.rst
@@ -51,165 +51,121 @@ OPTIONS
 The following options are applicable to most subcommands, and can be
 used to modify the core behavior and resources that swupd uses.
 
--  ``-h, --help``
+-h, --help    Display general help information. If put after a subcommand, it
+        will display help specific to that subcommand.
 
-   Display general help information. If put after a subcommand, it will
-   display help specific to that subcommand.
+-v, --version   Displays the version information of the swupd program, and exit.
+        It also displays compile options and copyright information.
 
--  ``-v, --version``
+-p <path>, --path=<path>   Optionally set the top-level directory for the
+        swupd-managed system. This can be used to point to a chroot installation
+        of the OS or a custom mount. If not specified this will default to ``/``.
 
-   Displays the version information of the swupd program, and exit. It also
-   displays compile options and copyright information.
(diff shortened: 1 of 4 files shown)


docs/swupd.1.rst
```diff
@@ -51,165 +51,121 @@ OPTIONS
 The following options are applicable to most subcommands, and can be
 used to modify the core behavior and resources that swupd uses.
 
--  ``-h, --help``
+-h, --help    Display general help information. If put after a subcommand, it
+        will display help specific to that subcommand.
 
-   Display general help information. If put after a subcommand, it will
-   display help specific to that subcommand.
+-v, --version   Displays the version information of the swupd program, and exit.
```
castulo, line 57: I have one comment that is repeated many times in this PR, although I am not 100% sure it is part of this PR or if it should be done in a separate PR. But we have a lot of inconsistencies where some flags that expect a value sometimes they show the value and some time they don't. I think all those should always show a value, for example "version" in this case. I think it should be:
`-v <version number>, --version=<version number> `

the same applies for example to  `-C <path>, --certpath <path>`, `-W <num of downloads>, --max-parallel-downloads <num of downloads>`, etc.

otaviobp: It's part of this patch. I'll review that. thanks for pointing this out

otaviobp:
V2: Fixing incorrect parameters in doc

docs/swupd.1.rst
otaviobp: The --version is correc the way it is because there's no parameters expected.
I fixed all other cases I found. If there's anything wrong left, please comment here.

castulo: haha, true, I chose a bad example because this --version does not expect parameters as you said 😝

PR #1498, opened 2020-04-22
Adding a --3rd-party flag to update

### LINKED ISSUE #1473: Add a flag to force 3rd-party update on update command
We can add a flag --3rd-party on update command to always update 3rd-party repositories if update is successful. So:
```
swupd update --3rd-party
```
Would update the system and if successful, update all 3rd-party repositories

castulo:
This commit adds a flag that can be used to update content from
3rd-party repositories after updating the regular content if
successfull.

Closes #1473

### DIFF
--- a/src/3rd_party_update.c
+++ b/src/3rd_party_update.c
@@ -249,31 +249,23 @@ static enum swupd_code update_repos(UNUSED_PARAM char *unused)
 	}
 }
 
-enum swupd_code third_party_update_main(int argc, char **argv)
+enum swupd_code third_party_execute_update(void)
 {
 	enum swupd_code ret_code = SWUPD_OK;
 	char *template_file = NULL;
 	char *template = NULL;
+	size_t template_len;
 	int steps_in_update;
 	int ret;
-	size_t template_len;
 
-	if (!parse_options(argc, argv)) {
-		print("\n");
-		print_help();
-		return SWUPD_INVALID_OPTION;
-	}
+	/* 3rd-party updates can be executed also from the update command
+	 * using the --3rd-party option, make sure a flag was not set in that
+	 * command before calculating steps */
+	cmdline_option_download_only |= update_get_option_download_only();
 
-	ret_code = swupd_init(SWUPD_ALL);
-	if (ret_code != SWUPD_OK) {
-		error("Failed swupd initialization, exiting now\n");
-		return ret_code;
-	}
-
-	/* set the command options */
-	update_set_option_version(cmdline_option_version);
-	update_set_option_download_only(cmdline_option_download_only);
-	update_set_option_keepcache(cmdline_option_keepcache);
+	/* the --update-search-file-index is not supported for 3rd-party
+	 * so set it to false in case it was set up to true by update */
+	update_set_option_update_search_file_index(false);
 
 	/*
 	 * Steps for update:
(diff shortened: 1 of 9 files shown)


src/3rd_party_update.c
```diff
@@ -299,34 +335,8 @@ enum swupd_code third_party_update_main(int argc, char **argv)
...
-	/* read the current template copy */
-	template_file = sys_path_join("%s/%s/%s", globals_bkp.path_prefix, SWUPD_3RD_PARTY_DIR, SWUPD_3RD_PARTY_TEMPLATE_FILE);
-	template = sys_mmap_file(template_file, &template_len);
-
-	if (!template || strncmp(template, SCRIPT_TEMPLATE, template_len) != 0) {
-		/* there is no template file, or the template changed,
-		 * all scripts need to be recreated */
-		info("The scripts that export binaries from 3rd-party repositories need to be regenerated\n\n");
-		ret_code = third_party_run_operation_multirepo(NULL, regenerate_all_wrapper_scripts, SWUPD_OK, "regenerate_scripts", steps_in_update);
+	ret_code = third_party_execute_update(steps_in_update);
```
otaviobp, line 338: This looks incorrect. Why do we need the number of steps from update in the 3rd-party update?

castulo:
V2:
- using the correct number of steps for 3rd-party update

src/update.c
```diff
@@ -675,6 +682,15 @@ enum swupd_code update_main(int argc, char **argv)
 		ret = check_update();
 	} else {
 		ret = execute_update();
+		if (cmdline_option_3rd_party) {
+			if (ret == SWUPD_OK) {
+				progress_finish_steps(ret);
+				info("\nUpdating content from 3rd-party repositories\n");
+				ret = third_party_execute_update(steps_in_3rd_party_update);
```
otaviobp, line 689: I keep with my comment. Why do you need to move the information on how many steps are needed on 3rd_party update to update.

src/update.c
```diff
@@ -675,6 +682,15 @@ enum swupd_code update_main(int argc, char **argv)
 		ret = check_update();
```
otaviobp, line 682: Not sure how useful it is, but for me it's a bit weird to have `swupd udpate --status --3rd-party` not running `swupd 3rd-party update --status` like any other parameter

src/update.c
castulo: I'm sorry I don't understand your question, `third_party_run_operation_multirepo()` requires the number of steps of the operation to be run as a parameter.

otaviobp: Yes. But you are not calling `third_party_run_operation_multirepo` here. You are calling `third_party_execute_update`. Can't you change `third_party_execute_update` to have no parameters and check the number of steps there instead of setting it here?

castulo: ah I see what you mean now. Yeah I thought of that at first, but we need to know if the user used the `--download` flag in order for us to calculate the number of steps accurately. I could send that as parameter to third_party_execute_update instead of sending the calculated number of steps.

src/update.c
castulo: cool, I didn't add it at first because I wanted to minimize the swupd commands that call a 3rd-party command, but I'll add it so it is consistent

castulo:
V3:

- Enabling the use of `update --status --3rd-party`
- Calculating the number of steps in the 3rd-party update entirely within 3rd_party_update.c

otaviobp:
looks good now

PR #1499, opened 2020-04-22
3rd-party: using "repository" for headers

castulo:
Headers are printed at the beginning of the output for every repository
in operations that run in multipl repos. This header was using the word
"repo" but we are trying to use the whole "repository" word every time
we printed it to the user. This commit changes that.

### DIFF
--- a/test/functional/3rd-party/3rd-party-bundle-list-basic.bats
+++ b/test/functional/3rd-party/3rd-party-bundle-list-basic.bats
@@ -30,9 +30,9 @@ global_setup() {
 
 	assert_status_is "$SWUPD_OK"
 	expected_output=$(cat <<-EOM
-		_______________________
-		 3rd-Party Repo: repo1
-		_______________________
+		_____________________________
+		 3rd-Party Repository: repo1
+		_____________________________
 
 		Installed bundles:
 		 - os-core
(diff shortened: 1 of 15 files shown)


test/functional/3rd-party/3rd-party-bundle-list-basic.bats
```diff
@@ -30,9 +30,9 @@ global_setup() {
 
 	assert_status_is "$SWUPD_OK"
 	expected_output=$(cat <<-EOM
-		_______________________
```
otaviobp, line 33: Isn't it better to filter this from the output so we don't need to add that to tests?

otaviobp: And in json_output the output is not very cool. We print one `_` per line

castulo: > Isn't it better to filter this from the output so we don't need to add that to tests?

yeah, that may be a good idea. We can always use `--identical` for those tests we really want to check all output.

castulo: > And in json_output the output is not very cool. We print one `_` per line

yes, it is not cool, but I didn't know how to fix this, currently there is no way of making exceptions of what we include in the json output. I am open to suggestions @otaviobp

otaviobp: I made a PR trying to fix that: #1510

castulo: +1 ☝️

castulo:
Rebased

PR #1507, opened 2020-04-23
Bug_fix: Fixing bug in diagnose/repair using the `--file` option

### LINKED ISSUE #1472: Swupd diagnose --file has some unusual behaviors
**Describe the bug**
I found 2 different unusual behavior of swupd diagnose family of commands and I decided to group them together because it looks like the root cause is the same.

1) When trying to diagnose one extra file, sometimes it works. Sometimes it doesn't. I am using --quiet for the examples because it's easier to explain, but the bug is not related to using --quiet.
```
$ touch /usr/bin/a1 /usr/bin/a2 /usr/bin/a3
$ swupd diagnose --extra-files-only --file /usr/bin/ --quiet; echo $?
/usr/bin/a3
/usr/bin/a2
/usr/bin/a1
1
$ swupd diagnose --extra-files-only --file /usr/bin/a1 --quiet; echo $?
0
$ swupd diagnose --extra-files-only --file /usr/bin/a2 --quiet; echo $?
/usr/bin/a2
1
$ swupd diagnose --extra-files-only --file /usr/bin/a3 --quiet; echo $?
0
```
I have no idea why it works only for a2.

2) When using `swupd diagnose --file FILE` swupd is diagnosing all files that starts with FILE and not files that are equal FILE, like the documentation states
```
$ rm /usr/bin/lsusb
$ swupd diagnose --file /usr/bin/ls --quiet; echo $?
/usr/bin/lsusb
1
```
The expected would be to fix /usr/bin/ls binary, but not /usr/bin/lsusb

castulo:
This PR fixes a couple of bugs found in the diagnose/repair functionality:
- Diagnose should not use partial matches for the `--file` option
- Swupd should still check for extra files even in no tracked files are found in the path specified by the user with the `--file` flag.

Closes #1472

### DIFF
--- a/src/verify.c
+++ b/src/verify.c
@@ -882,14 +882,29 @@ static struct list *keep_matching_path(struct list *all_files)
 	struct list *matching_files = NULL;
 	struct list *iter = NULL;
 	struct file *file;
+	char *tmp_dir = NULL;
+
+	/* when considering the item as a directory, make sure the item
+	 * finishes with a forward slash so we can match items accuratelly,
+	 * oherwise we can end up with items that partially match */
+	tmp_dir = str_or_die("%s%s", cmdline_option_file, cmdline_option_file[str_len(cmdline_option_file)] == '/' ? "" : "/");
 
 	for (iter = all_files; iter; iter = iter->next) {
 		file = iter->data;
-		if (strncmp(cmdline_option_file, file->filename, str_len(cmdline_option_file)) == 0) {
+
+		/* if the item entered by the user is a file, the match should
+		 * be exact, so look for an exact match first, if there is none,
+		 * look for a match treating the item as directory */
+		if (str_cmp(cmdline_option_file, file->filename) == 0) {
+			/* preserving the order is important */
+			matching_files = list_append_data(matching_files, file);
+		} else if (strncmp(tmp_dir, file->filename, str_len(tmp_dir)) == 0) {
 			/* preserving the order is important */
 			matching_files = list_append_data(matching_files, file);
 		}
 	}
+
+	free_and_clear_pointer(&tmp_dir);
 
 	return list_head(matching_files);
 }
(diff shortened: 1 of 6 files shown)


src/verify.c
```diff
@@ -882,15 +882,30 @@ static struct list *keep_matching_path(struct list *all_files)
 	struct list *matching_files = NULL;
 	struct list *iter = NULL;
 	struct file *file;
+	char *tmp_dir = NULL;
+
+	/* when considering the item as a directory, make sure the item
+	 * finishes with a forward slash so we can match items accuratelly,
+	 * oherwise we can end up with items that partially match */
+	tmp_dir = str_or_die("%s%s", cmdline_option_file, cmdline_option_file[str_len(cmdline_option_file)] == '/' ? "" : "/");
```
otaviobp, line 890: I like the approach on treating all files as a possible directory here, but this is inconsistent with the messages  `Limiting diagnose to the following file/directory`. If there's a file type change you are printing 'directories' for files and files for directories.

src/verify.c
```diff
@@ -882,15 +882,30 @@ static struct list *keep_matching_path(struct list *all_files)
...
 
 	for (iter = all_files; iter; iter = iter->next) {
 		file = iter->data;
-		if (strncmp(cmdline_option_file, file->filename, str_len(cmdline_option_file)) == 0) {
+
+		/* if the item entered by the user is a file, the match should
+		 * be exact, so look for an exact match first, if there is none,
+		 * look for a match treating the item as directory */
+		if (str_cmp(cmdline_option_file, file->filename) == 0) {
```
otaviobp, line 898: Here you are comparing the strings twice just because of one char. You can look only at that char

```
size = str_len(cmdline_option_file);
if (strncmp(cmdline_option_file, filename, size) == 0) {
   if (filename[size] == '/0' || filename[size] == '/') {
       //is a match
   }
}
```

castulo: nice approach

PR #1510, opened 2020-04-23
header: Print the header decoration all at once

otaviobp:
When printing the header, print all '_' chars at once to prevent
json-output and --debug to print that in multiple lines.

### DIFF
--- a/src/helpers.c
+++ b/src/helpers.c
@@ -915,13 +915,14 @@ fail:
 	return ret;
 }
 
-void print_pattern(const char *pattern, int times)
+static void print_char(const char c, int times)
 {
-	while (times > 0) {
-		info("%s", pattern);
-		times--;
-	}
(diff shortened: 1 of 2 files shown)


src/helpers.c
```diff
@@ -915,23 +915,24 @@ int get_value_from_path(char **contents, const char *path, bool is_abs_path)
...
 {
-	while (times > 0) {
-		info("%s", pattern);
-		times--;
-	}
-	info("\n");
+	char *str;
+
+	str = calloc(1, times + 1);
+	memset(str, c, times);
```
otaviobp, line 923: An option to prevent allocation more memory is to do this:
```printf("%.*s", times, "___________________________________________________________");```
But I prefered to keep it like this.

castulo: clever! 👏

PR #1512, opened 2020-04-24
API: Do not validate error messages in API tests

### LINKED ISSUE #1502: Remove all error messages parsing from API tests
The idea on having an API test is to test only what's going to be in our API, so error messages shouldn't be validated.
If there's need to check an error message it should be checked in another test on another category.

castulo:
The tests in test/functional/api are going to validate the --quiet
output is rarely changed, therefore only expected output should be
validated and not warnings or error messages.

Closes #1502

### DIFF
--- a/test/functional/api/api-mirror.bats
+++ b/test/functional/api/api-mirror.bats
@@ -49,7 +49,7 @@ global_setup() {
 
 }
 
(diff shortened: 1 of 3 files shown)


castulo:
Rebased

PR #1514, opened 2020-04-27
target_root: Use run_commmand to execute tar|tar command

### LINKED ISSUE #640: Path with a single quote fails bundle-add

otaviobp:
Porting code to better way on interacting with system commands using the
run_command function. As run_command don't support multi-thread for
performance reasons, we are now using a file instead of a pipe. As tar|tar
is now a fallback, this won't have any performance impacts on any case
where with no problems and errors.

Fixes #640 and #671

### DIFF
--- a/scripts/build_and_run_tests.bash
+++ b/scripts/build_and_run_tests.bash
@@ -42,7 +42,7 @@ run_checks_tartar() {
 
 	# Tar tar doasnt support hardlinks
 	# There's a known bug in tar tar code that doesn't support unusual file names
-	FILES="$(find test/functional  -name "*.bats" \( ! -path "*only_in_ci_system*" \) | sort | grep -v install-hardlink-symlink.bats | grep -v update-unusual-file-names.bats)"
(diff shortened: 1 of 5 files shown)


otaviobp:
Blocking the merge of this PR before next release. I would like to give some time to check the results of using new implementation of install files before having this in.

PR #1515, opened 2020-04-27
Disable "update --3rd-party" when applicable

castulo:
There is a configuration option to build swupd without the 3rd-party
functionality, this commit adds a guard to disable the code that adds
the "swupd update --3rd-party" flag which should be only available if
3rd-party is enabled.


lebensterben:
I'll update the zsh completion to remove(not adding) "3rd-party" stuff when necessary.

otaviobp:
> I'll update the zsh completion to remove(not adding) "3rd-party" stuff when necessary.

cool

PR #1521, opened 2020-04-28
Fail if memory allocation fails during print_char

castulo:


castulo:
There is another PR already fixing this. I'm closing this PR in favor of that one. (#1518)

PR #1524, opened 2020-04-28
Int: Multiple fixes on signed to unsigned conversions on several files

otaviobp:


otaviobp:
I'm planning a different move for macros.h so I'll close this

PR #1528, opened 2020-04-29
Shellcheck

otaviobp:


test/code_analysis/shellcheck.bats
```diff
@@ -9,6 +9,6 @@
 
 		sed 's/^@.*/func() {/' "$i" |
 		sed 's/^load.*/source test\/functional\/testlib.bash/' |
-		shellcheck -s bash -x -e SC1008 /dev/stdin
+		shellcheck -s bash -x -e SC1008,SC2119 /dev/stdin
```
castulo, line 12: Maybe we should add an explanation at the top of the file with all the exceptions that we ad here, so it is clearer why we decided to skip those checks. something like:

```
#!/usr/bin/env bats

# Author: Otavio Pontes

# Exceptions:
#
# SC2119: Use foo "$@" if function's $1 should mean script's $1.
# We need to skip this check because in many ocassions we use functions
# that are not expecting arguments, but still shellcheck will detect this
# as error because the show_help which is called from all functions uses $@
# SC1008: ....

@test "ANL001: Shellcheck on changed files" {
```

otaviobp:
V2:
 - Adding comment and updating all shellcheck bats scripts to use same code

PR #1533, opened 2020-04-30
Verifytime: Reduce dependencies from verify time binary

otaviobp:


src/helpers.c
```diff
@@ -461,6 +461,50 @@ void remove_trailing_slash(char *url)
...
+		return true;
+	}
+
+	if (err == -ENOENT) {
+		error("Failed to read system timestamp file\n");
+	}
+
+	time_str = time_to_string(system_time);
+	if (time_str) {
+		printf("Current time is %s\nAttempting to fix...\n", time_str);
```
castulo, line 489: I find this message a little confusing:
```
Current time is Mon May  4 09:39:22 PDT 2020
Attempting to fix...
```
First, in this sentence "Current time is X", is that the time that is incorrect that is going to be fixed? or is that the correct time that the wrong time is going to be corrected to?

Secondly, I think "attempting to fix..." should be "attempting to fix it...".

src/helpers.c
```diff
@@ -461,6 +461,50 @@ void remove_trailing_slash(char *url)
...
+		FREE(time_str);
+	}
+
+	if (set_time(clear_time)) {
+		time_str = time_to_string(clear_time);
+		if (time_str) {
+			info("Set system time to %s\n", time_str);
```
castulo, line 496: Maybe I am reading this incorrectly, but shouldn't this be "System time set to X" instead?

src/helpers.c
```diff
@@ -461,6 +461,50 @@ void remove_trailing_slash(char *url)
...
+			FREE(time_str);
+		}
+
+		// System time has been fixed
+		return true;
+	}
+
+	// System time is wrong
```
castulo, line 504: Shouldn't we warn about not being able to fix the time? After all the last message the user would have seen was "Attempting to fix it..."

src/helpers.c
otaviobp: Updated to:
```Fixing outdated system time from 'Tue May 05 09:03:42 2020' to 'Fri May 01 11:47:27 2020'```

otaviobp:
V2: Fixed the error/info messages

PR #1534, opened 2020-04-30
Add `--orphans` flag for bundle-list

### LINKED ISSUE #1530: Add `--orphans` flag for bundle-list
This issue is to add a flag to list all bundles that are not tracked or dependency of a tracked bundle.
This option should also be available for `3rd-party`.

`$ swupd bundle-list --orphans`

Related to #1529

castulo:
This commit adds the `--orphans` flag to `bundle-list` and `3rd-party bundle-list`. This flag will list the bundles that are no longer required by any tracked bundles.

Closes #1530

### DIFF
--- a/src/bundle_list.c
+++ b/src/bundle_list.c
@@ -28,12 +28,14 @@
 
 #define FLAG_DEPS 2000
 #define FLAG_STATUS 2001
+#define FLAG_ORPHANS 2002
 
 static bool cmdline_option_all = false;
 static char *cmdline_option_has_dep = NULL;
 static char *cmdline_option_deps = NULL;
 static bool cmdline_local = true;
 static bool cmdline_option_status = false;
+static bool cmdline_option_orphans = false;
 
 void bundle_list_set_option_all(bool opt)
 {
@@ -66,6 +68,11 @@ void bundle_list_set_option_status(bool opt)
 	cmdline_option_status = opt;
 }
 
+void bundle_list_set_option_orphans(bool opt)
+{
+	cmdline_option_orphans = opt;
+}
+
 static void free_has_dep(void)
 {
 	FREE(cmdline_option_has_dep);
@@ -89,6 +96,7 @@ static void print_help(void)
 	print("   -D, --has-dep=[BUNDLE]  List all bundles which have BUNDLE as a dependency (use --verbose for tree view)\n");
 	print("   --deps=[BUNDLE]         List BUNDLE dependencies (use --verbose for tree view)\n");
 	print("   --status                Show the installation status of the listed bundles\n");
+	print("   --orphans               List orphaned bundles\n");
 	print("\n");
 }
 
@@ -97,6 +105,7 @@ static const struct option prog_opts[] = {
 	{ "deps", required_argument, 0, FLAG_DEPS },
 	{ "has-dep", required_argument, 0, 'D' },
 	{ "status", no_argument, 0, FLAG_STATUS },
+	{ "orphans", no_argument, 0, FLAG_ORPHANS },
 };
 
 static bool parse_opt(int opt, char *optarg)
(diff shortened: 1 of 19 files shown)


castulo:
@otaviobp this PR is going to fail because it adds tests to the `api` test suite.

src/bundle_list.c
```diff
@@ -419,6 +445,91 @@ static enum swupd_code show_bundle_reqd_by(const char *bundle_name, bool server,
...
+		error("Cannot load official manifest MoM for version %i\n\n", version);
+		return SWUPD_COULDNT_LOAD_MOM;
+	}
+
+	// start with a list of the tracked bundles
+	// (bundles specifically installed by the user)
+	tracked_bundles = bundle_list_tracked();
+	if (!tracked_bundles) {
+		// this should never happen, swupd_init makes sure of it
+		UNEXPECTED();
```
otaviobp, line 474: The idea when I created the UNEXPECTED macro was for some unexpected behaviors that we try to recover and that wouldn't be able to be noticed unless we have an abort() call to warn us. If it's a situation where you return an error code, you don't need to use that. Just return the error code and the test is going to break.

src/bundle_list.c
```diff
@@ -419,6 +445,91 @@ static enum swupd_code show_bundle_reqd_by(const char *bundle_name, bool server,
...
+	info("Orphan bundles:\n");
+	for (iter = bundles; iter; iter = iter->next) {
+		bundle_manifest = mom_search_bundle(mom, (char *)iter->data);
+		if (bundle_manifest) {
+			name = get_printable_bundle_name(bundle_manifest->filename, bundle_manifest->is_experimental, cmdline_option_status && is_installed_bundle(bundle_manifest->filename), cmdline_option_status && is_tracked_bundle(bundle_manifest->filename));
+			info(" - ");
+			print("%s\n", name);
+			FREE(name);
+		} else {
+			UNEXPECTED();
```
otaviobp, line 517: This is a good usage of UNEXPECTED()

src/bundle_list.c
```diff
@@ -419,6 +445,91 @@ static enum swupd_code show_bundle_reqd_by(const char *bundle_name, bool server,
...
+	// create the list of required bundles by using tracked bundles as base
+	// and adding the dependencies of each tracked bundles to the list
+	ret = add_subscriptions(tracked_bundles, &subs, mom, true, 0);
+	if (ret != add_sub_NEW) {
+		// something went wrong or there were no includes, print a message and exit
+		if (ret & add_sub_ERR) {
+			error("Cannot load included bundles\n\n");
+			ret_code = SWUPD_COULDNT_LOAD_MANIFEST;
+		} else {
+			UNEXPECTED();
```
otaviobp, line 488: You don't need this here.

src/bundle_list.c
castulo: got it

castulo:
V2:
- Removed "prefix" from sys_ls()
- Removed get_dir_files_sorted() in favor of using sys_ls()
- Corrected usage of UNEXPECED()
- Fixed a problem with showing a message even if swupd failed to list
- Not using errno to check for errors

src/bundle_list.c
```diff
@@ -476,6 +589,9 @@ enum swupd_code bundle_list_main(int argc, char **argv)
 	progress_init_steps("bundle-list", steps_in_bundlelist);
 
 	ret = list_bundles();
+	if (cmdline_option_orphans) {
+		info("\nUse \"swupd bundle-add BUNDLE\" to no longer list BUNDLE and its dependencies as orphaned\n");
```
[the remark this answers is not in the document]
otaviobp: I still think it doesn't make sense to print this if the number of orphan bundles is 0

PR #1544, opened 2020-05-05
macros: Create wrapper for malloc and calloc functions

otaviobp:
Use always calloc to allocate memory and abort on errors

### DIFF
--- a/src/lib/macros.h
+++ b/src/lib/macros.h
@@ -1,6 +1,8 @@
 #ifndef __INCLUDE_GUARD_MACROS_H
 #define __INCLUDE_GUARD_MACROS_H
 
+#include <stdio.h>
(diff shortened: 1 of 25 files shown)


src/lib/macros.h
```diff
@@ -46,4 +48,19 @@
 		_ptr = NULL; \
 	} while (0)
 
+/**
+ * @brief allocate memory of indicated size and zero it.
+ *
+ * Abort on out of memory errors.
+ */
+static inline void *malloc_or_die(size_t size)
```
castulo, line 56: I find this name a little confusing because I would have assumed you were using malloc within this function. Do you think is worth using a different word to avoid this confusion? why not naming it calloc _or_die if the function is really using calloc?

PR #1555, opened 2020-05-07
Heuristics: Some consolidation on heuristics module

otaviobp:


otaviobp:
v2: Adding tests and one improvement

otaviobp:
V3: Fixing reported problems

Issue #1538, opened 2020-05-05
Create a new module statedir, to group all functions related to handling of statedir


otaviobp:
Create a statedir module with functions related operations exclusive to statedir, like cleaning staged dir, other cleaning operations and functions to build specific directories, like:

 - statedir_get_datadir()
 - statedir_get_stagedir()

PR #1558, opened 2020-05-08
Grouping functions related to handling of statedir (part 3)

castulo:
This PR adds functions to get manifest files from the statedir.

This PR is built on top of #1552 and #1557 so they need to be reviewed first.

Related to #1538


castulo:
Looks like at some point I made a mistake with the rebasing and this PR only has commits that already merged. I must have overwritten one of the PRs at some point. 🤔 
Anyway, since all the commit in the PR are already merged I'm canceling this PR and will rename PR #1559 as (part 3)

PR #1561, opened 2020-05-12
hashdump: Also sys_path_join when path_prefix is not set

otaviobp:
We don't need to prevent duplicating the "/" because sys_path_join()
removes that


otaviobp:
actually this PR doesn't make sense. I misunderstood the hashdump parameter when I wrote that.

PR #1565, opened 2020-05-12
Fixing memory leak in delta.c

castulo:
Fixing:
- Memory leak delta.c
- Dead code in thread_pool.c

### DIFF
--- a/src/swupd_lib/delta.c
+++ b/src/swupd_lib/delta.c
@@ -117,6 +117,7 @@ void apply_deltas(struct manifest *current_manifest)
 	DIR *dir = opendir(delta_dir);
 	if (!dir) {
 		/* No deltas available to apply. */
+		FREE(delta_dir);
 		return;
 	}
 


src/lib/thread_pool.c
```diff
@@ -91,12 +91,7 @@ struct tp *tp_start(int num_threads)
 	}
 
 	tp = malloc_or_die(sizeof(struct tp) + int_to_uint(num_threads) * sizeof(pthread_t *));
-
 	tp->num_threads = num_threads;
-	if (num_threads == 0) {
```
otaviobp, line 96: There's a fix for that on #1564

src/swupd_lib/delta.c
```diff
@@ -117,6 +117,7 @@ void apply_deltas(struct manifest *current_manifest)
 	DIR *dir = opendir(delta_dir);
 	if (!dir) {
 		/* No deltas available to apply. */
+		FREE(delta_dir);
```
otaviobp, line 120: +1 for this one

src/lib/thread_pool.c
castulo: I'll remove that from here then and will review #1564 😝

castulo:
V: only fixing the memory leak

Issue #1539, opened 2020-05-05
Save all downloaded files in a path dependent on the URL


otaviobp:
Everything you download should be saved in a path like ${STATEDIR}/cache/${url}/
For example, we could have the following files in the system:
 - /var/lib/swupd/cache/http_cdn.download.clearlinux.org_update/32000/Manifest.R-basic
 - /var/lib/swupd/cache/http_cdn.download.clearlinux.org_update/pack-R-basic-from-31900-to-32900.tar
 - /var/lib/swupd/cache/http_cdn.download.clearlinux.org_update/pack-R-basic-from-31900-to-32900.tar

PR #1568, opened 2020-05-14
Group manifest artifacts under the same directory in the statedir

castulo:
This PR groups all the manifest-related artifacts that swupd downloads under the same "manifest" directory, so we keep a better consistency within the statedir.

This PR is built on top of #1567 and it needs to be merged first.

Related to #1539

### DIFF
--- a/src/cmds/clean.c
+++ b/src/cmds/clean.c
@@ -466,5 +466,9 @@ enum swupd_code clean_statedir(bool dry_run, bool all)
 	}
 
 	/* NOTE: do not clean the state_dir/bundles directory */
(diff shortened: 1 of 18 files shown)


castulo:
@otaviobp This PR should be ready for review, it has tests failing because there were some changes in the API tests. But the changes in the API tests tests are safe, the location where manifests are changed so things like this had to be changed:
from
`sudo mkdir -p "$STATE1"/10`
to
`sudo mkdir -p "$STATE1"/manifest/10`

castulo:
Rebased

PR #1570, opened 2020-05-14
delta: Improve message on delta errors

### LINKED ISSUE #1425: Improve messaging on swupd update when delta file fails
When `swupd update` is run and there are delta files that can't apply, we print a warning error that is more focused on debugging the issue than explaining the problem to the user. Current message is
```
Warning: Couldn't use delta file /var/lib/swupd/delta/32750-32760-a30e825b9ac1ac220453ba191442d25e16687aee450476409cb4c304d094a803-a30e825b9ac1ac220453ba191442d
25e16687aee450476409cb4c304d094a803: 'from' file corrupted on system, consider running 'swupd verify --fix'
```

We could print something like:
```
Warning: /usr/lib/foo/bar was changed from the installed version. Replacing.
Consider running 'swupd repair' to restore any other files that were modified.
```

otaviobp:
Print a more useful error when we fail to locate a file to apply a
delta. Add a suggestion for the user to run `swupd repair` and make
it clear that the problem is in their system and not in the update
content.

Fixes #1425

### DIFF
--- a/src/swupd_lib/delta.c
+++ b/src/swupd_lib/delta.c
@@ -153,7 +153,6 @@ void apply_deltas(struct manifest *current_manifest)
 
 		struct list *ll = list_head(current_manifest->files);
 		char *found = NULL;
-		bool bad_on_sys = false;
 
 		for (; ll && !found; ll = ll->next) {
 			struct file *file = ll->data;
@@ -168,7 +167,7 @@ void apply_deltas(struct manifest *current_manifest)
 
 			if (!compute_hash_from_file(filename, hash) || !hash_equal(file->hash, hash)) {
 				FREE(filename);
-				bad_on_sys = true;
+				warn("File \"%s\" is missing or corrupted\n", file->filename);
 				continue;
 			}
 
@@ -176,12 +175,8 @@ void apply_deltas(struct manifest *current_manifest)
 		}
 
 		if (!found) {
-			warn("Couldn't use delta file %s\n", delta_file);
-			if (bad_on_sys) {
-				info("'from' file corrupted on system, consider running 'swupd verify --fix'\n");
-			} else {
-				info("no 'from' file to apply was found\n");
-			}
+			warn("Couldn't use delta file because original file is corrupted or missing\n");
+			info("Consider running \"swupd repair\" to fix the issue\n");
 			goto next;
 		}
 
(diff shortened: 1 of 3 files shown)


src/swupd_lib/delta.c
```diff
@@ -168,20 +167,15 @@ void apply_deltas(struct manifest *current_manifest)
 
 			if (!compute_hash_from_file(filename, hash) || !hash_equal(file->hash, hash)) {
 				FREE(filename);
-				bad_on_sys = true;
+				warn("File '%s' is missing or corrupted - consider running 'swupd repair'\n", file->filename);
```
castulo, line 170: In other parts of the code when we show this type of warning we do it in two steps:
```
warn("My warning\n");
info("My suggestion\n");
```
You can see an example of this here: https://github.com/clearlinux/swupd-client/blob/master/src/cmds/verify.c#L673
Also I would use "command" instead of 'command' as in the example above to keep it consistent.

otaviobp: cool, I followed your suggestion. Can you take another look

castulo: Looks good, the only thing you didn't do was using "" instead of `` like this:
`info("Consider running /"swupd repair/" to fix the issue\n");`

src/swupd_lib/delta.c
```diff
@@ -168,20 +167,16 @@ void apply_deltas(struct manifest *current_manifest)
...
 		}
 
 		if (!found) {
-			warn("Couldn't use delta file %s\n", delta_file);
-			if (bad_on_sys) {
-				info("'from' file corrupted on system, consider running 'swupd verify --fix'\n");
-			} else {
-				info("no 'from' file to apply was found\n");
-			}
+			warn("Couldn't use delta file becase original file is corrupted or missing\n");
```
phmccarty, line 178:
```suggestion
			warn("Couldn't use delta file because original file is corrupted or missing\n");
```

test/functional/update/update-verify-fix-path-missing-dir.bats
```diff
@@ -26,7 +26,9 @@ test_setup() {
 		Downloading packs for:
 		 - test-bundle
 		Finishing packs extraction...
-		Warning: Couldn.t use delta file .*
+		Warning: File '/foo/bar/baz/test-file' is missing or corrupted
+		Warning: Couldn't use delta file becase original file is corrupted or missing
```
phmccarty, line 30:
```suggestion
		Warning: Couldn't use delta file because original file is corrupted or missing
```

otaviobp:
V3: Fixing typo and replacing ' by "

PR #1574, opened 2020-05-15
Separating cache from state files in statedir

castulo:
This commit moves the cache directories and files into their own "cache"
directory in the statedir, that way it is separated from state files.

Related to #1539 


castulo:
closing in favor of #1582

PR #1577, opened 2020-05-19
Save cached files in a path dependent on the URL

### LINKED ISSUE #1539: Save all downloaded files in a path dependent on the URL

### LINKED ISSUE #1540: Save all extracted files (staged files) in a path dependent on the URL
Everything you download should be saved in a path like ${STATEDIR}/staged/${url}/
For example, we could have the following file in the system:
 - /var/lib/swupd/staged/http_cdn.download.clearlinux.org_update/aa726606daecf7b9f0af03903cd4c6180b50323f305fb30b0f9425f053c62aa1

castulo:
When downloading files, store them in a path that is dependent on the
content url to make it easier to share them for more target systems.

This PR is built on top of #1574 and needs to be merged first.

Closes #1539 
Closes #1540

### DIFF
--- a/src/swupd_lib/statedir.c
+++ b/src/swupd_lib/statedir.c
@@ -49,6 +49,9 @@
 /* Name of the directory that contains all the cache */
 #define CACHE_DIR "cache"
 
+/* Name of the directory that contains statedir of a 3rd-party repo */
+#define THIRD_PARTY_DIR "3rd-party"
+
 
 /* ************* */
 /* State section */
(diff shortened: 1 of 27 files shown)


src/swupd_lib/statedir.c
```diff
@@ -56,44 +66,128 @@ char *statedir_get_tracking_file(const char *bundle_name)
...
+}
+
+/* ************* */
+/* Cache section */
+/* ************* */
+
+static char *get_url(void)
+{
+	size_t i, j = 0;
+	char *url = strdup_or_die(globals.content_url);
```
otaviobp, line 91: You don't need to duplicate the url if you never change that.

src/swupd_lib/statedir.c
```diff
@@ -56,44 +66,128 @@ char *statedir_get_tracking_file(const char *bundle_name)
...
+	char *converted_url = malloc_or_die(str_len(url) + 1);
```
otaviobp, line 92: It's better to create another function `convert_url(const char *url)` so it can be possibly reused for other urls

src/swupd_lib/statedir.c
```diff
@@ -56,44 +66,128 @@ char *statedir_get_tracking_file(const char *bundle_name)
...
+
+	// replace ':' and '/' with '_', but do not allow
+	// multiple '_' in a row
```
otaviobp, line 95: I personally wouldn't avoiding multiple '_'. There's no problem on having that, reduces the chance on having url clashes and the code is easier.

src/swupd_lib/statedir.c
castulo: ahh good point, I guess at some point I was changing it (or planning to), I'll remove that

src/swupd_lib/statedir.c
castulo: oo, the only bad thing is that makes the name extra long (which already is)

castulo:
closing in favor of #1582

PR #1578, opened 2020-05-19
clean: Clean manifests in deprecated location

otaviobp:
Manifest directory has moved and we need to clean manifests in previous location.

### DIFF
--- a/src/cmds/clean.c
+++ b/src/cmds/clean.c
@@ -469,6 +469,13 @@ enum swupd_code clean_statedir(bool dry_run, bool all)
 	path = statedir_get_manifest_root_dir();
 	ret = clean_staged_manifests(path, dry_run, all);
 	FREE(path);
+	if (ret != SWUPD_OK) {
+		return ret;
+	}
+
+	// TODO: Remove after format bump to format 31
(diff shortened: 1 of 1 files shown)


src/cmds/clean.c
```diff
@@ -469,6 +469,13 @@ enum swupd_code clean_statedir(bool dry_run, bool all)
 	path = statedir_get_manifest_root_dir();
 	ret = clean_staged_manifests(path, dry_run, all);
 	FREE(path);
+	if (ret != SWUPD_OK) {
```
castulo, line 472: Isn't this supposed to be
`if (ret == SWUPD_OK)`?

If the provided "path" doesn't exist is going to return `SWUPD_COULDNT_LIST_DIR` and that means we need to carry on and try to remove from the old location. Am I missing something?

otaviobp: We need to always clean both. I just followed the standard on clean to fail on first error.
this shouldn't fail for that reason because the path is created on statedir setup.

castulo: ahh I see... you are right. So for the moment we will be cleaning both and eventually we'll remove this last check. Got it.

PR #1579, opened 2020-05-19
Fix syntax error in configure.ac

phmccarty:
The `configure` script generated from configure.ac tries to run a `]`
command, but that command doesn't exist.

The root cause is a syntax issue configure.ac: there is a stray `]` on
one of the lines.


castulo:
👌

PR #1582, opened 2020-06-05
Separating the cache from the data used by swupd

### LINKED ISSUE #490: Make /var/lib/swupd world-readable
We should make the state directory world-readable so we can do all `bundle-list` operations as non-root. I don't believe there's anything in there that users can't already find published on the update server.

### LINKED ISSUE #1537: Split Cache and state in 2 different directories (/var/lib/swupd and /var/cache/swupd)
Right now we save cache and state files in the same directory, /var/lib/swupd. The major problem on doing that is that cache is good to be reused in different chroots, but state is not. So state should be saved (as default) in a path related to path_prefix and cache should be saved on a path related to root system.

Tasks to complete that:
 - Create a new module statedir, to group all functions related to handling of statedir (#1538)
 - Save all cache files that are download from a server/mirror to ${STATEDIR}/cache/${url}/ (similar to what we use for 3rd-party, but on 3rd-party we can continue to use the repo_name). (#1539)
 - Save all files that are extracted on ${STATEDIR}/staged/${url}/  (#1540)
 - Create a new flag --datadir and save all data related to one specific installation on datadir. Datadir default should be ${PATH_PREFIX}/var/lib/swupd/  (#1541)
 - Save telemetry records on ${DATADIR}/telemetry
 - Save tracked bundles under ${DATADIR}/data/bundles
 - Review the correct location for delta/ download/ temp/ tmprenamedir/
 - Set the default for the --statedir to /var/lib/cachedir (future)
 - Change the read permissions for cache and data to user-readable, making sure we keep the permissions for staged very restricted. (#490)

### LINKED ISSUE #1538: Create a new module statedir, to group all functions related to handling of statedir

### LINKED ISSUE #1539: Save all downloaded files in a path dependent on the URL

### LINKED ISSUE #1540: Save all extracted files (staged files) in a path dependent on the URL

### LINKED ISSUE #1541: Create a new flag --datadir and save all data related to one specific installation on datadir.
Datadir default should be ${PATH_PREFIX}/var/lib/swupd/

So running this:

```swupd bundle-add --path /mnt/chroot --statedir /var/lib/swupd --datadir /mnt/chroot/var/lib/swupd vim```

is the same as this:

```swupd bundle-add --path /mnt/chroot vim```

And that would save all downloaded and extracted files to /var/lib/swupd, but all tracked bundles to /mnt/chroot/var/lib/swupd/

castulo:
Currently there is only one location, known as "statedir", where swupd stores data and cache. Normally it make sense to keep the data in a path that is relative to the path prefix of the target system, since the data is specific to that system, but cache can be independent of the system, and that way it can be reused for other target systems if desired.

This PR includes the following changes to accomplish this:

- It splits the cache and data in the internal memory into two different locations.
- It stores the cache in a path that is dependent of the url of the update content server
- Removes hardcoded references to files in the cache and data throughout the code so they can be relocated easily at any time if desired.
- Sets the read permissions of the data directory to be world readable.
- Sets the read permissions of all directories in the cache to be readable by root only with the exception of the `manifest` directory which is world readable.
- Updates the `swupd clean` and `swupd 3rd-party clean` commands so they remove cache from the new location.
- The `swupd clean` and `swupd 3rd-party clean` commands  will continue to clean cache from the old locations until the bump to format 31 to make sure we don't leave old files behind.
- It introduces the `--cachedir` and `--datadir` which can be used to set the locations of the cache and data respectively. If users want to set both paths to the same location they can still use `--statedir` to set both paths at once.
- Adds functional tests to validate the removal of cache from old and new locations.

Closes #1538
Closes #1539 
Closes #1540 
Closes #1541 
Closes #490 
Closes #1537


castulo:
It is expected to have the `API` job to fail because there were a couple of changes in the tests, the output hasn't changed though.

PR #1584, opened 2020-06-09
Allow paths that contain symlinks for cachedir

### LINKED ISSUE #1581: swupd cannot os-install into a path containing a symlink
**Describe the bug**
If the path that is passed to os-install contains a symlink, when swupd is trying to do an os-install, it fails right away saying that a symlink cannot be in this path.

**To Reproduce**
```
> $ sudo -E swupd os-install --wait-for-scripts --path=/home/clrbuilder/docker/base/chroot --version=33240 --statedir=/home/clrbuilder/docker/base/chroot/var/lib/swupd --format=30 --url=file:///home/update/www --no-boot-update --bundles os-core,os-core-update

Installing OS version 33240
Error: Cannot extract through symlink /home/clrbuilder/docker/base/chroot/var/lib/swupd/manifest/33240/Manifest.MoM
Error: Failed to retrieve 33240 MoM manifest
Error: Unable to download/verify 33240 Manifest.MoM
```

**Expected behavior**
On this machine, /home/clrbuilder is a symlink pointing to /srv/home/clrbuilder.  Swupd should be able to take the `realpath` from these values and use this instead when doing its operations.

 - Clear Linux OS Version: 33200, swupd 4.2.1
 - Platform: Dell R640, SSD RAID 10, Intel Xeon Gold 80 vCPUs

castulo:
When the path provided by a user for the --statedir (or --cachedir)
contains a symlink swupd will fail to extrar tar files since
libarchive does not support extracting files through symlinks.

This commit fixes the issue allowing users to use symlinks in the
--statediri/--cachedir path.

Closes #1581

### DIFF
--- a/src/lib/archives.c
+++ b/src/lib/archives.c
@@ -32,6 +32,7 @@
 #include "log.h"
 #include "macros.h"
 #include "strings.h"
+#include "sys.h"
 
 /* _archive_check_err(ar, ret)
  *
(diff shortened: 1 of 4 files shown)


castulo:
V3:
- checking for the realpath when setting the global statedir variable.

PR #1586, opened 2020-06-13
Testlib env variables

### LINKED ISSUE #1479: Testlib: include important environment variables in the testlib help
After sourcing testlib, if the user runs `testlib` from the terminal, the help of the library is displayed which includes the list of commands provided by the library. Testlib uses some environment variables to modify the way tests are run, for example running a test with DEBUG_TEST=true will print all debug messages of the test, or using  KEEP_ENV=true will prevent the environment from being removed after the test finishes. It would be helpful to list all those env variables and their use in the testlib help.

### LINKED ISSUE #1484: Testlib: zero_hash is a constant in the library so it should be ZERO_HASH
We need to replace zero_hash with ZERO_HASH since it is a constant env variable provided by swupd. Function and tests that use zero_hash need to be updated too.

### LINKED ISSUE #1489: testlib: Stop exporting env variables on functions other than set_env_variables
It's not possible to get on tests the information of any exported variable that is run on global_setup(). So there are some functions that won't work when used in global_setup() in files with multiple tests, like:
 - create_config_file()
 - create_third_party_repo()
 - create_mirror()

castulo:
This PR includes many updates for environment variables in testlib:
- Adds information about BATS env variables
- Makes variable names more consistent by following a local convention for variables
- Moves the location where variables are set and exported so they are easily identified
- Some environment variables were available to all tests within a test file while others were only available to the first test, this PR also makes all variables available to all tests within a file
- Refactors a the setup of tests to make it more straightforward and corrects some loading problems it had

Closes #1479 
Closes #1484 
Closes #1489 

It is still a work in progress. Not ready for review.

### DIFF
--- a/test/functional/3rd-party/3rd-party-allow-http.bats
+++ b/test/functional/3rd-party/3rd-party-allow-http.bats
@@ -12,13 +12,13 @@ global_setup() {
 
 	# use a web server for serving the content, this is necessary
 	# since the code behaves differently if the content is local (e.g. file://)
-	start_web_server -r -D "$TPWEBDIR"
+	start_web_server -r -D "$TP_WEB_DIR"
 
 }
 
 test_teardown() {
 
-	sudo rm -rf "$TARGETDIR"/etc/swupd
+	sudo rm -rf "$TARGET_DIR"/etc/swupd
 
 }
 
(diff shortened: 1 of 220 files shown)


castulo:
I was thinking into adding one more commit to this PR but I think I'm going to merge this as it is and will consider opening a new PR with the other change.

PR #1593, opened 2021-03-17
Fix swupd.zsh

lebensterben:
Surprisingly no one has reported this bug...


lebensterben:
BTW the CI is currently broken..

PR #1594, opened 2021-03-17
swupd.bash: Fixed bugs

lebensterben:
- Currently there's a unreported bug:
  - When the line is 'swupd 3rd-party -*' or 'swupd -*', pressing tab results in
    an error.
- This also fixes clearlinux/distribution#2241 by introducing a test first.
  Users would be notified to install `shells` bundle, which includes the
  required `bash-completion` function
- Updated the header:
  - Used a more standard and robust shebang, namely `/usr/bin/env bash`. (Note
    that before this PR we use `/bin/bash` where `/bin` is not standard binary
    installation path accoring to Linxu Filesystem Specification)
  - The boilerplate of license notice is updated.

### DIFF
--- a/swupd.bash
+++ b/swupd.bash
@@ -1,104 +1,117 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #   Software Updater - autocompletion script
 #
-#	  Copyright © 2018 Intel Corporation.
+#	  Copyright © 2016-2021 Intel Corporation.
+#	            © 2018-2021 Lucius Hu <lebensterben@users.noreply.github.com>.
 #
-#   This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


lebensterben:
It looks like there're many changes, actually not really.

https://github.com/lebensterben/swupd-client/blob/e19a957bbe76e506f3d3137f47c783ce4dab37dd/swupd.bash#L26-L29 is the new test introduced.

https://github.com/lebensterben/swupd-client/blob/e19a957bbe76e506f3d3137f47c783ce4dab37dd/swupd.bash#L59-L60 and https://github.com/lebensterben/swupd-client/blob/e19a957bbe76e506f3d3137f47c783ce4dab37dd/swupd.bash#L74-L75 are used to fix another bug.

PR #1602, opened 2022-09-01
Allow delta files without version prefix

fenrus75:


castulo (approved):

phmccarty (approved):
LGTM

src/lib/strings.h
```diff
@@ -19,12 +19,12 @@ extern "C" {
 /**
  * Helper to use strnlen() instead of strlen.
  */
-#define str_len(_str) strnlen(_str, INT64_MAX)
+#define str_len(_str) strlen(_str)
```
castulo, line 22: Using strlen is not allowed by the swupd coding rules, `strlen` is considered unsafe so strnlen has to be used instead. The same applies to strcmp. You can see the functions that are not allowed to be used in swupd here:
https://github.com/clearlinux/swupd-client/blob/master/test/code_analysis/warning_functions.bats#L11

castulo (changes requested):

test/code_analysis/warning_functions.bats
```diff
@@ -16,6 +16,8 @@
 	exceptions["dirname"]="sys.*"
 	exceptions["free"]="macros.h,verifytime.c,verifytime_main.c"
 	exceptions["calloc"]="macros.h,verifytime.c"
+	exceptions["strcmp"]="strings.h"
+	exceptions["strlen"]="strings.h"
```
castulo, line 20: I don't see why using strlen and strcmp instead of the "safer" options strnlen and strncpm. And in any case, if it is really necessary to use strlen it is better to just add an exception for the file that requires them (delta.c) instead of changing them for all files. Changing them in strings.h is going to cause those unsafe functions to be used everywhere, not just in delta.c.

fenrus75:
unfortunately this is fixing a security exposure; the str_len() macro is
far less secure than just using strlen()

so we are improving security by getting rid of this broken coding rule

castulo (changes requested):

test/code_analysis/warning_functions.bats
bryteise: As Arjan mentioned we are finding the strn* functions to be more problematic than the non-n versions, thus this change.

fenrus75: strnlen and friends have the human put in a length

strlen/strcpy and friends, with FORTIFY_SOURCE, have hte compiler put in the length of the buffer automatically.. which is more accurate.

strncpy and friends also have the nasty behavior of not 0-terminating the strings

castulo: I see. If that is the case, and the str_len and str_cmp macros are no longer needed, it may be better to just get rid of them instead of just using them as an alias for strlen and strcmp, and just remove those functions from https://github.com/clearlinux/swupd-client/blob/master/test/code_analysis/warning_functions.bats#L11. But that can probably done in a different PR.

bryteise: Yea, I was thinking about that but decided to keep the macro around for now. Who knows when things change and we need a new str* =P.

PR #1603, opened 2022-09-21
Announce overrides of content and version URLs

bwarden:
When using -v, -c, or -u to set alternate version or content URLs, print out a message so it's a little bit more obvious when you accidentally use the wrong option.


bryteise:
@bwarden Mind rebasing on latest. Hopefully CI will be happy with you.

bryteise:
Ah looks like there are some updates needed in the tests for the override string.

bryteise (approved):

PR #1605, opened 2022-10-05
Fix shellcheck directive placement

bryteise:
Signed-off-by: William Douglas <william.r.douglas@gmail.com>

### DIFF
--- a/test/real_content/swupd_update.bats
+++ b/test/real_content/swupd_update.bats
@@ -7,6 +7,7 @@ load "real_content_lib"
 	# shellcheck disable=SC2153
 	print "Install minimal system with oldest version (${VERSION[0]})"
 
+	# shellcheck disable=SC2153
 	run sudo sh -c "$SWUPD os-install $SWUPD_OPTS -V ${VERSION[0]} -F $FORMAT"
 	assert_status_is 0
 	check_version "${VERSION[0]}"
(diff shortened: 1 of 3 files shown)


test/real_content/swupd_update.bats
```diff
@@ -4,9 +4,9 @@ load "real_content_lib"
 
 # Test update in delta pack range
 @test "RCT001: Incremental Updates" {
-	# shellcheck disable=SC2153
```
castulo, line 7: I believe this directive has to remain here so shellcheck doesn't complaint about VERSION being used in the print statement.

bryteise: Yep. Do you know if this need to be on every use of the variable defined in the loaded file or just the first instance?

castulo: I'm a bit confused, the documentation suggests it would need to happen before every use of the variable:
"Directives that replace or are immediately after the shebang apply to the entire script. Otherwise, they are scoped to the command that follows it (including compound commands like function definitions, loops and case statements). A directive may only be applied to a complete command" (https://github.com/koalaman/shellcheck/wiki/directive#shell). But in these test files it would seem like just adding it before the first instance suffice.  I guess that as long as shellcheck is not complaining it is fine to just use it once.

bryteise: Yea, it seems fine with the single directive for however many uses. Maybe at some point later this will break again and I'll need to adjust.

phmccarty: If you add the directive to line 2 (after the shebang line), does that impact the entire script, as the shellcheck docs suggest?

PR #1607, opened 2023-01-27
Move packs to download synchronously

bryteise:
There are some problems with how we are using multiplexed curl causing huge slowdowns (2-5 times slower).

For now rather than rework the multiplexed curl code, switch to using the synchronous code path.

Signed-off-by: William Douglas <william.douglas@intel.com>


bryteise:
Still working on the migrating fullfiles but wanted to get the github checks running on packs first.

fenrus75 (approved):

PR #1608, opened 2023-02-07
Fix broken links

carlwiede:
I was reading up on what Clear Linux was and found some broken links so I went ahead and fixed them :)


bryteise (approved):

bryteise:
Thanks!

PR #1609, opened 2023-02-17
Fix tests

bryteise:
Update tests for new ubuntu CI container.


bryteise:
The API test failure is expected due to the test checking if there is a diff to any api test. The test changes but not the output check so it is not a real failure for what is being validated.

fenrus75 (approved):

PR #1621, opened 2024-01-20
Add APX flag support

bryteise:
Add support for new APX optimization level. Change increases the maximim potential optimization level but does not yet allow the files to be used (installed). get_opt_level_mask still needs to be updated to account for system supporting the new optimization level (and adding APX_SKIP_FILE support) before the APX files can be installed.

This change is being done at this time to allow mixer to create these files without swupd error messages during updates.


src/swupd_lib/manifest_parser.c
```diff
@@ -364,7 +376,7 @@ struct manifest *manifest_parse(const char *component, const char *filename, boo
 		}
 
 		bitmask_translation = OPTIMIZED_BITMASKS[(unsigned char)line[2]];
-		if (bitmask_translation > AVX512_3) {
+		if (bitmask_translation > APX_7) {
 			error("Skipping unsupported file optimization level\n");
```
fenrus75, line 380: given our previous lesson should we limit this message to a once kind of thing?

fenrus75 (approved):

src/swupd_lib/manifest_parser.c
bryteise: Good point, will change that.

fenrus75 (approved):

PR #1624, opened 2024-04-04
Rework how curl errors are checked

bryteise:
In cases where curl perform returns an error but is actually a 400 level response, swupd needs to check curl info to see this. If the info reveals a 400 level response, return the corresponding download status error to prevent useless download retries.

### DIFF
--- a/src/swupd_lib/curl.c
+++ b/src/swupd_lib/curl.c
@@ -456,6 +456,23 @@ enum download_status process_curl_error_codes(int curl_ret, CURL *curl_handle)
 			return DOWNLOAD_STATUS_ERROR;
 		}
 	} else { /* download failed but let our caller do it */
+		long response = 0;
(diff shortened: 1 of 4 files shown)


src/swupd_lib/curl.c
```diff
@@ -456,6 +456,23 @@ enum download_status process_curl_error_codes(int curl_ret, CURL *curl_handle)
 			return DOWNLOAD_STATUS_ERROR;
 		}
 	} else { /* download failed but let our caller do it */
+		long response = 0;
+		curl_easy_getinfo(curl_handle, CURLINFO_RESPONSE_CODE, &response);
```
bwarden, line 460: Fine for now, but when we refactor, we should move this call before this if/else and figure out how to condense these conditions better.

bwarden (approved):

PR #1625, opened 2024-04-04
Dont delete user content

bryteise:


src/swupd_lib/target_root.c
```diff
@@ -756,5 +814,26 @@ int target_root_remove_files(struct list *files)
...
+		dirs_iter = dirs_iter->next;
+		fullfile = sys_path_join("%s/%s", globals.path_prefix, file->filename);
+		ret = sys_rm(fullfile);
+		if (ret != 0) {
+			ret = move_to_timestamp(file->filename);
+			if (ret != 0) {
+				error("Could not move file to backup location: %s\n", file->filename);
+			}
+		}
+		if (ret != 0) {
```
bwarden, line 829: Squash this `if` into the previous?

src/swupd_lib/target_root.c
```diff
@@ -732,21 +784,27 @@ enum swupd_code target_root_install_files(struct list *files, struct manifest *m
...
 		file = iter->data;
 		iter = iter->next;
 		fullfile = sys_path_join("%s/%s", globals.path_prefix, file->filename);
-		if (sys_rm_recursive(fullfile) == -1) {
-			/* if a -1 is returned it means there was an issue deleting the
-			 * file or directory, in that case decrease the counter of deleted
-			 * files.
+		ret = sys_rm(fullfile);
+		if (ret == -ENOTEMPTY) {
+			count--;
```
bwarden, line 803: I don't understand why we're decrementing `count` here. Maybe add comments to explain the use and relationship of `count`, `total`, and `deleted`?

src/swupd_lib/target_root.c
```diff
@@ -330,6 +333,61 @@ static enum swupd_code install_dir(const char *fullfile_path, const char *target
 	return install_dir_using_tar(fullfile_path, target_file);
 }
 
+/* Backup target directory to a backup and timestamp prefixed name. */
+/* This function should only be used after the deletes are processed. */
+static enum swupd_code move_to_timestamp(const char *filename)
```
fenrus75, line 338: maybe call it "dirname" not "filename" to indicate we expect it to be a directory?

src/swupd_lib/target_root.c
```diff
@@ -330,6 +333,61 @@ static enum swupd_code install_dir(const char *fullfile_path, const char *target
...
+		ret = SWUPD_TIME_UNKNOWN;
+		goto out;
+	}
+
+	strftime(time_str, sizeof(time_str), "%s", timeinfo);
+	backup_dir = sys_path_join("%s/%s%s.%s", target_path, BACKUP_FILE_PREFIX, time_str, target_basename);
+
+	/* It shouldn't be possible to have an already created backup_dir
+	   but just in case, try and remove it and error out if unable to. */
+	ret = sys_rm_recursive(backup_dir);
```
fenrus75, line 367: shuold we actually check if this is there first? just incase we have some bug... then it only ever happens if it actually exists

fenrus75 (approved):

src/swupd_lib/target_root.c
bryteise: Sure, less to go wrong. I suspect if things are going wrong in this area though we are in a whole different world of hurt that'll show up all over in testing.

src/swupd_lib/target_root.c
```diff
@@ -330,6 +333,61 @@ static enum swupd_code install_dir(const char *fullfile_path, const char *target
...
+
+	ret = rename(target_file, backup_dir);
+	if (ret != 0) {
+		ret = SWUPD_COULDNT_RENAME_DIR;
+		goto out;
+	}
+
+out:
+	FREE(dir);
+	FREE(target_file);
```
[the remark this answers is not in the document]
bryteise: No, it calls basename under the hood and that's not something that should be free'd.

src/swupd_lib/target_root.c
```diff
@@ -632,6 +665,25 @@ static int rename_all_files_to_final(struct list *updates)
...
+	while (dirs_list) {
+		struct file *file;
+		file = dirs_list->data;
+		dirs_list = dirs_list->next;
+
+		ret = rename_staged_file_to_final(file);
+		if (ret != 0) {
+			ret = move_to_timestamp(file->filename);
+			if (ret != 0) {
+				error("Could not move file to backup location: %s\n", file->filename);
```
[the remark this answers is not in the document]
bryteise: No, this isn't "fatal" it would happen when a directory swupd thinks should be deleted but user content in it can't be renamed for some reason likely disk related I'd expect. The directory would be left with whatever user content was in it (that's what the filename would be in this case). Updates after will still work fine too but swupd repair --picky will try and delete it.

src/swupd_lib/target_root.c
bryteise: So I think what's going on here is that count is how many files the loop has processed so far and I basically am decrementing it because in this case we still need to process it in the next loop (there is an unconditional increment below).

deleted starts as total and is decremented each time a file fails to be deleted.

I can definitely add a bit of documentation around that, this is all for book keeping reporting to the user (deleted should match total for a successful operation).

src/swupd_lib/target_root.c
bryteise: Ah actually this was a mistake, the above if should have been checking if the error was another ENOTEMPTY instead of any error. Turning this into an else if though.

PR #1629, opened 2024-07-27
Test fixups and update for newer systemd

bryteise:
systemctl reports that the timer file is missing now as a return code 4 so update the unknown start number.

Also with this change SWUPD_NO is no longer always returned so add a new helper to check if a status is one of multiple options.

### DIFF
--- a/test/functional/api/api-autoupdate.bats
+++ b/test/functional/api/api-autoupdate.bats
@@ -16,7 +16,7 @@ test_setup() {
 
 	run sudo sh -c "$SWUPD autoupdate $SWUPD_OPTS --quiet"
 
-	assert_status_is "$SWUPD_NO"
+	assert_status_in "$SWUPD_NO" "$SWUPD_SUBPROCESS_ERROR"
 	assert_output_is_empty
 
 }
--- a/test/functional/testlib.bash
+++ b/test/functional/testlib.bash
@@ -4823,6 +4823,38 @@ assert_status_is() { # assertion
 
 }
 
+assert_status_in() { # assertion
+
+	local expected_statuses="$*"
+	validate_param "$expected_statuses"
+
+	if [ -z "$status" ]; then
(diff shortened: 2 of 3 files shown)


bryteise:
@bwarden So I'm inclined to merge this though the API detection change is real. The underlying systemctl command seems to have changed how it reports missing unit files so our reporting should update as well. That said, the test for running `systemctl is-enabled/is-active` isn't great so maybe something more could be done for the test if I'm ignoring an error anyway.

test/functional/api/api-autoupdate.bats
```diff
@@ -16,7 +16,7 @@ test_setup() {
 
 	run sudo sh -c "$SWUPD autoupdate $SWUPD_OPTS --quiet"
 
-	assert_status_is "$SWUPD_NO"
+	assert_status_in "$SWUPD_NO" "SWUPD_SUBPROCESS_ERROR"
```
bwarden, line 19:
```suggestion
	assert_status_in "$SWUPD_NO" "$SWUPD_SUBPROCESS_ERROR"
```

test/functional/testlib.bash
```diff
@@ -4823,6 +4823,37 @@ assert_status_is() { # assertion
...
+	if [ -z "$status" ]; then
+		echo "The \$status environment variable is empty."
+		echo "Please make sure this assertion is used inside a BATS test after a 'run' command."
+		return 1
+	fi
+
+        local found=0
+	for expected_status in $expected_statuses; do
+		if [ "$status" -eq "$expected_status" ]; then
+			found=1
```
bwarden, line 4840:
```suggestion
			found=1
			break
```

test/functional/api/api-autoupdate.bats
bryteise: Oops, vision sadness.

PR #1630, opened 2024-08-12
Add error handling for curl_easy_getinfo

bryteise:
Set response explicitly in case of curl_easy_getinfo error (though in practice the response isn't modified and so program behavior is the same as before this change).

### DIFF
--- a/src/swupd_lib/curl.c
+++ b/src/swupd_lib/curl.c
@@ -416,6 +416,7 @@ CURLcode swupd_download_file_close(CURLcode curl_ret, struct curl_file *file)
 enum download_status process_curl_error_codes(int curl_ret, CURL *curl_handle)
 {
 	char *url;
+	int info_ret;
 	if (curl_easy_getinfo(curl_handle, CURLINFO_EFFECTIVE_URL, &url) != CURLE_OK) {
 		url = "<not available>";
 	}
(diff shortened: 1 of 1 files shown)


src/swupd_lib/curl.c
```diff
@@ -457,7 +457,9 @@ enum download_status process_curl_error_codes(int curl_ret, CURL *curl_handle)
 		}
 	} else { /* download failed but let our caller do it */
 		long response = 0;
-		curl_easy_getinfo(curl_handle, CURLINFO_RESPONSE_CODE, &response);
+		if (curl_easy_getinfo(curl_handle, CURLINFO_RESPONSE_CODE, &response) != CURLE_OK) {
+			response = -1; // Force error
```
bwarden, line 461: I wonder if we should go ahead and decode the error code with `curl_easy_strerror` to produce an informative message here.
https://curl.se/libcurl/c/libcurl-errors.html#CURLEOK

src/swupd_lib/curl.c
```diff
@@ -457,7 +457,9 @@ enum download_status process_curl_error_codes(int curl_ret, CURL *curl_handle)
 		}
 	} else { /* download failed but let our caller do it */
 		long response = 0;
-		curl_easy_getinfo(curl_handle, CURLINFO_RESPONSE_CODE, &response);
+		if (curl_easy_getinfo(curl_handle, CURLINFO_RESPONSE_CODE, &response) != CURLE_OK) {
```
bwarden, line 460:
```suggestion
		if ((curl_ret = curl_easy_getinfo(curl_handle, CURLINFO_RESPONSE_CODE, &response)) != CURLE_OK) {
				error("curl_easy_getinfo: %s\n", curl_easy_strerror(curl_ret));
		
```

bryteise: I don't know if we want to be changing curl_ret as this is used in other parts of the function. I agree we might be able to get better error messages though. Let me try another patch.

bwarden (approved):

PR #1638, opened 2024-10-17
Update github actions

bryteise:


bwarden (approved):
Took me a minute to understand this is to satisfy dependabot _and_ keep it from opening a new PR for every minor version.

